PR #7, opened 2018-10-14
Fix: make_banner.sh

Nold360:
Hi,

i had some trouble building the latest toolchain. There were two problems in "make_banner.sh" for me.
First, line 13 "git describe" didn't give me any value back. I've forced it using "--always".

Also the if-block in [line 47](https://github.com/Nold360/KallistiOS/blob/72be57894c7c9aab7dd69f5dd085ca05cf79ac5e/kernel/arch/dreamcast/kernel/make_banner.sh#L4) fails on a distribution where /bin/sh is not simply a link to /bin/bash (like debian).

### DIFF
--- a/kernel/arch/dreamcast/kernel/make_banner.sh
+++ b/kernel/arch/dreamcast/kernel/make_banner.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 # Re-creates the banner.h file for each compilation run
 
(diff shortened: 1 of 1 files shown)


ljsebald (approved):

ljsebald (changes requested):
A better approach here would be changing the problematic test to use standard POSIX syntax instead, which has already been done on the SourceForge tree (which for some reason, I forgot to bring over to this mirror of the repo).

ljsebald:
I've merged the first change from this pull request, while bringing over an alternative fix for the other one.

Thank you for your contribution!

PR #12, opened 2018-10-23
Avoid compiled "addons" libraries removal when cleaning

sizious:
When doing `make clean` on **KallistiOS** root, all objects are removed including compiled binaries.
The removal of compiled libraries should be done only with the `make distclean` command.

### DIFF
--- a/addons/Makefile.prefab
+++ b/addons/Makefile.prefab
@@ -25,6 +25,6 @@ linklib:
 clean: defaultclean
          
 defaultclean: clean_subdirs
-	-rm -f $(OBJS) $(KOS_BASE)/addons/lib/$(KOS_ARCH)/$(TARGET) $(LOCAL_CLEAN)
+	-rm -f $(OBJS) $(LOCAL_CLEAN)
 
 include $(KOS_BASE)/Makefile.rules


ljsebald:
I'm not sure I see why there's a need to change the behavior here. The entire addons tree takes literally seconds to build and is often more closely tied to the kernel code than is (for instance) things in kos-ports.

If the ports tree still placed its compiled libraries here, there might be more of a point, but that hasn't been the case for a while now...

sizious:
This request was made because I usually install a newer **KallistiOS** release with the following command: `make && make clean`. I'm doing so because I like to clean all temporary objects (e.g. `*.o` files) after compiling the project.

Then, you said me the following: [the libraries in the addons directory are all compiled along with KOS. No need to compile them separately](http://dcemulation.org/phpBB/viewtopic.php?f=29&t=104696&p=1055256#p1055262).

The issue is when you run `make clean` in the **KallistiOS** root (like I did), the compiled binaries in `addons/` are gone, but not the **KallistiOS** main libraries. `libkallisti` are removed only when running `make distclean` (and this behaviour is correct to me).

This pull request homogenize the `clean` and `distclean` commands between the main **KallistiOS** library and **addons** libraries.

PR #16, opened 2019-07-17
[utils/dc-chain/download.sh] prefer HTTPS over FTP

duckinator:
Change all the FTP URLs to HTTPS for security reasons.

Conveniently, all of them use the same URLs, just different protocols. :)


ljsebald:
Sorry for the late reply, but thanks for the contribution. It has been merged.

PR #20, opened 2020-01-11
Add a thd_each function for iterating the current threads

Kazade:
Adds a method for iterating the current threads so their context can be inspected.


ljsebald:
Merged and pushed to the main repo.

PR #24, opened 2020-04-07
GCC 9 fix

einsteinx2:
Fixes building with deps, updates the deps to higher versions, and switches to install-strip to drastically reduce final toolchain size.


ljsebald:
Merged, thanks.

PR #25, opened 2020-04-08
Prefer curl over wget in dc-chain download script

einsteinx2:
Curl works better when building Docker images as wget prints to many lines while curl prints over the same line.

All I did was switch the order of the checks in download.sh and added the --progress-bar flag that is used in the kos-ports Makefile to get the same style of output.


ljsebald:
Fine by me. Committed.

PR #30, opened 2020-06-06
Fixed sfx vol 255 crash

andressbarajas:
Fixes audio crash issue when trying to play sound effects using
`snd_sfx_play(hnd, vol, pan)`
with vol = 255.

Streaming sounds doesn't suffer from this issue and I can see why.  We hard set the value while for sound effects we "ramp up" to the value.  Still not sure why that ramp up for-loop crashes the Aica for 255 only but why do we need to ramp up anyway?  I see in the sound example provided at http://mc.pp.se/dc/files/sound.tar.gz we hard set the value and even in libs3m (which doesn't exist anymore in current kos-ports) we hard set it as well https://github.com/losinggeneration/kos-ports/blob/master/libs3m/s3mplay/aica.c#L161.

Not sure where ramping up came from but it looks like that approach no longer works.

### DIFF
--- a/kernel/arch/dreamcast/sound/arm/aica.c
+++ b/kernel/arch/dreamcast/sound/arm/aica.c
@@ -96,7 +96,6 @@ void aica_play(int ch, int delay) {
 
     uint32 freq_lo, freq_base = 5644800;
     int freq_hi = 7;
-    uint32 i;
     uint32 playCont;
 
     /* Stop the channel (if it's already playing) */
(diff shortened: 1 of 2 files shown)


andressbarajas:
I found out what the ramp up for-loop is for.  It's to prevent cracking when stopping and playing an audio file.  Which normally isn't an issue for sfx since they happen so fast but for streaming it would be nice to be able to stop/pause and play without hearing a crack sound. Please wait and I'll see if I can get the for-loop or another approaching working.

andressbarajas:
@ljsebald Alright, I figured out somewhat why the for-loop failed. Its because i is declared as an unsigned int (int => unsigned int commit: https://github.com/KallistiOS/KallistiOS/commit/9a8efc21a0bbb1d3e70d2913f48c15cbb0eb6348#diff-29e49bb2c0f8126f9b51ec6f8c1a37d8).  When volume is 255,`i` eventually gets set to -1 which causes an infinite for-loop.  Still...even after fixing this the ramp up code doesn't really seem to be doing anything. You can still hear pops/cracks when pausing and playing streams(I tested ramp up on streams as well).  So I think we are good with just hard setting the value as is.

PR #37, opened 2020-06-24
Ramdisk bug fix and Romdisk comment fix

andressbarajas:
Have ramdisk header match romdisk  (MAX_RAM_FILES publicly available in fs_ramdisk.h like MAX_RD_FILES for fs_romdisk.h)


andressbarajas:
Fixed fs_ramdisk_attach() where it was cutting off the first character of the filename.  The function now accepts filename in the formats: "/temp.ogg" & "temp.ogg"

ljsebald:
The MAX_RAM_FILES and MAX_RD_FILES probably should have better names if they're in public headers like that, to be perfectly honest. Also, they should probably be moved to kos/opts.h, since that's what that header was designed for...

Care to rename them and move them there as part of this? Perhaps naming them something like FS_ROMDISK_MAX_FILES and FS_RAMDISK_MAX_FILES?

andressbarajas:
I also moved MAX_ISO_FILES (now FS_CD_MAX_FILES) to opts.h as well

PR #40, opened 2020-06-30
Controller Attach/Detach Callbacks

andressbarajas:
Much like cont_btn_callback, added ability to have a callback executed for attaching/detaching a maple that supports the specified functions([uint32 MAPLE_FUNCs](https://github.com/KallistiOS/KallistiOS/blob/c9ff49a6b0ed6c74a8b5bd7a162a7857002cf626/kernel/arch/dreamcast/include/dc/maple.h#L142)). 0 for all maple devices.

I haven't really dug into using maple devices all that much so I don't know the best way to do things.  Im guessing we always have to scan manually for changes for maple devices(attached/detached) and thought this would be a nice to have since KOS already handles attaching/detaching anyway.

I also noticed that when reattaching a VMU to a controller, the controller gets detached and attached and then the VMU gets attached which I found interesting. This intended?

Was easy to code, just wanted to open a dialog if we may want to include this.


ljsebald:
With regard to the VMU causing the controller to detach and reattach, that's the hardware doing that. If you ever looked at a commercial game when plugging in a VMU while on the save/load screen, you'd probably see the same thing happen (IIRC, PlanetWeb demonstrates it quite well).

As for the concept at hand here, I don't necessarily think it is a bad idea to have this, but I have to think a bit on how it might be best implemented. I don't really like the idea of only having one of these for all function types (or whatever subset you set)...

andressbarajas:
My thought process was that we only need one because in the callback you could look at dev->info.functions and act on it accordingly. But Im excited to see what you come up with.

ljsebald (approved):

Issue #15, opened 2019-06-04
vmu_pkg and vmu_hdr structs are inconsistent


Protofall:
I accidentally posted this on a mirror, so re-posting here.

<http://gamedev.allusion.net/docs/kos-2.0.0/structvmu__pkg.html>
<http://gamedev.allusion.net/docs/kos-2.0.0/structvmu__hdr.html>

The pkg struct is converted into the hdr struct for easy usage with VMU save files, however almost every variable has a different type or number of elements depending on which struct you look at and usually the pkg vars are larger than the hdr vars. This can cause issues if you make a vmu_pkg struct with a 20 char "app_id" name and try to use it to make a save file, the last 4 chars will be cut off since the header's var is 4 chars shorter.

From what I can tell, this appears to be an oversight of sorts. Maybe they used to be the same, but then an update changed one struct, but not both. I believe the vmu_hdr struct is correct since it is the header format <http://mc.pp.se/dc/vms/fileheader.html>

andressbarajas:
No oversight. Explanation here https://github.com/KallistiOS/KallistiOS/pull/41

Protofall:
Makes sense. All g

PR #41, opened 2020-06-30
Fix vmu_pkg to match vmu_hdr property length

andressbarajas:
Addresses the [issue](https://github.com/KallistiOS/KallistiOS/issues/15) pointed out by Protofall.  Just made sure that property lengths/types matched between vmu_pkg and vmu_hdr

### DIFF
--- a/kernel/arch/dreamcast/include/dc/vmu_pkg.h
+++ b/kernel/arch/dreamcast/include/dc/vmu_pkg.h
@@ -33,13 +33,13 @@ __BEGIN_DECLS
     \headerfile dc/vmu_pkg.h
 */
 typedef struct vmu_pkg {
-    char        desc_short[20];     /**< \brief Short file description */
-    char        desc_long[36];      /**< \brief Long file description */
-    char        app_id[20];         /**< \brief Application ID */
-    int         icon_cnt;           /**< \brief Number of icons */
-    int         icon_anim_speed;    /**< \brief Icon animation speed */
-    int         eyecatch_type;      /**< \brief "Eyecatch" type */
-    int         data_len;           /**< \brief Number of data (payload) bytes */
+    char        desc_short[16];     /**< \brief Short file description */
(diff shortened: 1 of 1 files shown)


ljsebald:
The reason they're longer in the package structure than in the header structure is to allow for NUL terminating of the strings while still respecting the alignment of the structure (at least I think that was the idea). If I recall correctly, the strings do not have to be NUL terminated when being saved (so a name with an actual length of 16 is perfectly valid -- you don't have to waste one of those spaces on a NUL character).

andressbarajas:
That makes sense, but why the four extra bytes everywhere? desc_short & desc_long are padded with space while app_id is NUL terminated.  Would it be fine then to change it to:

```
char        desc_short[17];     /**< \brief Short file description */
char        desc_long[33];      /**< \brief Long file description */
```
and keep
`char        app_id[16]; `  
The same because it is NUL terminated?

QuzarDC:
> 
> 
> That makes sense, but why the four extra bytes everywhere? desc_short & desc_long are padded with space while app_id is NUL terminated. Would it be fine then to change it to:
> 
> ```
> char        desc_short[17];     /**< \brief Short file description */
> char        desc_long[33];      /**< \brief Long file description */
> ```
> 
> and keep
> `char app_id[16]; `
> The same because it is NUL terminated?

They have an extra 4 bytes to maintain their 32-bit alignment. Various systems in the DC (and thus in KOS) require 32 bit alignment to function properly.

andressbarajas:
@QuzarDC Totally makes sense.  Thanks for explanation. I'll close maybe one of you can close the issue by Protofall?

ljsebald:
To be fair, with the exception of strings stacked right up next to each other, GCC will automatically put in the extra alignment bytes anyway...

That said, I should probably add some comments somewhere explaining why those fields are the way they are, since there's obviously at least some confusion over it.

PR #42, opened 2020-07-02
Added vmu functionality

andressbarajas:
1. Separated out XBM to 1-bitmap logic to its own utility function.
2. Added vmu_draw_lcd_xbm() so the user can set a single vmu screen using xbm image instead of all vmu screens using vmu_set_icon()
2. Added vmu_set_icon_shape(), vmu_set_custom_color() & vmu_use_custom_color() to use features that exist in the root block of the VMU flash memory referenced [HERE](http://mc.pp.se/dc/vms/flashmem.html)
3. Added vicon_find_icon() and #defines offsets for VMU icons that exist in the bios. See bottom of page [Here](http://submarine.org.uk/info/biosfont/) to see the icons.


ljsebald:
I apologize for the late review on this (I know I've been saying I'd do it for a while now)... Real life and all...

Anyhow, everything looks good functionality-wise (thanks for the new stuff!), but I'd prefer it if the stuff added in the biosfont stuff all started with bfont. By that, I mean changing the VICON_* constants to be BFONT_ICON_*, and the vicon_find_icon to be bfont_find_icon instead. That way things stay relatively consistent with what is already there.

PR #43, opened 2020-07-07
Update types to match newlib 3.3.0

tsowell:
Hi, I noticed that localtime_r() only returns the correct date the first time it was called.

For example, this code:
```
  struct tm tm;                                                                 
  time_t t;                                                                    
  t = time(0);                                                                  
  localtime_r(&t, &tm);                                         
  printf("%d\n", tm.tm_year);                                                   
  localtime_r(&t, &tm);                                         
  printf("%d\n", tm.tm_year);                                                   
  localtime_r(&t, &tm);                                         
  printf("%d\n", tm.tm_year);
```

Gives me these results:
120
4611
2978

The problem is that newlib has changed its default time_t to long long, but KOS has it hard-coded to long in include/sys/_types.h

Strangely, making time_t match in either direction (32-bit/32-bit or 64-bit/64-bit in KOS/newlib) breaks other stuff in newlib (printf, notably).  I compared the sizes of every type in _types.h between KOS and newlib and found that ino_t had also changed, from long to short.  Changing both of them at once to match newlib gets everything working again, so I think we were lucky newlib worked like this at all - things must have aligned just right!

### DIFF
--- a/include/sys/_types.h
+++ b/include/sys/_types.h
@@ -108,7 +108,7 @@ typedef unsigned long __id_t;
 #endif
 
 #ifndef __ino_t_defined
-typedef unsigned long __ino_t;
+typedef unsigned short __ino_t;
 #endif
 
 #ifndef __pid_t_defined
@@ -127,7 +127,7 @@ typedef unsigned short __nlink_t;
 typedef long        __suseconds_t;  /* microseconds (signed) */
 typedef unsigned long   __useconds_t;   /* microseconds (unsigned) */
 
-#define _TIME_T_ long
+#define _TIME_T_ long long
 typedef _TIME_T_    __time_t;
 
 #ifndef __clockid_t_defined


ljsebald:
I'm not quite sure I see why making it 64-bit all around by itself (without messing with the type of ino_t) would break printf(). That just doesn't make sense at all...

I'm not going to merge this just yet, only because "officially" the supported version for everything is still the older dc-chain defaults. Not to mention that I do intend to continue to support the older versions of Newlib for at least a while... Perhaps there's a way to condition these changes on the newer version of Newlib by using some of the Newlib macros to determine what version of Newlib is actually in use?

tsowell:
> I'm not quite sure I see why making it 64-bit all around by itself (without messing with the type of ino_t) would break printf(). That just doesn't make sense at all...

The best explanation I could come up with was that there were both ino_t and time_t elements in a struct that were affecting the placement of other elements, and that it worked by coincidence when both were incorrect.

> I'm not going to merge this just yet, only because "officially" the supported version for everything is still the older dc-chain defaults. Not to mention that I do intend to continue to support the older versions of Newlib for at least a while... Perhaps there's a way to condition these changes on the newer version of Newlib by using some of the Newlib macros to determine what version of Newlib is actually in use?

Sounds like the right way forward.  Sorry, I assumed the current dc-chain versions were the official ones.  Are these the official ones, from the FAQ, "sh-elf GCC 4.7.3 in kos threads mode, sh-elf binutils 2.23.2, newlib 2.0.0 (patched for KOS), and KOS the Git repository."?  Thanks!

ljsebald:
Sorry for the late reply, but yes, those are the "officially" supported versions still at this point.

ljsebald:
Commit d50dd457bbdde503bd4928ae61d5ce05ff41646f changed KOS' time_t to 64 bits on Newlib 3.0.0 and newer (since that appears to be when Newlib made the change).

Commit 4cc401b276210dbc471fa5953cd4f63e319422a8 changes Newlib to use 32 bits for ino_t in our patches.

Everything seems to compile cleanly with these changes, although I haven't actually tested them properly just yet...

PR #47, opened 2020-09-13
feat: support basic version of genromfs on windows

mrneo240:
- no support for symlinks
- basic romfs images only
- cleans warnings for compilers newer than 2001


mrneo240:
I just realized the change at the beginning of moving netinit to linux only might break cygwin compile. 
someone please check & verify

PR #54, opened 2021-06-18
Generate separate kernel lib files per subarch

Tchan0:
As-is, the kernel lib file libkallisti.a is overwritten whenever you build for another subarch.

With this patch, separate kernel lib files are generated, depending on the subarch:
- pristine -> libkallisti.a (remains as-is, to minimize the impact on side-projects)
- navi -> libkallistinavi.a
- naomi -> libkallistinaomi.a

### DIFF
--- a/doc/environ.sh.sample
+++ b/doc/environ.sh.sample
@@ -19,6 +19,17 @@ export KOS_ARCH="dreamcast"
 #  "naomi" - a NAOMI or NAOMI 2 arcade board
 export KOS_SUBARCH="pristine"
 
+
+#suffix used to generate subarch-specific lib files
+case $KOS_SUBARCH in
+  "navi" )
+    export KOS_SUBARCH_SUFFIX="navi" ;;
+  "naomi" )
+    export KOS_SUBARCH_SUFFIX="naomi" ;;
+  *)
+    export KOS_SUBARCH_SUFFIX="" ;;
+esac
+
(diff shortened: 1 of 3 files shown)


ljsebald:
I'm not sure I particularly like the idea of naming the library differently per subarch... Yes, the environ script and the gnu_wrappers can (and do) hide all of it from the user, but it just doesn't sit right with me for some reason...

Let me think a bit on this one before I act on it.

Tchan0:
No problem, I'm also a little bit torn about it to be honest.

A little bit of context: I want to compile my app both for Dreamcast & Naomi.

Yesterday, I only saw 2 solutions for that:
1) either duplicate the whole kos directory, with the hassle of keeping those in sync...
   I find it also strange to have to keep the same source code on 2 places.
   
2) or differentiate the lib output by giving them other names.
   This would solve the issues I see with 1), but doesn't work well eg for the /examples yet, since only 1 output is kept.
   (but since I'm building my app, I didn't really care about the examples at that point)
   
I think a solution 3) might be cleaner:
3) keep the lib names the same, but foresee separate /build subdirectories where a difference needs to be made.
	(or only create /build subdirectories for the naomi/navi where needed, and keep the rest as-is)

Tchan0:
Abandoning this pull request - having 2 completely separate folders is the best way to go.

PR #70, opened 2022-03-03
Fix DHCP issues

Kazade:
These changes allow the dns-client example to work over dc-tool-ser and dc-tool.

The changes are:

 - Fixes a crash when closing a UDP socket when there are multiple packets to free
 - Allow sending UDP packets from 0.0.0.0 (for DHCP)
 - Make a DHCP request when running over dcload-ip and use the passed IP address as the "requested address". This ensures that networking is properly initialised with gateway, DNS, etc. and the IP is properly leased.
 - Ensure that DHCPDISCOVER options are always at least 64 bytes to comply with https://www.rfc-editor.org/rfc/rfc951
 - Read static vs DHCP setting correctly from the flashrom
 - If the method is static, use what we can from the flashrom to init the connection (BBA + LAN)
 - Read block 0xC6 from the flashrom (could be a separate PR, but I was already fiddling with this stuff)

### DIFF
--- a/kernel/net/net_dhcp.h
+++ b/kernel/net/net_dhcp.h
@@ -147,7 +147,7 @@ int net_dhcp_init(void);
 
 void net_dhcp_shutdown(void);
 
-int net_dhcp_request(uint32* required_address);
+int net_dhcp_request(uint32 required_address);
 
 __END_DECLS
 
--- a/kernel/net/net_dhcp.h
+++ b/kernel/net/net_dhcp.h
@@ -147,7 +147,7 @@ int net_dhcp_init(void);
 
 void net_dhcp_shutdown(void);
 
-int net_dhcp_request(void);
+int net_dhcp_request(uint32* required_address);
 
 __END_DECLS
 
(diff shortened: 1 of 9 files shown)


kernel/net/net_dhcp.h
```diff
@@ -147,7 +147,7 @@ int net_dhcp_init(void);
 
 void net_dhcp_shutdown(void);
 
-int net_dhcp_request(void);
+int net_dhcp_request(uint32* required_address);
```
ljsebald, line 150: I'm curious why the parameter here is a pointer. Seeing how it is used, I think it would be better to just pass the normal packed form of an IPv4 address here. Especially since passing 0 would make the current behavior free, without needing the `?` that was added into the function.

Kazade: Fixed.

DC-SWAT:
@ljsebald merge please. It's good fixes.

ljsebald:
@Kazade can merge this whenever he feels it is ready. He has permission to do so, as a contributor to the project, after all.

DC-SWAT:
> @Kazade can merge this whenever he feels it is ready. He has permission to do so, as a contributor to the project, after all.

I got it, thank you!

@Kazade Did you finish it or have issues?

Kazade:
So I think these are ready, I just wanted to make sure the flashrom stuff is really REALLY tested before I merge :)

Kazade:
Hmm, on second thought the Driving Strikers demos have been using this fine for a while now and I know it fixed a bunch of issues for various people. I'll merge and follow up if I find anything that needs fixing :)

PR #76, opened 2022-06-16
Fixed controller capability Masks

ArtemioUrbina:
Confirmed with a several official controllers in hardware, using the 240p Test Suite Controller Test


asiekierka:
... Huh. That's just an endianness swap, isn't it?

ArtemioUrbina:
> ... Huh. That's just an endianness swap, isn't it?

It didn't work for me otherwise

PR #80, opened 2022-12-19
Add fix for drive reactivation when using custom BIOS mods

darcagn:
Fix allows KOS applications to work properly when using custom BIOSs like Dreamboot or Dreamshell loader BIOS.
Method well tested in DreamShell for years. Added code comments.

### DIFF
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -447,11 +447,21 @@ int cdrom_init() {
                      *bios = (uint32 *)0xa0000000;
 
     /* Reactivate drive: send the BIOS size and then read each
-       word across the bus so the controller can verify it. */
-    *react = 0x1fffff;
-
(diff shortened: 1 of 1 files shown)


QuzarDC:
Is the 0xe6ff value documented somewhere? If all official BIOS use 0xe3ff then it would make more sense to me to check if the first byte is *not* the official known value rather than rely on the magic value of the mod.

darcagn:
> Is the 0xe6ff value documented somewhere? If all official BIOS use 0xe3ff then it would make more sense to me to check if the first byte is _not_ the official known value rather than rely on the magic value of the mod.

0xe3ff is the value in all known official Dreamcast firmwares, but it cannot be assumed it is the only valid value.
In the case of 0xe6ff, this is borrowed from a House of the Dead 2 Naomi proto which includes a smaller bootstrap that passes the Holly check, and all custom BIOS images are using this bootstrap to bypass the majority of the Holly check. Therefore we should specifically check for that case, not the other way around.

DC-SWAT:
Now I can remove this code from KOS patching in DreamShell :)

PR #81, opened 2022-12-31
Support 32MB RAM modifications

darcagn:
Patches introduce:
- Kernel changes to modify stack for 32MB RAM modifications, along with related functions
- Example memory-testing program demonstrating use of above mentioned functions
- Patches for GCC 4.7.4 and 9.3.0 (including Apple Silicon patch) supporting stack changes
- Patches for GCC 4.7.4 and 9.3.0 adding builtin defines __KOS_GCC_PATCHED__, __KOS_GCC_PATCHLEVEL__, and __KOS_GCC_AMX3_32MB__. 

Thanks go to tsowell, without whom this wouldn't have been possible, as well as GyroVorbis and cepawiel for support.

### DIFF
--- a/examples/dreamcast/basic/Makefile
+++ b/examples/dreamcast/basic/Makefile
@@ -12,6 +12,7 @@ all:
 	$(KOS_MAKE) -C stacktrace
 	$(KOS_MAKE) -C mmu
 	$(KOS_MAKE) -C stackprotector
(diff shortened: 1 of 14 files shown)


darcagn:
Closing PR as we have figured out a way to handle the 32MB detection better than the MCR bit, allowing NAOMI to be included in the code path. Will resubmit better patch.

PR #85, opened 2023-01-07
Build Containers with Github Actions

cepawiel:
Hello Everyone!

Github is offering free hours on Github Action and I think we should be
taking more advantage of it. I've put together an actions file that builds
the toolchains and tests that KOS, Ports, and the examples compile. Optionally
it can also push containers to the Github Container Repository. Personally I believe we 
should be providing containers, with CI usage being a huge benefit.

I there there could also be a benefit to the new developer experience by using 
prebuilt containers. I've written a few samples here (https://github.com/cepawiel/KOS-Docker-Examples)
that show various samples of using the containers. One such easy to use example is with VSCode Dev 
Containers.

There's certainly some polish left to add, but wanted to get feedback before I do so. Below I've included some more information about how the containers and builds are put together.

Container Layouts:

kos-sh4-toolchain:
This container includes the  sh-elf and arm-eabi toolchains
in their normal location under /opt/toolchains/dc/

*target*-kos:
Container is built on the toolchain image but also includes
and compiles KOS at /opt/toolchains/dc/

Brief description of Github Actions Steps:
1. Setup Job
    First we check to see if the GITHUB_TOKEN is valid to determine if
    images can be pushed to the container repo. If not we can still build
    just cant pull down images.

    Then we load in the json files containing the toolchain and target configs.
    Then I put them into the GHA env var so they can be accessed in different jobs.

2. Build Toolchains
    For each toolchain defined in toolchains.json the docker build command is ran.
    The containers are only saved to the cache at this point. The reason this is
    separated from the following step is due to the fact that the toolchain is shared
    between the two targets and we can then share the cached toolchain.

3. Build KOS + Ports + Examples
    For each cached toolchain we build KOS + Ports for each target and optionally
    the examples if the targets.json is configured to do so. Some of the examples do build
    for NAOMI here so I was considering a way to specify a list of example directories to 
    attempt to build. If examples are built the artifacts are uploaded to the job. 

4. Publish Containers to Github Container Registry
    If GITHUB_TOKEN is set in the repo secrets and the ports/examples built then this 
    job is run. Docker is called to build the container at two stages which is it able to
    quickly pull from the build cache. It publishes two containers at the moment. A toolchain
    container which contains only the two toolchains and is intended to be used when wanting
    to use your own KOS. And a KOS container that contains an already compiled copy of KOS.
    I also have tested working one for ports, just felt it might belong in the ports repo.
    Any thoughts/feelings? If here we'll have to manually/cron check for ports updates. Same issue on
    the reverse if its in ports.


ljsebald:
I'm going to hold off on this for now... For a few reasons:
1. I'm not convinced that we will actually always stay within the free limits on GitHub actions. Yes, I'm sure most of the time we would... But with things set up to be done on every push (or any similar schedule), I would be very worried that there may be times when we would not.
2. The GitHub repository for KOS is still just a mirror of what is on Sourceforge, ultimately. I do not like tying us to any one provider more than necessary.

I'd be more willing to consider setting something like this up at a later date for tag events -- but certainly not on every push.

I am by no means saying that something like this is a bad idea (so please don't interpret it that way), but I think that with how long it takes to build a toolchain container and all each push, we could run into serious trouble if there was some point where we were doing a lot of development during a month. Plus, the ad-hoc nature of how development often happens with KOS (look at me in particular, with my constant commits/pushes directly to the master branch), doesn't really lend itself to this approach -- especially building the toolchain every time.

PR #88, opened 2023-01-25
Return -1 and set ENOMEM if sbrk fails

Kazade:
**This still needs testing**

This patch updates mm_sbrk to return -1 and set errno according to the spec. This is a change in behaviour as it means applications that run out of memory will no longer panic.

The benefits of not panicking are:

 - C++ can throw an appropriate exception
 - Languages like Lua can run a garbage collection sweep if malloc returns NULL

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,6 +43,7 @@ Joe Fenton: 2016
 Stefan Galowicz: 2016, 2017
 Luke Benstead: 2020, 2021
 Eric Fradella: 2023
+Falco Girgis: 2023
 
 Files with Specific licenses:
 -----------------------------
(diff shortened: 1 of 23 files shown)


gyrovorbis:
Assigning this one to me. I was asked to verify the behavior with a Lua emergency garbage collection sweep and/or a std::bad_alloc exception upon running out of RAM.

gyrovorbis:
Validated that this thing works like a charm and the behavior is as expected between both 16MB and 32MB modded DCs, with the latter able to fulfill requests to 2x the size.

I've taken the liberty of making my test bed into another example while I was at it, to remind people that they can now do this and recover from it. The demo was done in C++, just because the stdlib adds extra complexity above a malloc() failure, and I wanted to ensure the exception path and C++ stdlib stuff was working as expected.

ljsebald (approved):

PR #89, opened 2023-01-26
VMU driver support for date/time and buttons

gyrovorbis:
1. Added a higher-level convenience wrapper, vmu_beep(), around vmu_beep_raw(), for generating buzzer square waves
2. Added support for setting and getting the datetime structure from the VMU
3. Added support for polling VMU button states 
    - VMU buttons states are actually maintained and can be reported via the Maple protocol when the VMU is plugged in. There now exist 3rd party controllers, such as the RetroFighters StrikerDC gamepad, which expose these buttons and can be used for homebrew games
    - to avoid unnecessary Maple overhead for games that don't care, the buttons have to be enabled via vmu_set_buttons_enabled(), which dynamically adds the periodic polling callback to the driver
    - to further reduce unnecessary overhead, only the first unit in each controller is polled, since the second is assumed to be inaccessible


kernel/arch/dreamcast/include/dc/maple/vmu.h
```diff
@@ -149,6 +151,37 @@ int vmu_block_write(maple_device_t * dev, uint16 blocknum, uint8 *buffer);
 */
 void vmu_set_icon(const char *vmu_icon);
 
+typedef struct vmu_time {
+    uint16_t year;     // 0-????
+    uint8_t month;     // 1-12
+    uint8_t day;       // 1-32
```
QuzarDC, line 157: When would day go up to 32?

gyrovorbis: LOL WTF. Durrrr…. Good catch. ;)

PR #90, opened 2023-01-29
GCC12 Toolchain Support (with C++20) 

gyrovorbis:
GCC12 is the latest version of the GCC toolchain, which importantly adds support for the C++20 language standard, which includes "concepts" and a myriad of other useful features. Additionally, it seems to consistently produce smaller ELF files and .text segments plus increased performance on the pvrmark benchmark. It also has a fix for a known GCC9 compiler
crash.

EDIT: While we were at it, we also bumped Binutils + Newlib up to the latest version. The newest version of Newlib is only about 2 weeks old, too! That puts the DC toolchain on building with the latest and greatest everything.

- added config.mk.latest.sample as the config file template, using GCC version 12.2.0, Binutils 2.40.0, Newlib version 4.3.0
- added a patch for the GCC12 compiler
    * required a fix for libobjc configuration
- hunted down and backported a CRITICAL bugfix from GCC13/GCCRS which saved SH4 from being completely broken in GCC12 (darc + colton)
- added a tiny change to _types.h for time_t + daddr_t, as required by Newlib
- updated main README file in the dc-chain directory
- tested examples to make sure nothing new is broken
- tested C++20 "concepts" and "constraints" on DC
- tested new toolchain with GDB and source-level debugging in Qt Creator
- have consistently used this toolchain for all of libGimbal work and unit tests (800+) and have never run into any issues with it
- fixed duplicate symbol issues with pvrmark, quadmark, KGL benchmarks, etc using the ALREADY USED symbol name "start" for a variable. We use it in crt1 and it's declared extern. No idea why any other toolchain would've been okay with this.


gyrovorbis:
Just a note to anybody who has been following this thread from the various Discord links, mailing lists, and forum posts: I think this is pretty much ready to go. I will be talking to BlueCrab tomorrow about a couple small things then will take this out of draft.

Been posting a lot more detail about my testing and the ONE SINGLE COMPILER BUG I did manage to find in this thread: https://dcemulation.org/phpBB/viewtopic.php?p=1059783#p1059783

If you're pulling this down early to play with GCC12, please please let us know if you have any problems, we really want to keep track of issues that come up.

include/sys/_types.h
```diff
@@ -128,7 +132,9 @@ typedef unsigned short __nlink_t;
 typedef long        __suseconds_t;  /* microseconds (signed) */
 typedef unsigned long   __useconds_t;   /* microseconds (unsigned) */
 
-#if __NEWLIB__ > 3
+#if __NEWLIB__ >= 4
```
ljsebald, line 135: This one might be a bit of a problem where we have to detect on minor version too... After all, the declaration was for > 3, not >= 3. So, it would only be detecting the same thing as your >= 4.

PR #91, opened 2023-02-01
Added CMake Toolchain file + environment variable

gyrovorbis:
Many users have expressed an interest in being able to target the DC and use KOS with a CMake-based build system. They are typically either confused as to how to integrate it with our Make-based build system or are forced to reinvent the wheel.

I think it's a good idea to provide a top-level toolchain file for interfacing with our build system from a CMake project. That way we can keep it versioned, in-sync, ensure its flags are optimal, and can add it to the environment like the rest of the build system paths.


utils/cmake/dreamcast.toolchain.cmake
```diff
@@ -0,0 +1,106 @@
...
+
+##### Configure CMake System #####
+set(CMAKE_SYSTEM_NAME Generic)
+set(CMAKE_SYSTEM_VERSION 1)
+set(PLATFORM_DREAMCAST TRUE)
+
+##### Configure Cross-Compiler #####
+set(CMAKE_CROSSCOMPILING TRUE)
+set(CMAKE_C_COMPILER $ENV{KOS_CC_BASE}/bin/sh-elf-gcc)
+set(CMAKE_CXX_COMPILER $ENV{KOS_CC_BASE}/bin/sh-elf-g++)
```
losinggeneration, line 38: This applies to a lot most of these set operations, but is there a reason the more specific environment variables aren't being used here (`KOS_CC` & `KOS_CCPLUS` for these two)? Maybe I've just been out of the loop for too long, but are there people with projects not using a typical `environ.sh` and instead using a much more minimal set of environment variables (at quick glance: `KOS_BASE`, `KOS_CC_BASE`, & optionally `KOS_SUBARCH` & `KOS_CMAKE_TOOLCHAIN`)?

gyrovorbis: Oh, we have just been having a whole huge discussion about this exact thing in the SImulatnt discord server. I've been doing some testing by swapping to those two to see if we could unify them, which does have benefits in terms of maintenance and consistency, but a couple things to note here:

1) I'm trying to verify from at least 1 IDE (Qt Creator for me) and command-line. Qt Creator seems to get really angry with kos-cc/kos-c++ and claims its an invalid compiler that cannot generate a binary. It has something to do with the configuration step where it's calling it with a bunch of extra args. Gonna play with it more to figure it out.
2) This makes some opinionated decisions about optimization level, adding debug symbols, and turning on -WExtra, which others might not be cool with. HOWEVER, it does seem as though you can simply pass the opposite flags from CMake and have them properly override the defaults. I'm still trying to do a bunch of testing, because I'd hate to find out some default set here screws the project
3) General spookiness. None of the regular things in Cmake like CMAKE_C_FLAGS/CMAKE_CXX_FLAGS, etc etc will actually have values representative of what is getting passed to the compiler and may lead to some confusion/irritation. I'm no CMake purist, so I don't know that I care, but just putting that out there in case someone does.

Still looking into all of this right now.

losinggeneration: 1. I don't think kos-cc/kos-c++ are strictly required since they're just shell script wrappers around sh-elf-gcc and adding the flags from the environment. In the case of the lines 37 & 38, the values `$ENV{KOS_CC_BASE}/bin/sh-elf-gcc` & `$ENV{KOS_CC}` are the same value in a standard kos toolchain & `environ.sh`.
2. My suggestion would be to go with the default flags from the base toolchain/environ/makefiles and add some of these more opinionated optimizations as optional options that can be turned on based on the projects' specific needs. Not sure if this works with toolchain included files though. If it does, this would look something like
```cmake
option(EXTRA_DEBUG "Build with extra debug options" off)

if(EXTRA_DEBUG)
    set(EXTRA_FLAGS ${EXTRA_FLAGS} -DFRAME_POINTERS -fno-omit-frame-pointer)
endif(EXTRA_DEBUG)
```
3. I think what I did for CrabEmu's CMake build, way back, was similar to what I suggested in 2. I built up flags in an extra variable, but instead of setting `CMAKE_C_FLAGS` or `CMAKE_CXX_FLAGS` directly, I used `add_definitions`. A quick look at the documentation looks like `add_compile_options` might be better, and optionally `add_compile_definitions` for `-D` flags. I'll admit that it's been awhile since I've been deep into CMake, so I can't really say if that's the best strategy here. 
4. This should be pretty easy to test manually with a small c file, then compile with `-fno-builtins` normally, then again with `-fmsrra -fmsca -ffast-math`, and finally with `-fno-builtins -fmsrra -fmsca -ffast-math`, optionally maybe one more with `-fno-builtins` last, and see if the objects are significantly different based on the order of the flags. In theory, GCC overrides based on the rightmost flag (and I don't have time right this second to verify that in the docs.) If it would be better to just not trust that the compiler potentially respects flag order for overrides, something that could be potentially done would be using a `foreach` to build up a temporary variable with the options while checking for, and skipping, `-fno-builtins`. Similar to 2, defaulting to the basic kos behavior, and using `option(FAST_MATH ` to trigger this behavior to skip `-fno-builtins` and switch in the other flags.

gyrovorbis: Thank you for this, I'm about to have some free time available to get back to this PR, and you've given good feedback and direction for what to look into and test. I'll keep you posted!

utils/cmake/dreamcast.toolchain.cmake
```diff
@@ -0,0 +1,106 @@
...
+
+set(CMAKE_EXECUTABLE_SUFFIX ".elf")
```
cepawiel, line 40: I've been testing with this toolchain file and have come across that this doesn't apply properly. According to this (https://discourse.cmake.org/t/how-to-specify-executable-file-type-for-arm/4896/3) setting "CMAKE_EXECUTABLE_SUFFIX" get overwritten when set in the toolchain file. Seems like our options are:

1) Set "CMAKE_EXECUTABLE_SUFFIX_C", "CMAKE_EXECUTABLE_SUFFIX_CXX", and "CMAKE_EXECUTABLE_SUFFIX_ASM" variables to ".elf" which do not overwritten and does fix this, but seems to be an unrecommended (but common for bare-metal) configuration.

2) The other option is to set "CMAKE_SYSTEM_NAME" to "Generic-ELF" from "Generic" which is just Generic but with the elf suffix appended to executables. The the downside is that it was merged in for CMake 3.23 (https://gitlab.kitware.com/cmake/cmake/-/merge_requests/6630) which has only been out about a year.

I've tested both of these spit out executables with the correct ".elf" extension, just a question of if we want to do it the new correct way, or the older hack which will support more versions of CMake.

utils/cmake/dreamcast.toolchain.cmake
```diff
@@ -0,0 +1,106 @@
...
+set(CMAKE_ASM_FLAGS_RELEASE "")
+
+##### Configure Include Directories #####
+set(CMAKE_SYSTEM_INCLUDE_PATH "${CMAKE_SYSTEM_INCLUDE_PATH} $ENV{KOS_BASE}/include $ENV{KOS_BASE}/kernel/arch/dreamcast/include $ENV{KOS_BASE}/addons/include $ENV{KOS_BASE}/../kos-ports/include")
+
+INCLUDE_DIRECTORIES(
+    $ENV{KOS_BASE}/include
+    $ENV{KOS_BASE}/kernel/arch/dreamcast/include
+    $ENV{KOS_BASE}/addons/include
+    $ENV{KOS_BASE}/../kos-ports/include
```
cepawiel, line 85: This could reference KOS_PORTS env var since we use it in LINK_DIRECTORIES
`$ENV{KOS_PORTS}/include`

utils/cmake/dreamcast.toolchain.cmake
```diff
@@ -0,0 +1,106 @@
...
+# Frame pointers are enabled in debug builds as these are required for 
+# stack traces and GDB. They are disabled in release.
+# 
+# The original toolchain file was created by Kazade for the Simulant 
+# engine who has graciously allowed the rest of the scene to warez it. 
+#
+# Special thanks to Matt Godbolt of Compiler Explorer, for supporting
+# the SH toolchains, which allowed us to figure out the precise magical
+# incantations of compiler flags for fast math.  
+
```
cepawiel, line 29: We should add checks here to make sure the ENV vars we use are defined, otherwise you get errors from incorrect paths when we could display meaningful errors.

```
#### Check Required Environment Variables ####
if(NOT DEFINED ENV{KOS_BASE})
  message(FATAL_ERROR "Environment Variable KOS_BASE not set")
endif()

if(NOT DEFINED ENV{KOS_CC_BASE})
  message(FATAL_ERROR "Environment Variable KOS_CC_BASE not set")
endif()

if(NOT DEFINED ENV{KOS_SUBARCH})
  message(FATAL_ERROR "Environment Variable KOS_CC_BASE not set")
endif()

if(NOT DEFINED ENV{KOS_PORTS})
  message(FATAL_ERROR "Environment Variable KOS_CC_BASE not set")
endif()
```

utils/cmake/dreamcast.toolchain.cmake
```diff
@@ -0,0 +1,106 @@
...
+        -D_arch_sub_naomi
+    )
+else()
+    ADD_DEFINITIONS(-D_arch_sub_pristine)
+endif()
+
+##### Configure Build Flags #####
+add_compile_options(-ml -m4-single-only -ffunction-sections -fdata-sections)
+
+set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -DFRAME_POINTERS -mfsrra -mfsca -fno-omit-frame-pointer")
```
cepawiel, line 63: Can fix the repetition of these flags using generator expressions.

```
##### Configure Build Flags #####
add_compile_options(-ml -m4-single-only -ffunction-sections -fdata-sections)

set(ENABLE_DEBUG_FLAGS   $<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>)
set(ENABLE_RELEASE_FLAGS $<OR:$<CONFIG:Release>,$<CONFIG:MinSizeRel>>)

add_compile_options(
    "$<${ENABLE_DEBUG_FLAGS}:-DFRAME_POINTERS;-fno-omit-frame-pointer>"
    "$<${ENABLE_RELEASE_FLAGS}:-fomit-frame-pointer>"
    )

set(CMAKE_ASM_FLAGS "")
set(CMAKE_ASM_FLAGS_RELEASE "")
```

For handling the SH4 Math flags the best I could come up with was having an include-able helper file with a function to add the math flags to the compile. The Vita SDK includes a similar helper file https://github.com/vitasdk/samples/blob/master/hello_world/CMakeLists.txt#L21. 

The issue being that the `-mfsrra` and `-mfsca` flags are not available on the GCC4.7.4 (legacy) toolchain. We can't check if they're avaiable in the toolchain file, as CMake doesn't run any checks on the compiler until after the toolchain file finishes loading, so we don't know the GCC version or supported compiler flags inside the toolchain file. 

This helper file could be included alongside the toolchain file in KOS. See https://github.com/cepawiel/tutorials/blob/main/dreamcast.cmake#L117 for an example. I do believe I'm missing a few of the additional math flags required, just was a quick example based on the flags shown here. Also allows us to include other helper functions like creating a romdisk and a CDI (also in linked file).

utils/cmake/dreamcast.toolchain.cmake
```diff
@@ -0,0 +1,106 @@
...
+    $ENV{KOS_BASE}/include
+    $ENV{KOS_BASE}/kernel/arch/dreamcast/include
+    $ENV{KOS_BASE}/addons/include
+    $ENV{KOS_BASE}/../kos-ports/include
+)
+
+##### Configure Libraries #####
+set(CMAKE_SYSTEM_LIBRARY_PATH "${CMAKE_SYSTEM_LIBRARY_PATH} $ENV{KOS_BASE}/addons/lib/dreamcast $ENV{KOS_PORTS}/lib")
+
+if($ENV{KOS_SUBARCH} MATCHES naomi)
```
cepawiel, line 91: I think we should extract the shared platform flags and just leave the text start and linker file flags in the if. Something like:

```
if($ENV{KOS_SUBARCH} MATCHES naomi)
  add_link_options(-Wl,-Ttext=0x8c020000 -T$ENV{KOS_BASE}/utils/ldscripts/shlelf-naomi.xc)
else()
  add_link_options(-Wl,-Ttext=0x8c010000 -T$ENV{KOS_BASE}/utils/ldscripts/shlelf.xc)
endif()

add_link_options(-ml -m4-single-only -Wl,--gc-sections -nodefaultlibs)
```

utils/cmake/dreamcast.toolchain.cmake
```diff
@@ -0,0 +1,106 @@
...
+endif()
+
+LINK_DIRECTORIES(
+    $ENV{KOS_BASE}/addons/lib/dreamcast
+    $ENV{KOS_PORTS}/lib
+)
+
+add_link_options(-L$ENV{KOS_BASE}/lib/dreamcast -L$ENV{KOS_BASE}/addons/lib/dreamcast)
+LINK_LIBRARIES(-Wl,--start-group -lstdc++ -lkallisti -lc -lgcc -Wl,--end-group m)
+LINK_LIBRARIES(c gcc kallisti stdc++)
```
cepawiel, line 104: I don't think we actually need this last line. I can't find anything that fails to link when removing it and just seems to add the flags multiple times. The start and end group should be enough as far as I can tell.

utils/cmake/dreamcast.toolchain.cmake
gyrovorbis: WELL NO WONDER this behavior seemed to be either not working or not working consistently for me. Thank you! 

You bring up a good point about which CMake version minimum we're targeting here...

utils/cmake/dreamcast.toolchain.cmake
gyrovorbis: Agreed, good point.

utils/cmake/dreamcast.toolchain.cmake
gyrovorbis: Also agreed, that's cleaner and would let you manually set them.

gyrovorbis:
OKAY, SO. A few things have happened

1) There's just no way in hell we're going to be able to use the GNU KOS wrappers here. It seems both QtCreator and one other major IDE just freak out when they are used in place of an actual compiler binary.... So with that out of the way...

2) The way you're supposed to be setting the .ELF extension for our C and C++ executables is a newfangled thing to do "properly" and requires a newer version of CMake in the file. You can do it the "bad and shunned" way with older versions, but they explicitly tell you not to do that anymore... Does anyone have a problem with the minimal required version being 3.23? It's still like a year old, and all my stuff is newer, so I don't mind... Neither does Colton, but polling the audience here

3) The damn fast math flags can no longer be a part of the toolchain, because they don't work on our legacy 4.7 toolchains, and thanks to CMake being Cmake, it's impossible to determine compiler version within a toolchain file, due to file resolution/include ordering. This stuff had to get added to an external function.

4) @cepawiel Added a bunch of external functionality that doesn't belong in the toolchain file for feature parity with what we have going on with Make. I thought it was a good idea to include these, because I'd rather do this CMake stuff right the first time than half ass it and then wind up with everyone doing their own thing again. This repo: https://github.com/cepawiel/tutorials/blob/main/CMakeLists.txt features a bunch of tutorials/examples using more-or-less this current cmake configuration if someone wanted to look at it in an actual project.

Anyway, let me know what you guys think.

cepawiel:
I updated a few small changes that were breaking the environment variable checks as well as changed the spacing to match across the files.

Tested this with GLdc and outside of a `-DNDEBUG` flag (https://gitlab.com/simulant/GLdc/-/blob/master/toolchains/Dreamcast.cmake#L64) I had to move into the CMakeList.txt, it worked fine. Tested with a few examples and didn't find any issues.

ljsebald (approved):

PR #101, opened 2023-02-07
Improved cache management

DC-SWAT:
Perhaps it will be useful to someone and I will remove it from the DreamShell KOS patch.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/cache.h
+++ b/kernel/arch/dreamcast/include/arch/cache.h
@@ -1,7 +1,8 @@
 /* KallistiOS ##version##
 
    arch/dreamcast/include/cache.h
-   (c)2001 Dan Potter
+   Copyright (C) 2001 Dan Potter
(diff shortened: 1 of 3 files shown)


ianmicheal:
Do you have an example of using the improved cache functions for testing and use cases?

DC-SWAT:
> Do you have an example of using the improved cache functions for testing and use cases?

I did not do comparative testing, although probably should be.
In fact, I removed unnecessary operations in invalidate and flush a long time ago and they are works in DS and loaders without issues.
I did not find any documentation for these operations, and the authors of the Demul emulator told me that this is nonsense. For the rest, I added what was simply not enough for completeness here just for fun.

DC-SWAT:
I removed allocation stuff because I remembered that I had issues with them and they were not used anywhere, so I decided not to create new issues here and removed it. Everything else is tested a years and not have any issues.

PR #104, opened 2023-02-10
Fixed and improved reading for SCIF SPI and use it for SD.

DC-SWAT:
I added optimized reading function from DreamShell and fix scif_spi_read_byte because it's not working for me in general.
Also added example for speed testing.
My SanDisk Extreme microSD 16GB without changes give me 630 KB/s, with changes 710 KB/s.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/scif.h
+++ b/kernel/arch/dreamcast/include/dc/scif.h
@@ -3,6 +3,7 @@
    dc/scif.h
    Copyright (C) 2000,2001,2004 Dan Potter
    Copyright (C) 2012 Lawrence Sebald
+   Copyright (C) 2023 Ruslan Rostovtsev (SWAT)
 
 */
 
@@ -13,6 +14,7 @@
 
     \author Dan Potter
     \author Lawrence Sebald
+    \author Ruslan Rostovtsev (SWAT)
 */
 
 #ifndef __DC_SCIF_H
@@ -180,11 +182,17 @@ void scif_spi_write_byte(uint8 b);
 /** \brief  Read a byte from the SPI device.
 
     This function reads a byte from the SPI device, one bit at a time. The
-    timing follows that of the scif_spi_rw_byte() function.
+    timing better a bit that of the scif_spi_rw_byte() function.
 
     \return                 The byte returned from the card.
 */
 uint8 scif_spi_read_byte(void);
+
+/** \brief  Read a data from the SPI device.
+
+    This function reads data from the SPI device with optimizations.
+*/
+void scif_spi_read_data(uint8 *buffer, size_t len);
 
 __END_DECLS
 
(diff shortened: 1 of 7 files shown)


kernel/arch/dreamcast/include/dc/scif.h
```diff
@@ -180,12 +182,18 @@ void scif_spi_write_byte(uint8 b);
 /** \brief  Read a byte from the SPI device.
 
     This function reads a byte from the SPI device, one bit at a time. The
-    timing follows that of the scif_spi_rw_byte() function.
+    timing better a bit that of the scif_spi_rw_byte() function.
```
andressbarajas, line 185: Did you mean to say:

timing is a bit better than that of the scif_spi_rw_byte() function.

andressbarajas:
Tried it with my SanDisk 32 GB SDHC (non micro) and there was no difference in speeds.  Gonna order the sd card mentioned and play with other SD Card DC serial adaptors.

DC-SWAT:
> Tried it with my SanDisk 32 GB SDHC (non micro) and there was no difference in speeds. Gonna order the sd card mentioned and play with other SD Card DC serial adaptors.
> 
> Which one of these below is the SD card you are using?
> 
> https://www.amazon.com/SanDisk-Extreme-Mobile-Gaming-microSD/dp/B089M5KV4Y/ https://www.amazon.com/Sandisk-Extreme-MicroSDHC-UHS-I-SDSDQXL-032G-A46A/dp/B00G5R75AU/

I use SanDisk Extreme 16GB and Samsung 64GB. Both got improvements.
Check it with SD speedtest example.

QuzarDC:
There should be improvement due, at least, to the removal of the SD_WAITs.

It might be useful, for producing more reliable averages, to do the same read multiple times and average the result. At the same time, the results of each read could be compared with each other to confirm that they produce the same result every time.

andressbarajas:
^ Yes, great idea to improve the example test.

Im just sad the speed of reading my 32GB pales in comparison to DC-SWATs..:

SD card read took 2165 ms (242.165 KB/sec)

DC-SWAT:
> There should be improvement due, at least, to the removal of the SD_WAITs.
> 
> It might be useful, for producing more reliable averages, to do the same read multiple times and average the result. At the same time, the results of each read could be compared with each other to confirm that they produce the same result every time.

Removed useless SD_WAITs, used 4-byte writing to RAM, unroll loop a bit, improved pins managment and opcode caching.
All this little by little improved the reading as a whole. Test results are always almost identical for me.

DC-SWAT:
> ^ Yes, great idea to improve the example test.
> 
> Im just sad the speed of reading my 32GB pales in comparison to DC-SWATs..:
> 
> SD card read took 2165 ms (242.165 KB/sec)

You use SD card with very bad response, try another.

PR #106, opened 2023-02-12
Change default SCIF baud rate to 115200 and add support for external clock

DC-SWAT:
It seems to me that the time has passed when one should be afraid to use 115200 :) I've been using this value just fine for many years now with various adapters.
In fact, in some cases even works 500000. But this is not good as default value of course. I use it for dc-load-serial if can't use dc-load-ip by some reason.

### DIFF
--- a/kernel/arch/dreamcast/hardware/scif.c
+++ b/kernel/arch/dreamcast/hardware/scif.c
@@ -193,23 +193,21 @@ int scif_init_fake() {
     return 0;
 }
 
-/* Initialize the SCIF port; baud_rate must be at least 9600 and
-   no more than 57600. 115200 does NOT work for most PCs. */
+/* Initialize the SCIF port; */
 /* recv trigger to 1 byte */
 int scif_init() {
     int i;
     unsigned char scbrr2;
     unsigned short scsmr2;
 
-    /* int fifo = 1; */
-
     /*  If dcload-serial is active, then do nothing here, or we'll
         screw that up. */
     if(dcload_type == DCLOAD_TYPE_SER)
         return 0;
 
-    /* Disable interrupts, transmit/receive, and use internal clock */
-    SCSCR2 = 0;
+    /* Disable interrupts, transmit/receive,
+       and use internal or external clock */
+    SCSCR2 = serial_baud ? 0x0 : 0x02;
 
     /* Enter reset mode */
     SCFCR2 = 0x06;
@@ -226,7 +224,7 @@ int scif_init() {
         scsmr2 = 2;
         scbrr2 = (50000000 / (512 * serial_baud)) - 1;
     }
-    else {
+    else if (serial_baud) {
         scsmr2 = 3;
         scbrr2 = (50000000 / (2048 * serial_baud)) - 1;
     }
@@ -255,7 +253,8 @@ int scif_init() {
     SCLSR2 = 0;
 
     /* Enable transmit/receive */
-    SCSCR2 = 0x30;
+    /* If external clock is used set CKE1 bit */
+    SCSCR2 = serial_baud ? 0x30 : 0x32;
 
     /* Wait a bit for it to stabilize */
     for(i = 0; i < 800000; i++)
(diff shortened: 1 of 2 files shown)


losinggeneration:
Did you mean for this to be a changed default for [dcload-serial](https://github.com/KallistiOS/dcload-serial/blob/master/Makefile.cfg#L68)?

darcagn:
> Did you mean for this to be a changed default for [dcload-serial](https://github.com/KallistiOS/dcload-serial/blob/master/Makefile.cfg#L68)?

He's been working on DreamShell's isoldr, which logs to serial without dcload-serial, so I'm pretty sure this is what he's intended.

DC-SWAT:
> Did you mean for this to be a changed default for [dcload-serial](https://github.com/KallistiOS/dcload-serial/blob/master/Makefile.cfg#L68)?

I use serial port for logging without dcload-serial.
KOS debug output put to serial by default.

losinggeneration:
That's fair. I'm going to be annoying and ask if there are any downsides here?

The only thing I could think it could potentially affect would be hardware projects using the serial port and expecting 57600.

DC-SWAT:
> That's fair. I'm going to be annoying and ask if there are any downsides here?
> 
> The only thing I could think it could potentially affect would be hardware projects using the serial port and expecting 57600.

I do not know of such projects that could break because of this change.
This value here was determined only by the fact that in those days, 20 years ago, not all PC COM ports supported this baud rate normally :)

losinggeneration:
> > That's fair. I'm going to be annoying and ask if there are any downsides here?
> > The only thing I could think it could potentially affect would be hardware projects using the serial port and expecting 57600.
> 
> I do not know of such projects that could break because of this change. And I don't think they exist. This value here was determined only by the fact that in those days, 20 years ago, not all PC COM ports supported this baud rate normally :)

I was originally thinking the midi adapter that I saw someone made recently based on the official one, but I overlooked that midi is at 31250 anyway and would need to set it manually. 

Anyway, might be worth updating the comment for [scif_init](https://github.com/KallistiOS/KallistiOS/blob/525cbdac17317acfc2e4937e2efd7b99c85e73a3/kernel/arch/dreamcast/hardware/scif.c#L196-L197) to reflect things today rather than 20 years ago :)

DC-SWAT:
> > > That's fair. I'm going to be annoying and ask if there are any downsides here?
> > > The only thing I could think it could potentially affect would be hardware projects using the serial port and expecting 57600.
> > 
> > 
> > I do not know of such projects that could break because of this change. And I don't think they exist. This value here was determined only by the fact that in those days, 20 years ago, not all PC COM ports supported this baud rate normally :)
> 
> I was originally thinking the midi adapter that I saw someone made recently based on the official one, but I overlooked that midi is at 31250 anyway and would need to set it manually.
> 
> Anyway, might be worth updating the comment for [scif_init](https://github.com/KallistiOS/KallistiOS/blob/525cbdac17317acfc2e4937e2efd7b99c85e73a3/kernel/arch/dreamcast/hardware/scif.c#L196-L197) to reflect things today rather than 20 years ago :)

I fixed it and add support for external clock.

kernel/arch/dreamcast/hardware/scif.c
```diff
@@ -193,23 +193,20 @@ int scif_init_fake() {
     return 0;
 }
 
-/* Initialize the SCIF port; baud_rate must be at least 9600 and
-   no more than 57600. 115200 does NOT work for most PCs. */
+/* Initialize the SCIF port; */
 /* recv trigger to 1 byte */
 int scif_init() {
-    int i;
```
ljsebald, line 200: Why did this variable get removed? There's no reason to change this to defining the variable `i` in both of the for loops below, and it does not match with the style of the rest of the code in KOS.

DC-SWAT: Sorry for that, I'll get back it.
I played with the values of the cycles and apparently played too much ))

PR #108, opened 2023-02-18
Updated GCC dependencies for testing config.mk

gyrovorbis:
- As requested by Sizious for DreamSDK

### DIFF
--- a/utils/dc-chain/config.mk.testing.sample
+++ b/utils/dc-chain/config.mk.testing.sample
@@ -67,10 +67,10 @@ sh_mpc_tarball_type=gz
 sh_isl_tarball_type=bz2
 
 # GCC dependencies for ARM
-arm_gmp_ver=6.2.1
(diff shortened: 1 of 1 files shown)


utils/dc-chain/config.mk.testing.sample
```diff
@@ -67,10 +67,10 @@ sh_mpc_tarball_type=gz
 sh_isl_tarball_type=bz2
 
 # GCC dependencies for ARM
-arm_gmp_ver=6.1.0
-arm_mpfr_ver=3.1.4
-arm_mpc_ver=1.0.3
-arm_isl_ver=0.18
+arm_gmp_ver=6.2.1
```
sizious, line 70: To me, you don’t have to do that for the ARM GCC because it’s still GCC 8

gyrovorbis: Good point! Reverted!

PR #109, opened 2023-02-20
Implement a new FS driver for /dev/random and /dev/urandom

Kazade:
This helps with portability of libraries which expect it to be there.


kernel/libc/newlib/newlib_getentropy.c
```diff
@@ -0,0 +1,35 @@
...
+
+        for(int j = 0; j < block_size; ++j) {
+            b ^= *--src;
+        }
+
+        *dst = b;
+        ++dst;
+    }
+    return 0;
+}
```
ljsebald, line 35:
```suggestion
int getentropy(void *ptr, size_t len) {
    const int block_size = 128;
    struct timeval tv;
    uint8_t *src = ((uint8_t *)_arch_mem_top);
    uint8_t *dst = ptr;
    size_t i;
    int j;
    uint8_t b;

    gettimeofday(&tv, NULL);

    /* We read backwards from the end of available memory
       and XOR in blocks into the key array, while XORing with the current time.
       If anyone has a better idea for generating entropy then send a patch :) */
    for(i = 0; i < len; ++i) {
        b = tv.tv_usec & 0xff;

        for(j = 0; j < block_size; ++j) {
            b ^= *--src;
        }

        *dst = b;
        ++dst;
    }

    return 0;
}
```

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -116,6 +116,7 @@ int  __attribute__((weak)) arch_auto_init() {
     fs_pty_init();          /* Pty */
     fs_ramdisk_init();      /* Ramdisk */
     fs_romdisk_init();      /* Romdisk */
+    fs_random_init();  /* /dev/urandom */
```
ljsebald, line 119:
```suggestion
    fs_romdisk_init();      /* Romdisk */
    fs_random_init();       /* /dev/urandom */
```

ljsebald:
I also think it might be a better idea to perhaps make this a bit more generic. Rather than having two separate filesystems for what are essentially individual files, perhaps it's better to have a `/dev` filesystem that just happens to have those two files in it. That way, if we are to expand this in the future, we won't be just adding a bunch of extra filesystem handlers for each file we want to add in `/dev`.

ljsebald:
Not sure how far along/close to ready this is... But I figured I'd make one comment now... Is `getentropy` ever called/used by Newlib proper? If not, then it probably belongs in `kernel/libc/koslib`, not in the `kernel/libc/newlib` directory.

gyrovorbis:
I just happened to be testing some crazy newfangled C++20 stuff, and it looks like we have a WHOLE bunch of newfangled randomness utilities within `<random>`: https://cplusplus.com/reference/random/

So I gave it a try with this little thing:
```
#include <random>
int main(int argc, char* argv[]) { 
    auto val = std::mt19937(std::random_device()())
    return 0;
}
```
...and guess what Newlib is complaining about...
```
opt/toolchains/dc/sh-elf/lib/gcc/sh-elf/12.2.0/../../../../sh-elf/bin/ld: /opt/toolchains/dc/sh-elf/lib/gcc/sh-elf/12.2.0/../../../../sh-elf/lib/libstdc++.a(random.o): in function `__ZNSt12_GLOBAL__N_117__libc_getentropyEPv':
(.text._ZNSt12_GLOBAL__N_117__libc_getentropyEPv+0x24): undefined reference to `_getentropy'
/opt/toolchains/dc/sh-elf/lib/gcc/sh-elf/12.2.0/../../../../sh-elf/bin/ld: /opt/toolchains/dc/sh-elf/lib/gcc/sh-elf/12.2.0/../../../../sh-elf/lib/libstdc++.a(random.o): in function `__ZNSt13random_device7_M_initERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE':
(.text._ZNSt13random_device7_M_initERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEE+0x84): undefined reference to `_getentropy'
/opt/toolchains/dc/sh-elf/lib/gcc/sh-elf/12.2.0/../../../../sh-elf/bin/ld: /opt/toolchains/dc/sh-elf/lib/gcc/sh-elf/12.2.0/../../../../sh-elf/lib/libc.a(libc_a-arc4random.o): in function `_getentropy_fail':
/opt/toolchains/dc/kos/utils/dc-chain/build-newlib-sh-elf-4.3.0.20230120/sh-elf/newlib/../../../newlib-4.3.0.20230120/newlib/libc/stdlib/arc4random.h:71: undefined reference to `_getentropy'
```
It looks as though his `_getentropy()` is going to be used by newlib and will be even more useful than expected!

Kazade:
> Not sure how far along/close to ready this is... But I figured I'd make one comment now... Is `getentropy` ever called/used by Newlib proper? If not, then it probably belongs in `kernel/libc/koslib`, not in the `kernel/libc/newlib` directory.

Yeah it's used by newlib. I think this PR is ready to go but let me know if you see anything else.

PR #114, opened 2023-02-27
Fix out of memory example compiling for gcc 4.7.4

andressbarajas:
This example wont compile for GCC 4.7.4 because C++11 is experimental and we need to turn it on with -std=c++11.


gyrovorbis:
You're right. I didn't realize both <cstdint> and vector<T>::shrink_to_fit() were C++11 and beyond only. Good call.

gyrovorbis (approved):
This is the correct change to make for every toolchain we support. Should be good to go.

PR #120, opened 2023-03-02
Flush stdout at Kernel Shutdown

gyrovorbis:
Since the GCC12 "testing" toolchain was introduced, we've been noticing that in many scenarios, such as the "hello" example, no output is ever flushed to the terminal without an explicit call to `fflush(stdout)`. 

- Standard C doesn't dictate how often the stdout buffer gets flushed
- As far as I can see, KOS is never explicitly forcing stdout to flush before exiting the program
- I believe we should be forcing a flush upon kernel exit, so that printf() and friends' output is at least never lost upon normal program execution

### DIFF
--- a/kernel/arch/dreamcast/kernel/init.c
+++ b/kernel/arch/dreamcast/kernel/init.c
@@ -175,6 +175,8 @@ int  __attribute__((weak)) arch_auto_init() {
 }
 
 void  __attribute__((weak)) arch_auto_shutdown() {
+    fflush(stdout);
+
 #ifndef _arch_sub_naomi
     fs_dclsocket_shutdown();
     net_shutdown();


QuzarDC:
The only minor improvement I could think would be if the shutdowns were ordered in a way that others that may trigger prints happen first (fs stuff, peripherals?), and the ones that facilitate the output happen last (net, dcload?). Since fs_vmu_shutdown can throw a shutdown error, should this try to happen after that?

gyrovorbis:
> The only minor improvement I could think would be if the shutdowns were ordered in a way that others that may trigger prints happen first (fs stuff, peripherals?), and the ones that facilitate the output happen last (net, dcload?). Since fs_vmu_shutdown can throw a shutdown error, should this try to happen after that?

OMG it's Quzar!!! What's up, man!? So good to hear from you. Lots of epic stuff happening lately.

You know what, I think you might be right. In my head, I was thinking we wanted to fflush() before everything shut down, but yeah,  guess we should do it at the last minute before that path is uninitialized to capture anything else in the shutdown process. Good point.

PR #122, opened 2023-03-02
Properly call newlib's exit to cleanup

cepawiel:
Routes arch_exit() calls and main() returns though newlib's exit() which should properly flush stdout as well as call functions registered with atexit() which include C++ static deconstructors. 

EDITED BY FALCO:

Basically this fix is to address a pretty bad issue that was discovered upon realizing destructors for static objects in C++ were never getting called. We were essentially calling the incorrect entry point into newlib when exiting KOS, causing it to never call its own internal chain of functions registered to `atexit()` internally. At VERY LEAST this chain included C++ static destructors and a call for flushing stdout upon shutdown; however, @cepawiel has reported that there were as many as like 5-7 entries into that list, so who knows what else was not being properly shut down.

The funny thing is that an application calling `exit(code)` directly from the standard C library for early termination would've resulted in the correct shutdown sequence being called, and proper uninitialization. 

While we were in here, we also wanted to clean up some of the K&R-style variable argument, blank argument list signatures like `void functionThatTakesNoArgs()` to do the "right thing" and be type-safe C as `void functionThatTakesNoArgs(void)`...

Note that in doing "the right thing," `arch_exit(void)` now EXPLICITLY takes no arguments. Yes, we were concerned about breaking something (that shouldn't really have built to begin with).The one use-case we always see around the scene is it being registered as a callback for all button presses, with pretty much this exact incantation:
```
    // Guard against an untoward exit during testing.
    cont_btn_callback(0, CONT_START | CONT_A | CONT_B | CONT_X | CONT_Y,
        (void (*)(unsigned char, long  unsigned int))arch_exit);
```        
Luckily we seem to always be doing that explicit cast, so it doesn't matter what the signature for the function is anyway. Also note that C++ sees an empty argument list as equivalent to `(void)` as well... so we feel pretty confident nothing should really be impacted by cleaning this up. (other than the single example that was fixed, which curiously lacked the cast)

Notice that `arch_exit()` is really just a wrapper around `exit(0)` now. `exit(...)` MUST do the right thing in standard C anyway, so this should really be the favored way to be exiting from now on... portable, standard, control over the exit code, not exposing KOS's internals. lol.


gyrovorbis:
@QuzarDC Check this one out. I converted mine to draft, because we found another bug (C++ static destructors not being called) that led us down some deep rabbit hole of ultimately finding out we were never properly calling Newlib's top-level exit function, to which it actually had registered its own fflush() call! lol.

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -246,9 +248,8 @@ void arch_set_exit_path(int path) {
 }
 
 /* Does the actual shutdown stuff for a proper shutdown */
-void arch_shutdown() {
+void arch_shutdown() {
```
gyrovorbis, line 251: should be `void arch_shutdown(void)`, cuz we're in C.

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -280,7 +281,17 @@ void arch_shutdown() {
 }
 
 /* Generic kernel exit point (configurable) */
-void arch_exit() {
+__noreturn void arch_exit() {
+    exit(0);
```
gyrovorbis, line 285: Maybe make a note that we're explicitly returning 0 here, which signifies a successful shutdown, because once return values matter, this is not the way you exit with an error code. This call should really just be used for graceful exit (or never used directly again).

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -292,29 +303,22 @@ void arch_exit() {
...
             break;
     }
 }
 
 /* Called to shut down the system and return to the debug handler (if any) */
-void arch_return() {
-    /* Shut down */
-    arch_shutdown();
-
+__noreturn void arch_return() {
```
gyrovorbis, line 312: Any way that while you're here, you can just add (void) to all of these function prototypes? I'd like to start getting that stuff fixed to be like... proper C. lol.

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -280,7 +281,17 @@ void arch_shutdown() {
...
 
 /* Generic kernel exit point (configurable) */
-void arch_exit() {
+__noreturn void arch_exit() {
+    exit(0);
+}
+
+/* Return point from newlib's _exit() */
+__noreturn void post_newlib_exit(int ret_code) {
+    dbglog(DBG_INFO, "arch: exit value %d\n", ret_code);
```
gyrovorbis, line 290: I like that you're explicitly printing this, but maybe it should be like "arch: exit return code: %d", to make it more clear where the value is coming from, what it represents, and how you can change it.

kernel/libc/newlib/newlib_exit.c
```diff
@@ -7,6 +7,8 @@
 
 #include <arch/arch.h>
 
+extern void post_newlib_exit(int ret_code) __noreturn;
+
 void _exit(int code) {
```
gyrovorbis, line 12: I know you don't have to do it, but can we declare this extern? I'd like to start making things "extern" that are coming from newlib or are getting called from newlib.

gyrovorbis (commented):
Looking good! I can't leave these comments in-line in the code, but couple other things:

- line 25 should disappear
- lines 33-35 should all have explicit (void) arguments and should be declared "extern." No, they don't have to, but it's less confusing that they're external to KOS. I'd like to get into that practice when it comes to KOS/newlib and KOS/libc functions.

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -22,17 +22,15 @@
...
 #if __GNUC__ == 4
 #define _init init
 #define _fini fini
 #endif
 
-void _init(void);
-void _fini(void);
-void __verify_newlib_patch();
+extern void _init(void);
+extern void _fini(void);
```
gyrovorbis, line 32: BlueCrab, we thought ti would be a good idea to mark everything that's pulled in externally from newlib as explicitly "extern," to make it clearer the symbols are not part of KOS.

gyrovorbis (dismissed):
I approve!

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -225,16 +223,18 @@ int arch_main() {
...
 
     /* Call the user's main function */
     rv = main(0, NULL);
 
     /* Call kernel exit */
-    arch_exit();
+    exit(rv);
 
+    /* should never actually return here */
     return rv;
```
ljsebald, line 238: Probably should get rid of the return with a value, since the function returns void now. Even though it shouldn't ever get there...
```suggestion
    /* should never actually return here */
```

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -246,9 +246,8 @@ void arch_set_exit_path(int path) {
 }
 
 /* Does the actual shutdown stuff for a proper shutdown */
-void arch_shutdown() {
+void arch_shutdown(void) {
```
ljsebald, line 249:
```suggestion
void arch_shutdown(void) {
```

kernel/arch/dreamcast/kernel/init.c
gyrovorbis: Good catch, btw,

andressbarajas:
Whats left to do in this PR?

cepawiel:
I can fix the merge conflicts in the morning, but otherwise we were just trying to think of a better name for the `arch_newlib_exit()` function.

andressbarajas:
How about: "arch_exit_handler" since it handles the process of exiting the program based on the specified exit path?

cepawiel:
> How about: "arch_exit_handler" since it handles the process of exiting the program based on the specified exit path?

I think that's better than what it was! Renamed it while I was fixing the merge conflicts

gyrovorbis:
@ljsebald What do you think about the new name? `arch_exit_handler()`?

ljsebald (approved):

PR #124, opened 2023-03-03
Parallax RasterMelt only works with png

andressbarajas:
Thanks to Ian Michael:  https://discord.com/channels/488332893324836864/661640634095894549/663620005794938880

I was able to figure out why this is broken which makes me think libParallax image loading is broken for jpegs...


ljsebald:
Going to close this for now... Since this isn't really a fix -- there's got to be something wrong with libjpeg, so we should fix that instead of working around it this way.

PR #126, opened 2023-03-10
Fix dc-chain target dependencies

cepawiel:
Adds Pre-Requisites for the previous build step as well as the associated patches

Binutils now depends on Binutils patches
GCC Stage 1 now depends on Binutils and GCC Patches
Newlib now depends on GCC Stage 1 and Newlib Patches
GCC Stage 2 now depends on Newlib and GCC Patches 

This now allows make to be called with more than one job without failing and will actually build the SH4 and ARM toolchains concurrently.


cepawiel:
Closing due to fix being merged in #233

Issue #118, opened 2023-03-01
C++11 Concurrency Issues with GCC12 (and possibly GCC9)


gyrovorbis:
It looks as though our Newlib configuration for modern C++ threads, mutexes, and general concurrency is somehow incomplete. Given the following trivial thread usage:

```
#include <stdio.h>
#include <thread>

KOS_INIT_FLAGS(INIT_MALLOCSTATS);


void thd() {
    printf("asdf\n");
}

int main(int argc, char **argv) {
    std::thread thd1(thd);

    return 0;
}
```
You will be met with the following compiler error:
```
out_of_memory.cc: In function ‘int main(int, char**)’:
out_of_memory.cc:12:25: error: no matching function for call to ‘std::thread::thread(void (&)())’
   12 |     std::thread thd1(thd);
      |                         ^
In file included from /opt/toolchains/dc/sh-elf/sh-elf/include/c++/12.2.0/thread:43,
                 from out_of_memory.cc:2:
/opt/toolchains/dc/sh-elf/sh-elf/include/c++/12.2.0/bits/std_thread.h:156:5: note: candidate: ‘std::thread::thread(std::thread&&)’
  156 |     thread(thread&& __t) noexcept
      |     ^~~~~~
/opt/toolchains/dc/sh-elf/sh-elf/include/c++/12.2.0/bits/std_thread.h:156:21: note:   no known conversion for argument 1 from ‘void()’ to ‘std::thread&&’
  156 |     thread(thread&& __t) noexcept
      |            ~~~~~~~~~^~~
/opt/toolchains/dc/sh-elf/sh-elf/include/c++/12.2.0/bits/std_thread.h:120:5: note: candidate: ‘std::thread::thread()’
  120 |     thread() noexcept = default;
      |     ^~~~~~
/opt/toolchains/dc/sh-elf/sh-elf/include/c++/12.2.0/bits/std_thread.h:120:5: note:   candidate expects 0 arguments, 1 provided
```
So basically there is no available constructor template for std::thread, even though the class itself is there. I'm almost positive this is some Newlib configuration issue with our KOS threading model not defining and mapping everything the __GTHREAD model it was based on was doing. This is probably a relatively straightforward thing which may just be a few #defines and/or mapping a few things to kthreads. 

I'm creating this issue, because I have a bunch on my plate with ElysianVMU stuff right now, and want to finish up the TLS stuff first, but this could be some low-hanging toolchain fruit that would yield great gains for C++ users.

PR #128, opened 2023-03-11
Seperate Pass 1 and Pass 2 GCC folders

cepawiel:
Using the same build folder for both was causing an issue in Pass 2 where our gthr-kos.h file was failing to replace the gthr-default.h file. This was causing issues with libstdc++ feature detection failing to figure out our threading support and thus things like std::thread and std::mutex wouldn't work properly on the GCC9 & GCC12 toolchains (and caused a failure to build libstdc++ on GCC13).

Tested on GCC4.7, GCC9.2, GCC12.2, and the upcoming GCC13 release which all now build libstdc++ supporting our threads

Edit: This fixes issue #118

### DIFF
--- a/utils/dc-chain/scripts/clean.mk
+++ b/utils/dc-chain/scripts/clean.mk
@@ -21,8 +21,9 @@ clean_patches_stamp:
 clean: clean_patches_stamp
 	-rm -rf build-newlib-$(sh_target)-$(newlib_ver)
 	-rm -rf build-newlib-$(arm_target)-$(newlib_ver)
-	-rm -rf build-gcc-$(sh_target)-$(sh_gcc_ver)
-	-rm -rf build-gcc-$(arm_target)-$(arm_gcc_ver)
+	-rm -rf build-gcc-$(sh_target)-$(sh_gcc_ver)-pass1
+	-rm -rf build-gcc-$(sh_target)-$(sh_gcc_ver)-pass2
+	-rm -rf build-gcc-$(arm_target)-$(arm_gcc_ver)-pass1
 	-rm -rf build-binutils-$(sh_target)-$(sh_binutils_ver)
 	-rm -rf build-binutils-$(arm_target)-$(arm_binutils_ver)
 	-rm -rf build-$(gdb_name)
--- a/utils/dc-chain/scripts/clean.mk
+++ b/utils/dc-chain/scripts/clean.mk
@@ -23,7 +23,7 @@ clean: clean_patches_stamp
 	-rm -rf build-newlib-$(arm_target)-$(newlib_ver)
 	-rm -rf build-gcc-$(sh_target)-$(sh_gcc_ver)-pass1
 	-rm -rf build-gcc-$(sh_target)-$(sh_gcc_ver)-pass2
-	-rm -rf build-gcc-$(arm_target)-$(arm_gcc_ver)-pass1
+	-rm -rf build-gcc-$(arm_target)-$(arm_gcc_ver)
 	-rm -rf build-binutils-$(sh_target)-$(sh_binutils_ver)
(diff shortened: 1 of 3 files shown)


utils/dc-chain/scripts/gcc-pass1.mk
```diff
@@ -5,9 +5,9 @@
 # Initially adapted from Stalin's build script version 0.3.
 #
 
-$(build_gcc_pass1) $(build_gcc_pass2): build = build-gcc-$(target)-$(gcc_ver)
+$(build_gcc_pass1): build = build-gcc-$(target)-$(gcc_ver)-pass1
 $(build_gcc_pass1) $(build_gcc_pass2): src_dir = gcc-$(gcc_ver)
-$(build_gcc_pass1): log = $(logdir)/$(build)-pass1.log
+$(build_gcc_pass1) $(build_gcc_pass2): log = $(logdir)/$(build).log
```
sizious, line 10: If I understand well, you want to have both pass1 and pass2 in the same log file?
Interesting approach but I think it could be interesting to have a clear separation in the log file to distinguish pass1 and pass2. What do you think?

utils/dc-chain/scripts/clean.mk
```diff
@@ -21,8 +21,9 @@ clean_patches_stamp:
 clean: clean_patches_stamp
 	-rm -rf build-newlib-$(sh_target)-$(newlib_ver)
 	-rm -rf build-newlib-$(arm_target)-$(newlib_ver)
-	-rm -rf build-gcc-$(sh_target)-$(sh_gcc_ver)
-	-rm -rf build-gcc-$(arm_target)-$(arm_gcc_ver)
+	-rm -rf build-gcc-$(sh_target)-$(sh_gcc_ver)-pass1
+	-rm -rf build-gcc-$(sh_target)-$(sh_gcc_ver)-pass2
+	-rm -rf build-gcc-$(arm_target)-$(arm_gcc_ver)-pass1
```
darcagn, line 26: Should the arm build have -pass1 on it if there's only one pass to begin with?

sizious: Today it's the case indeed but I don't think it's necessary

utils/dc-chain/scripts/gcc-pass1.mk
cepawiel: This actually creates 2 logs still, because the build variable name is being changed to include the pass number. If we keep the log variable as-is then the logs are named `build-gcc-sh-elf-9.3.0-pass1-pass1.log` and `build-gcc-sh-elf-9.3.0-pass2-pass2.log`.

sizious: Thanks for the explanation, maybe there is something easy to do for avoiding pass keyword repeating?

utils/dc-chain/scripts/clean.mk
sizious: Yes indeed I think it would be OK like that 👍🏻

utils/dc-chain/scripts/gcc-pass1.mk
cepawiel: Sorry I believe I worded my reply a bit oddly. With the proposed changes the following logs files for GCC would be output:
```
build-gcc-sh-elf-9.3.0-pass1.log
build-gcc-sh-elf-9.3.0-pass2.log
```
If line 10 above was not modified then it would produce the double pass keyword output files mentioned above. I simply was trying to say that my intention with the line 10 change was to fix the double keyword in the file name and not to combine the log files. I think having them separate makes the most sense as well :slightly_smiling_face: 

Please do correct me if that was not the pass keyword repetition you're talking about

ljsebald (approved):

PR #134, opened 2023-03-12
Link-Time Optimizations (LTO)

pcercuei:
Add support for compiling KallistiOS and programs with Link-Time Optimizations (LTO).

Add `-flto=auto` to the default KOS_CFLAGS, as well as `-fipa-pta` which is a worthwhile optimization.

Tested with the GCC13 toolchain. This *might* work with the GCC9 toolchain, and will most likely fail with the old GCC 4.x toolchain.

### DIFF
--- a/doc/CHANGELOG
+++ b/doc/CHANGELOG
@@ -159,6 +159,7 @@ KallistiOS version 2.1.0 -----------------------------------------------
 - *** Add timespec_get C11 function to koslib's libc [LS]
 - DC  Added timer_ns_gettime64() and performance counters [AB]
 - *** Added check to allow strict C++17+ to use timespec_get [FG]
+- *** Add support for compiling with LTO [Paul Cercueil == PC]
 
 KallistiOS version 2.0.0 -----------------------------------------------
 - DC  Broadband Adapter driver fixes [Megan Potter == MP]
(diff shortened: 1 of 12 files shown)


ljsebald:
@pcercuei Can you make a version of this PR without the changes to the environ scripts, but rather just a note in the documentation (or even a comment in the environ script) that says what to change to enable LTO?

Considering your comment that it may not work with the GCC 9.x toolchain and that it will not work with the 4.x one, I would rather that we don't enable these things by default if it's going to break most people's tools.

pcercuei:
@ljsebald Done + rebased.

doc/CHANGELOG
```diff
@@ -5,6 +5,7 @@ versions (where they are missing from early versions, assume that they
 should be attributed to Megan Potter).
 
 KallistiOS version 2.1.0 -----------------------------------------------
+- *** Add support for compiling with LTO [Paul Cercueil = PC]
```
ljsebald, line 8: New entries in the CHANGELOG go at the bottom of the list for that version, not the top (otherwise we'd be constantly moving the = entries around.

pcercuei: Updated.

environ_base.sh
```diff
@@ -28,10 +28,10 @@ export KOS_LIBS="-Wl,--start-group -lkallisti -lc -lgcc -Wl,--end-group"
 export KOS_CC="${KOS_CC_BASE}/bin/${KOS_CC_PREFIX}-gcc"
 export KOS_CCPLUS="${KOS_CC_BASE}/bin/${KOS_CC_PREFIX}-g++"
 export KOS_AS="${KOS_CC_BASE}/bin/${KOS_CC_PREFIX}-as"
-export KOS_AR="${KOS_CC_BASE}/bin/${KOS_CC_PREFIX}-ar"
+export KOS_AR="${KOS_CC_BASE}/bin/${KOS_CC_PREFIX}-gcc-ar"
```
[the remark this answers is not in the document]
ljsebald: Per a discussion on Discord, it does indeed appear that these are still required, despite that SO post.

ljsebald (dismissed):

ljsebald (approved):

PR #135, opened 2023-03-13
Removed all Compiler Warnings when Building KOS

gyrovorbis:
There are a bunch of dumb and mostly harmless warnings in a bunch of files when building KOS, and I think newer GCC versions are starting to get even more whiny about certain things that used to slide. It's been on my agenda to fix them.

- bunch of signed vs unsigned comparisons warnings
- bunch of "X can be used uninitialized" warnings
- several variables declared and never used
- errno setting when erasing a directory in `fs_fat.c` looks like it may have actually been set wrong

Note that my initial work on these warnings is with GCC12, because I think it's the angriest one, but I will make sure they're resolved in the other toolchains as well. 

EDIT: Done. @cepawiel just verified warnings are all gone with legacy and stable toolchains as well. THANK YOU!


addons/libkosfat/directory.h
```diff
@@ -14,6 +14,8 @@ __BEGIN_DECLS
 
 #include "fatfs.h"
 
+#define PACKED __attribute__((packed))
+
```
ljsebald, line 18: Did you mean to change the two instances of `__attribute__((packed))` down below to use this? Otherwise, why is this needed?

kernel/libc/koslib/realpath.c
```diff
@@ -26,11 +26,10 @@
  * SUCH DAMAGE.
  */
 
+#include <kos/limits.h>
```
ljsebald, line 29: Why'd this need to be moved?

andressbarajas: Looks like he was going from bigger to smaller depending on the length of the includes.  I do this too.

ljsebald (approved):

PR #141, opened 2023-03-31
Update kos/time.h header for C++ support

gyrovorbis:
- Notice the timespec stuff was only made available in C++17
- C++ is being built with __STRICT_ANSI__ apparently which made the code conditionally #ifdef'd out


gyrovorbis:
For anyone following, it was conditionally compiled out, because I was compiling with -std=c++XX, without the GNU extensions, which should and does enable __STRICT_ANSI__, so that's right, and this would've worked when building for C++ with GNU extensions enabled.

ljsebald (approved):

PR #143, opened 2023-04-01
Added ARM driver stream.drv to .gitignore

gyrovorbis:
Surely I'm not the only one who keeps accidentally almost committing this thing? Assuming the binary actually belongs in the repo, if it should ever need be updated, it can be done with `git add -f` in the future.


gyrovorbis:
Okay, I'm an idiot. The file has to be removed from the repo for .gitignore to matter. 

Does anybody else feel like it shouldn't be in here? Is it here just so you don't need the ARM toolchain? A few of us keep running into issues with almost committing it on accident, since it's updated with KOS and is always tracked.

PR #146, opened 2023-04-02
Integrate NSTimer into Newlib and C/C++ Date/Time Implementation

gyrovorbis:
The new performance counter API brought us a nanosecond timer in KOS. This nanosecond timer can immediately be leveraged to increase the resolution of standard C and C++ time/chrono functions which support the extra resolution. The primary goal is to allow developers at the application-level to leverage the NS timer's increased accuracy transparently, by just using standard library calls in platform-independent code.

NOTE: While I was in there working with the timer/performance counters, I decided to do some minor cleanup and refactoring. Most notably, I believe the performance counters deserve their own separate header and source file, rather than just being tacked onto the bottom of the timer files, because they represent a separate peripheral and hardware feature of the SH4, which encompass more than just timing... They're also a separate "doxygen" module.

I'm also taking the opportunity to flesh out the Doxygen documentation of the timer API better, now that we're trying to update it and clean it up.

EDIT:

Here's a list of timing/precision enhancements I've implemented:
1) `timer_usec_gettime64()` is no longer lossily using msec by default. It defaults to nsec before gracefully falling back to msec when not available.
2) Newlib's POSIX `gettimeofday()` is now using the usec timing, so it won't be lossy by default
3) Out implementation of C11's `timespec_get()` is now using the full nanosecond resolution by default, but will gracefully fall back to msec resolution if you tie up the first performance counter.

I've also taken the liberty of steroiding out the classic "simple clock" C date/time demonstration (but in C++) demo to flex:
1) C time: `time()`, `timespec_get()`, `clock()`
2) POSIX time: `gettimeofday()`
3) C++ time: `std::chrono`


gyrovorbis:
Just an update on the progress here... for myself and anyone else... I'm trying to make the Newlib stdlib implementations use the high-resolution NS timer provided by the performance counters when available now, for higher resolution than the current MS timer implementation...

I modified the CRAP out of the simple DC clock example to flex all of the various ways which date/time can be queried in C, C++, and POSIX: https://twitter.com/falco_girgis/status/1642412123807514624

...however, I was running into an issue with timing... I don't understand what's wrong with my logic, but it does not appear that the NS timer is actually maintaining time correctly. I could verify that it was nowhere near correct when making part of newlib use the MS timer and part use the NS timer... The NS timer would consistently run slow.

I'm not sure whether I'm doing something stupid or whether there's a bug in the NS Timer/performance counter code, or (god forbid), if we've made a bad assumption about even being able to use it as an NS timer or not... but it's going to take a little bit more work to investigate exactly what's going on.

andressbarajas:
I know we chatted about this before in discord but putting it here for the record.  According to  https://github.com/sega-dreamcast/dreamhal/blob/3620f5b96af2668d8c065bf78abf219224d236b8/modules/perfctr.h#LL30C57-L35C13:

There is also a second mode that counts cycles according to a ratio between the CPU frequency and the system bus clock, and it increments the counter by 12 every bus cycle. This second mode is detailed in the description for PMCR_CLOCK_TYPE in this file, and it is recommended for use when the CPU frequency is not a runtime constant.

It looks to be more accurate and also takes into account overclocked DC's.

QuzarDC:
Was trawling through some old threads and wanted to raise this issue here, rather than as a new PR to not interfere much. Seems it got lost in the shuffle when originally brought up: https://dcemulation.org/phpBB/viewtopic.php?t=105777

gyrovorbis:
> Was trawling through some old threads and wanted to raise this issue here, rather than as a new PR to not interfere much. Seems it got lost in the shuffle when originally brought up: https://dcemulation.org/phpBB/viewtopic.php?t=105777

OMG. THANK YOU for posting this. I had no idea, and it seems like you already looked into it and figured out a solution too. Will make sure to address it!

@andressbarajas Thanks for the info, going to use that method then.

Just finished unloading the big truck after my move, but have to unpack everything and set up my dev stuff again, but after that this PR will be my top priority to finish.

gyrovorbis:
Alrighty, @ljsebald, we've hit an impasse with this one, and it looks like the future of this whole endeavor is going to be up to your wisdom.

Sooo... The whole point of this PR was to use the new performance counter-based nanosecond timer for backing C, C++, and POSIX date/time calls to up the resolution of our timing... This was a relatively simple change to do, as the performance timers have an "elapsed time" mode, which just counts clock or BUS ratio cycles...

Problem is, I could not get the damn thing to EVER properly keep time. I made a whole test bed and was sure to compare its timing to our own MS timers to ensure accuracy, and it was just always FAR off. (Pardon the language, I'm still 12):
IMG_8030

Fast forward a few weeks, and both @andressbarajas and @cepawiel have pitched in to help the cause of getting to the bottom of this... and to make a long story short, guess what breaks them: https://github.com/KallistiOS/KallistiOS/blob/ad6524dbf2e6094933dd8fd81a044dc42dca2efe/kernel/arch/dreamcast/include/arch/arch.h#L284

"Sleep" basically prevents them from counting cycles, and commenting out this single line has it all working perfectly: https://github.com/KallistiOS/KallistiOS/blob/master/kernel/thread/thread.c#L196

We have also diligently gone back over the Moop and TapamN discussions on performance counters, and surprise, it looks like both of them have also encountered this and have independently reached the same conclusion.

So now how to proceed beyond here is what we aren't sure about. Obviously sleeping when we aren't using the processor is the right thing to do. How bad would it be to not do that? Is it worth not doing to increase our timing precision?

pcercuei:
I'd suggest the high-res timers to be explicitly enabled with a flag in KOS_INIT_FLAGS() for instance. Then in that case, the processor would never `sleep`.

I don't think not sleeping is a big deal - the SH-4 eats only a few watts at full blast, it only marginally contributes to the overall power usage of the console.

PR #147, opened 2023-04-02
Fix overflow warning

andressbarajas:
Fix an overflow warning in gdb_stub


gyrovorbis:
You fixed a BUG with that, technically. ;)

ljsebald (approved):

PR #149, opened 2023-04-03
64-Bit Store Queue Copy + SQ Doxygen Module

gyrovorbis:
- Added sq_cpy64(), which was apparently floating around in the scene for years and was originally from the (fantastic) SMS Plus emulator, so credit goes to Pascal Bosquet & Heliophobe. Thanks to Ian Michael for making me aware of it.
- Took the time to flesh out a new "Store Queues" top-level Doxygen module, taking care to add warnings for alignment requirements to everything and to give a slightly more detailed introduction to the SQs for newbs who have no idea what they are.

### DIFF
--- a/kernel/arch/dreamcast/hardware/sq.c
+++ b/kernel/arch/dreamcast/hardware/sq.c
@@ -72,6 +72,48 @@ void * sq_cpy(void *dest, const void *src, int n) {
     return dest;
 }
 
+/* copies n bytes from src to dest, dest must be 64-byte aligned */
+void* sq_cpy64(void *dest, const void *src, int n)
+{
+
+  uint32 *sq;
+  uint32 *d, *s;
+  
+  d = (uint32 *)(0xe0000000 | (((uint32)dest) & 0x03ffffe0));
+  s = (uint32 *)(src);
+  
+  
+  *((volatile unsigned int*)0xFF000038) = ((((uint32)dest)>>26)<<2)&0x1c;
+  *((volatile unsigned int*)0xFF00003C) = ((((uint32)dest)>>26)<<2)&0x1c;
+  
+  n >>= 6;
+  while (n--) 
+  {
+    // sq0 
+    sq = d;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    __asm__("pref @%0" : : "r" (d));
+    d += 8;
+    
+    // sq1 
+    sq = d;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    __asm__("pref @%0" : : "r" (d));
+    d += 8;
+  }
+
+  *((uint32 *)(0xe0000000)) = 0;
+  *((uint32 *)(0xe0000020)) = 0;
+
+  return dest;
+}
+
 /* fills n bytes at s with byte c, s must be 32-byte aligned */
 void * sq_set(void *s, uint32 c, int n) {
     unsigned int *d = (unsigned int *)(void *)
--- a/kernel/arch/dreamcast/hardware/sq.c
+++ b/kernel/arch/dreamcast/hardware/sq.c
@@ -5,6 +5,7 @@
 */
 
 #include <dc/sq.h>
+#include <stdint.h>
 
 /*
     Functions to clear, copy, and set memory using the sh4 store queues
@@ -112,6 +113,19 @@ void* sq_cpy64(void *dest, const void *src, int n)
   *((uint32 *)(0xe0000020)) = 0;
 
   return dest;
+}
+
+/* copies n bytes from src to dest (in VRAM), dest must be 32-byte aligned */
+void * sq_cpy_pvr(void *dst, const void *src, size_t len) {
+   //Set PVR DMA register
+   *(volatile int *)0xA05F6888 = 1;
+   
+   //Convert read/write area pointer to DMA write only area pointer
+   void *dmaareaptr = (void*)(((uintptr_t)dst & 0xffffff) | 0x11000000);
+   
+   sq_cpy(dmaareaptr, src, len);
+
+   return dst;
 }
 
 /* fills n bytes at s with byte c, s must be 32-byte aligned */
--- a/kernel/arch/dreamcast/include/dc/sq.h
+++ b/kernel/arch/dreamcast/include/dc/sq.h
@@ -75,12 +75,13 @@ void sq_clr(void *dest, int n);
     \param  n               The number of bytes to copy (multiple of 32).
     \return                 The original value of dest.
 
-    \sa sq_cpy64()
+    \sa sq_cpy64(), sq_cpy_pvr()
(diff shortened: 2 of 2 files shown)


QuzarDC:
I'd suggest that the function be adjusted to match the other versions. Specifically, use the QACR defines, use the dest pointer d the same way in the loop, and tidy it up in the same way at the end. Would make it more readable in context (at least in as much as clearly seeing the difference between it and the other).

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -72,6 +72,48 @@ void * sq_cpy(void *dest, const void *src, int n) {
...
+  
+  d = (uint32 *)(0xe0000000 | (((uint32)dest) & 0x03ffffe0));
+  s = (uint32 *)(src);
+  
+  
+  *((volatile unsigned int*)0xFF000038) = ((((uint32)dest)>>26)<<2)&0x1c;
+  *((volatile unsigned int*)0xFF00003C) = ((((uint32)dest)>>26)<<2)&0x1c;
+  
+  n >>= 6;
+  while (n--)
```
gyrovorbis, line 90: Note to self: BlueCrab commented on this algorithm never prefetching from the source buffer... This is something to play with when I am set up to profile with the perf counters.

gyrovorbis:
> 

Good call on all fronts. Will do!

gyrovorbis:
So it has also been brought to my attention that TapamN has shared this JUICY little tidbit on the forums: https://www.dreamcast-talk.com/forum/viewtopic.php?f=2&t=16230&hilit=voxel

```
void modified_sq_cpy_pvr32(void *dst, void *src, size_t len) {
   //Set PVR DMA register
   (volatile int *)0xA05F6888 = 1;
   
   //Convert read/write area pointer to DMA write only area pointer
   void *dmaareaptr = ((uintptr_t)dst & 0xffffff) | 0x11000000;
   
   sq_cpy(dmaareaptr, src, len);
}
```
This is for writing to VRAM quickly using the SQs. He shared some pretty eye-opening timing analysis using a benchmark he wrote:
```
memcpy: 20.80 ms
KOS sq_cpy: 8.24 ms
Modified sq_cpy: 3.89 ms
```
I think we should absolutely include his version in KOS, as this is an incredible performance benefit for a common operation that he's shared with us...

I've also seen from Compiler Explorer that our existing sq_copy() seems to be generating superior ASM by using the __builtin_prefetch() intrinsic rather than inline assembly for prefetching... I intend to profile this in order to confirm, but perhaps there are other areas for improvement, especially with more modern toolchains and tools.

While we're revisiting the SQ API--which hasn't really been touched in a decade--does anybody have anything else they'd like to contribute? Now that we have the performance counters, we plan to do detailed profiling and analysis, and are in a good position to optimize and verify this stuff. We all know how important this SQ stuff is to performance... so if any of you have been sitting on any gems, please do share with the class. You will get full credit. Lets make this API all it can be. We want to make the documentation better and also provide an example/benchmark for newbies to see first-hand why they should be leveraging some of this stuff and how to do so too.

andressbarajas:
I did google search on some of those addresses: 

https://dreamcast-talk.com/forum/viewtopic.php?f=5&t=14727&start=70
https://dcemulation.org/phpBB/viewtopic.php?t=21563
https://dreamcast-talk.com/forum/viewtopic.php?f=5&t=14727&sid=8304c7bfa995b443f7041c49632ddae5&start=80#p156774

gyrovorbis:
> I did google search on some of those addresses:
> 
> https://dreamcast-talk.com/forum/viewtopic.php?f=5&t=14727&start=70 https://dcemulation.org/phpBB/viewtopic.php?t=21563 https://dreamcast-talk.com/forum/viewtopic.php?f=5&t=14727&sid=8304c7bfa995b443f7041c49632ddae5&start=80#p156774

Very nice! Ian just sent me this one, which it looks like you just found too:
```
void pvr_tex_lmemset32(pvr_ptr_t dst, int l, size_t len) {
   //Set PVR DMA registers
   volatile int *pvrdmacfg = (int*)0xA05F6888;
   pvrdmacfg[1] = pvrdmacfg[0] = 0;
   
   //Convert read/write area pointer to DMA write only area pointer
   void *dmaareaptr = ((uintptr_t)dst & 0xffffff) | 0x11000000;
   
   sq_set32(dmaareaptr, l, len);
}
```

Also I realized we can tag people as "authors" on a per-function granularity, so hopefully people will be more motivated to contribute if they have anything... Darc is about to get pushes to this repo auto-updating the Doxygen pages on Dreamcast.wiki, so they'll show up there too.

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -72,6 +73,61 @@ void * sq_cpy(void *dest, const void *src, int n) {
...
+  *((uint32 *)(0xe0000000)) = 0;
+  *((uint32 *)(0xe0000020)) = 0;
+
+  return dest;
+}
+
+/* copies n bytes from src to dest (in VRAM), dest must be 32-byte aligned */
+void * sq_cpy_pvr(void *dst, const void *src, size_t len) {
+   //Set PVR DMA register
+   *(volatile int *)0xA05F6888 = 1;
```
Kazade, line 121: Can we alias these registers somewhere so it's clear what the magic numbers are? Same for the 64 byte copy

gyrovorbis: Very very very good point.

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -72,6 +73,61 @@ void * sq_cpy(void *dest, const void *src, int n) {
...
+  n >>= 6;
+  while (n--) 
+  {
+    // sq0 
+    sq = d;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    *sq++ = *s++; *sq++ = *s++;
+    __asm__("pref @%0" : : "r" (d));
```
DC-SWAT, line 99: You can use dcache_pref_block here

DC-SWAT, line 99: You can use dcache_pref_block here

gyrovorbis: Good idea!

andressbarajas:
Can you replace all instances of uint32 with uint32_t?

kernel/arch/dreamcast/hardware/sq.c
andressbarajas, line 121: That address is wrong.  Here is updated Tapamn code from a voxel.7z provided here(https://www.dreamcast-talk.com/forum/viewtopic.php?f=2&t=16230&hilit=voxel&start=10#p178783):

#define MEM_MODE_VRAM64	0
#define MEM_MODE_VRAM32	1
void sq_cpy_pvr_fast(void *dst, void *src, size_t len, unsigned mem_mode) {
   //Set PVR DMA register
   *(volatile int *)0xA05F6884 = mem_mode;
   
   //Convert read/write area pointer to DMA write only area pointer
   uint32_t dmaareaptr = ((uintptr_t)dst & 0xffffff) | 0x11000000;
   
   sq_cpy((void *)dmaareaptr, src, len);
}

gyrovorbis:
Closing this bad boy out on account of @andressbarajas taking it over for me. Everyone be sure to follow up on his PR: https://github.com/KallistiOS/KallistiOS/pull/320

PR #150, opened 2023-04-03
Cleanup of register access in video.c

QuzarDC:
This was a set of changes I had taken notes to do long ago. The idea is to redo all the register access in video.c, which sets up basic framebuffer operations, to use the register and memory macros that are already defined and shared with pvr.h. A number of the registers used had never been defined before so pvr.h is also being updated.

All changes here should be transparent to the operations as they are merely switching direct access by an array into the same abstracted by macro.

The one change that I rolled into here that should make a difference is that fb_console will attempt to use the current video settings, if they exist, to set its initialization rather than a default (which it still falls back to).

I don't currently have a DC set up to test the correctness, but it should be possible immediately to test with any FB-based/text demo as the registers being wrong would break video.

Setting as draft to ensure there can be both review and testing before this moves. I will be tying this to an issue to standardize register access and naming.


kernel/arch/dreamcast/include/dc/video.h
```diff
@@ -47,9 +47,13 @@ __BEGIN_DECLS
 */
 #define PM_RGB555   0       /**< \brief RGB555 pixel mode (15-bit) */
 #define PM_RGB565   1       /**< \brief RGB565 pixel mode (16-bit) */
+#define PM_UNK      2       /**< \brief Unknown pixel mode (?-bit) */
```
DC-SWAT, line 50: Look here:

https://github.com/DC-SWAT/DreamShell/blob/master/firmware/isoldr/loader/kos/dc/video.h#L49

DC-SWAT: I use it for screenshot feature in ISO Loader.

andressbarajas:
Does this mean that there is another video mode to be supported here:

https://github.com/KallistiOS/KallistiOS/blob/master/kernel/arch/dreamcast/util/screenshot.c

QuzarDC:
> Does this mean that there is another video mode to be supported here:
> 
> https://github.com/KallistiOS/KallistiOS/blob/master/kernel/arch/dreamcast/util/screenshot.c

Yes, but the reason I updated vid_clear was because it didn't previously have a default case when switching on pm. Screenshot already handles that case. But that is quite lazy of me. I'll see about updating that too.

DC-SWAT:
It will be great if you clean up the video modes and add native widescreen support:

https://github.com/sega-dreamcast/dreamhal/blob/3620f5b96af2668d8c065bf78abf219224d236b8/startup/startup_support.c#L2741

QuzarDC:
> It will be great if you clean up the video modes and add native widescreen support:
> 
> https://github.com/sega-dreamcast/dreamhal/blob/3620f5b96af2668d8c065bf78abf219224d236b8/startup/startup_support.c#L2741

I've made a note of that for the future. It shouldn't be too difficult a process to import a few of those modes. For this PR I was looking for it to be primarily stylistic for readability/documentation (especially since I don't currently have a test environment set up).

gyrovorbis:
Just for getting rid of the magic numbers alone, I THANK YOU. GODSPEED!

QuzarDC:
I've shifted this out of draft. I was able to compile and run the video examples and some of my old ports with these changes in place and the video seemed to work fine. If anyone has a setup with 50hz, RCA connections, or something other than VGA, it would be useful as the whole of the core video driver was fiddled with. Hopefully though, any fat-fingered mistake would break it all.

gyrovorbis:
Is there any way that while we're at it, we can change the following prototypes:
```
void vid_empty();
void vid_waitvbl();
```
to have explicitly `(void)` argument lists? 

In plain C, an empty argument list is not implicitly void, as it is within C++. It means an "unspecified" number of arguments... So essentially we're losing type checking and type-safety on our API when we do that within C, as the compiler will no longer validate arguments.

This is an issue throughout the codebase that I've spoken with BlueCrab about. We've been slowly fixing the declarations to void argument lists as we've been touching files like this.

QuzarDC:
> Is there any way that while we're at it, we can change the following prototypes:
> 
> ```
> void vid_empty();
> void vid_waitvbl();
> ```
> 
> to have explicitly `(void)` argument lists?
> 
> In plain C, an empty argument list is not implicitly void, as it is within C++. It means an "unspecified" number of arguments... So essentially we're losing type checking and type-safety on our API when we do that within C, as the compiler will no longer validate arguments.
> 
> This is an issue throughout the codebase that I've spoken with BlueCrab about. We've been slowly fixing the declarations to void argument lists as we've been touching files like this.

Sure, that one I rolled in since it should have no run-time impact. Also went ahead and did it for all the empty function arguments in video.c/h

gyrovorbis (commented):
I just reviewed it and only saw that one little thing. We should wait for Crabby's input, but WOW, this is SO MUCH better.

kernel/arch/dreamcast/hardware/video.c
```diff
@@ -586,33 +585,33 @@ void vid_set_mode_ex(vid_mode_t *mode) {
...
+    PVR_SET(PVR_BITMAP_X, mode->bitmapx);
     data = mode->bitmapy;
 
     if(mode->flags & VID_PAL) {
         data++;
     }
 
     data = (data << 16) | mode->bitmapy;
-    regs[0x3C] = data;
+    PVR_SET(PVR_BITMAP_X, data);
```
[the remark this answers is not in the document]
QuzarDC: Fixed, yup that was wrong.

QuzarDC:
> OH YEAH. Almost forgot. @ljsebald is going to want you to update the CHANGELIST located in the `doc` folder.

I'd not done that since it doesn't much add feature, but looking through I see others like that (including one of my own old cleanups :P ) so sure.

QuzarDC (commented):
Just a comment that I figured the last commit (adding spaces in the long list of register defines) would both make it easier to read/review and provide some visual clues for as-of-yet undefined functionality. While certainly not every gap in the address space points to some undiscovered thing, I think this is the kind of thing that could flow into #151 as a standard way we define out these groups.

ljsebald (dismissed):

ljsebald (approved):

PR #155, opened 2023-04-20
Add ability to specify git repos for dc-chain sources

cepawiel:
These changes to dc-chain will allow the use of git repositories to download sources instead of using URLs to tarballs. The purpose of this is to be able to easily build versions of the toolchain with upstream versions of the sources in order to keep on top of new releases (i.e. GCC14, GCCRS). 

Currently the variables in `config.mk` the options for `sh_gcc` would are set as follows:
```
sh_gcc_ver=12.2.0
sh_gcc_tarball_type=xz
```
This will continue to function as it has been, but by changing the `sh_gcc_download_type` we can specify a git repo instead:
```
sh_gcc_ver=rs
sh_gcc_download_type=git
sh_gcc_git_repo=https://github.com/Rust-GCC/gccrs.git
sh_gcc_git_branch=master
```
In which case the `tarball_type` variable also becomes optional. The `sh_gcc` is just used as an example here, it should also work perfectly fine with `arm_gcc`, `newlib`, or any other package where the version and tarball type can be specified. 

From most tested to un-tested:
- [X] Linux
- [X] MacOS (Arm64) 
- [x] Windows

### DIFF
--- a/utils/dc-chain/README.md
+++ b/utils/dc-chain/README.md
@@ -135,17 +135,18 @@ Doing so will use your custom versions of **GMP**, **MPC**, **MPFR** and
 if you have trouble using the `contrib/download_prerequisites` script provided
 with GCC.
 
-Please note that you have the possibility to specify the `tarball_type`
-extensions you want to download too; this may be useful if a package
-changes its extension on the servers. For example, for GCC `4.7.4`, there is no
-`xz` tarball file, so you may change this to `gz`.
+Please note that you have the possibility to specify the tarball extensions
+you want to download using `download_type`; this may be useful if a
+package changes its extension on the servers. For example, for GCC `4.7.4`, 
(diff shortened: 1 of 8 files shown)


darcagn:
Tested Windows functionality with mingw-w32, it works as expected. Marking Windows as tested.

gyrovorbis:
Was there anything left to verify or do before this PR goes out of draft?

darcagn:
I've built both the GCC master branch (14.0.0) and GCC-RS toolchains from git with this successfully, including with the tarball_type fallback. Approving!

darcagn (approved):

sizious:
What could be great also is to have the possibility to download RC tarballs like @darcagn explained in another item

ljsebald (approved):

PR #157, opened 2023-04-21
Optimised memcpy and memset for Newlib

DC-SWAT:
- Prefetch and allocate CPU cache with instructions like `pref` and `movca.l`
- Support 64-bit copy via FPU registers
- Superscalar and low latency instruction ordering

This is the best implementation that I could find some time ago for the ISO loader firmware.
I liked the result so much that I built them into the DreamShell Core too with postfix "_sh4".
It seems to me that it would be good place them to be directly in the Newlib.


ljsebald:
Unfortunately, I cannot accept this PR.

The code for those functions is licensed under the LGPLv2.1 with no special linking clause (per the header comments in the memset implementation), which would essentially require that everything that uses our Newlib would need to be licensed under compatible terms (which is way beyond the normal BSD license terms that we have for everything). This is not a situation I want us to be in, so, I can't accept the PR.

sizious:
Done :)
Currently, the port is hosted [here](https://github.com/sega-dreamcast/libfast) but feel free to suggest another location.

Issue #156, opened 2023-04-20
Unable to build KallistiOS on GCC 4.x since LTO enabled


sizious:
**It’s now impossible to build KallistiOS using GCC 4.x since LTO (Link Time Optimization) is enabled.**

Some exports aren't properly defined, as shown [here](https://github.com/cepawiel/dreamcast-toolchain-containers/actions/runs/4598237777/jobs/8122028216#step:9:621). As explained by @pcercuei:

> `fdopen()` is in `stdio.h` and needs `_POSIX_C_SOURCE` to be defined to POSIX `2001` minimum. The generated code validates those conditions but we need to fix [genexports](https://github.com/KallistiOS/KallistiOS/tree/master/utils/genexports).

pcercuei:
genexports works fine. The problem is that this won't compile on the GCC4 toolchain:
```
#define _POSIX_C_SOURCE 200809
#include <stdio.h>

int main(void) {
    fdopen(0, NULL);
    return 0;
}
```

gyrovorbis:
I've opened a PR addressing this issue here: https://github.com/KallistiOS/KallistiOS/pull/159

gyrovorbis:
Should be fixed now, Crabby just approved my PR.

PR #159, opened 2023-04-23
Fixed building with GCC-4.7.4 after LTO changes

gyrovorbis:
This PR is to address the issue found here: https://github.com/KallistiOS/KallistiOS/issues/156.

- The compiler/Newlib config for 4.7.4 was such that the POSIX function "fdopen" was never getting declared in stdio.h when generating the exports file
- Added a preprocessor check for GCC 4 that explicitly declares the function's prototype
- Also fixed a minor issue with the Makefile in the "stackprotector" not defining a rule for "clean" on GCC4, which made top-level "make clean" calls fail out on that example

Verification:
- [x] GCC-4.7.4
- [x] GCC-9.3.0
- [x] GCC-12.2.0

### DIFF
--- a/examples/dreamcast/basic/stackprotector/Makefile
+++ b/examples/dreamcast/basic/stackprotector/Makefile
@@ -4,6 +4,9 @@ KOS_CFLAGS += -fstack-protector-all
 
 GCC_MAJOR = $(basename $(basename $(KOS_GCCVER)))
 
+clean:
+	-rm -f $(TARGET) $(OBJS)
+
 ifeq ($(GCC_MAJOR), 4) 
 
 all dist:
@@ -14,9 +17,6 @@ else
 all: rm-elf $(TARGET)
 
 include $(KOS_BASE)/Makefile.rules
-
-clean:
-	-rm -f $(TARGET) $(OBJS)
 
 rm-elf:
 	-rm -f $(TARGET)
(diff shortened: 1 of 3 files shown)


examples/dreamcast/basic/stackprotector/Makefile
```diff
@@ -4,6 +4,9 @@ KOS_CFLAGS += -fstack-protector-all
 
 GCC_MAJOR = $(basename $(basename $(KOS_GCCVER)))
 
+clean:
```
ljsebald, line 7: It'd probably make more sense to revert this and add clean to the `all dist:` line inside the 4.x block.

utils/genexports/genexports.sh
```diff
@@ -30,6 +30,11 @@ for i in $includes; do
 	echo "#include <$i>" >> $outpfile
 done
 
+echo '/* Newlib with GCC 4.7.4 will not export fdopen from stdio.h. */' >> $outpfile
+echo '#if __GNUC__ == 4' >> $outpfile
+echo '	extern FILE* fdopen(int fd, const char* mode);' >> $outpfile
```
ljsebald, line 35: Rather than doing this, we should be able to get the definition from Newlib's stdio.h by undefining `__STRICT_ANSI__` before we the include block for GCC 4.x...

Also, it's not technically the GCC version doing this -- it's the Newlib version, so technically this is wrong anyway.

gyrovorbis: Yeah, fair. I was just worried about doing that, because I wasn't sure what the repercussions would be in terms of the rest of the headers getting conditionally compiled differently from how they're compiled with the rest of KOS, using `-std=c99`...

...but if there was a problem, it wouldn't even build, so I guess I'd know immediately anyway.

ljsebald (approved):

PR #160, opened 2023-04-24
Update copyright notice in GCC and Newlib patches

darcagn:
This patch updates copyright notices in the GCC and Newlib patches to properly credit the author.


gyrovorbis (approved):
Wow. Good catch. Guess they slipped through the cracks.

Issue #163, opened 2023-04-25
Deprecate or Fix Cooperative Threading


QuzarDC:
In trying to use [examples/dreamcast/libdream/320x240/320x240.c](https://github.com/KallistiOS/KallistiOS/blob/1ea31b794b17c758087938cc97b529428fcf1d53/examples/dreamcast/libdream/320x240/320x240.c) to validate changes to video, I found that it happened to override default the KOS INIT which sets Preemptive mode and defaults to cooperative. This caused a problem where the usleep at the end of the example would never return.

At the very least the functionality of usleep has to be corrected to work in cooperative threading mode. At best, we should have some thorough testing of it to root out other issues and properly support it. As an alternative though, just ditch it.

QuzarDC:
I believe I managed to fix the specific usage of coop mode in the example provided above here: https://github.com/KallistiOS/KallistiOS/pull/164 The issue of testing/fixing/ditching stands though, as it's likely that coop mode still breaks various hardware usages.

gyrovorbis:
I cast my vote to deprecate it.

1) It has been been broken for so long, and I don't see anybody complaining.
2) It complicates the internals, potentially needs serious work to fix, and as you said, will require some testing/validation examples at very least to verify it's fixed and allow us to maintain it
3) There are alternatives. For example, C++20 has brought coroutines, and as newfangled as that is, they actually work on DC and are meant to model that kind of cooperative multitasking. Then we have developers like Kazade in Simulant who have basically rolled their own coroutine implementation.

ljsebald:
Honestly, I'd be on-board with deprecating it at this point, unless someone really has a good reason to keep it around.

I doubt there is much out there that really ever used it (especially in the last 10 years), since it wasn't the default.

ljsebald:
Closed with #176.

PR #164, opened 2023-04-25
Using timer when not in preempt

QuzarDC:
This seems to fix the behavior of usleep in https://github.com/KallistiOS/KallistiOS/blob/1ea31b794b17c758087938cc97b529428fcf1d53/examples/dreamcast/libdream/320x240/320x240.c which is in coop threading mode. The idea is just to use the alternate timer method for sleeping when in any mode but preempt (as opposed to just when threading is not initted).

### DIFF
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -655,7 +655,7 @@ static void thd_timer_hnd(irq_context_t *context) {
    sleep because it eases the load on the system for the other
    threads. */
 void thd_sleep(int ms) {
-    if(thd_mode == THD_MODE_NONE) {
+    if(thd_mode != THD_MODE_PREEMPT) {
         timer_spin_sleep(ms);
         return;
     }
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -655,7 +655,7 @@ static void thd_timer_hnd(irq_context_t *context) {
    sleep because it eases the load on the system for the other
    threads. */
 void thd_sleep(int ms) {
-    if(thd_mode != THD_MODE_PREEMPT) {
+    if(thd_mode == THD_MODE_NONE) {
         timer_spin_sleep(ms);
         return;
     }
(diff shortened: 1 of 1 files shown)


ljsebald:
This does technically fix the problem of sleeping breaking entirely, but unfortunately, it also breaks the common assumption of cooperative multitasking -- that the process that is going to sleep gives up the CPU.

The process sleeping is holding onto the CPU in this case, which is not desirable.

QuzarDC:
> This does technically fix the problem of sleeping breaking entirely, but unfortunately, it also breaks the common assumption of cooperative multitasking -- that the process that is going to sleep gives up the CPU.
> 
> The process sleeping is holding onto the CPU in this case, which is not desirable.

Oh, hah. Of course, that makes a lot of sense. I made another fix candidate. The idea is to pass at the end of the sleep after registering the gen wait which normally would just hang.

Need to write a bit of testing to see that it would actually sleep properly with another thread around and wait the appropriate amount of time before resuming execution.

ljsebald:
I don't see how the second version here would fix the problem at hand?

My understanding of the situation is as follows:
1. In cooperative mode, the primary kernel timer is not set up to wake the scheduler.
2. The main thread tries to sleep, which causes the thread to be de-scheduled to wait in `genwait_wait()`, adding it to the timer queue. The only other two threads in the system are the reaper and the idle task. The reaper task is blocked waiting for a signal that it is needed, so the only task available to run is the idle task.
3. The idle task is scheduled and immediately calls `arch_sleep()`, which causes the CPU to sleep until an IRQ.
4. On next IRQ, the idle task is still in control (it has not asked to be de-scheduled), and immediately calls `arch_sleep()` again. This just repeats forever.

What might fix the situation (if my understanding there is correct) is having the idle task do a `thd_block_now(&thd_current->context);` each pass through the loop after `arch_sleep()`.

QuzarDC:
That's how I had followed it too, but I presumed I had misunderstood and at some point it had to get to the thread reschedule after the gen wait because adding that *did* stop a program in coop mode from hanging on usleep. Though I'm totally on board with the consensus from #163 of just ditching coop mode entirely, I would like to understand why this managed to fix it (or maybe I just made a mistake somewhere?).

ljsebald:
Mooted by #176.

Issue #151, opened 2023-04-03
Standardize register access/interface


QuzarDC:
All of the various hardware subsystems and drivers need to read/write to registers. Due to them being written by different people at different times, or being brought in from differing origins, they can differ drastically in how they are addressed, defined, or accessed. If this were standardized, any amount of the code would be more readable and serve as better documentation.

The Maple code might be an exemplar of how to go about this:
In https://github.com/KallistiOS/KallistiOS/blob/master/kernel/arch/dreamcast/include/dc/maple.h#L68

We define a base address, then follow it with defines that have the offset of each register from that base:

`#define MAPLE_BASE      0xa05f6c00          /**< \brief Maple register base */
#define MAPLE_DMAADDR   (MAPLE_BASE+0x04)   /**< \brief DMA address register */`

A macros are then defined:
`#define maple_read(A) ( *((vuint32*)(A)) )

/** \brief  Maple memory write macro. */
#define maple_write(A, V) ( *((vuint32*)(A)) = (V) )`

That is used like this:
`maple_write(MAPLE_RESET1, MAPLE_RESET1_MAGIC);`

On the other end of the scale is the video/framebuffer system:
In https://github.com/KallistiOS/KallistiOS/blob/master/kernel/arch/dreamcast/hardware/video.c#L410

We make a pointer to the registers base:
`static vuint32 *regs = (uint32*)0xA05F8000;`

Then set them all by raw offsets to that base with raw magic numbers:
`    /* Blank screen and reset display enable (looks nicer) */
    regs[0x3A] |= 8;    /* Blank */
    regs[0x11] &= ~1;   /* Display disable */`
    
Sort of halfway between is something like https://github.com/KallistiOS/KallistiOS/blob/master/kernel/arch/dreamcast/include/dc/ubc.h#L32 :
`#define BARA (*((vuint32*)0xFF200000))  /**< \brief BARA register. */
#define BASRA (*((vuint8*)0xFF000014))  /**< \brief BASRA register. */`

Here the register addresses are each in a define, but not built off each other, not named as verbosely, and the casting is built-in.

Reading and writing is then handled by inline functions without intermediary macros:
`static inline void ubc_break_data_write(uint32 address) {
    BASRA = 0;      /* ASID = 0 */
    BARA = address;     /* Break address */`
    
    
[...]

PR #167, opened 2023-04-27
RTC cleanup, documentation, setting date/time

gyrovorbis:
I have been referencing the RTC a bunch with the date/time work for Newlib and now the NS timer stuff with the perf counters that I'm working on... I noticed the docs could be fleshed out better and there was no way to SET the date/time. 

- Improved Doxygen documentation for RTC, adding more information on it and a new module
- Added support for setting the date/time
- Removed magic numbers
- Added a #define for configuring retry attempts for read/write failures
- Added explicit `(void)` argument lists for type checking parameters properly in plain C
- Tested on my own DC, ensuring that the cached boot time was updated properly as well and that standard C date/time functions behave properly with change

### DIFF
--- a/kernel/arch/dreamcast/include/arch/rtc.h
+++ b/kernel/arch/dreamcast/include/arch/rtc.h
@@ -83,7 +83,7 @@ time_t rtc_unix_secs(void);
 */
 int rtc_set_unix_secs(time_t time);
 
-/** \brief   Get the time somce the sytem was booted.
+/** \brief   Get the time since the sytem was booted.
     \ingroup rtc
 
     This function retrieves the cached RTC value from when KallistiOS was started. As
--- a/kernel/arch/dreamcast/include/arch/rtc.h
+++ b/kernel/arch/dreamcast/include/arch/rtc.h
@@ -92,7 +92,7 @@ __BEGIN_DECLS
 
 /** \brief Timestamp write enable
 
-    \ref{RTC_CTRL_ADDR} value to be written in order to unlock
+    #RTC_CTRL_ADDR value to be written in order to unlock
     writing to the timestamp registers.
 */
 #define RTC_CTRL_WRITE_EN         (1 << 0)
--- a/kernel/arch/dreamcast/include/arch/rtc.h
+++ b/kernel/arch/dreamcast/include/arch/rtc.h
@@ -28,7 +28,7 @@ __BEGIN_DECLS
 #include <time.h>
 
 /** \defgroup rtc Real-Time Clock
-    \brief Real-Time Clock Management 
+    \brief Real-Time Clock (RTC) Management
 
     Provides an API for fetching and managing the date/time using
     the Dreamcast's real-time clock. All timestamps are in standard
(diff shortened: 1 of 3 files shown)


kernel/arch/dreamcast/kernel/rtc.c
```diff
@@ -22,13 +25,20 @@
  */
 
 #include <arch/rtc.h>
+#include <arch/timer.h>
 #include <dc/g2bus.h>
 
+#define RTC_CONTROL_ADDR        0xa0710008  /* SFR with write-enable bit */
+#define RTC_SECONDS_HIGH_ADDR   0xa0710000  /* Upper 16 bits of timestamp */
+#define RTC_SECONDS_LOW_ADDR    0xa0710004  /* Lower 16 bits of timestamp */
```
QuzarDC, line 33: I'd suggest that the three register values be in address order (CTRL last), and be moved into the header file. With them in the header they can get doxy comments added, which would help tracking and hw documentation. Then could also note the behavior that is in comments of the set time function with the CTRL being write-only and self clearing when the high_addr is written to.

Though it's possibly overly pedantic, I might also have '#define RTC_CTRL_ADDR_WRITE_ENABLE (1 << 0)' or similar.
Or having RTC_CONTROL_ADDR be RTC_WRITE_ENABLE then define the value as RTC_WRITE_ENABLE_SET. In any case tagging in #151 for visibility there.

kernel/arch/dreamcast/kernel/rtc.c
```diff
@@ -52,23 +63,66 @@ time_t rtc_unix_secs() {
...
+
+    /* Try 3 times to ensure we didn't write a value then have 
+       the clock increment itself before the next. */
+    for(i = 0; i < RTC_RETRY_COUNT; i++) { 
+        /* Write the least-significant 16-bits first, because 
+           writing to the high 16-bits will lock RTC writes. */
+        g2_write_32(RTC_SECONDS_LOW_ADDR, (adjusted) & 0xffff);
+        g2_write_32(RTC_SECONDS_HIGH_ADDR, (adjusted >> 16) & 0xffff);
+
+        /* Read the time back again, to ensure it was written properly. */
```
QuzarDC, line 92: Should this just call rtc_unix_secs? If the reading is fragile, then couldn't it be possible that the write succeeds properly, but the read to verify it is broken without doing the looping? Then we loop over and do 2 invalid writes and maybe still not validate?

If the writing is fragile, as the comments suggest, I could also see a situation where on the last retry the write succeeds then we only get one chance to read correctly.

kernel/arch/dreamcast/include/arch/rtc.h
```diff
@@ -23,29 +27,78 @@ __BEGIN_DECLS
...
 #include <time.h>
 
-/** \brief  Get the current date/time.
+/** \defgroup rtc Real-Time Clock
+    \brief Real-Time Clock Management 
+
+    Provides an API for fetching and managing the date/time using
+    the Dreamcast's real-time clock. All timestamps are in standard
+    Unix format, with an epoch of January 1, 1970. Due to the fact
+    that there is no time zone data on the RTC, all times are expected
```
QuzarDC, line 36: Not sure if it should be here or in the brief, but there should be at least once place where RTC is glossed. So "Real-Time Clock (RTC) Management"

QuzarDC (commented):
Overall approve, but the looping in the set seems off to me (as noted in my comments).

kernel/arch/dreamcast/kernel/rtc.c
gyrovorbis: I had actually considered moving the registers to the header file, but was worried you guys might think that was beyond the scope of what this driver should be doing...

BUT I ACTUALLY TOTALLY AGREE! That's a fantastic idea. I think every header should document the hardware publicly as well, like that for our KOS hardware documentation in doxygen!

Agreed on what you're saying about having a define for the actual write enable bit. I'll get on this.

kernel/arch/dreamcast/include/arch/rtc.h
gyrovorbis: Good call!

ljsebald (approved):

PR #168, opened 2023-04-28
Ensure (void) instead of () for argument-less function declarations

QuzarDC:
As title. It is technically incorrect to have just (), as that can be interpreted as 'the arguments for this are undefined'. There are a number of places where changes were made aside from just replacing () with (void):

1) In kernel/arch/dreamcast/fs/fs_dclsocket.c: irq_enable was allowed to be called as it was prototyped as () and only declared in asm. irq_enable does not take any arguments, so this was never working correctly. Instead irq_restore was intended, which takes the return from irq_disable.

2) In kernel/arch/dreamcast/include/dc/fs_dcload.h the ending header guard comment name was corrected to match the starting one.

3) In kernel/arch/dreamcast/kernel/gdb_stub.c the declarations with the return type split on its own line were modified to follow the coding style of the rest of KOS.

4) In the koslib libc stubs kernel/libc/koslib/scandir.c, kernel/libc/koslib/seekdir.c, and kernel/libc/koslib/telldir.c correct argument lists were provided instead of the empty one. This required including dirent.h to get the 'DIR' typedef. They additionally required the introduced parameters to be void casted to prevent unused warnings.

5) Following from the void casting for 4, /kernel/net/net_thd.c /kernel/net/net_ipv4_frag.c and /kernel/net/net_dhcp.c were updated to remove their use of __attribute__((unused)) in favor of void casting the parameters.

Finally, this was all done by hand and through some searching so it's possible some were missed. I avoided doing this in the malloc internals as they seem to mostly be intentional.


andressbarajas:
To suppress the new warnings, I think you can add lines like:

(void)unused_parameter;

To each function, for each used variable.

QuzarDC:
> To suppress the new warnings, I think you can add lines like:
> 
> (void)unused_parameter;
> 
> To each function, for each unused parameter.

I've updated it to use that. I wasn't aware we did that in the codebase, but found a few in /net/ using it. I also corrected the one other warning, which was related to having implicitly defined the dirent struct.

ljsebald (changes requested):
Instead of using `__attribute__((unused))`, can you use the `(void)parameter;` method for marking parameters as known unused?

A comment in the thread seems to suggest you did this, but I still see the ones using the `__attribute__` in the PR.

QuzarDC:
> Instead of using `__attribute__((unused))`, can you use the `(void)parameter;` method for marking parameters as known unused?
> 
> A comment in the thread seems to suggest you did this, but I still see the ones using the `__attribute__` in the PR.

Sure. I had searched the codebase and saw that other places were using 'unused' so I used that. Previously I had simply let it have the warning. Just to be sure, should it not be used anywhere? If so I'll go ahead and do the same cleanup in the few /net/ files that have it.

gyrovorbis:
I vote for replacing all of the `__attribute__((unused))` with the `(void)parameter` technique, because the former relies on nonstandard compiler directives while the latter is just plain C.

ljsebald (approved):

PR #169, opened 2023-04-28
Fixed missing header from rtc.c

gyrovorbis:
- stdint.h was missing, but it only manifested on GCC-4.7.4.

SORRY, GUYS!


darcagn (approved):
Tsk, tsk.

PR #170, opened 2023-04-29
Add exit code as argument to dcload exit syscall

cepawiel:
Currently this is not supported in dcload, but was tested with lxdream-nitro as it has support for a return code being passed in the r5 register.


gyrovorbis:
LONG OVERDUE!!!! :)

ljsebald (approved):

PR #172, opened 2023-05-01
Remove checked in stream.drv file.

jpeach:
The generic sounds driver "stream.drv" gets built from source, so we don't need to keep the resulting binary checked in.


darcagn:
Hi, James! Thanks for the contribution. However, it's intended to be this way -- see this bit from doc/README:
> You do not necessarily need an ARM version of GCC to build KOS for the DC, as precompiled binaries are provided of the necessary code. As long as you don't intend to modify the sound driver, you don't need the ARM tools.

That being said, the behavior around the sound driver could definitely be improved (as it annoyingly shows up in `git status` as a changed file and sometimes accidentally gets included with PRs), and in fact we've discussed the best way to do so a few times in the past few weeks after a similar PR was submitted, but this was only a casual conversation and no action has been taken yet. I will close this PR but open a new issue so we can track and resolve this annoyance.

PR #176, opened 2023-05-03
Remove Co-Op threading mode

QuzarDC:
As discussed in #163 this seeks to just remove coop mode as an option in KOS. There were very few places where it was referenced or worked against, with the presumption most everywhere being that the system is in pre-emptive mode. It may make sense to change the internal names for thd_mode to something like 'initted' used in other function, as the only purpose it has currently is to note if threading has been enabled and hasn't been shutdown yet.

There *may* also be some other places where conditional considerations were being made about the threading mode without explicitly calling out the mode (nothing was using get_mode) I tried looking around some and couldn't find any. 

I left a stub warning in case someone does pass INIT_THD_PREEMPT they get a warning, so that should be good enough to provide notice. To my knowledge there's not a good way to mark a define as deprecated. I would suppose that we leave that kos_flags bit reserved for now and if we implement more to go above it before reclaiming it (we've got plenty of bits spare).

### DIFF
--- a/include/kos/thread.h
+++ b/include/kos/thread.h
@@ -227,6 +227,7 @@ typedef struct kthread_attr {
     @{
 */
 #define THD_MODE_NONE       -1  /**< \brief Threads not running */
+#define THD_MODE_COOP       0   /**< \brief Cooperative mode (deprecated) */
 #define THD_MODE_PREEMPT    1   /**< \brief Preemptive threading mode */
 /** @} */
 
--- a/include/kos/thread.h
+++ b/include/kos/thread.h
@@ -23,10 +23,7 @@ __BEGIN_DECLS
     \brief  Threading support.
 
     This file contains the interface to the threading system of KOS. Timer
-    interrupts are used to reschedule threads within the system while in
-    preemptive mode. There is also some support for a cooperative threading
-    mode (where each thread must manually give up its timeslice to swap out
-    threads).
+    interrupts are used to reschedule threads within the system.
 
     The thread scheduler itself is a relatively simplistic priority scheduler.
     There is no provision for priorities to erode over time, so keep that in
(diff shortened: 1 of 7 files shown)


ljsebald (changes requested):
I'd rather if we approached this a little bit differently, at least for now...
1. Don't remove the `thd_set_mode()`/`thd_get_mode()` functions just yet. Just mark them deprecated.
2. Change the code of `thd_set_mode()` to log a warning if a user tries to do cooperative mode, and essentially do nothing else.
3. Leave the default init flags to include the preempt flag, and change the warning to be if it isn't specified.

I don't have a problem with removing the parameter to `thd_init()` as that isn't meant to be called by users.

Numbers 2 and 3 on this list will catch anyone who does try to use cooperative mode and at least warn them, rather than causing a compile error (2) or just silently enabling preemption when they aren't expecting it (3).

QuzarDC:
> I'd rather if we approached this a little bit differently, at least for now...
> 
>     1. Don't remove the `thd_set_mode()`/`thd_get_mode()` functions just yet. Just mark them deprecated.
> 
>     2. Change the code of `thd_set_mode()` to log a warning if a user tries to do cooperative mode, and essentially do nothing else.
> 
>     3. Leave the default init flags to include the preempt flag, and change the warning to be if it isn't specified.
> 
> 
> I don't have a problem with removing the parameter to `thd_init()` as that isn't meant to be called by users.
> 
> Numbers 2 and 3 on this list will catch anyone who does try to use cooperative mode and at least warn them, rather than causing a compile error (2) or just silently enabling preemption when they aren't expecting it (3).

Good call. The only difference my update should have from that is that, as described via discord, I have set mode return a warning regardless of what mode they are choosing, to help highlight that it just shouldn't be used anymore.

include/kos/thread.h
```diff
@@ -224,14 +221,12 @@ typedef struct kthread_attr {
...
-    The threading system will always be in one of the following modes. This
-    represents the type of scheduling done by the system (or the special case of
-    threads not having been initialized yet).
+    The threading system will always be in one of the following modes. This 
+    represents either pre-emptive scheduling or an un-initialized state.
 
     @{
 */
 #define THD_MODE_NONE       -1  /**< \brief Threads not running */
-#define THD_MODE_COOP       0   /**< \brief Cooperative threading mode */
```
ljsebald, line 234: This should probably stay too, but with the comment saying that it isn't supported anymore.

ljsebald (approved):

PR #177, opened 2023-05-04
Fix dist target for examples

cepawiel:
This change fixes two issues with the `make dist` target for the examples:

1) `dreameye/basic` had extra `.elf`'s appended to the target variable on both the run and dist targets

2) `network/httpd` was missing a target for dist


QuzarDC (approved):
Looks good to me. Clear and simple cleanup. Good catch!

PR #178, opened 2023-05-05
Cleanup of asic.c/h

QuzarDC:
Through static code analysis I was clued to a few issues in asic.c that could cause illegal array access. asic_evt_set_handler was incorrectly checking the max evtreg and evt. Decided to just do a fairly comprehensive cleanup, similar to #150 . The core changes are as follows:

1) Rewrite register access and listing to follow better standards outlined in other subsystems (working towards #151 ).
2) Reduce data types in the various functions to better describe how the system works (for instance, use a uint16 for the codes rather than uint32).
3) Reduce the size of the asic handler table to match its valid values. The irqs available are 0,1,2 but for whatever reason the table was of size 4 instead of 3.
4) Modify the workings of ASIC_IRQ_DEFAULT to just be defined to its value instead of checking in runtime.
5) Use asserts to error check. Nothing using these routines was ever checking for any kind of error value, and there's not really any kind of behavior that could be used to recover from such an error anyways.
6) Reduce magic numbers. There's still one in handler_irq9. It might make sense to make a second convenience macro like ASIC_EVT_REG_ADDR to address the ack registers by number too.


kernel/arch/dreamcast/include/dc/asic.h
```diff
@@ -24,12 +24,12 @@
...
-#include <arch/types.h>
+#include <stdint.h>
 
 /* All event codes are two 8-bit integers; the top integer is the event code
-   register to look in to check the event (and to acknolwedge it). The
-   register to check is 0xa05f6900+4*regnum. The bottom integer is the
-   bit index within that register. */
+    register to look in to check the event (and to acknolwedge it). The
+    register to check is 0xa05f6900+4*regnum. The bottom integer is the
+    bit index within that register. */
```
ljsebald, line 32: These should not be spaced in an extra space like this... They should line up with the first character of the word `All` in the first line.
```suggestion
   register to look in to check the event (and to acknolwedge it). The
   register to check is 0xa05f6900+4*regnum. The bottom integer is the
   bit index within that register. */
```

QuzarDC: Done.

QuzarDC:
Just to add an additional note, I wasn't able to think of any great ways to test this since it's so internal. It would seem very much like most things would break if anything were off by much. And it should be almost identical to the original functionally.

ljsebald (approved):

PR #188, opened 2023-05-10
Unbounded indexing fix in isp_settings example

QuzarDC:
In testing using the isp_settings example, I found that my main Dreamcast had the unusual value of 18 set for methods. This caused the example to try printing from an out of bounds place and died.

My change checks the size of the pretty names array and doesn't try to use it if the actual value is out of bounds.

### DIFF
--- a/examples/dreamcast/network/isp-settings/isp-settings.c
+++ b/examples/dreamcast/network/isp-settings/isp-settings.c
@@ -27,7 +27,7 @@ int main(int argc, char **argv) {
                 "Static",
             };
             /* Check that cfg.method is set to a known value */
-            if(cfg.method<(sizeof(methods) / sizeof(methods[0])))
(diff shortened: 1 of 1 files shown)


examples/dreamcast/network/isp-settings/isp-settings.c
```diff
@@ -26,7 +26,11 @@ int main(int argc, char **argv) {
                 "PPPoE"
                 "Static",
             };
-            printf("Method:   %s\n", methods[cfg.method]);
+            /* Check that cfg.method is set to a known value */
+            if(cfg.method<(sizeof(methods) / sizeof(methods[0])))
```
ljsebald, line 30:
```suggestion
            if(cfg.method < (sizeof(methods) / sizeof(methods[0])))
```

ljsebald (changes requested):

ljsebald (approved):

PR #189, opened 2023-05-10
Allow overriding stack size / early init function

pcercuei:
Update the GCC-13 patch to allow setting a different stack size.

Add a `early_init` weak function. If provided, the function will be called. This is used in my case to copy data appended to the raw binary to its final location in RAM; the advantage vs. romdisks, is that the data appended to the raw binary will eventually be reused by the .bss section and the heap.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/arch.h
+++ b/kernel/arch/dreamcast/include/arch/arch.h
@@ -182,6 +182,14 @@ extern void * __kos_romdisk;
 /** \brief  State that you don't want a romdisk. */
 #define KOS_INIT_ROMDISK_NONE   NULL
 
+/** \brief  Register a single function to be called very early in the boot
+ *          process, before the BSS section is cleared.
+ *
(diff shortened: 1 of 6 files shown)


ljsebald (changes requested):
There's a few things that prevent me from merging this PR as it stands right now.
1. I have reservations about the added function being called `early_init`. I think the function needs to be named a less-likely-to-be-used name than that. I'm not sure what to suggest here for that, perhaps someone else could chime in with a suggestion?
2. The patches need to be updated for all the supported versions of GCC, not just GCC 13.
3. The early init function really needs to be documented somewhere...

gyrovorbis:
Trying to think of some names...

Considering what it's doing is actually adding a polymorphic override point for the `arch` subsystem's initialization, maybe we should also put it in that namespace? Maaaaybe...

- `arch_preinit()`
- `arch_custom_init()`
- `arch_user_init()`

or something?

`

ljsebald (approved):

PR #192, opened 2023-05-14
Remove Insight from dc-chain

cepawiel:
I've been unable to get Insight to build properly and was initially attempting to fix it, but have had no success. It would need some work to match the rest of our builds, as it currently does not support any patching. 

I cannot imagine anyone is building it with dc-chain as it currently will download the archive `insight-6.8-1a.tar.bz2` and then fail to even attempt to build as its expecting the `a` to also be in the extracted directory name which is incorrect as the extracted directory is named `insight-6.8-1`.  This release that we're attempting to build was released in 2009 as best I can tell.


ljsebald (approved):

ljsebald:
Being that Insight appears to have been last updated over a decade ago, I don't see a problem removing it.

If anyone wants to bring it back, then feel free to comment and volunteer to keep it working.

PR #193, opened 2023-05-15
Update GDB to use _download_type config

cepawiel:
This change allows specifying a git repo to pull sources from, matching the behavior of Binutils, GCC, & Newlib.

Also included in this change is some changes to the downloading tool detection to download `config.guess` so GDB can be downloaded and built without running `download.sh` to obtain it. 

~~As mentioned in #192, I've been unable to build Insight and thus cannot test it fully. The source is downloaded and unpacked into the directory. I will update this PR should we go forward with the removal.~~ GDB does downloads & builds properly on Linux with the legacy, stable, and testing `config.mk` files.

I have some additional changes to support downloading the sources for Binutils, GCC, & Newlib in the Makefile using the same methods, but will need to do some more testing on them before they're PR ready.


cepawiel:
I also added a configuration option `gnu_mirror` to config.mk which allows overriding the default mirror which was found to be intermittently slow during testing depending on which mirror we were directed to.

darcagn:
The three sample config files and phony.mk need newlines at the end of the files.
Also the sample configs contain a typo: "overriden" should be "override" or "overridden" (not sure which you were going for)

utils/dc-chain/config.mk.stable.sample
```diff
@@ -103,7 +103,7 @@ makejobs=-j2
 # here is to build C, C++, Objective C, and Objective C++. You may want to take
 # out the latter two if you're not worried about them and/or you're short on
 # hard drive space.
-pass2_languages=c,c++,objc,obj-c++
+pass2_languages=c,c++
```
ljsebald, line 106: Why was this changed?

utils/dc-chain/config.mk.testing.sample
```diff
@@ -103,7 +103,7 @@ makejobs=-j2
 # here is to build C, C++, Objective C, and Objective C++. You may want to take
 # out the latter two if you're not worried about them and/or you're short on
 # hard drive space.
-pass2_languages=c,c++,objc,obj-c++
+pass2_languages=c,c++
```
ljsebald, line 106: Also, why was this changed?

utils/dc-chain/scripts/common.sh
```diff
@@ -114,8 +114,8 @@ export ARM_ISL_TARBALL_URL=gcc.gnu.org/pub/gcc/infrastructure/isl-${ARM_ISL_VER}
 # Retrieve the web downloader program available in this system.
 export IS_CURL=0
 export WEB_DOWNLOADER=
-curl_cmd=`get_make_var curl_cmd scripts/init.mk`
-wget_cmd=`get_make_var wget_cmd scripts/init.mk`
+curl_cmd="curl --fail --location -C - -O"
+wget_cmd="wget -c"
```
ljsebald, line 118: Why was this changed?

utils/dc-chain/scripts/phony.mk
```diff
@@ -6,8 +6,9 @@
 #
 
 .PHONY: all
+.PHONY: fetch-all fetch-gdb
 .PHONY: patch patch-sh4 patch-arm $(patch_sh4_targets) $(patch_arm_targets)
 .PHONY: $(patch_binutils) $(patch_gcc) $(patch_newlib) $(patch_kos)
 .PHONY: build build-sh4 build-arm $(build_sh4_targets) $(build_arm_targets) clean
 .PHONY: build-binutils build-newlib build-gcc-pass1 build-gcc-pass2 fixup-sh4-newlib
-.PHONY: gdb install_gdb build_gdb unpack_gdb
+.PHONY: gdb install_gdb build_gdb
```
ljsebald, line 14: Missing a newline here at the end.

ljsebald (changes requested):

utils/dc-chain/scripts/common.sh
cepawiel: These were changed due to converting the variables in the Makefile to functions which now take parameters of the URL and optionally the output filename:
```
wget_cmd = wget -c $(if $(2),-O $(2)) '$(1)'
curl_cmd = curl --fail --location  -C - $(if $(2),-o $(2),-O) '$(1)' 
```

Directly pulling this out of the Makefile into the bash script like was done previously would no longer work so I just set them to the previous values that would be obtained from the Makefile to keep the script functioning as it was.

utils/dc-chain/config.mk.stable.sample
cepawiel: Unintentionally committed, was just during testing while there was issues with the ObjC build. Will revert these unintended changes.

ljsebald (approved):

PR #198, opened 2023-05-16
Fix libobjc building after regression in 'Allow overriding stack size / early init function' PR

darcagn:
This PR updates startup.s with the changes that were supposed to be added in the "Allow overriding stack size / early init function" but were only added in the GCC patches directly.

Additionally, after that PR, the following errors occur:
```
undefined reference to `_arch_stack_16m'
undefined reference to `_arch_stack_32m'
```
And then libobjc fails to compile. 

It was @cepawiel 's suggestion to add the following to the `startup.s`:
```
.weak   _arch_stack_16m
.weak   _arch_stack_32m
```
Colton was unsure if this was the correct/proper way to fix the problem, but I am submitting it as a PR as this fixed the issue for me and we can discuss it here.

### DIFF
--- a/kernel/arch/dreamcast/kernel/startup.s
+++ b/kernel/arch/dreamcast/kernel/startup.s
@@ -18,6 +18,9 @@
 .globl __arch_old_fpscr
 .globl __arch_mem_top
 
+.weak   _arch_stack_16m
+.weak   _arch_stack_32m
+
 _start:
 start:
 	! Disable interrupts (if they're enabled)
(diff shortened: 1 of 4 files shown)


gyrovorbis:
Based on the description of the issue and what you guys did, I’m pretty sure this is right. 

I have a barebones libobjc sanity check that just creates a class and a few instances dynamically using the C runtime API somewhere that I can check in once I get my stuff set back up.

gyrovorbis (approved):

ljsebald (approved):

ljsebald:
Seems reasonable to me and if you say it fixes the problem, I have no reason to disagree.

PR #202, opened 2023-05-19
Fixed Objective-C + added C runtime example

gyrovorbis:
Objective-C support has been incomplete and downright broken in KOS. BlueCrab has done a good job providing the initial patchwork for the threading required to use it; however, it doesn't actually work and we have no examples... I've been working with Andrew Apperly to port the Apple Foundation framework to DC using GNUStep-base, and we encountered some toolchain-level issues, which I will be addressing. 

Firstly there's a terrible bug in the Makefile within the libobjc folder, where there `install-strip` target was never actually implemented... meaning if you enabled objective-c as a language when building GCC, you would build it, then never actually install it, building it for nothing.. 

Secondly, there are no examples for using it at all or any validation we can use to ensure libobjc works properly with our toolchains. I'm adding a small example using the libobjc runtime only.

NOTE: Without the Foundation framework, Objective-C is pretty boring and lacks a lot of the functionality typically expected of the language, such as NSObject, NSString, etc. We're working on that... :)

- Added ObjC + ObjC++ Makefile rules
- Modified GCC patches to implement an `install-strip` rule for libobjc
- added an example that dynamically creates a class and calls a method using only the Objective-C runtime API (libobjc)

EDIT: Verified to work with legacy, stable, and testing toolchains!

### DIFF
--- a/doc/CHANGELOG
+++ b/doc/CHANGELOG
@@ -170,6 +170,7 @@ KallistiOS version 2.1.0 -----------------------------------------------
 - DC  Removed cooperative threading mode -- threading is always preemptive [DH]
 - DC  Added early user init function [PC]
 - *** Added Objective-C example testing/demonstrating C Runtime [FG]
+- *** Updated Objective-C examples to include more thorough testing of the runtime aspects of the language [Andrew Apperley == AA]
 
 KallistiOS version 2.0.0 -----------------------------------------------
 - DC  Broadband Adapter driver fixes [Megan Potter == MP]
--- a/doc/CHANGELOG
+++ b/doc/CHANGELOG
@@ -169,6 +169,7 @@ KallistiOS version 2.1.0 -----------------------------------------------
 - DC  Cleaned up asic functions and corrected potential bugs [DH]
 - DC  Removed cooperative threading mode -- threading is always preemptive [DH]
 - DC  Added early user init function [PC]
+- *** Added Objective-C example testing/demonstrating C Runtime [FG]
 
 KallistiOS version 2.0.0 -----------------------------------------------
 - DC  Broadband Adapter driver fixes [Megan Potter == MP]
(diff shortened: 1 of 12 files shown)


andrewapperley (approved):

gyrovorbis:
@andrewapperley Here is the guy who has gone through hell helping me port GNUStep's `libs-base` to DC as our implementation of Apple's "Foundation" framework, btw, which is the main low-level core stdlib for all of ObjC's power, and is what ObjC developers typically rely on.

We still have some issues with it, but we're working through them. Excited to have it on DC, because it's actually one of the most powerful standard libraries I've ever seen, encompassing a WHOLE lot of stuff. We've also successfully mixed it with modern C++20 using Objective-C++ as well... lots and lots of power. lol.

examples/dreamcast/objc/runtime/runtime.m
```diff
@@ -0,0 +1,123 @@
...
+    printIVarsForPerson(person2);
+    printf("Person iVars End:: \n");
+    fflush(stdout);
+
+    if (!person1.bestFriend) {
+        return EXIT_FAILURE;
+    }
+
+    return EXIT_SUCCESS;
+}
```
darcagn, line 123: _knock, knock_
It's the newline police. Open up!

doc/CHANGELOG
```diff
@@ -169,6 +169,8 @@ KallistiOS version 2.1.0 -----------------------------------------------
 - DC  Cleaned up asic functions and corrected potential bugs [DH]
 - DC  Removed cooperative threading mode -- threading is always preemptive [DH]
 - DC  Added early user init function [PC]
+- *** Added Objective-C example testing/demonstrating C Runtime [FG]
+- *** Updated Objective-C examples to include more thorough testing of the runtime aspects of the language [Andrew Apperley == AA]
```
ljsebald, line 173: This line is too long. Wrap it at 80 characters.

gyrovorbis: @andrewapperley I got it, but take note!!

PR #209, opened 2023-05-22
Use threads for button callback

QuzarDC:
The current implementation of the button callback in controller.c merely calls the callback function. While this is very useful for simple things, the fact that it runs the callback during an interrupt causes a number of issues. It limits the scope of what such a callback can do, which has been quite exemplified by our standard usage of it to call arch_exit. 

The idea here is to raise the possibility of configuring the maple button callback routine to spin off a new thread with the function rather than run it entirely in the context of the interrupt that triggered the callback.

It may be the case that this should be an optional behavior alongside the current one. I'm not entirely sure the scope of usage of this functionality and how it could be negatively impacted by the change. I can confirm though that it works with a few different tests in multi-threaded environments that otherwise trigger various errors regarding mutex_lock or thd_join being called in an interrupt.


gyrovorbis:
Cool! I wasn't even aware of the issues this was creating, but I can definitely see how being able to call this in more contexts could be useful. 

Can't really comment much on the side effects of the implementation or do much other than minor C suggestions ,but the solution looks purdy. :)

QuzarDC:
@pcercuei (zcrc in discord) raised the notion that this may merely be a stopgap to the implementation of threaded interrupts. With that in mind, I reworked this to use a standing thread created when the callback in installed rather than being created for the interrupt. In addition, it now cleans up after itself and can be un-set by setting to NULL.

It raised the question though of 'why have this, and not other interrupt callbacks spin off a thread?'. Compared to the majority of interrupt handlers though, this functionality is presented as something to be integrated into what would normally be user-land code. It's possible that the SIP and snd_stream callbacks would be candidates for such a change, as they are similar in that respect. The DMA callbacks might, too, but they are all one step closer into the interrupt handler than the ones mentioned so far.

Finally, the behavior of it now schedules the thread next to run immediately. While this makes it very simple, it also doesn't give the scheduling benefits that just sending it to the queue might.

QuzarDC:
As an update/clarification to the last comment, I wanted to add the three big questions I have about this change:
1)Do we want it? Perhaps we should just be noting that the callback happens in an interrupt, and shouldn't be used for stuff like arch_exit.
2)If we do, is this the right scope to set something like this up in? A thread could be used at the interrupt level or in a few places up the way.
3)If so, is this the right way to do it? This could be the template of redoing dma callbacks or the snd_stream one so getting the structure/layout/details right would be useful.

QuzarDC:
Marking this back as 'ready for review'. The change should be entirely transparent to users but make the button callback work significantly closer to how it's expected to work. The only reason I had marked it as draft was to sort of push to get some more discussion before merging it as this is sort of a 'minimal simplest' version of the implementation, and we don't really have anything that works like it currently in the codebase.

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -29,6 +32,10 @@ static cont_btn_callback_t btn_callback = NULL;
 static uint8_t btn_callback_addr = 0;
 static uint32_t btn_callback_btns = 0;
 
+static kthread_t * btn_callback_thd = NULL;
```
ljsebald, line 35:
```suggestion
static kthread_t *btn_callback_thd = NULL;
```

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -29,6 +32,10 @@ static cont_btn_callback_t btn_callback = NULL;
...
+static uint8 btn_callback_arg_addr = 0;
+static uint32 btn_callback_arg_btns = 0;
```
ljsebald, line 37:
```suggestion
static uint8_t btn_callback_arg_addr = 0;
static uint32_t btn_callback_arg_btns = 0;
```

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -41,11 +48,54 @@ int cont_has_capabilities(const maple_device_t *cont, uint32_t capabilities) {
                    & capabilities) == capabilities) : -1;
 }
 
+static void * btn_callback_wrapper(void* args) {
+
```
ljsebald, line 52:
```suggestion
static void *btn_callback_wrapper(void *args) {
    (void)args;

```

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -41,11 +48,54 @@ int cont_has_capabilities(const maple_device_t *cont, uint32_t capabilities) {
...
+    }
+
+    return NULL;
+}
+
+void cont_btn_callback_shutdown(void) {
+
+    /* This means either the callback is shutting down the 
+       whole system, or some jerk called this in the callback. */
+    if(thd_get_current()->tid == btn_callback_thd->tid) return;
```
ljsebald, line 65:
```suggestion
    /* This means either the callback is shutting down the 
       whole system, or some jerk called this in the callback. */
    if(thd_get_current()->tid == btn_callback_thd->tid)
        return;
```

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -41,11 +48,54 @@ int cont_has_capabilities(const maple_device_t *cont, uint32_t capabilities) {
...
+
+    thd_destroy(btn_callback_thd);
+    btn_callback_thd = NULL;
+    btn_callback = NULL;
+    btn_callback_addr = 0;
+    btn_callback_btns = 0;
+    btn_callback_arg_addr = 0;
+    btn_callback_arg_btns = 0;
+
+    return;
```
ljsebald, line 75:
```suggestion
```

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -41,11 +48,54 @@ int cont_has_capabilities(const maple_device_t *cont, uint32_t capabilities) {
...
+}
+
 /* Set a controller callback for a button combo; set addr=0 for any controller */
 void cont_btn_callback(uint8_t addr, uint32_t btns, cont_btn_callback_t cb) {
+
```
ljsebald, line 80:
```suggestion
```

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -41,11 +48,54 @@ int cont_has_capabilities(const maple_device_t *cont, uint32_t capabilities) {
...
     btn_callback_btns = btns;
     btn_callback = cb;
+
+    btn_callback_thd = thd_create(0, btn_callback_wrapper, NULL);
+
+    /* This may require an update to thd_create to be able to send custom flags. 
+       Otherwise it might run before it gets removed. */
+    thd_remove_from_runnable(btn_callback_thd);
+
+    thd_set_label(btn_callback_thd, "cont_reply cb");
```
ljsebald, line 98:
```suggestion
    thd_set_label(btn_callback_thd, "cont_btn_callback");
```

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -142,4 +193,6 @@ int cont_init(void) {
 
 void cont_shutdown(void) {
     maple_driver_unreg(&controller_drv);
+
```
ljsebald, line 196:
```suggestion
```

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -41,11 +48,54 @@ int cont_has_capabilities(const maple_device_t *cont, uint32_t capabilities) {
...
     btn_callback_addr = addr;
     btn_callback_btns = btns;
     btn_callback = cb;
+
+    btn_callback_thd = thd_create(0, btn_callback_wrapper, NULL);
+
+    /* This may require an update to thd_create to be able to send custom flags. 
+       Otherwise it might run before it gets removed. */
+    thd_remove_from_runnable(btn_callback_thd);
```
ljsebald, line 96: It's actually not safe to call this with interrupts enabled, as it could corrupt the runqueue if a thread scheduling interrupt comes in in the middle of this... Let me ponder the best way to deal with this (probably with another attribute in the `kthread_attr_t` struct).

kernel/arch/dreamcast/include/dc/maple/controller.h
```diff
@@ -223,7 +223,8 @@ typedef void (*cont_btn_callback_t)(uint8_t addr, uint32_t btns);
     \param  addr            The controller to listen on (or 0 for all ports). 
                             This value can be obtained by using maple_addr().
     \param  btns            The buttons bitmask to match.
-    \param  cb              The callback to call when the buttons are pressed.
+    \param  cb              The callback to call when the buttons are pressed, 
+                            or NULL to un-set the current callback.
```
ljsebald, line 227: The part above this in the comment probably should get updated, since it says that the callback will be called in an interrupt context, which is no longer true after this PR is merged.

Also, it'd probably be a good idea to also note that only one of these can be set up, since otherwise a user might think they can schedule as many as they want (I seem to remember at least one topic on DCEmulation where someone was trying to make one of these per controller or something before).

ljsebald (changes requested):

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -86,14 +136,15 @@ static void cont_reply(maple_frame_t *frm) {
...
+        if(btn_callback_thd && (thd_get_current()->tid != btn_callback_thd->tid)) {
             if(!btn_callback_addr ||
                     (btn_callback_addr &&
                      btn_callback_addr == maple_addr(frm->dev->port, frm->dev->unit))) {
                 if((cooked->buttons & btn_callback_btns) == btn_callback_btns) {
-                    btn_callback(maple_addr(frm->dev->port, frm->dev->unit),
-                                 cooked->buttons);
+                    btn_callback_arg_addr = maple_addr(frm->dev->port, frm->dev->unit);
+                    btn_callback_arg_btns = cooked->buttons;
+                    thd_schedule_next(btn_callback_thd);
```
ljsebald, line 147: Technically, this call here is fine... Since this is in an interrupt handler.

kernel/arch/dreamcast/include/dc/maple/controller.h
QuzarDC: Updated the notes on this with both parts. I forgot when I fast-forwarded that Falco had added in the notes about the interrupt context.

pcercuei:
My thoughts on this PR: It would be better to have a *generic* mechanism with an API that allows you to register a handler. The kernel would then create a thread for you, that would call your callback when notified to do so by the main IRQ handler (kind of a mutex+cond, but that would work in IRQ context).

The main IRQ handler would then ACK and mask the IRQ that fired, and unpause the interrupt thread. Only after the registered callback returns should the IRQ be unmasked, and the interrupt thread is back into the "paused" state, waiting for a new interrupt.

Then this PR would just be a matter of registering a button callback with this new threaded interrupt mechanism.

pcercuei:
> We'd still be blocked through by the process of creating non-running threads, which isn't currently supported, so I'll hold this open for feedback from @ljsebald on that for now while working on a draft of a generic version.

Don't we have "genwait" for that? We don't need to create a non-running thread, we can create a running thread that's blocked until the main IRQ handler unblocks it.

pcercuei:
RFC / untested: https://github.com/KallistiOS/KallistiOS/commit/5cdfbf3acf960bf802f08005527c15125dcb6503

QuzarDC:
Just going to close this up for now as it'd have to be rewritten mostly. Will take up the effort in v2.2.0 and see about expanding the functionality as well (allow installing multiple such callbacks. perhaps allow other filtering mechanisms).

PR #211, opened 2023-05-23
Make bin2o a little more robust.

jpeach:
- use a private directory for temporary files
- set shell error checking flags
- error on unsupported $KOS_ARCH settings

### DIFF
--- a/utils/bin2o/bin2o
+++ b/utils/bin2o/bin2o
@@ -1,47 +1,66 @@
-#!/bin/sh
+#! /usr/bin/env bash
 
 # bin2o (heh)
 # script to convert a binary data to an elf object file
 # (c)2000 Megan Potter
+
+# Unfortunately pipefail is not POSIX, see https://github.com/koalaman/shellcheck/issues/2555.
+set -o pipefail
+set -o nounset
+set -o errexit
+
+error() {
+        echo "bin2o:" "$@" 1>&2
+}
 
 if [ $# != 3 ]; then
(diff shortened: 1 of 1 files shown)


gyrovorbis:
Anyone else think we should go ahead and remove the options for PS2 and GBA while we're at it? I didn't even realize they were still here, and we don't have any support for them anymore in the source tree...

jpeach:
> 



> Anyone else think we should go ahead and remove the options for PS2 and GBA while we're at it? I didn't even realize they were still here, and we don't have any support for them anymore in the source tree...

Does that mean that $KOS_ARCH is also obsolete now?

gyrovorbis:
> Does that mean that $KOS_ARCH is also obsolete now?
Pretty sure it does.

But pinging the Crabby one and Quzar to have them weigh-in. 

@ljsebald @QuzarDC

QuzarDC:
The options for gba and ps2 were officially deprecated quite some time ago. I think the KOS_ARCH was left to keep the framework ready for additional architectures to be supported again.

I, for one, would love if ps2 or x86 were reintegrated.

ljsebald:
I'd prefer not removing `$KOS_ARCH` for the same reason that @QuzarDC said it wasn't removed for -- the framework is there to bring other arches in, but there was nobody there to maintain them. If someone really wanted to, then I'd like it to be still possible to do so.

I don't think it really needs to be removed from this tool at the moment either, so I'm going to merge this as-is.

ljsebald (approved):

PR #213, opened 2023-05-23
Broken Socket Write

QuzarDC:
This is a simple example that I had been discussing with others in the discord. It was an attempt to write a simple transmission speed test for testing bba changes. Unfortunately I found that it just didn't want to work. It seems that in this case, where there is a tight loop around a socket write, it very quickly breaks and will just loop forever.

It seems that this has to do with some sort of stack corruption that is happening due to something with the timing. If a sleep or printf is added into the write loop, then it works fine.

The idea of this is to make it visible publicly. If any solution is found, then the code change can be in this PR and it will retain the original sample that triggered it (I'm not much advocating to add the example to our provided examples, though perhaps it could be the basis for a regression test?).


ljsebald:
I think it is a good idea to have a testsuite at some point for regression testing... Definitely something we should discuss more.

pcercuei:
https://github.com/KallistiOS/KallistiOS/blob/master/kernel/arch/dreamcast/hardware/network/broadband_adapter.c#L520-L521

Not sure that's your problem (didn't test) but it doesn't look very good either.

QuzarDC:
> https://github.com/KallistiOS/KallistiOS/blob/master/kernel/arch/dreamcast/hardware/network/broadband_adapter.c#L520-L521
> 
> Not sure that's your problem (didn't test) but it doesn't look very good either.

Tested with that and it makes no difference, though it is clearly wrong. I'll put in a PR to fix that hole.

It seems that the behavior has also changed since the last time I had tested. Previously it would either give no output forever or give a general exception indicating stack corruption. Now it consistently gives the following:
```
All threads (may not be deterministic):
addr            tid     prio    flags   wait_timeout    state     name
00000000        0       0       00000000        0                   zombie
--end of list--
kernel panic: couldn't find a runnable thread
arch: aborting the system
```

This could merely be due to slight changes in the binary, but may be a consequence of some specific change.

andressbarajas (commented):
Missing some dashes in makefile

QuzarDC:
> Missing some dashes in makefile

Thanks. Not sure this would ever actually be committed though. The point is more to have a standing demo to get some community support to fixing the underlying problem.

examples/dreamcast/network/badsock/Makefile
```diff
@@ -0,0 +1,35 @@
...
+
+$(TARGET): $(OBJS)
+	kos-cc -o $(TARGET) $(OBJS)
+
+run: $(TARGET)
+	$(KOS_LOADER) $(TARGET)
+	
+dist: $(TARGET)
+	-rm -f $(OBJS)
+	$(KOS_STRIP) $(TARGET)
```
andressbarajas, line 35:
```suggestion
	$(KOS_STRIP) $(TARGET)

```
Adding a newline

examples/dreamcast/network/badsock/main.c
```diff
@@ -0,0 +1,350 @@
...
+        do_dirlist(buf, hs);
+    }
+    else {
+        /* Starting at the second character, try to read as a dec number and put the end to leftover */
+        unsigned int size = strtol(buf + 1, NULL, 10);
+        
+        /* Couldn't pull a number */
+        if(size == 0) size=1000;
+
+        unsigned int o = size;
```
andressbarajas, line 225:
```suggestion
       int o = size;
```
This needs to be an int because it will cause underflow issues with o -= r;

This still doesnt fix the infinite loop issue thought :(

andressbarajas:
@QuzarDC Okay, I just tested this against zcrcs mutex fix and everything seems to work now. So, make the underflow fix above and also merge in the latest master.  It should the issue!

QuzarDC:
Updated the example to avoid the underflow and performed various tests. Was unable to recreate the original issue. Don't think it's really worthwhile to add this as its own example since it's just a narrower version of httpd, so will just close it out.

Good work all, hunted down quite a few bugs on the way to the real solution.

PR #217, opened 2023-05-26
Better handle arch that isn't dreamcast

QuzarDC:
This is a series of changes that came amount from trying to reinstate the gba arch into kos. Since we haven't had any tested working alternative arches in almost 20 years, a number of things have drifted into the global that should be restricted to the arch. The changes include:

-Updating some net files to use _t types as they shouldn't need to be able to see the arch-specific others.
-Cleaning up kos/limits. Both removing a number of unnecessary references to it, def protecting the values in it, in case they are already provided by a standard limits, and renaming to match standard names.
-Moved the inclusion of rtc functions to not be arch_dreamcast only, as they are exported in the kernel exports and not the arch ones. It may be the case that this should go the other way, but I believe any future arch would likely have rtc.
-Moving lock.h for faked newlib locking into the standard fs. This way it's available even if newlib isn't being patched. Since it's protected from double-inclusion, this should be safe. Also removed it from the diff patch and have it just copying over.
-Reduced the scope of includes in kos/thread. It was previously including arch/arch and arch/types for no good reason. Unfortunately this had a cascading effect. A number of files and examples were not including arch/arch (which is needed for KOS_INIT stuff) and were just incidentally getting it via kos/thread. This may cause some new warnings/errors for people not using <kos.h>
-With the reduced scope of kos/thread, it's no longer necessary to copy the dc includes into newlib.
-Adjust newlib patch to disable syscalls via a flag rather than patching them out.

None of this should change any functionality of existing dreamcast stuff.


kernel/libc/pthreads/pthread_mutex.c
```diff
@@ -6,6 +6,8 @@
 #include <errno.h>
 #include <assert.h>
 
+#include <kos/mutex.h>
```
QuzarDC, line 9: This ensures access to MUTEX_TYPE_NORMAL which isn't directly in the other includes.

include/kos.h
```diff
@@ -66,11 +66,11 @@ __BEGIN_DECLS
 #include <arch/exec.h>
 #include <arch/stack.h>
 #include <arch/byteorder.h>
+#include <arch/rtc.h>
```
QuzarDC, line 69: This matches the definition in the exports files. The two exported rtc functions are present in the main exports instead of the arch_dreamcast specific one.

kernel/libc/koslib/inet_ntoa.c
```diff
@@ -20,7 +20,7 @@ char *inet_ntoa(struct in_addr addr) {
            values. This works, regardless of the endianness of the host system
            because the specs require the address passed in here to be in
            network byte order (big endian). */
-        part = ((uint8 *) &addr.s_addr)[i];
+        part = ((uint8_t *) &addr.s_addr)[i];
```
QuzarDC, line 23: These are very specifically because the 'uint8' is only defined in the arch/types, whereas _t are in the general types. Either way we are trying to move towards _t everywhere, to my understanding.

kernel/libc/pthreads/pthread_thd.c
```diff
@@ -2,6 +2,8 @@
 #include <errno.h>
 #include <assert.h>
 
+#include <kos/thread.h>
```
QuzarDC, line 5: This is to ensure that the kthread_t in pthread_create is defined.

examples/dreamcast/sd/speedtest/sd-speedtest.c
```diff
@@ -50,8 +50,8 @@ static void __attribute__((__noreturn__)) wait_exit() {
 
 int main(int argc, char *argv[]) {
     kos_blockdev_t sd_dev;
-    uint64 begin, end, timer, average;
-    uint64 sum = 0;
+    uint64_t begin, end, timer, average;
```
QuzarDC, line 53: Lets arch/types be removed, though we still have to add arch/arch for KOS_INIT.

kernel/arch/dreamcast/fs/fs_dcload.c
```diff
@@ -20,6 +20,7 @@ printf goes to the dc-tool console
 #include <dc/fs_dcload.h>
 #include <kos/thread.h>
 #include <arch/spinlock.h>
+#include <arch/arch.h>
```
QuzarDC, line 23: Needed for kos_init_flags. Previously it was accidentally chaining in from kos/thread

kernel/arch/dreamcast/include/arch/arch.h
```diff
@@ -21,7 +21,7 @@
 #include <kos/cdefs.h>
 __BEGIN_DECLS
 
-#include <dc/video.h>
```
QuzarDC, line 24: This was mysterious. All I can guess is that some time in the distant past it was needed. At this point it was just a roundabout way to get to arch/types.

kernel/libc/newlib/newlib_getentropy.c
```diff
@@ -10,6 +10,8 @@
 #include <unistd.h>
 #include <sys/time.h>
 
+#include <arch/arch.h>
```
QuzarDC, line 13: Needed for _arch_mem_top. It was somehow being pulled in before by one of the other includes.

utils/dc-chain/patches/newlib-4.3.0.20230120-kos.diff
```diff
@@ -243,374 +243,11 @@ diff --color -ruN newlib-4.3.0.20230120/newlib/libc/stdlib/assert.c newlib-4.3.0
  #ifndef HAVE_ASSERT_FUNC
  /* func can be NULL, in which case no function information is given.  */
  void
-@@ -72,3 +74,5 @@
+@@ -72,3 +74,7 @@
```
QuzarDC, line 246: This was in order to move the newlib patch extern out of syscalls.c, since that's no longer needed in the patch. Git's diff doesn't make it that clear.

utils/dc-chain/patches/newlib-4.3.0.20230120-kos.diff
```diff
@@ -243,374 +243,11 @@ diff --color -ruN newlib-4.3.0.20230120/newlib/libc/stdlib/assert.c newlib-4.3.0
...
     __assert_func (file, line, NULL, failedexpr);
    /* NOTREACHED */
  }
 +#endif
-+
-diff --color -ruN newlib-4.3.0.20230120/newlib/libc/sys/sh/ftruncate.c newlib-4.3.0.20230120-kos/newlib/libc/sys/sh/ftruncate.c
---- newlib-4.3.0.20230120/newlib/libc/sys/sh/ftruncate.c	2023-02-02 15:26:13.675093134 -0600
```
QuzarDC, line 253: This and all the syscalls are removed in favor of the disable syscalls flag.

utils/dc-chain/patches/newlib-4.3.0.20230120-kos.diff
```diff
@@ -243,374 +243,11 @@ diff --color -ruN newlib-4.3.0.20230120/newlib/libc/stdlib/assert.c newlib-4.3.0
...
--#include "sys/syscall.h"
-+/* Nothing here */
- 
--int
--ftruncate (int file, off_t length)
--{
--  return __trap34 (SYS_ftruncate, file, length, 0);
--}
-diff --color -ruN newlib-4.3.0.20230120/newlib/libc/sys/sh/sys/lock.h newlib-4.3.0.20230120-kos/newlib/libc/sys/sh/sys/lock.h
---- newlib-4.3.0.20230120/newlib/libc/sys/sh/sys/lock.h	1969-12-31 18:00:00.000000000 -0600
```
QuzarDC, line 267: This has been placed into sys/lock.h inside of KOS. The patch.mk file now copies it to this location, as it was already doing for other files.

utils/dc-chain/scripts/newlib.mk
```diff
@@ -43,21 +44,19 @@ fixup-sh4-newlib-apply:
...
-# arch/arch.h requires dc/video.h
+# kos/thread.h requires arch/irq.h
 	cp $(kos_base)/include/pthread.h $(newlib_inc)
 	cp $(kos_base)/include/sys/_pthread.h $(newlib_inc)/sys
 	cp $(kos_base)/include/sys/sched.h $(newlib_inc)/sys
+	cp $(kos_base)/include/sys/lock.h $(newlib_inc)/sys
 ifndef MINGW32
 	ln -nsf $(kos_base)/include/kos $(newlib_inc)
 	ln -nsf $(kos_base)/kernel/arch/dreamcast/include/arch $(newlib_inc)
-	ln -nsf $(kos_base)/kernel/arch/dreamcast/include/dc   $(newlib_inc)
```
QuzarDC, line 54: The dc folder is just no longer needed as kos/thread no longer chains into needed anything from there.

QuzarDC:
I added comments explaining individual changes not realizing that they'd all have to be individually clicked and resolved first. Hopefully that just helps make it more digestible.

ljsebald:
I feel like we should probably move the init flags stuff out of <arch/arch.h> and into its own header file... Something like <kos/init.h>, which could chain in a <arch/init_flags.h> file or something like that for arch-specific init stuff.

Thoughts?

QuzarDC:
> I feel like we should probably move the init flags stuff out of <arch/arch.h> and into its own header file... Something like <kos/init.h>, which could chain in a <arch/init_flags.h> file or something like that for arch-specific init stuff.
> 
> Thoughts?

I think that's more a symptom of a larger issue. Without other arches to live as examples, it's quite difficult to tell what should and shouldn't be at that level.

That being said, you're absolutely right. The majority of what it describes are things that are common to KOS outside of the DC-specific arch. And that split makes sense where we may want too, to define a split 'x bits for KOS x bits for the arch' or such. Since the names are defined in KOS there's no worry about moving around what the actual flags are.

I would ask though, that perhaps this become a new issue (unless you think the changes here interfere with it too significantly) to be addressed by a different PR. I'm guessing it was sparked through by how 'wrong' it feels that the examples get arch/arch included in them? But this was always happening, it was just being laundered by other headers.

ljsebald:
> > I feel like we should probably move the init flags stuff out of <arch/arch.h> and into its own header file... Something like <kos/init.h>, which could chain in a <arch/init_flags.h> file or something like that for arch-specific init stuff.
> > Thoughts?
> 
> I think that's more a symptom of a larger issue. Without other arches to live as examples, it's quite difficult to tell what should and shouldn't be at that level.
> 
> That being said, you're absolutely right. The majority of what it describes are things that are common to KOS outside of the DC-specific arch. And that split makes sense where we may want too, to define a split 'x bits for KOS x bits for the arch' or such. Since the names are defined in KOS there's no worry about moving around what the actual flags are.
> 
> I would ask though, that perhaps this become a new issue (unless you think the changes here interfere with it too significantly) to be addressed by a different PR. I'm guessing it was sparked through by how 'wrong' it feels that the examples get arch/arch included in them? But this was always happening, it was just being laundered by other headers.

It was indeed sparked by the examples getting <arch/arch.h> included in them... I guess we can merge this now, and one of us (possibly me later today) can throw together the new headers...

ljsebald (approved):

PR #219, opened 2023-05-26
Improve fipr() function

pcercuei:
Update the __fipr() macro to use better registers.

With this change, the inner body of the fipr() function passes from being 11 opcodes long, to being just 3 opcodes long.

See it in Compiler Explorer: https://godbolt.org/z/qj3P8x4ba


gyrovorbis:
Well I'll be damned! Love the fact you whipped out Compiler Explorer to prove it. ;)

ljsebald (approved):

PR #220, opened 2023-05-26
Clean up makefiles

andressbarajas:
Clean up example makefiles


gyrovorbis:
Good work! I was just noticing how inconsistent they were after adding the last example. lol.

QuzarDC:
So, perhaps a silly question, but ... is there a good reason we don't add a lot of these pieces into Makefile.rules? That way there's no worry about having 100 copies that are all the same.

andressbarajas:
I dont see any pieces that are 100% shared across all examples.

ljsebald:
I'd really rather that we didn't remove the `-` characters before the `rm` commands. That is there by design, and should stay.

andressbarajas:
I reverted the removal of the dashes and added them to rm in dist:

andressbarajas:
Also as a note:

Using -rm in a Makefile: When you use -rm instead of rm in a Makefile, any errors encountered during the removal process are ignored, and the make process continues even if the removal fails. This allows the build process to proceed without interruptions caused by non-critical removal errors.

examples/dreamcast/2ndmix/Makefile
```diff
@@ -13,10 +13,10 @@ KOS_ROMDISK_DIR = romdisk_boot
 include $(KOS_BASE)/Makefile.rules
 
 clean: rm-elf
-	rm -f $(OBJS)
+	-rm -f $(OBJS)
```
jpeach, line 16: Not really an issue, but if you like you could use the `$(RM)` variable here (it expands to `rm -f`).

https://www.gnu.org/software/make/manual/html_node/Implicit-Variables.html#index-RM

andressbarajas: I think we want to keep -rm instead of rm.

ljsebald (approved):

PR #225, opened 2023-05-28
Fix some warnings

andressbarajas:
This branch is to fix warning that popup when we remove -fno-builtin from environ_base.sh

### DIFF
--- a/addons/libkosext2fs/fs_ext2.c
+++ b/addons/libkosext2fs/fs_ext2.c
@@ -1584,7 +1584,7 @@ static int fs_ext2_symlink(vfs_handler_t *vfs, const char *path1,
     if(len < 60) {
         /* We can make a fast symlink. */
         strncpy((char *)inode->i_block, path1, 59);
-        ((char *)inode->i_block)[len - 1] = '\0';
+        ((char *)inode->i_block)[60] = '\0';
         inode->i_size = (uint32_t)len;
     }
     else {
--- a/addons/libkosext2fs/fs_ext2.c
+++ b/addons/libkosext2fs/fs_ext2.c
@@ -1584,7 +1584,7 @@ static int fs_ext2_symlink(vfs_handler_t *vfs, const char *path1,
     if(len < 60) {
         /* We can make a fast symlink. */
         strncpy((char *)inode->i_block, path1, 59);
-        ((char *)inode->i_block)[60] = '\0';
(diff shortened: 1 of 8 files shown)


addons/libkosext2fs/fs_ext2.c
```diff
@@ -1583,7 +1583,10 @@ static int fs_ext2_symlink(vfs_handler_t *vfs, const char *path1,
     /* Will the link fit in the inode? */
     if(len < 60) {
         /* We can make a fast symlink. */
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstringop-truncation"
         strncpy((char *)inode->i_block, path1, 60);
+#pragma GCC diagnostic pop
```
pcercuei, line 1589: So your "fix" is to hide the bug?

andressbarajas: Yes and no. There is another "fix" in this PR that does the same for VMU filename where there is no NULL terminator. I made the same "fix" here just to bring focus to the part of the code that generates the warning because I know BlueCrab ,  who wrote this code, would know if should suppress it or actually "fix" it.

kernel/fs/fs_utils.c
```diff
@@ -109,14 +114,14 @@ ssize_t fs_load(const char * src, void ** out_ptr) {
...
-    if(left > 0) {
-        *out_ptr = realloc(data, total);
-
-        if(*out_ptr == NULL)
-            *out_ptr = data;
+    if (left > 0) {
+        void * new_data = realloc(data, total);
+        if (new_data != NULL) {
+            data = new_data;
+        }
```
ljsebald, line 121: In addition to the below stylistic fixes, please move the declaration of `new_data` to the beginning of the function.

```suggestion
    if(left > 0) {
        void * new_data = realloc(data, total);
        if(new_data != NULL) {
            data = new_data;
        }
```

ljsebald:
I'm really not a big fan of having a bunch of these diagnostic pragmas all over the place. I'd really like to find a better way to solve the warnings than doing that, because it looks really messy to me...

andressbarajas:
I'll mark the two that I think are unavoidable and try and handle the rest using another method.

andressbarajas:
Removed just about all pragmas except for the export.

ljsebald (approved):

PR #238, opened 2023-06-04
Fix underscores missed in last dc-chain change

cepawiel:
I had some last minute changes from underscores to dashes that I left in my Staged changed and missed in PR #233


darcagn (approved):
Fixes the problem for me!

PR #240, opened 2023-06-04
Split GCC patching behavior into diff file and copying files from tree

darcagn:
As things currently stand, the KOS patches to GCC are large diff files containing multiple individual patches. These individual patches can be categorized into two groups:

- Patches that alter existing version-specific GCC files, including (e.g. in GCC 13.1.0):
```
gcc/config/sh/sh-c.cc
gcc/configure
libgcc/config/sh/t-sh
libgcc/configure
libobjc/configure
libobjc/Makefile.in
libstdc++-v3/config/cpu/sh/atomicity.h
libstdc++-v3/configure
```

- Patches that create new files or completely overwrite an existing file. These files are universal and not specific to the GCC version being patched. These include:
```
libgcc/config/sh/crt1.S
libgcc/config/sh/fake-kos.S
libgcc/config/sh/gthr-kos.h
```

This draft PR proposes reducing maintenance overhead by removing the latter files from the diff patching process and altering dc-chain script behavior to instead copy them over from the KOS tree. Changes to crt1.S (aka startup.s) in particular are relatively frequent, and as it stands, necessitate regenerating all the GCC patches when doing so. With this proposed change, changes may be made to startup.s in tree and the dc-chain patch will handle the rest with no additional actions required. Additionally, the remaining gcc-specific diff files will be much smaller and easier to manipulate (although it is much less likely to even need to do so after the change).

I'm making this a draft for two reasons:
- I put fake-kos and gthr-kos in `kernel/arch/dreamcast/kernel/` because that's where startup.s already resides. Is there a better location for these files (whether that be somewhere in the KOS tree or somewhere in the dc-chain directory?). Thoughts @ljsebald ?
- @cepawiel has been our resident toolchain make script expert lately. I'd like him to at least review and have an opportunity to comment on this patch before merging.


cepawiel:
Looks great, this should make it much easier to generate the patches! I think the messages about potentially requiring a toolchain rebuild is a really good idea to have in the files. The only thing I can think of is that `gthr-kos.h` and `fake-kos.s` could probably live outside the Dreamcast folder as I don't believe there's anything DC specific in there?

darcagn:
> Looks great, this should make it much easier to generate the patches! I think the messages about potentially requiring a toolchain rebuild is a really good idea to have in the files. The only thing I can think of is that `gthr-kos.h` and `fake-kos.s` could probably live outside the Dreamcast folder as I don't believe there's anything DC specific in there?

Thanks for the review! You're right, I hadn't considered that those files may be directly applicable to building GCC for other platforms. Maybe `kernel/arch/dreamcast/kernel` for startup.s, `kernel/thread` for gthr-kos and fake-kos? Let's see what @ljsebald thinks. They might also do well to have better filenames (while in the KOS tree that is, before they're copied over and named gthr-kos and fake-kos within GCC).

QuzarDC:
> Looks great, this should make it much easier to generate the patches! I think the messages about potentially requiring a toolchain rebuild is a really good idea to have in the files. The only thing I can think of is that `gthr-kos.h` and `fake-kos.s` could probably live outside the Dreamcast folder as I don't believe there's anything DC specific in there?

This was going to be my comment as well. I had similarly moved lock.h out of the newlib patch and into our main tree in #217 . When fiddling with the gba arch, I didn't consider these, but they look related to any number of warnings I got trying to build KOS with the DevKitPro toolchain. `fake-kos.s` looks like it should be convertible into a .h file that is platform independent, but I'm not familiar enough to understand it well.

ljsebald:
Well, `startup.s` has its original location in `kernel/arch/dreamcast/kernel/`, and I see no reason why it should move from there.

As for the other files (`gthr-kos.h` and `fake-kos.s`), my gut feeling on them is that since they're *really* related to the toolchain, they probably should live in the dc-chain directory somewhere, rather than in the kernel tree.

ljsebald (approved):

PR #242, opened 2023-06-06
Newlib lock.h fix

QuzarDC:
Whelp, I broke some stuff. lock.h was being copied into the wrong include folder (into our own instead of newlib's internal) and at the wrong time (after building newlib instead of before).

This creates a new target to match the gcc fixup to continue using lock.h from the KOS sources, instead of being diff patched in, but it gets fix-up'd (copied over).


darcagn (approved):
It works!

PR #243, opened 2023-06-06
Update older newlib patches to move lock.h, disable syscalls cleanly.

QuzarDC:
Bring the changes from #217 into older versions. Without this they may also throw warnings/errors about lock.h not being patchable.


QuzarDC:
I was able to properly build the stable toolchain in master without this without issue. I'll defer to @darcagn for being able to test them out, but I didn't see any issue with having the lock.h supplied both by the fixup copy and patch diff.

darcagn (approved):
4.7.4 and 9.3.0 now build without any errors and run pvrmark and its printfs properly without the issues I was seeing earlier.

PR #246, opened 2023-06-10
Defines for memory areas

QuzarDC:
The naming of these, placement of the new include, comments, etc will need adjustment, but I at least wanted to start the conversation on it. The idea is to just de-magic-fy the places where we force or strip away addressing of memory access.


kernel/arch/dreamcast/include/dc/memmap.h
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+/** \file   dc/memmap.h
+    \brief  Defines of basic memory map values.
+
+    Various addresses and masks are not tied to any single subsystem, but 
+	instead are generalized to the whole system.
+
+    \author Donald Haase
+*/
```
ljsebald, line 15:
```suggestion
/** \file   dc/memmap.h
    \brief  Constants for areas of the system memory map.

    Various addresses and masks are not tied to any single subsystem, but 
    instead are generalized to the whole system.

    \author Donald Haase
*/
```

ljsebald (changes requested):
I have a couple of suggestions for this PR...
1. I suggest moving the header to the include/arch directory and just calling it memory.h or mem.h (I prefer the former, but the latter is fine too).
2. Add defines for all of the areas of memory, even if we aren't using them right now. That would specifically mean adding defines for P0, P3, and U0. Also, this would mean adding a define for the P0/U0 size, since it is different.
3. Instead of `MEM_AREA_Px_MASK`, I'd suggest `MEM_AREA_Px_BASE`, since it more accurately reflects what the constants are.
4. Add the new header file to kos.h, even though it will get included through pvr.h.

QuzarDC:
> I have a couple of suggestions for this PR...
> 
>     1. I suggest moving the header to the include/arch directory and just calling it memory.h or mem.h (I prefer the former, but the latter is fine too).
> 
>     2. Add defines for all of the areas of memory, even if we aren't using them right now. That would specifically mean adding defines for P0, P3, and U0. Also, this would mean adding a define for the P0/U0 size, since it is different.
> 
>     3. Instead of `MEM_AREA_Px_MASK`, I'd suggest `MEM_AREA_Px_BASE`, since it more accurately reflects what the constants are.
> 
>     4. Add the new header file to kos.h, even though it will get included through pvr.h.

I'll take it back and polish it with all that. All makes sense. I was feeling a bit fuzzy on exactly how this might work out generally, but knew it would be useful for demystifying some of the DMA/SQ internals, so I really appreciate the feedback.

QuzarDC:
From the first one, the following changes have been made:
1) Moved and renamed the file as suggested.
2) Added to kos.h. I put it in the dreamcast-exclusive region as it seemed to be more in-line with the gdb and mmu headers that are there. I think it might be worth moving all the arch ones out of the dreamcast ifdef since it would seem the whole point is for those to be expected of any arch.
3) Renamed defines as directed.
4) Expanded the defines to cover the whole memory map.
5) Removed my use of defines in mmu.c. I don't believe I was accurately mapping the magic numbers there to the proper defines. For instance, 0x80000000 is used in mmu.c, seemingly to ensure that an address in P0/U0 is in use *not* to test against P1 (as I had been setting).

Of note/questions for review and discussion:
1) The comments may be a bit verbose for all the P4 stuff. It might also make sense to have their define names better note that they are all part of the P4 region or otherwise change the prefix on them.
2) A number of the P4 sub-range defines could be used in cache.s, startup.s, and itlb.s, but they'd have to become *.S and
3) With the rest of the P4 stuff, it may make sense to include all the SH control registers here as well. These are currently spread out in their individual sub-system headers (ubc.h, sq.h) or, more often, directly in the main code that is using them as magic numbers (ser_console.c, timer.c, init.c, mmu.c, irq.c, gdb_stub.c, g2bus.c, etc etc).
4) It may make sense to note this all as specifically SH4 memory defines? From what I understand none of these are dreamcast specific.

kernel/arch/dreamcast/include/arch/memory.h
```diff
@@ -0,0 +1,217 @@
...
+
+*/
+
+/** \file   arch/memory.h
+    \brief  Constants for areas of the system memory map.
+
+    Various addresses and masks are not tied to any single subsystem, but 
+    instead are generalized to the whole system.
+
+    These values are drawn from the Hitatchi SH7750 Series Hardware Manual rev 6.0
```
ljsebald, line 14:
```suggestion
    These values are drawn from the Hitatchi SH7750 Series Hardware Manual rev 6.0.
```

kernel/arch/dreamcast/include/arch/memory.h
```diff
@@ -0,0 +1,217 @@
...
+#define MEM_AREA_CACHE_MASK 0x1fffffff
+
+/** \brief U0 memory region (cachable).
+    \ingroup memory
+
+    This is the base user mode memory address. It is cacheable as determined 
+    by the WT bit of the cache control register. By default KOS sets this to 
+    copy-back mode.
+
+    KOS runs in priviledged mode, so this is here merely for completeness.
```
ljsebald, line 52:
```suggestion
    KOS runs in privileged mode, so this is here merely for completeness.
```

kernel/arch/dreamcast/include/arch/memory.h
```diff
@@ -0,0 +1,217 @@
...
+
+*/
+#define MEM_AREA_U0_BASE    0x00000000
+
+/** \brief P0 memory region (cachable).
+    \ingroup memory
+
+    This is the base priviledged mode memory address. It is cacheable as determined
```
ljsebald, line 60:
```suggestion
    This is the base privileged mode memory address. It is cacheable as determined 
```

kernel/arch/dreamcast/include/arch/memory.h
```diff
@@ -0,0 +1,217 @@
...
+    \defgroup p4mem
+    \ingroup memory
+
+    This offset maps to on-chip I/O channels.
+
+*/
+#define MEM_AREA_P4_BASE    0xe0000000
+
+/** \brief Store Queue (SQ) memory base.
+    \ingroup memory
```
ljsebald, line 106:
```suggestion
    \ingroup memory
    \ingroup p4mem
```

kernel/arch/dreamcast/include/arch/memory.h
```diff
@@ -0,0 +1,217 @@
...
+    \ingroup p4mem
+
+    This offset maps to the SQ memory region. RW to addresses from 
+    0xe0000000-0xe3ffffff follow SQ rules.
+
+    \see dc\sq.h
+
+*/
+#define MEM_AREA_SQ_BASE    0xe0000000
+
```
ljsebald, line 126: Why is this here twice?
```suggestion
```

ljsebald (changes requested):

ljsebald:
> 1. The comments may be a bit verbose for all the P4 stuff. It might also make sense to have their define names better note that they are all part of the P4 region or otherwise change the prefix on them.

Reasonable. If you'd like to adjust, feel free.

> 2. A number of the P4 sub-range defines could be used in cache.s, startup.s, and itlb.s, but they'd have to become *.S

Might be a good idea, but I'd say move it to another PR.

> 3. With the rest of the P4 stuff, it may make sense to include all the SH control registers here as well. These are currently spread out in their individual sub-system headers (ubc.h, sq.h) or, more often, directly in the main code that is using them as magic numbers (ser_console.c, timer.c, init.c, mmu.c, irq.c, gdb_stub.c, g2bus.c, etc etc).

Reasonable. If you want to tackle that, have at it. As with the above, probably would be better to roll that into another PR (or perhaps even multiple PRs, one per subsystem). It might be a good idea to keep things separated out as SH4-general and DC specific, perhaps in two separate files (which leads to the answer to the below question as well).

> 4. It may make sense to note this all as specifically SH4 memory defines? From what I understand none of these are dreamcast specific.

Seems reasonable as a comment at the top of the file, probably.

QuzarDC:
Ok. Hopefully the final commit of the PR.  After going over them a few times, figured to not bother renaming some of the longer defines. They're likely only ever going to get referenced each once (if that) in assembly files and clarity and specificity trumps 'ew, that name is so long'. 

Made some comment edits to reflect my point 4 above, then also updated the include-protection to match the current path.

ljsebald (approved):

PR #255, opened 2023-07-22
Add patches and update testing configuration for GCC 13.2.0

darcagn:
GCC 13.2.0 is scheduled to be released on July 27, 2023. The release candidate was tested and it performs as expected. 
Please review this patch in advance, but do not actually merge until the GCC 13.2.0 release announcement, when packages are uploaded.


ljsebald (approved):
Approved ahead of the release of GCC 13.2.0.

PR #256, opened 2023-07-30
Update binutils to 2.41 for SH and ARM toolchains for testing config

darcagn:
After upgrade, everything looks normal for SH toolchain. 
For ARM toolchain, a newly built `stream.drv` file has the same md5 as `stream.drv.prebuilt` so no surprises there.


darcagn:
Closing as this update is rolled up in my other PR.

PR #266, opened 2023-08-14
Implement POSIX clock_gettime() and friends

gyrovorbis:
So several people have been asking about `clock_gettime()` as the POSIX successor to `gettimeofday()` (which we support). I looked at the docs and didn't see any reason we couldn't support most of it, including even setting the clock... and we already support all of C, C++, and some of POSIX time()... I figured this might be useful for people porting and also this really does seem like the most accurate/desirable of all of the low-level time calls...

Anyway, double check what I've done here:

- CLOCK_REALTIME: getting the standard date/time as we do for `time()` and `timespec_get()` using the initial RTC time + our time-since-boot timer offset
- CLOCK_MONOTONIC: considering this the amount of time that has elapsed since KOS has been running, which means it returns the current time-since-boot 
- CLOCK_PROCESS_CPUTIME_ID: Since we only have one process in KOS, and that process started when KOS started, I'm making this return the same thing as CLOCK_MONOTONIC?
- CLOCK_THREAD_CPUTIME_ID: Okay, maybe we should do something different here... I was going to attempt to follow up this PR with implementing per-thread CPU timing, but in the meantime should we support this? I'm returning the same thing as CLOCK_THREAD_CPUTIME_ID, which... I guess isn't really true. 

References: https://linux.die.net/man/3/clock_gettime

### DIFF
--- a/include/kos/time.h
+++ b/include/kos/time.h
@@ -27,6 +27,8 @@ struct timespec;
 
 extern int timespec_get(struct timespec *ts, int base);
 
+#ifndef __STRICT_ANSI__
+
 #ifndef _POSIX_TIMERS
 #define _POSIX_TIMERS 1
 #endif 
@@ -47,6 +49,8 @@ extern int timespec_get(struct timespec *ts, int base);
 extern int clock_getres(__clockid_t clk_id, struct timespec *res);
 extern int clock_gettime(__clockid_t clk_id, struct timespec *tp);
 extern int clock_settime(__clockid_t clk_id, const struct timespec *tp);
+
+#endif /* !defined(__STRICT_ANSI__) */
 
 __END_DECLS
 
--- a/include/kos/time.h
+++ b/include/kos/time.h
@@ -27,6 +27,27 @@ struct timespec;
 
 extern int timespec_get(struct timespec *ts, int base);
 
+#ifndef _POSIX_TIMERS
+#define _POSIX_TIMERS 1
+#endif 
+#ifdef _POSIX_MONOTONIC_CLOCK
+#define _POSIX_MONOTONIC_CLOCK 1
+#endif
+#ifdef _POSIX_CPUTIME
+#define _POSIX_CPUTIME 1
+#endif
+#ifdef _POSIX_THREAD_CPU_TIME
+#define _POSIX_THREAD_CPUTIME 1
+#endif
+
+#define CLOCK_MONOTONIC          (CLOCK_REALTIME + 1)
+#define CLOCK_PROCESS_CPUTIME_ID (CLOCK_REALTIME + 2)
+#define CLOCK_THREAD_CPUTIME_ID  (CLOCK_REALTIME + 3)
+
+extern int clock_getres(__clockid_t clk_id, struct timespec *res);
+extern int clock_gettime(__clockid_t clk_id, struct timespec *tp);
+extern int clock_settime(__clockid_t clk_id, const struct timespec *tp);
+
 __END_DECLS
 
 #endif /* !defined(__STRICT_ANSI__) || (__STDC_VERSION__ >= 201112L) || (__cplusplus >= 201703L) */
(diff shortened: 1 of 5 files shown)


include/kos/time.h
```diff
@@ -27,6 +27,31 @@ struct timespec;
 
 extern int timespec_get(struct timespec *ts, int base);
 
+#ifndef __STRICT_ANSI__
```
ljsebald, line 30:
```suggestion
#if !defined(__STRICT_ANSI__) || (_POSIX_C_SOURCE >= 199309L)
```

include/kos/time.h
```diff
@@ -27,6 +27,31 @@ struct timespec;
...
+
+#define CLOCK_MONOTONIC          (CLOCK_REALTIME + 1)
+#define CLOCK_PROCESS_CPUTIME_ID (CLOCK_REALTIME + 2)
+#define CLOCK_THREAD_CPUTIME_ID  (CLOCK_REALTIME + 3)
+
+extern int clock_getres(__clockid_t clk_id, struct timespec *res);
+extern int clock_gettime(__clockid_t clk_id, struct timespec *tp);
+extern int clock_settime(__clockid_t clk_id, const struct timespec *tp);
+
+#endif /* !defined(__STRICT_ANSI__) */
```
ljsebald, line 53:
```suggestion
#endif /* !defined(__STRICT_ANSI__) || (_POSIX_C_SOURCE >= 199309L) */
```

kernel/libc/posix/clock_gettime.c
```diff
@@ -0,0 +1,67 @@
...
+
+int clock_gettime(clockid_t clk_id, struct timespec *ts) {
+    if(!ts) {
+        errno = EFAULT;
+        return -1;
+    }
+
+    switch(clk_id) {
+    case CLOCK_REALTIME:
+        return timespec_get(ts, TIME_UTC) == TIME_UTC? 0 : -1;
```
ljsebald, line 39:
```suggestion
        return timespec_get(ts, TIME_UTC) == TIME_UTC ? 0 : -1;
```

kernel/libc/posix/clock_gettime.c
```diff
@@ -0,0 +1,67 @@
...
+    case CLOCK_MONOTONIC:
+    case CLOCK_PROCESS_CPUTIME_ID:
+    case CLOCK_THREAD_CPUTIME_ID: {
+        uint32_t secs, msecs;
+        timer_ms_gettime(&secs, &msecs);
+        ts->tv_sec = secs;
+        ts->tv_nsec = msecs * 1000 * 1000;
+        return 0;
+    }
```
ljsebald, line 48: Please move these variables to the top of the function and get rid of this scope.

ljsebald (changes requested):
In addition to these few small things here, the indentation is not correct in the switches in clock_gettime.c. Cases should be indented in themselves (and their body indented in another level from there, of course). Also, there should be blank space between the end of one case's body and the beginning of the next case/default (i.e, after the return on line 24).

ljsebald (changes requested):
One more thing beyond the tiny spacing thing here... The kos/time.h file needs to be rearranged a bit, since right now the POSIX stuff is inside the C11 block. Please do the following things:
1. Move the include of <kos/cdefs.h> and the `__BEGIN_DECLS` outside of the `#if !defined(__STRICT_ANSI__)...` block (that is to say move lines 19-21 up to above the current line 17.
2. Move the ending `(__STRICT_ANSI__)...` guard from the bottom of the file (current line 57) up to right above where the new POSIX one was added (current line 30).

I think that's all that is needed from there, if I'm looking at this correctly.

gyrovorbis:
> In addition to these few small things here, the indentation is not correct in the switches in clock_gettime.c. Cases should be indented in themselves (and their body indented in another level from there, of course). Also, there should be blank space between the end of one case's body and the beginning of the next case/default (i.e, after the return on line 24).

Sounds good, and works in my test setup with the rearranged `#ifdef`. (I'm using a totally bastardized version of the C++ "simple clock" example which sources the date/time from clock_gettime()).

ljsebald (approved):

PR #268, opened 2023-08-14
Implemented posix_memalign() addition to stdlib.h

gyrovorbis:
This was another popular request for the sake of porting other codebases and compatibility with KOS. I modeled the way it's handled after how BlueCrab added the <kos/time.h> private header, only with <kos/stdlib.h>, which is where the function is declared on POSIX platforms.

In terms of implementation, it's a little bit of validation logic that then gets (size adjusted and) forwarded to C11's aligned_alloc(). Ran a bunch of tests locally on my Dreamcast, allocating weird sizes, 0 sizes, non-pow2 sizes, etc and everything looks good. Masked off respective low bits and ensured they were 0 for proper alignment.

### DIFF
--- a/include/kos/stdlib.h
+++ b/include/kos/stdlib.h
@@ -15,7 +15,8 @@
 #ifndef __KOS_STDLIB_H
 #define __KOS_STDLIB_H
 
-#if !defined(__STRICT_ANSI__)
+#if !defined(__STRICT_ANSI__) || (_POSIX_C_SOURCE >= 200112L) || \
+    (_XOPEN_SOURCE >= 600)
 
 #include <kos/cdefs.h>
 
@@ -25,5 +26,6 @@ extern int posix_memalign(void **memptr, size_t alignment, size_t size);
 
 __END_DECLS
 
-#endif /* !defined(__STRICT_ANSI__) */
+#endif /* !defined(__STRICT_ANSI__) || (_POSIX_C_SOURCE >= 200112L) ||
+          (_XOPEN_SOURCE >= 600) */
 #endif /* !__KOS_STDLIB_H */
--- a/include/kos/stdlib.h
+++ b/include/kos/stdlib.h
@@ -1,0 +1,30 @@
+/* KallistiOS ##version##
+
+   kos/time.h
+   Copyright (C) 2023 Falco Girgis
+*/
+
+/* 
+   Add select POSIX extensions to stdlib.h which are not present within Newlib.
+*/ 
+
+#ifndef _STDLIB_H_
+#error "Do not include this file directly. Use <stdlib.h> instead."
+#endif /* !_STDLIB_H_ */
+
+#ifndef __KOS_STDLIB_H
+#define __KOS_STDLIB_H
+
+#if !defined(__STRICT_ANSI__) || (__STDC_VERSION__ >= 201112L) || (__cplusplus >= 201703L)
+
+#include <kos/cdefs.h>
+
+__BEGIN_DECLS
+
+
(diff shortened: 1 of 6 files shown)


include/kos/stdlib.h
```diff
@@ -0,0 +1,30 @@
...
+*/ 
+
+#ifndef _STDLIB_H_
+#error "Do not include this file directly. Use <stdlib.h> instead."
+#endif /* !_STDLIB_H_ */
+
+#ifndef __KOS_STDLIB_H
+#define __KOS_STDLIB_H
+
+#if !defined(__STRICT_ANSI__) || (__STDC_VERSION__ >= 201112L) || (__cplusplus >= 201703L)
```
ljsebald, line 18: I think this might need to be adjusted, since this will make the symbol appear in strict C11/C++17 mode, which I don't think is correct.

gyrovorbis: Yeah, good point. So technically this is supposed to be available when `_POSIX_C_SOURCE >= 200112L`, but Newlib only has it defined up to 2009, and I didn't want to redefine it or anything since we don't actually support everything from 20011... So I was just going to enable the code whenever strict ANSI was not enabled.

gyrovorbis:
> Indentation appears to be incorrect in the posix_memalign.c file. Everything's indented 3 spaces instead of 4 as the rest of KOS is.
Ughgghgh I don't even know how Sublime text decided to change it only on that one file when my global settings are set to 4... Good catch. lol.

include/kos/stdlib.h
```diff
@@ -0,0 +1,29 @@
...
+
+#include <kos/cdefs.h>
+
+__BEGIN_DECLS
+
+extern int posix_memalign(void **memptr, size_t alignment, size_t size);
+
+__END_DECLS
+
+#endif /* !defined(__STRICT_ANSI__) */
```
ljsebald, line 28:
```suggestion
#if !defined(__STRICT_ANSI__) || (_POSIX_C_SOURCE >= 200112L) || \
    (_XOPEN_SOURCE >= 600)

#include <kos/cdefs.h>

__BEGIN_DECLS

extern int posix_memalign(void **memptr, size_t alignment, size_t size);

__END_DECLS

#endif /* !defined(__STRICT_ANSI__) || (_POSIX_C_SOURCE >= 200112L) ||
          (_XOPEN_SOURCE >= 600) */
```

kernel/libc/posix/posix_memalign.c
```diff
@@ -0,0 +1,43 @@
...
+    size_t new_size = size;
+
+    if(align_rem)
+        new_size += (alignment - align_rem);
+
+    return new_size;
+}
+
+int posix_memalign(void **memptr, size_t alignment, size_t size) {
+    assert(memptr);
```
ljsebald, line 27:
```suggestion
    if(!memptr) {
        return EFAULT;
    }
```

kernel/libc/posix/posix_memalign.c
```diff
@@ -0,0 +1,43 @@
...
+
+    if(!size) {
+        *memptr = NULL;
+        return 0;
+    }
+
+    size = aligned_size(size, alignment);
+    *memptr = aligned_alloc(alignment, size);
+
+    return *memptr? 0 : ENOMEM;
```
ljsebald, line 42:
```suggestion
    return *memptr ? 0 : ENOMEM;
```

ljsebald (changes requested):

ljsebald (approved):

PR #274, opened 2023-08-25
Add spinlock_trylock macro

ljsebald:
Add a `spinlock_trylock` macro to `<arch/spinlock.h>` that attempts to lock the lock, but doesn't spin if it cannot.

Note that this uses a GCC extension for implementing the macro. At some point I might reconsider and make it an inline function, but for now since everything else in that file is macros, I figured it should be kept that way.


gyrovorbis (approved):
Just wrote an example/test to flex it, and it's working great. Will add that as another PR.

PR #276, opened 2023-08-25
Simplify kthread_once_t into a simple variable rather than a struct

ljsebald:
Change the `kthread_once_t` type from a two element struct to a `volatile int`. There really isn't a good reason for keeping it as a two-element struct, as we don't actually gain anything useful from having two elements that we can't get from a single simple variable. Plus, this is more in-line with how this is handled on Linux, so it can't be all that wrong...

Also, this has a tiny update to the once_test example to ensure that we don't just happen to get the right answer by chance.


gyrovorbis (approved):
The `once` KOS example works fine with it. Then to do some extra testing, I modified it to do horrible things like made the thread entry-point recursive, the once function recursive, added sleeps in the once function, removed the spinlocks, hammered at the thread function 999 times from the main thread... Nothing I did could ever get the once function to be called multiple times. Looks solid. 

Screenshot from 2023-08-24 23-57-30

PR #281, opened 2023-08-28
Simplify the structures of some sync primitives.

ljsebald:
- Remove useless `initialized` members
- Renumber `MUTEX_TYPE_NORMAL` to 0, providing `MUTEX_TYPE_OLDNORMAL` as an alias with a value of 1 (the old value for it)
- Clean up some formatting issues


QuzarDC (approved):
Since you're actively working on these pieces, it doesn't matter much for this PR, but I'd suggest at some point more clearly marking that the 'dummy' in condvar (and possibly also the 'dynamic' in sem) is vestigial rather than the result of a hidden requirement. Stuff to dump out in KOS 3.0 or whatever.

PR #283, opened 2023-08-29
New pthreads implementation

ljsebald:
- Remove `kernel/libc/pthreads` (and its header file)
- Add new addon library `libpthread` with a much more complete/compliant implementation of pthreads.


QuzarDC:
> Is there anything you need from us or that we can offer to help with to move the pthreads along? We're totally actually using this PR for the Objective-C Foundation/GNUStep work, and it's working out for us.
> 
> EDIT: The least I can do is resolve the dumb merge conflicts for you. lol.

Perhaps a small factor was to possibly roll the kthreads update that might be needed for this https://github.com/KallistiOS/KallistiOS/pull/209#discussion_r1314068934 into here.

gyrovorbis:
Just leaving a note for the future, apparently obj.c ran into a missing pthread API function while continuing his work on the Objective-C foundation port. He needed `pthread_mutex_trylock()` and has temporarily worked around it. Not sure if this was just WIP or if it wasn't intended to be part of the scope, but wanted to make a note here.

addons/libpthread/pthread_mutex_trylock.c
```diff
@@ -0,0 +1,22 @@
...
+   Copyright (C) 2023 Lawrence Sebald
+
+*/
+
+#include "pthread-internal.h"
+#include <pthread.h>
+#include <errno.h>
+#include <kos/mutex.h>
+
+int pthread_mutex_lock(pthread_mutex_t *mutex) {
```
ljsebald, line 13:
```suggestion
int pthread_mutex_trylock(pthread_mutex_t *mutex) {
```

ljsebald:
> Just leaving a note for the future, apparently obj.c ran into a missing pthread API function while continuing his work on the Objective-C foundation port. He needed `pthread_mutex_trylock()` and has temporarily worked around it. Not sure if this was just WIP or if it wasn't intended to be part of the scope, but wanted to make a note here.

Well, that's definitely a mistake, as there is supposed to be a pthread_mutex_trylock function, but I accidentally just named it pthread_mutex_lock instead. I've gone ahead and fixed that issue.

gyrovorbis:
Just a heads up, I rebased back with master, fixed the CHANGELOG conflict plus one other dumb one with `pthread.h` which should've been automatic... and there was a teeeensy little build error that I fixed in `barrier.c`, and now everything works great regarding the barriers! Also reran your general tests for pthreads, and everything is looking good there too.

include/kos/barrier.h
```diff
@@ -0,0 +1,42 @@
...
+#ifndef __KOS_BARRIER_H
+#define __KOS_BARRIER_H
+
+#include <sys/cdefs.h>
+__BEGIN_DECLS
+
+/** \file   kos/barrier.h
+    \brief  Definitions thread barriers.
+
+    TODO: Write barrier description.
```
andressbarajas, line 16:
```suggestion
    Thread barriers are used to synchronize the progress of multiple threads. A 
    barrier causes threads to wait until a specified number of threads have reached
    a certain execution point, ensuring a consistent state across different 
    execution paths.

    The implementation uses a mutex and a condition variable to manage the 
    synchronization process. Threads increment a waiting counter upon reaching the 
    barrier and sleep on the condition variable if they are not the last expected 
    thread. When the final thread reaches the barrier, it triggers a broadcast on 
    the condition variable to wake up all waiting threads. The barrier can be reset 
    for reuse or destroyed as needed.

    This synchronization primitive is essential for scenarios in parallel 
    programming where tasks executed by multiple threads must reach a certain point 
    before any can proceed, ensuring data consistency and coordination among 
    threads.
```

ljsebald: Thanks for writing up that description. I'm gonna go ahead and add that in, but without the second paragraph, as I don't think the user needs to know necessarily how it is implemented internally (and that way we don't have to change that description if the internal implementation were to ever change).

ljsebald:
I'm un-drafting this now. I honestly am not sure why I haven't done so already at this point, beyond testing... Which has been extensively done by everyone else by this point, so I guess all is well...

ljsebald:
Closing since this was merged as #991.

PR #288, opened 2023-09-02
environ.sh.sample cleanup + KOS debugging + optimizations

gyrovorbis:
I'm ashamed to say that I've only just stumbled upon @ljsebald's beautiful `kos/include/kos/opts.h` header file, containing `#defines` for various compile-time configurable values in KOS. I realized that if I didn't know they existed, probably most people don't know either... So I thought it would be a good idea to expose this stuff as the `KOS_DEBUG` level within the environ.sample.sh which the user can easily configure.

While I was going about adding that, I realized we're probably overdue for a cleanup of the file in general. I got rid of a bunch of old comments that applied to ancient toolchains (pre GCC 4.7) that KOS doesn't support anymore. I added a few more configuration options and tried to mirror the way we handle dc-chain's config.mk. 

- Cleaned up environ.sh.sample to more closely reflect the model we're using for the dc-chain config files. 
     * more optimization flags added 
     * more detailed descriptions added 
     * legacy comments removed 
     * KOS debug levels exposed * KOS filesystem limits exposed
- Actually enabling KOS debug levels has exposed a bunch of warnings which I had to take care of in pvr_mem.c, pvr_mem_core.c, malloc.c
- opts.h was modified to only define default values if none were specified already

! - Everything tested and works great EXCEPT enabling KOS's malloc() verbose debugging hangs. The root cause still needs to be determined.

### DIFF
--- a/doc/environ.sh.sample
+++ b/doc/environ.sh.sample
@@ -1,69 +1,71 @@
-# KallistiOS environment variable settings
+# KallistiOS Environment Variable Settings
 #
-# This is a sample script. Configure to suit your setup. Some possible
-# alternatives for the values below are included as an example.
+# This is a sample script for configuring and customizing your
+# KOS buid environment. Modify it to suit your setup. Several 
+# settings are provided with alternative values or may be enabled
+# optionally. 
 #
-# This script should be sourced in your current shell environment (probably
-# by bashrc or something similar).
+# This script is typically sourced in your current shell environment 
+# (probably by .bashrc, .bash_profile, or something similar), so that
+# the KOS environment is set up automatically for each shell session.
 #
 
-# Build architecture. Set the major architecture you'll be building for.
+# Build Architecture 
+# Set the major architecture you'll be building for.
 # The only option here is "dreamcast" as of KOS 2.0.0.
 export KOS_ARCH="dreamcast"
 
-# Build sub-architecture. If you need a particular sub-architecture, then set
-# that here; otherwise use "pristine".
+# Build Sub-Architecture 
+# Defines the sub architecture your configuration
+# is targetting or uses an existing value that
+# can be set externally via your IDE.
+#
 # Possible subarch options include:
-#  "pristine" - a normal Dreamcast console or HKT-0120 devkit
-#  "naomi" - a NAOMI or NAOMI 2 arcade board
-# You can also pre-define it in eg the build config of your IDE
+#  "pristine" - Dreamcast console or HKT-0120 devkit (default)
+#  "naomi"    - NAOMI or NAOMI 2 arcade board
+# 
 if [ -z "${KOS_SUBARCH}" ] ; then
     export KOS_SUBARCH="pristine"
 else
     export KOS_SUBARCH
 fi
 
-# KOS main base path
+# KOS root path
 export KOS_BASE="/opt/toolchains/dc/kos"
+
+# KOS-ports path
 export KOS_PORTS="${KOS_BASE}/../kos-ports"
 
 # Make utility
 export KOS_MAKE="make"
-#export KOS_MAKE="gmake"
 
 # CMake toolchain
 export KOS_CMAKE_TOOLCHAIN="${KOS_BASE}/utils/cmake/dreamcast.toolchain.cmake"
 
-# Load utility
-export KOS_LOADER="dc-tool -x"				# dcload, preconfigured
-# export KOS_LOADER="dc-tool-ser -t /dev/ttyS0 -x"	# dcload-serial
-
 # Genromfs utility
 export KOS_GENROMFS="${KOS_BASE}/utils/genromfs/genromfs"
-#export KOS_GENROMFS="genromfs"
 
-# Compiler prefixes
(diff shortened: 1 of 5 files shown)


kernel/arch/dreamcast/hardware/pvr/pvr_mem.c
```diff
@@ -157,8 +157,8 @@ void pvr_mem_print_list(void) {
 
     printf("pvr_mem_print_list block list:\n");
     LIST_FOREACH(ctl, &block_list, list) {
-        printf("  unfreed block at %08lx size %d, allocated by thread %d/%08lx\n",
-               ctl->block, ctl->size, ctl->thread, ctl->addr);
+        printf("  unfreed block at %08lx size %lu, allocated by thread %d/%08lx\n",
+               (unsigned long)ctl->block, ctl->size, ctl->thread, (unsigned long)ctl->addr);
```
gyrovorbis, line 161: fit to 80 columns.

doc/environ.sh.sample
```diff
@@ -1,102 +1,137 @@
...
-# KOS main base path
+# KOS root path
 export KOS_BASE="/opt/toolchains/dc/kos"
+
+# KOS-ports path
 export KOS_PORTS="${KOS_BASE}/../kos-ports"
 
 # Make utility
 export KOS_MAKE="make"
-#export KOS_MAKE="gmake"
```
gyrovorbis, line 32: readd with note for BSD users

gyrovorbis:
Alrighty, so after talking to @ljsebald, the scope has slightly changed here:

- He didn't think exposing `KOS_DBG` and the filesystem default limits via the `environ.sh` script was a good idea. Removed those configuration options
- Went ahead and finalized all cleanup and documentation of every setting in `environ.sh`
- Readded the line for optionally using `gmake` as the make utility
- Went through `pvr_mem.c`, ensuring the 80 column character limit is respected
- Still got rid of a bunch of compiler warnings from using the non-default `KOS_DBG` levels
- The way that I've made these values only take effect if they weren't already defined will still allow a power user to set them via Makefile or their `environ.sh` without having to go in and manually modify KOS headers.

*** Since we are no longer exposing the `KOS_DBG` levels to the user, the fact that level 2 (with verbose malloc logging) crashes currently is not an issue with this PR and can be fixed later now that we know it's a problem. ***

doc/environ.sh.sample
```diff
@@ -1,102 +1,173 @@
-# KallistiOS environment variable settings
+# KallistiOS Environment Settings
 #
-# This is a sample script. Configure to suit your setup. Some possible
-# alternatives for the values below are included as an example.
+# This is a sample script for configuring and customizing your
+# KOS buid environment. Modify it to suit your setup. Several
```
sizious, line 4: Typo: buid => build

doc/environ.sh.sample
```diff
@@ -1,102 +1,173 @@
...
-# Build sub-architecture. If you need a particular sub-architecture, then set
-# that here; otherwise use "pristine".
-# Possible subarch options include:
-#  "pristine" - a normal Dreamcast console or HKT-0120 devkit
-#  "naomi" - a NAOMI or NAOMI 2 arcade board
-# You can also pre-define it in eg the build config of your IDE
+# Build Sub-Architecture 
+#
+# Defines the sub architecture your configuration
+# is targetting or uses an existing value that
```
sizious, line 23: Typo: targetting => targeting

doc/environ.sh.sample
```diff
@@ -1,102 +1,173 @@
...
 # NOTE: This also requires -fno-builtin be removed from KOS_CFLAGS to take effect!
+#
 # export KOS_CFLAGS="${KOS_CFLAGS} -ffast-math -ffp-contract=fast -mfsrra -mfsca"
 
-# Everything else is pretty much shared. If you want to configure compiler
-# options or other such things, look at this file.
+# Additional Tools Path
+#
+# If not already set, add "bin" directory to PATH variable, which is where
+# addititional tools such as dc-tool-ip and dc-tool-serial are typically
```
sizious, line 160: Typo: addititional => additional

doc/environ.sh.sample
```diff
@@ -1,102 +1,173 @@
...
+#
+export KOS_CFLAGS="${KOS_CFLAGS} -O2"
+
+# Additional Optimizations
+#
+# Uncomment this to enable what has been found emperically to be 
+# the optimal set of additional flags for release build performance 
+# on the current stable toolchain.
+#
+#export KOS_CFLAGS="${KOS_CFLAGS} -freorder-blocks-algorithm=simple -flto=auto"
```
darcagn, line 127: Since certain kos-ports and examples won't work with `-flto=auto` that should be mentioned here.

gyrovorbis: Good point!

doc/environ.sh.sample
gyrovorbis: Thank you so much for catching my dumb typos. Can't have them in such an important file as one of the first things the user sees. :)

ljsebald (approved):

PR #289, opened 2023-09-03
bba: don't shut it down if it hasnt been initialized

snickerbockers:
KOS crashes during shutdown if there isn't a BBA plugged in because bba_if.if_stop and bba_if.if_shutdown will both be NULL.

This is a problem if you don't have one plugged in and you're trying to debug using dcload-serial because it will reboot the dreamcast instead of returning to the bootloader.


gyrovorbis (approved):
Woooow. Good find!

ljsebald (approved):

PR #307, opened 2023-09-17
Vmu extra blocks

andressbarajas:
Add two functions to VMU API:

1. Check the enabled status of a VMUs extra blocks.
2. Enable/Disable the status of the VMUs extra blocks.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/vmu.c
+++ b/kernel/arch/dreamcast/hardware/maple/vmu.c
@@ -182,7 +182,7 @@ int vmu_has_extra_blocks(maple_device_t *dev) {
     return 0;
 }
 
-int vmu_use_extra_blocks(maple_device_t *dev, int enable) {
+int vmu_toggle_241_blocks(maple_device_t *dev, int enable) {
     vmu_root_t root;
 
     if(vmufs_root_read(dev, &root) < 0)
--- a/kernel/arch/dreamcast/hardware/maple/vmu.c
+++ b/kernel/arch/dreamcast/hardware/maple/vmu.c
@@ -179,9 +179,6 @@ int vmu_has_extra_blocks(maple_device_t *dev) {
     if(root.blk_cnt == 241)
         return 1;
 
-    if(vmufs_root_write(dev, &root) < 0)
-        return -1;
-
     return 0;
 }
 
--- a/kernel/arch/dreamcast/hardware/maple/vmu.c
+++ b/kernel/arch/dreamcast/hardware/maple/vmu.c
@@ -170,6 +170,35 @@ int vmu_get_buttons_enabled(void) {
     return !!vmu_drv.periodic;
 }
 
+int vmu_has_extra_blocks(maple_device_t *dev) {
+    vmu_root_t root;
+
+    if(vmufs_root_read(dev, &root) < 0)
+        return -1;
+
+    if(root.blk_cnt == 241)
+        return 1;
+
+    if(vmufs_root_write(dev, &root) < 0)
+        return -1;
+
+    return 0;
+}
+
+int vmu_use_extra_blocks(maple_device_t *dev, int enable) {
+    vmu_root_t root;
+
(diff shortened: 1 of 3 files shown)


gyrovorbis (dismissed):
Looks good. I'm very happy to see this, because there's only one mainstream tool out there that can unlock these blocks, and I'd like to see the KOS VMU stuff support all it has to offer. :)

gyrovorbis (approved):

kernel/arch/dreamcast/hardware/maple/vmu.c
```diff
@@ -170,6 +170,32 @@ int vmu_get_buttons_enabled(void) {
     return !!vmu_drv.periodic;
 }
 
+int vmu_has_extra_blocks(maple_device_t *dev) {
```
ljsebald, line 173:
```suggestion
int vmu_has_241_blocks(maple_device_t *dev) {
```

kernel/arch/dreamcast/hardware/maple/vmu.c
```diff
@@ -170,6 +170,32 @@ int vmu_get_buttons_enabled(void) {
...
+    if(vmufs_root_read(dev, &root) < 0)
+        return -1;
+
+    if(root.blk_cnt == 241)
+        return 1;
+
+    return 0;
+}
+
+int vmu_use_extra_blocks(maple_device_t *dev, int enable) {
```
ljsebald, line 185:
```suggestion
int vmu_toggle_241_blocks(maple_device_t *dev, int enable) {
```

kernel/arch/dreamcast/include/dc/maple/vmu.h
```diff
@@ -87,13 +87,47 @@ __BEGIN_DECLS
...
+
+    \param  dev             The device to check the status of.
+
+    \retval 1               On success: extra blocks are enabled
+    \retval 0               On success: extra blocks are disabled
+    \retval -1              On failure
+
+    \sa vmu_has_extra_blocks
+*/
+int vmu_has_extra_blocks(maple_device_t *dev);
```
ljsebald, line 104:
```suggestion
*/
int vmu_has_241_blocks(maple_device_t *dev);
```

kernel/arch/dreamcast/include/dc/maple/vmu.h
```diff
@@ -87,13 +87,47 @@ __BEGIN_DECLS
...
+
+    \param  dev             The device to enable/disable 41 blocks.
+    \param  enable          Values other than 0 enables. Equal to 0 disables.
+
+    \retval 0               On success
+    \retval -1              On failure
+
+    \sa vmu_use_extra_blocks
+*/
+int vmu_use_extra_blocks(maple_device_t *dev, int enable);
```
ljsebald, line 122:
```suggestion
*/
int vmu_toggle_241_blocks(maple_device_t *dev, int enable);
```

ljsebald (changes requested):
I think that it would be more appropriate to be more exacting in the names of these functions. Feel free to comment if anyone has a better idea than what I've suggested here...

gyrovorbis:
> I think that it would be more appropriate to be more exacting in the names of these functions. Feel free to comment if anyone has a better idea than what I've suggested here...

Actually, I think you might be right... Watch us find out we can do VMU sizes bigger than that (TapamN had the BIOS detecting 400 blocks), or watch us add something for VM2 that makes this naming become ambiguous... I think this is the best way to not have our API bite us in the ass at the end.

I also agree with the `vmu_use_extra_blocks()`, being changed to `vmu_toggle_extra_blocks()`, on account of the former sounding like it's only enabling them. I also use the "toggle" nomenclature in my codebase for when the API can either enable or disable something based on a bool, plus this is what Qt does at least.... lol.

ljsebald (approved):

megavolt85:
ladies and gentlemen, there is an excellent command MAPLE_COMMAND_GETMINFO, which provides all the necessary information about VMU, why these crutches?

PR #308, opened 2023-09-17
Basic reintroduction of gba port

QuzarDC:
Pulled the GBA arch files out of the last branch they existed in and did work to get them compiling. This is based around the toolchains provided by devkitpro. Got it to a point where it compiles but then won't link due to various issues (mostly friction between our newlib expectations and what's provided in devkitpro. Many of the fixes/changes to get this working more smoothly were already committed in #217 so this is largely the environs and somewhat updated stubs.

I'm not entirely sure I'll pick this back up again to completion but wanted to make the work available in case others did.

Since it might help out, this is a link to the last commit with the multiple arches: https://github.com/KallistiOS/KallistiOS/tree/2301649311ec1eeb4efad496d11d8829c80b7985


gyrovorbis:
I wonder if there are any interested parties within the GB community we can reach out to... tell them how much cool stuff the KOS kernel supports and will have across platforms, and see if they're interested in helping get it going again.

QuzarDC:
I think one of the bigger hurdles might be that KOS would just be too big. From my perspective, it's a good way to work towards getting the options in to slim it down on the DC, but I can imagine it being a huge pain. Might make sense instead to focus on the PS2 since that has more resources and whatnot.

But I did just reinstall devkitARM and will be taking a whack at it again. At the very least the current work won't be lost if someone wants to pick it up a long time from now. Maybe too, just the work towards integrating with the devkitpro series gives an inroad towards supporting GC/Wii/GP/etc

QuzarDC:
Rebased onto master and added the `gba/include/arch/kos.h` file to match our current usage. Would probably need a *lot* of other adjustments to match the current state of the arch API though

darcagn:
Yeah, it's probably not really worth updating/rebasing this PR. I'm actively playing with the GBA port, and I'll probably just end up recreating it new and merely reference this one if necessary, as there's really not much here anyway.

QuzarDC:
> Yeah, it's probably not really worth updating/rebasing this PR. I'm actively playing with the GBA port, and I'll probably just end up recreating it new and merely reference this one if necessary, as there's really not much here anyway.

Got it.

PR #313, opened 2023-09-23
Rework once controls, once again.

ljsebald:
1. Make them into a tristate value: not done, in progress, and done.
2. Use a condvar to wake up any threads that try to use the same once control before the first caller's function finishes.
3. Make sure another once control can be used while one's init routine is being run. Even if that is done in another thread that the init routine from the first is waiting on somehow (looking at you libstdc++).

### DIFF
--- a/kernel/thread/once.c
+++ b/kernel/thread/once.c
@@ -4,39 +4,66 @@
    Copyright (C) 2009, 2023 Lawrence Sebald
 */
 
-#include <malloc.h>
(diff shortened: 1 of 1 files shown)


gyrovorbis:
Just tested, and yep, libstdc++ is pleased! Looks good!
Screenshot from 2023-09-23 13-47-06

gyrovorbis (approved):

PR #314, opened 2023-09-25
LICENSE, README, COPYRIGHT File and Path Updates

gyrovorbis:
EDIT: This PR is being changed to address updating the LICENSE, README.kos, and copyright files, plus adjusting their file paths to be in the expected location (for example, the LICENSE file has to be in the root to show up on the right in GitHub).
 
- I've had a few people mention not seeing a license associated with the project, so I moved it into the root
- Removed a couple lines which no longer seemed relevant (pre-1.0 version of KOS mentioned). Updated the path to the README.kos file.

EDIT: Since it's not immediately obvious in the diff, and touching this file is a big deal for someone to do, here's what I removed. Lines 4-5:
```
NOTE: The licensing terms have CHANGED since 0.7! Please read over this
document again if you read the one for 0.7.
```


QuzarDC:
Some of the other doc files like README.KOS repeat the text here, and the other README files point to look here (didn't change the name, but I think they sort of presumed that this was there in the same folder). Not sure if it should just point to here or perhaps also be edited to reflect these edits.

Most of these docs files need updates really. There are carveouts in the licensing for lwip, for instance, and the licenses for some of the early ports is mentioned despite them now being more well described and separated out in their own repo.

gyrovorbis:
> Some of the other doc files like README.KOS repeat the text here, and the other README files point to look here (didn't change the name, but I think they sort of presumed that this was there in the same folder). Not sure if it should just point to here or perhaps also be edited to reflect these edits.
> 
> Most of these docs files need updates really. There are carveouts in the licensing for lwip, for instance, and the licenses for some of the early ports is mentioned despite them now being more well described and separated out in their own repo.

You're right. I was afraid to change too much in one commit, and wasn't sure if anyone agreed that this stuff needed attention... I'll take this PR back into draft and see about taking care of the rest of it as well... good point.

gyrovorbis:
@darcagn has showed me his WIP branch where he's addressing basically everything ever within the `doc` directory, going above and beyond what the scope of this change was. I'm closing this PR out, because he's doing a better job with it than I was.

PR #315, opened 2023-09-27
C++20 Concurrency Test/Example

gyrovorbis:
I spent a substantial amount of time testing and verifying concurrency in modern C++. Right now we have everything that modern C++ even supports tested and verified to be working. Considering we don't even have anything using modern C++ in KOS, and we found at least one issue with C++'s `std::async()` not working properly, I think it would be a good idea for me to KOS-ify the testbed I wrote as an example/verification check.

This way we have one single ELF we can run to validate the status of the entirety of concurrency in C++ after KOS threading, Newlib, or toolchain changes.


examples/dreamcast/cpp/concurrency/concurrency.cpp
```diff
@@ -0,0 +1,743 @@
...
+        std::cerr << "\n\n****** C++ Concurrency Test: FAILURE *****\n\t" 
+                  << except.what() << "\n" << std::endl;
+
+        return EXIT_FAILURE;
+    }
+
+    std::cout << "\n\n***** C++ Concurrency Test: SUCCESS *****\n" << std::endl;
+
+    return EXIT_SUCCESS;
+}
```
ljsebald, line 743:
```suggestion
}

```

PR #316, opened 2023-09-27
Updating Dynamic ELF Library/Module Loading

gyrovorbis:
I'm taking a look at dynamic library and module loading for the first time after several of us have taken an interest in it. The D language apparently uses this kind of stuff widely, other console homebrew communities such as N64 and PSP are doing it, and we have no examples or validation tests for ensuring we don't break anything... which is bad, because we know @DC-SWAT and DreamShell depend on it. I was going to address a bunch of library-related things at once with this PR.

- Clean up Doxygen modules and documentation: We're polluting the global Doxygen module namespace with a billion ELF-related things that should all be in a single module, and it's making the Doxygen modules harder to navigate
- Add missing ELF data: The GDC maintainer mentioned that we're missing some structures (program header) for managing dynamic ELFs, which are going to be needed later
- Clean up code: I noticed many things, such as a debug macro which should use variadics, inconsistent struct naming (using the `_t` suffix without typedefs), and extra places where error checking/validation should be happening but isn't. There's also leftovers from when KOS supported GP32 and GBA in there.
- Additional API support for doing things such as iterating over all currently loaded libraries
- Stop requiring KOS's library entry points (lib_get_name(), lib_get_version(), lib_open(), lib_close()). They are a great idea, and the way libraries work is very nice; however, we cannot expect every single dynamic library to have all of these functions, especially if the user is building some multi-level project  in which one dependency which isn't even their code is a dynamic library. This requirement for these methods is intrusive and isn't necessary. We can gracefully ignore these methods, by checking if they're NULL and ignoring them if they aren't present. This will make the system more flexible and still keep compatibility with the KOS-style versioned/named libraries.
- Dynamic TLS: Our TLS model right now necessitates that everything be statically linked. This is also the first step for supporting the full dynamic model; however, more work must be done to support TLS within dynamically loaded libraries.
- Provide an example: I have a feeling nobody is using this fantastic feature of KOS simply because there are no examples. We need an example of dynamically discovering and loading a set of libraries as "plugins" which all implement some common interface
- Provide a standard POSIX API around libraries: It would be very trivial to implement `dl_open()`, `dl_sym()`, and friends as very minimal wrappers over the native API. Several developers such as myself and some of the PSP guys I've been talking to would use this cross-platform API at the application layer to target multiple devices, including the Dreamcast. We might as well for compatibility reasons.

### DIFF
--- a/include/kos/elf.h
+++ b/include/kos/elf.h
@@ -1,12 +1,14 @@
 /* KallistiOS ##version##
 
    kos/elf.h
-   Copyright (C)2000,2001,2003 Megan Potter
+   Copyright (C) 2000, 2001, 2003 Megan Potter
+   Copyright (C) 2023 Falco Girgis
 
 */
 
 /** \file   kos/elf.h
     \brief  ELF binary loading support.
+    \ingroup elf
 
     This file contains the support functionality for loading ELF binaries in
     KOS. This includes the various header structures and whatnot that are used
@@ -15,6 +17,7 @@
     within KOS.
 
     \author Megan Potter
+    \author Falco Girgis
 */
 
 #ifndef __KOS_ELF_H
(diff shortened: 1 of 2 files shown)


gyrovorbis:
On second though, looking at @QuzarDC's GBA reinstatement PR, the removal of the platform-dependent `#ifdefs` around the ELF architecture/platform was a mistake. This was common code that actually could handle dynamic ELF loading on all 3 architectures. I'll readd.

QuzarDC:
> On second though, looking at @QuzarDC's GBA reinstatement PR, the removal of the platform-dependent `#ifdefs` around the ELF architecture/platform was a mistake. This was common code that actually could handle dynamic ELF loading on all 3 architectures. I'll readd.

Just to copy in the discussion from discord: even if not maintaining the actual values needed by the other archs, the setting to SH should be something wrapped with the arch ifdef. That way it ensures that we're not hardcoding for the SH outside of the dreamcast arch branch.

PR #318, opened 2023-10-01
la_init(): early exit if LAN adapter not detected

Tchan0:
Align LAN's la_init() with the BBA code:
- if not present, do not register, and do not read flashrom
- remove a 'no device detected' dbglog that would be reduntant


gyrovorbis:
Wow... good catch!

ljsebald (approved):

PR #320, opened 2023-10-06
Sq refresh plus speed

andressbarajas:
I hijacked https://github.com/KallistiOS/KallistiOS/pull/149 with @gyrovorbis blessing.  

Added:

1. Fast path to sq_cpy if void* src is 8-byte aligned.
 - Adds about ~3 instructions for 8-byte align check.
 - I saw a speed up of 1.5x at 64 bytes, 2.45x at 256 bytes
2. Added TapamN sq_cpy_pvr(). Also benefits from 1.
 - For comparison, pvr_txr_load(sq_cpy wrapper for pvr textures) vs sq_cpy_pvr for a 64 x 64 texture coped to the pvr.  sq_cpy_pvr was ~9+ times faster.  Im sure with bigger texture sizes the benefit would be even greater looking at the details of 1.
3. Added TapamN sq_set_pvr(). Uses the lower 16-bits  of c to set texture to (RGB565) color.
4. Created macros for the repetitive SQ setup stuff.
5. Started using __builtin_prefetch where I could.  It does change the order of the asm output to optimize it.
6. Removed duplicate code with sq_set* functions

Tagging @QuzarDC @DC-SWAT @Kazade who commented on @gyrovorbis PR.


kernel/arch/dreamcast/include/dc/sq.h
```diff
@@ -22,73 +36,178 @@
...
+
+    \param  dest            The address to begin clearing at (32-byte aligned).
+    \param  n               The number of bytes to clear (multiple of 32).
+*/
+void sq_clr(void *dest, int n);
+
+/** \brief   PVR register governing texture memory write data size
+    \ingroup store_queues
+*/
+#define SQ_PVR_LMMODE (*(volatile uint32_t *)(void *)0xa05f6884)
```
ljsebald, line 164: This probably has no good reason to be exposed publicly, especially not here (since it is not a store queue related register). Please move it to the .c file.

ljsebald:
I have some reservations about blindly setting the LMMODE register in the code that does it, as it may interact with ongoing PVR DMA transfers that might be going on in the background (which also manipulate that register)... Unlikely that someone would be using both DMA and the store queues at the same time? Sure...

kernel/arch/dreamcast/include/dc/sq.h
andressbarajas: Done

andressbarajas:
I was also thinking about the unlikeliness of that happening.  I added that to the warnings for sq_cpy_pvr() and sq_set_pvr().  I'll look into disallowing the use of the functions while a DMA transfer is going on.

andressbarajas:
@ljsebald Okay, I looked at pvr_dma.c and implemented the same approach they did.  We exit early with some debug logging and return the dest param.

ljsebald:
> @ljsebald Okay, I looked at pvr_dma.c and implemented the same approach they did. We exit early with some debug logging and return the dest param.

Thanks. I'll try to get this reviewed more thoroughly in the next few days.

DC-SWAT:
Slightly unobvious behavior in sq_cpy. It seems to be written that SQ will be used, but in fact FPU can be used.
Can the FPU be faster when copying to external memory like SPU or PVR?

andressbarajas:
I believe that is possible and Im looking into it now but it might be some time before I have an example. Im looking at g2bus stuff at the moment.

DC-SWAT:
> I believe that is possible and Im looking into it now but it might be some time before I have an example. Im looking at g2bus stuff at the moment.

Of course this is possible, but there is a performance question here.

andressbarajas:
I don't have an answer for you because I haven't tried it yet.

gyrovorbis:
> > I believe that is possible and Im looking into it now but it might be some time before I have an example. Im looking at g2bus stuff at the moment.
> 
> Of course this is possible, but there is a performance question here.

So he has profiled them individually and has found that the approach that uses double-precision floating point loads is far faster (given proper alignments), but you do bring up a good point in terms of side-effects and clobbering all of those registers... Would you prefer these be separate functions rather than one single function which can take two paths?

DC-SWAT:
> > > I believe that is possible and Im looking into it now but it might be some time before I have an example. Im looking at g2bus stuff at the moment.
> > 
> > 
> > Of course this is possible, but there is a performance question here.
> 
> So he has profiled them individually and has found that the approach that uses double-precision floating point loads is far faster (given proper alignments), but you do bring up a good point in terms of side-effects and clobbering all of those registers... Would you prefer these be separate functions rather than one single function which can take two paths?

Thank you, you answered my question. If this really works faster than SQ for SPU memory, then I will use it in the iso loader when emulating CDDA. This can be an alternative mode even to the two existing ones DMA and SQ.

As for separation, yes. It confuses me a little that the name of the function indicates that the SQ will be used, but in fact it may not use it at all.
I would rather create another function fpu_cpy() and combine them both into a third function like fast_cpy() or something.

andressbarajas:
The SQ is always used in both cases. The fpu is just used to prep the data faster before sending it off using the SQ. 

The memcpy_fast library that you use does the same thing. Should probably name that fpu_cpy instead.

DC-SWAT:
> > > > I believe that is possible and Im looking into it now but it might be some time before I have an example. Im looking at g2bus stuff at the moment.
> > > 
> > > 
> > > Of course this is possible, but there is a performance question here.
> > 
> > 
> > So he has profiled them individually and has found that the approach that uses double-precision floating point loads is far faster (given proper alignments), but you do bring up a good point in terms of side-effects and clobbering all of those registers... Would you prefer these be separate functions rather than one single function which can take two paths?
> 
> I think the confusion is that it makes it seem like SQs are not being used at all and somehow the FPU is being used to transfer the data entirely. I had a lot of trouble parsing it out because of my unfamiliarity with the style of asm as well.

Thats right! So sorry, I read this from smartphone first time and confused. Now I understand how it works, it's really good improvements!

andressbarajas:
I can see the misunderstanding. Its the pref on the [out] that fires off the store queue.  I added the `/* Fire off store queue */` comment to make it clear but I guess I could of made it more clear.

DC-SWAT:
This gave me the idea to combine the process of PCM channel splitting and SQ transmission with a FPU into one optimized process in CDDA playback emulation (SQ mode). Also minus one memory copying and allocation, niice. Why didn't I think of this before? :D

DC-SWAT:
Does need a change `fpscr` here to use paired single precision load/store mode for 64-bit transfering?

andressbarajas:
Yes, fschg does that for us.

DC-SWAT:
> Yes, fschg does that for us.

Oh, that's right. I just was confused by the method of switching this in memcpy.

andressbarajas:
That is correct but I think we are fine as long as we use m4-single-only

kernel/arch/dreamcast/include/dc/sq.h
```diff
@@ -22,73 +36,177 @@
...
+
+    This function is similar to sq_cpy(), but it has been
+    optimized for writing to a destination residing within VRAM.
+
+    \note
+    TapamN has reported over a 2x speedup versus the regular
+    sq_cpy() when using this function to write to VRAM.
+
+    \warning
+    !!! Not be used at the same time as PVR DMA transfers !!!
```
ljsebald, line 173:
```suggestion
    This function cannot be used at the same time as a PVR DMA transfer.
```

kernel/arch/dreamcast/include/dc/sq.h
```diff
@@ -22,73 +36,177 @@
...
+void * sq_cpy_pvr(void *dest, const void *src, int n);
+
+/** \brief   Set a block of PVR memory to a 16-bit value.
+    \ingroup store_queues
+
+    This function is similar to sq_set16(), but it has been
+    optimized for writing to a destination residing within VRAM.
+
+    \warning
+    !!! Not be used at the same time as PVR DMA transfers !!!
```
ljsebald, line 195:
```suggestion
    This function cannot be used at the same time as a PVR DMA transfer.
```

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -2,165 +2,180 @@
...
 
-    /* Fill both store queues with c */
-    d[0] = d[1] = d[2] = d[3] = d[4] = d[5] = d[6] = d[7] =
-        d[8] = d[9] = d[10] = d[11] = d[12] = d[13] = d[14] = d[15] = c;
+/* Copies n bytes from src to dest (in VRAM), dest must be 32-byte aligned */
+void * sq_cpy_pvr(void *dest, const void *src, int n) {
+    if(PVR_DMA_DEST != 0) {
+        dbglog(DBG_ERROR, "sq_cpy_pvr: Previous DMA has not finished\n");
+        errno = EINPROGRESS;
+        return dest;
```
ljsebald, line 150:
```suggestion
        return NULL;
```

I think just returning a NULL pointer here and not manipulating errno is a better idea.

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -2,165 +2,180 @@
...
+
+    return dest;
+}
+
+/* Fills n bytes at PVR dest with short c, dest must be 32-byte aligned */
+void * sq_set_pvr(void *dest, uint32_t c, int n) {
+    if(PVR_DMA_DEST != 0) {
+        dbglog(DBG_ERROR, "sq_set_pvr: Previous DMA has not finished\n");
+        errno = EINPROGRESS;
+        return dest;
```
ljsebald, line 169:
```suggestion
        return NULL;
```

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -2,165 +2,180 @@
...
    Copyright (C) 2001 Andrew Kieschnick
+   Copyright (C) 2023 Falco Girgis
+   Copyright (C) 2023 Andy Barajas
 */
 
-#include <arch/memory.h>
 #include <dc/sq.h>
+#include <kos/dbglog.h>
+
+#include <errno.h>
```
ljsebald, line 12:
```suggestion
```

ljsebald (changes requested):
A few minor cleanups I'd like to request...

kernel/arch/dreamcast/hardware/sq.c
andressbarajas: Sounds good.  I just used the same approach from here: https://github.com/KallistiOS/KallistiOS/blob/1526b24db29723634ae096a139f70a918eb428fd/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c#L88

ljsebald (approved):

PR #323, opened 2023-10-12
Update GNU configs before dc-chain build

cepawiel:
This change downloads the newest versions of config.guess & config.sub into the dc-chain directory and then uses those files to replace any found in the downloaded sources. This should fix issues when building older releases on newer machines that the old scripts wouldn't recognize properly. Specifically this addresses the issue with GCC4.7 failing to build on ARM64 without updated config scripts.

TODO:
- [ ] Make sure `find` command is available on all supported dc-chain systems


sizious:
I'm interested to contribute to this one, I have some changes that I want to implement in `dc-chain`. 😊

cepawiel:
@sizious Didn't see this until now, sorry about that. Feel free to push to my branch and or fork, whatever is easiest for you. I was having trouble testing the build on Windows which is why I still had this in draft. Let me know if you need any testing or further help!

darcagn:
These changes look good and so I resolved conflicts and marked ready for review so we can get movement on this PR.

QuzarDC (approved):
All ready. Just wanted to get an ok from @cepawiel since it had been left as a draft.

cepawiel:
No objections from me!

PR #326, opened 2023-10-18
Optimized separating stereo PCM 16-bit for streams

DC-SWAT:
I decided to share my optimized functions from DreamShell for separating stereo PCM 16-bit.
Also I reduce one memory copying for mono PCM.


kernel/arch/dreamcast/include/dc/sound/pcm_split.h
```diff
@@ -0,0 +1,40 @@
...
+
+/** \brief  Separetes stereo PCM samples to 2 mono channel.
+
+    All arguments must be 32-byte aligned.
+*/
+void snd_pcm16_split(uint32_t *data, uint32_t *left, uint32_t *right, uint32_t size);
+
+/** \brief  Separetes stereo PCM samples to 2 mono channel with SQ transfer.
+
+    All arguments must be 32-byte aligned.
```
gyrovorbis, line 32:
```suggestion
    Splits a buffer containing 2 interleaved channels of 16-bit PCM samples into 2 separate buffers of 16-bit PCM samples by using the store queues for data transfer.
    
    \warning 
    All arguments must be 32-byte aligned.
    
    \warning 
    The store queues must be configured for transferring to the left and right destination buffers beforehand (QACRO <= left, QACRO1 <= right). 
    
    \param data   Source buffer of interleaved stereo samples
    \param left   SQ-masked left destination buffer address
    \param right  SQ-masked right destination buffer address
    \param size   Size of the source buffer in bytes
    
    \sa snd_pcm16_split()
```

kernel/arch/dreamcast/sound/snd_stream.c
```diff
@@ -148,16 +151,13 @@ static void process_filters(snd_stream_hnd_t hnd, void **buffer, int *samplecnt)
...
-/* Performs stereo seperation for the two channels; this routine
-   has been optimized for the SH-4. */
 static void sep_data(void *buffer, int len, int stereo) {
     register int16  *bufsrc, *bufdst;
     register int    x, y, cnt;
 
     if(stereo) {
         bufsrc = (int16*)buffer;
-        bufdst = sep_buffer[0];
+        bufdst = (int16*)sep_buffer[0];
```
gyrovorbis, line 160:
```suggestion
        bufdst = (int16 *)sep_buffer[0];
```
@ljsebald always makes me change that. :)

kernel/arch/dreamcast/sound/snd_stream.c
```diff
@@ -172,7 +172,7 @@ static void sep_data(void *buffer, int len, int stereo) {
 
         bufsrc = (int16*)buffer;
         bufsrc++;
-        bufdst = sep_buffer[1];
+        bufdst = (int16*)sep_buffer[1];
```
gyrovorbis, line 175:
```suggestion
        bufdst = (int16 *)sep_buffer[1];
```

gyrovorbis (commented):
Looking good. Thanks for this!

kernel/arch/dreamcast/sound/snd_stream.c
DC-SWAT: I always write like this by the way. But here it was different; I followed the style of the environment.

kernel/arch/dreamcast/include/dc/sound/pcm_split.h
```diff
@@ -0,0 +1,64 @@
...
+
+    Splits a buffer containing 2 interleaved channels of 16-bit PCM samples into 2 separate buffers of 16-bit PCM samples.
+
+    \warning 
+    All arguments must be 32-byte aligned.
+    
+    \param data   Source buffer of interleaved stereo samples
+    \param left   Destination buffer for left mono samples
+    \param right  Destination buffer for right mono samples
+    \param size   Size of the source buffer in bytes
```
andressbarajas, line 35:
```suggestion
    \param size   Size of the source buffer in bytes (must be an even number)
```

kernel/arch/dreamcast/include/dc/sound/pcm_split.h
```diff
@@ -0,0 +1,64 @@
...
+    \warning 
+    All arguments must be 32-byte aligned.
+    
+    \warning 
+    The store queues must be configured for transferring to the left and right destination buffers beforehand (QACRO <= left, QACRO1 <= right). 
+    
+    \param data   Source buffer of interleaved stereo samples
+    \param left   SQ-masked left destination buffer address
+    \param right  SQ-masked right destination buffer address
+    \param size   Size of the source buffer in bytes
```
andressbarajas, line 54:
```suggestion
    \param size   Size of the source buffer in bytes (must be an even number)
```

kernel/arch/dreamcast/include/dc/sound/pcm_split.h
gyrovorbis: Don't they also have to be divisible by 4? Since it's 2x interleaved 16-bit samples?

andressbarajas: Actually it should probably a multiple of 32 since the asm code divides the number of bytes by 32.

andressbarajas:
```suggestion
    \param size   Size of the source buffer in bytes (must be divisible by 32)
```

kernel/arch/dreamcast/include/dc/sound/pcm_split.h
andressbarajas:
```suggestion
    \param size   Size of the source buffer in bytes (must be divisible by 32)
```

DC-SWAT: Yep, as I mark - "All arguments must be 32-byte aligned".

ljsebald:
I'm curious why we're exposing what definitely seems to be internal functions into a new public header with the two pcm_split functions...

If there's not a good reason to make them public, then I'd prefer if we didn't do so (that is to say, move the header file out of the arch/dreamcast/include/dc/sound directory and instead put it in with the code in arch/dreamcast/sound, and removing them from the exports). If there is a good reason to expose them publicly, then I suggest that they be added into the sound/sound.h header, rather than making a new header file just for them.

DC-SWAT:
> I'm curious why we're exposing what definitely seems to be internal functions into a new public header with the two pcm_split functions...
> 
> If there's not a good reason to make them public, then I'd prefer if we didn't do so (that is to say, move the header file out of the arch/dreamcast/include/dc/sound directory and instead put it in with the code in arch/dreamcast/sound, and removing them from the exports). If there is a good reason to expose them publicly, then I suggest that they be added into the sound/sound.h header, rather than making a new header file just for them.
> 
> I've missed most of the discussion lately on Discord, so I apologize if this has been discussed there...

I know that there is a lot of code that uses AICA outside of KOS streams, so I wanted to make this functions public. But I agree that a separate header is probably not needed for this.

ljsebald (approved):

PR #327, opened 2023-10-20
SH4 cache improvements

DC-SWAT:
Added purge the entire data/operand cache based on TapamN's idea.
Instead of processing a large amount of data, we can re-allocate the entire cache (max 16KB), which will lead to a purge of what was in it. The only requirements is that we need a buffer for 16 KB.
Also I removed prefetch a range (is not an efficient), update block prefetch and add block allocation.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/cache.h
+++ b/kernel/arch/dreamcast/include/arch/cache.h
@@ -87,32 +87,34 @@ void dcache_purge_range(uint32 start, uint32 count);
 */
 void dcache_purge_all(uint32 start, uint32 count);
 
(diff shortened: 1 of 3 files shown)


gyrovorbis (approved):
Looks great to me. Code and Doxygen look good. Happy to have something like this, since I see the SH-Linux guys have some sort of purge all as well (not this nice, though).

Going to leave it for @ljsebald to also look at.

ljsebald (approved):
While I don't normally like to approve PRs that outright remove functionality, this is new enough and had a niche enough use-case that I don't see the harm in doing so.

PR #333, opened 2023-10-28
PVR: Add option to specify additional OPBs for TA overflow

kilgariff:
# OPB Overflow Support

Adds a new parameter to `pvr_init_params_t` that allows the user to specify a number of "overflow OPBs". These are additional OPBs preallocated so that the TA can use them if too much geometry is sent to fit in one OPB. This is a fix for the situation where rendered output has some geometry missing or flickering in certain tiles.

The direction new OPBs are added by the TA has been modified. Previously it used descending addresses (growing down) and now it uses ascending addresses (growing up). This is for consistency with the vertex buffer allocation and to prevent confusion for future maintainers.

_Note_: I added a new parameter to the end of `pvr_init_params_t` so that existing code using aggregate initialization wouldn't break (it would become 0). Unfortunately if existing code were to initialize each member separately, it could still break if they don't first memset to 0 as it would leave the new member uninitialized.

## VBlank Interrupt Swapped

The wrong interrupt was being used to signal VBlank. The HBlank interrupt was used, but it had been configured to only fire on line zero making it work effectively like VBlank. This has been swapped for the actual VBlank interrupt.

## PVR Error Codes Added

The PVR generates error interrupts that can signal buffer overflows and similar. These have been added along with an `#ifdef` guard allowing  the user to define `DBG_PVR_ERRORS` if they want to print errors from the PVR interrupt handler routine.

## Sources for Register and Interrupt Definitions

HBlank interrupt control register behaviour:
https://github.com/flyinghead/flycast/blob/c1f0a5a15f07ba82ccb974e5c7dc82206324b0a3/core/hw/pvr/pvr_regs.cpp#L208

Interrupt definitions:
https://github.com/flyinghead/flycast/blob/c1f0a5a15f07ba82ccb974e5c7dc82206324b0a3/core/types.h#L46

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
@@ -191,7 +191,16 @@ void pvr_allocate_buffers(pvr_init_params_t *params) {
     /* Look at active lists and figure out how much to allocate
        for each poly type */
     opb_total_size = 0;
+
+    /**
+     * Previously, we specified 1 << 20 to say that the OPB grows "down" when the TA needs more than one.
+     * To make it grow "up" instead (increasing addresses), we set 0 as the default value.
+    */
+#if 0
     pvr_state.list_reg_mask = 1 << 20;
+#else
+    pvr_state.list_reg_mask = 0;
+#endif
 
     for(i = 0; i < PVR_OPB_COUNT; i++) {
         pvr_state.opb_size[i] = WORDS_TO_BYTES(params->opb_sizes[i]);   /* in bytes */
@@ -245,10 +254,13 @@ void pvr_allocate_buffers(pvr_init_params_t *params) {
         /* N-byte align */
         outaddr = APPLY_ALIGNMENT(outaddr);
 
-        /* Object Pointer Buffers */
+        /* Object Pointer Blocks */
         buf->opb = outaddr;
         buf->opb_size = opb_total_size;
-        outaddr += opb_total_size;
+
+        /* Allocate extra space for overflow (when one OPB isn't big enough) */
+        buf->opb_overflow_count = params->opb_overflow_count;
+        outaddr += opb_total_size * buf->opb_overflow_count;
 
         /* Set up the opb pointers to each section */
         opb_size_accum = 0;
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
@@ -260,7 +260,7 @@ void pvr_allocate_buffers(pvr_init_params_t *params) {
 
         /* Allocate extra space for overflow (when one OPB isn't big enough) */
         buf->opb_overflow_count = params->opb_overflow_count;
-        outaddr += opb_total_size * buf->opb_overflow_count;
+        outaddr += opb_total_size * (1 + buf->opb_overflow_count);
 
         /* Set up the opb pointers to each section */
         opb_size_accum = 0;
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
@@ -192,10 +192,10 @@ void pvr_allocate_buffers(pvr_init_params_t *params) {
        for each poly type */
     opb_total_size = 0;
 
-    /**
-     * Previously, we specified 1 << 20 to say that the OPB grows "down" when the TA needs more than one.
-     * To make it grow "up" instead (increasing addresses), we set 0 as the default value.
-    */
+    /* Previously, we specified 1 << 20 to say that the OPB grows "down" when
+       the TA needs more than one. To make it grow "up" instead (increasing addresses),
+       we set 0 as the default value.
+     */
 #if 0
     pvr_state.list_reg_mask = 1 << 20;
 #else
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_init_shutdown.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_init_shutdown.c
@@ -154,7 +154,7 @@ int pvr_init(pvr_init_params_t *params) {
     asic_evt_set_handler(ASIC_EVT_PVR_RENDERDONE_TSP, pvr_int_handler);
     asic_evt_enable(ASIC_EVT_PVR_RENDERDONE_TSP, ASIC_IRQ_DEFAULT);
 
-#ifdef DBG_PVR_ERRORS
+#ifdef PVR_RENDER_DBG
     /* Hook up interrupt handlers for error events */
     asic_evt_set_handler(ASIC_EVT_PVR_ISP_OUTOFMEM, pvr_int_handler);
     asic_evt_enable(ASIC_EVT_PVR_ISP_OUTOFMEM, ASIC_IRQ_DEFAULT);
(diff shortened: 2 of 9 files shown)


kernel/arch/dreamcast/include/dc/asic.h
```diff
@@ -40,18 +40,33 @@ __BEGIN_DECLS
...
+#define ASIC_EVT_PVR_OPAQUEDONE           0x0007  /**< \brief Opaque list completed */
+#define ASIC_EVT_PVR_OPAQUEMODDONE        0x0008  /**< \brief Opaque modifiers completed */
+#define ASIC_EVT_PVR_TRANSDONE            0x0009  /**< \brief Transparent list completed */
+#define ASIC_EVT_PVR_TRANSMODDONE         0x000a  /**< \brief Transparent modifiers completed */
+
+#define ASIC_EVT_PVR_DMA                  0x0013  /**< \brief PVR DMA complete */
+#define ASIC_EVT_PVR_PTDONE               0x0015  /**< \brief Punch-thrus completed */
+
+/* Uncomment to print a message if an error is received in the PVR interrupt handler */
+/* #define DBG_PVR_ERRORS */
```
ljsebald, line 61: This block should probably go in `include/kos/opts.h`. Also, for uniformity, the macro should probably be called `PVR_RENDER_DBG`. This will also have to be accounted for in pvr_irq.c.

kernel/arch/dreamcast/include/dc/asic.h
```diff
@@ -40,18 +40,33 @@ __BEGIN_DECLS
...
-#define ASIC_EVT_PVR_PTDONE         0x0015  /**< \brief Punch-thrus completed */
-#define ASIC_EVT_PVR_PRIMOUTOFMEM   0x0202  /**< \brief Out of primitive memory */
-#define ASIC_EVT_PVR_MATOUTOFMEM    0x0203  /**< \brief Out of matrix memory */
+
+#define ASIC_EVT_PVR_RENDERDONE_VIDEO     0x0000  /**< \brief Video render stage completed */
+#define ASIC_EVT_PVR_RENDERDONE_ISP       0x0001  /**< \brief ISP render stage completed */
+#define ASIC_EVT_PVR_RENDERDONE_TSP       0x0002  /**< \brief TSP render stage completed */
+#define ASIC_EVT_PVR_VBLANK_BEG           0x0003  /**< \brief VBLANK begin interrupt */
+#define ASIC_EVT_PVR_VBLANK_END           0x0004  /**< \brief VBLANK end interrupt */
+#define ASIC_EVT_PVR_HBLANK_BEG           0x0005  /**< \brief HBLANK begin interrupt */
```
ljsebald, line 49: I'd prefer if we didn't abbreviate begin here... This will also have to be accounted for in pvr_irq.c and vblank.c.
```suggestion
#define ASIC_EVT_PVR_VBLANK_BEGIN         0x0003  /**< \brief VBLANK begin interrupt */
#define ASIC_EVT_PVR_VBLANK_END           0x0004  /**< \brief VBLANK end interrupt */
#define ASIC_EVT_PVR_HBLANK_BEGIN         0x0005  /**< \brief HBLANK begin interrupt */
```

kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
```diff
@@ -96,17 +100,46 @@ void pvr_int_handler(uint32 code) {
...
             pvr_state.render_completed = 1;
             pvr_sync_stats(PVR_SYNC_RNDDONE);
             break;
-        case ASIC_EVT_PVR_VBLINT:
+        case ASIC_EVT_PVR_VBLANK_BEG:
             pvr_sync_stats(PVR_SYNC_VBLANK);
             break;
     }
 
+#ifdef DBG_PVR_ERRORS
```
ljsebald, line 114:
```suggestion
#ifdef PVR_RENDER_DBG
```

kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
```diff
@@ -96,17 +100,46 @@ void pvr_int_handler(uint32 code) {
...
             break;
-        case ASIC_EVT_PVR_RENDERDONE:
+        case ASIC_EVT_PVR_RENDERDONE_TSP:
             //DBG(("irq_renderdone\n"));
             pvr_state.render_busy = 0;
             pvr_state.render_completed = 1;
             pvr_sync_stats(PVR_SYNC_RNDDONE);
             break;
-        case ASIC_EVT_PVR_VBLINT:
+        case ASIC_EVT_PVR_VBLANK_BEG:
```
ljsebald, line 109:
```suggestion
        case ASIC_EVT_PVR_VBLANK_BEGIN:
```

kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
```diff
@@ -124,7 +157,7 @@ void pvr_int_handler(uint32 code) {
 
     if(!pvr_state.to_texture[bufn]) {
         // If it's not a vblank, ignore the rest of this for now.
-        if(code != ASIC_EVT_PVR_VBLINT)
+        if(code != ASIC_EVT_PVR_VBLANK_BEG)
```
ljsebald, line 160:
```suggestion
        if(code != ASIC_EVT_PVR_VBLANK_BEGIN)
```

kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
```diff
@@ -12,6 +12,10 @@
 #include <arch/cache.h>
 #include "pvr_internal.h"
 
+#ifdef DBG_PVR_ERRORS
```
ljsebald, line 15:
```suggestion
#include <kos/opts.h>
#include "pvr_internal.h"

#ifdef PVR_RENDER_DBG
```

kernel/arch/dreamcast/hardware/pvr/pvr_init_shutdown.c
```diff
@@ -147,8 +151,22 @@ int pvr_init(pvr_init_params_t *params) {
     asic_evt_enable(ASIC_EVT_PVR_TRANSMODDONE, ASIC_IRQ_DEFAULT);
     asic_evt_set_handler(ASIC_EVT_PVR_PTDONE, pvr_int_handler);
     asic_evt_enable(ASIC_EVT_PVR_PTDONE, ASIC_IRQ_DEFAULT);
-    asic_evt_set_handler(ASIC_EVT_PVR_RENDERDONE, pvr_int_handler);
-    asic_evt_enable(ASIC_EVT_PVR_RENDERDONE, ASIC_IRQ_DEFAULT);
+    asic_evt_set_handler(ASIC_EVT_PVR_RENDERDONE_TSP, pvr_int_handler);
+    asic_evt_enable(ASIC_EVT_PVR_RENDERDONE_TSP, ASIC_IRQ_DEFAULT);
+
+#ifdef DBG_PVR_ERRORS
```
ljsebald, line 157:
```suggestion
#ifdef PVR_RENDER_DBG
```

kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
```diff
@@ -191,7 +191,16 @@ void pvr_allocate_buffers(pvr_init_params_t *params) {
     /* Look at active lists and figure out how much to allocate
        for each poly type */
     opb_total_size = 0;
+
+    /**
```
ljsebald, line 195: I don't think this is meant to be a doxygen comment...

```suggestion
    /*
```

ljsebald (changes requested):

ljsebald (approved):

PR #335, opened 2023-11-02
Modularizing KOS via KOS_INIT_FLAGS()

gyrovorbis:
Following zcrc's established mechanism for using KOS_INIT_FLAGS() to prevent unused drivers from getting linked into the final binary, I've broken out the maple drivers into individual flags as well.

So what this PR entails:
- Addition of `kos/init_base.h` which has common `#define`s for `KOS_INIT_FLAG` management and utilities which are used in both architecture-specific and independent init headers.
- Updated `cdefs.h` to include some other useful attributes like `__weak`, `__likely`, `__unlikely`, etc.
- Doxygen cleanup for init flags
- updated a few places that were relevant to this change to use the new attributes from `cdefs.h`

The first thing you might notice is the reliance on macros in `init_base.h` now to handle the weak symbols... I understand macros are typically a pain in the ass and are not the way to go, and that I'm pretty macro happy... but I honestly think they're justified in this scenario for reducing the boilerplate for adding new init flags, ensuring consistent naming of symbols, and reducing surface area for errors involved with using them... If you guys disagree, lmk.


include/kos/init.h
```diff
@@ -85,14 +108,24 @@ extern void * __kos_romdisk;
...
+#define INIT_QUIET          0x00000010  /**< \brief Disable dbgio */
+#define INIT_EXPORT         0x00000020  /**< \brief Export kernel symbols */
+#define INIT_CONTROLLER     0x00000040  /**< \brief Enable Controller maple driver */
+#define INIT_KEYBOARD       0x00000080  /**< \brief Enable Keyboard maple driver */
+#define INIT_MOUSE          0x00000100  /**< \brief Enable Mouse maple driver */
+#define INIT_LIGHTGUN       0x00000200  /**< \brief Enable Lightgun maple driver */
+#define INIT_VMU            0x00000400  /**< \brief Enable VMU maple driver */
+#define INIT_PURUPURU       0x00000800  /**< \brief Enable Puru Puru maple driver */
+#define INIT_SIP            0x00001000  /**< \brief Enable Sound input maple driver */
+#define INIT_DREAMEYE       0x00002000  /**< \brief Enable DreamEye maple driver */
```
ljsebald, line 126: These definitely should not be in the arch-independent file. They should be in kernel/arch/dreamcast/include/arch/init_flags.h

gyrovorbis: Oh yeah, good call!

gyrovorbis:
image
Adding notes to myself for what all was discussed in discord.

gyrovorbis:
@pcercuei took this and then did a bunch of extra work that I hadn't gotten around to doing in his follow-up: "Modularizing KOS, my take." That's the PR we should be looking at now.

PR #341, opened 2023-11-05
Rework romdisks

pcercuei:
- KOS_INIT_ROMDISK() is now useless; it is automatically added to the romdisk.o object file. The macro remains just for compatibility with older source code.
- If a generated romdisk.o file is linked, romdisk support will be initialized when KallistiOS starts. Otherwise, the romdisk init/shutdown code (as well as any unused dependency) will be garbage-collected automatically.
- Loading external romdisks should still be supported, but unless a romdisk was built-in, you'll need to manually call fs_romdisk_init().

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -67,5 +67,7 @@ romdisk.img:
 	$(KOS_GENROMFS) -f romdisk.img -d $(KOS_ROMDISK_DIR) -v -x .svn -x .keepme
 
 romdisk.o: romdisk.img
-	$(KOS_BASE)/utils/bin2o/bin2o romdisk.img romdisk romdisk.o
+	$(KOS_BASE)/utils/bin2o/bin2o romdisk.img romdisk romdisk_tmp.o
+	$(KOS_CC) -o romdisk.o -r romdisk_tmp.o $(KOS_LIB_PATHS) -Wl,--whole-archive -lromdiskbase
+	rm romdisk_tmp.o
 endif
--- a/doc/CHANGELOG
+++ b/doc/CHANGELOG
@@ -193,6 +193,8 @@ KallistiOS version 2.1.0 -----------------------------------------------
 - DC  Added vmu functions to check/enable/disable the extra 41 blocks [AB]
 - *** Added driver for the SH4's Watchdog Timer peripheral [FG]
 - DC  Added Moop powered fast path to sq_cpy, added TapamN pvr related sq functions [AB]
+- DC  Garbage-collect network stack [PC]
+- DC  Rework romdisks [PC]
 
 KallistiOS version 2.0.0 -----------------------------------------------
 - DC  Broadband Adapter driver fixes [Megan Potter == MP]
--- a/include/kos/init.h
+++ b/include/kos/init.h
@@ -48,16 +48,19 @@ __BEGIN_DECLS
     void (*bba_la_init_weak)(void) = ((flags) & INIT_NET) ? bba_la_init : NULL; \
     extern void bba_la_shutdown(void); \
     void (*bba_la_shutdown_weak)(void) = ((flags) & INIT_NET) ? bba_la_shutdown : NULL; \
+    extern int fs_romdisk_init(void); \
+    int (*fs_romdisk_init_weak)(void) = ((flags) & INIT_FS_ROMDISK) ? fs_romdisk_init : NULL; \
+    extern int fs_romdisk_shutdown(void); \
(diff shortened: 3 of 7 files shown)


gyrovorbis:
Got any numbers on how much .text or .rodata or heap space this saves by any chance?

pcercuei:
@gyrovorbis About 4.5 KiB ;D
Not sure in which section.

The point of this PR is less for automatic garbage collection, more for removing the requirement of having KOS_INIT_ROMDISK() every time you want to link a romdisk.

gyrovorbis:
> @gyrovorbis About 4.5 KiB ;D Not sure in which section.
> 
> The point of this PR is less for automatic garbage collection, more for removing the requirement of having KOS_INIT_ROMDISK() every time you want to link a romdisk.

Hey man, that's still something! But I actually agree, if you can get rid of the need for it without breaking backwards compatibility, *plus* get a few extra KB... sounds nice!

ljsebald (changes requested):
Can you please add a new `INIT_FS_ROMDISK` flag to the list of init flags and make that so that it will cause fs_romdisk_init and fs_romdisk_shutdown to be called in the normal init/shutdown, like what was done with the export stuff?

pcercuei:
> Can you please add a new `INIT_FS_ROMDISK` flag to the list of init flags and make that so that it will cause fs_romdisk_init and fs_romdisk_shutdown to be called in the normal init/shutdown, like what was done with the export stuff?

You don't like that this is done automatically by just linking a romdisk? ...

ljsebald:
> You don't like that this is done automatically by just linking a romdisk? ...

The automatic part is fine. I mean for programs that don’t want a linked romdisk, but only want to mount one later.

pcercuei:
Updated.

We now have a `INIT_FS_ROMDISK`, enabled by default, that controls whether or not the init/shutdown callbacks are called by KOS' init/shutdown code.

Linking a romdisk.o will cause the "/rd" VFS entry to be mounted. If not linked, "/rd" is not mounted, and if the code does not call `fs_romdisk_mount` anywhere else, the VFS romdisk functions will be garbage-collected.

ljsebald:
I think you still slightly misunderstood what I was saying. I wasn't saying that the romdisk init flag should be enabled by default. Just that there should be a flag there, like what was added for the export stuff. Just like with the export, it doesn't have to be enabled by default.

What I was basically suggesting was as follows:
1. Keep the auto-init when you have a romdisk linked (including only mounting /rd if one is linked in and calling the init as needed).
2. Add an init flag so that if someone wants to have a romdisk loaded later, they don't have to explicitly call the fs_romdisk_init function. **This doesn't have to be on by default, just as the exports one is not on by default now.**
3. Having a romdisk linked in should imply the romdisk init flag.

pcercuei:
(copy-paste of my answer on Discord, for reference)

That's fine, I understood it like that. But then it complexifies the code, because you need one weak function for the romdisk init set from an init flag, and one weak function for the romdisk init from the romdisk.o. So I believe it makes more sense to have one `INIT_FS_ROMDISK` flag enabled by default, that you can turn OFF if you want, and in which case romdisks won't work at all.

Note that you can't have two weak functions of the same name, because then I think GCC will pick the first one it sees. So if the one from the init flags resolves to NULL, and it picks that one, it won't work properly.

I have a commit that does it like that, if you want to have a look: https://github.com/pcercuei/KallistiOS/commit/af21f162c3920ecc3e3b99522b9e44b958bfe523

The other reason why I thought it wasn't great, is that you can remove the `INIT_FS_ROMDISK` from the init flags, link a romdisk, and it will initialize it and mount it to `/rd` properly. So it was a bit misleading.

As for the current version: setting the `INIT_FS_ROMDISK` flag without linking a `romdisk.o` will pull in `fs_romdisk_init` and `fs_romdisk_shutdown`, but since `fs_romdisk_mount` is never called, a huge part of the romdisk VFS code will be garbage-collected. Only when you link with romdisk.o (which will provide the `fs_romdisk_mount_weak`) will you have all the romdisk VFS code compiled in. And of course, if you use an external romdisk, your code manually calls `fs_romdisk_mount`, so the romdisk VFS code is kept as well.

Honestly though, this PR is not really about garbage-collecting the romdisk VFS code, it's very small (about 4.5 KiB), more about simplifying the usage of romdisk by removing the need to use the `KOS_INIT_ROMDISK` macro.

ljsebald (approved):

PR #342, opened 2023-11-06
Cache Refresh

andressbarajas:
Renamed dcache_purge_all() to dcache_purge_all_with_buffer().

Add two functions: 

dcache_flush_all(void)
dcache_purge_all(void)

When dcache_flush_range() or dcache_purge_range() have ranges bigger than the dcache 16kb, we use dcache_flush_all()/dcache_purge_all() instead.  This speeds up flushes over ranges 256x256x2 by 2.6x and  512x512x2 by 10.5x.

dcache_purge_all_with_buffer() is still the fastest method but requires a 8/16kb buffer.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/cache.h
+++ b/kernel/arch/dreamcast/include/arch/cache.h
@@ -89,7 +89,7 @@ void dcache_flush_range(uintptr_t start, size_t count);
     This function flushes all the data/operand cache, forcing a write-
     back on all of the cache blocks that are marked as dirty.
 
-    \warning
+    \note
     dcache_flush_range() is faster than dcache_flush_all() if the count
     param is 107520 or less.
 */
@@ -107,11 +107,12 @@ void dcache_purge_range(uintptr_t start, size_t count);
 
 /** \brief  Purge all the data/operand cache.
 
-    This function flushes all the data/operand cache, forcing a write-
-    back on all cache blocks that are marked as dirty and invalidates 
-    all of the cache blocks.
+    This function flushes the entire data/operand cache, ensuring that all 
+    cache blocks marked as dirty are written back to memory and all cache 
+    entries are invalidated. It does not require an additional buffer and is 
+    preferred when memory resources are constrained.
 
-    \warning
+    \note
     dcache_purge_range() is faster than dcache_purge_all() if the count
     param is 39936 or less.
 */
@@ -119,11 +120,22 @@ void dcache_purge_all(void);
 
 /** \brief  Purge all the data/operand cache with buffer.
 
-    This function flushes all the data/operand cache, forcing a write-
-    back and invalidate on all of the cache blocks using a buffer.
+    This function performs a purge of all data/operand cache blocks by 
+    utilizing an external buffer to speed up the write-back and invalidation 
+    process. It is always faster than dcache_purge_all() and is recommended 
+    where maximum speed is required.
 
-    \param  start           The physical address for temporary buffer (32-byte aligned)
-    \param  count           The number of bytes of temporary buffer (8 KB or 16 KB)
+    \note While this function offers a superior purge rate, it does require
+    the use of a temporary buffer. So use this function if you have an extra 
+    8/16 kb of memory laying around that you can utilize for no other purpose 
+    than for this function.
+
+    \param  start           The physical address for temporary buffer (32-byte 
+                            aligned)
+    \param  count           The size of the temporary buffer, which can be 
+                            either 8 KB or 16 KB, depending on cache 
+                            configuration - 8 KB buffer with OCRAM enabled, 
+                            otherwise 16 KB.
 
 */
 void dcache_purge_all_with_buffer(uintptr_t start, size_t count);
--- a/kernel/arch/dreamcast/include/arch/cache.h
+++ b/kernel/arch/dreamcast/include/arch/cache.h
@@ -27,23 +27,6 @@ __BEGIN_DECLS
 #include <stdint.h>
 #include <arch/types.h>
 
-/** \brief  Obtain a non-cached pointer.
-
-    Modifies an address to enable non-cached access. Uncached RAM is very 
-    slow, and bypassing the cache only helps in very specific circumstances.
-
-    \author TapamN
- */
-#define NONCACHED(addr) (typeof (&(addr)[0]))(((uintptr_t)(addr)) | 0x20000000)
-
-/** \brief  Obtain a cached pointer.
-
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/kernel/cache.s
```diff
@@ -74,99 +69,178 @@ flush_loop:
...
-	mov.l	l1align,r0
-	and	r0,r4
+	! Divide byte count by 32 
+	mov     #-5, r1
+	shad    r1, r5           
+
+	! Compare with 512
+	mov.w   cache_lines, r2
+	cmp/hi  r2, r5
+	bt      _dcache_flush_all  ! If lines > 512, jump to _dcache_flush_all
```
DC-SWAT, line 111: What if you flush buffer that almost not in cache with size like 515 cache lines. So a flush_all still will be faster? I mean, maybe a gap is needed?

andressbarajas: Yes, you are correct! I just tested flushing 16kb range and 16kb range +32. There is a huge time difference(60,000 ns) that occurs when we start relying on the hard coded 512 check. I can up the check(# >512) and find the sweet spot. For flushing typical ranges like 128*128*2, 256*256*2 or 512*512*2, dcache_flush_all() would be faster.

ljsebald:
A couple of quick suggestions (without looking too deeply at the PR just yet, so I may have some more later)...
1. Can the comment for `dcache_purge_all` and `dcache_purge_all_with_buffer` have a note as to why someone might choose one over the other. Noting that the buffered approach is faster, but requires the buffer to exist is a useful thing. Also, if it could be clarified why an 8kb versus 16kb buffer is needed, that'd be appreciated (yes, that probably should've been done on the PR that added the function).
2. The `\warning`s added on the two comments should probably be `\note`s instead.

kernel/arch/dreamcast/include/arch/cache.h
```diff
@@ -23,12 +24,29 @@
...
 #include <arch/types.h>
 
+/** \brief  Obtain a non-cached pointer.
+
+    Modifies an address to enable non-cached access. Uncached RAM is very 
+    slow, and bypassing the cache only helps in very specific circumstances.
+
+    \author TapamN
+ */
+#define NONCACHED(addr) (typeof (&(addr)[0]))(((uintptr_t)(addr)) | 0x20000000)
```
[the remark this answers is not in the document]
andressbarajas: I think that was the purpose when TapamN wrote these: Ignoring mem region but translating an address between cached/uncached.  Does translating to a different memory region (outside of P2 - uncached) affect the access speed ? Ex. from P1 to P0?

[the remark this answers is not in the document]
DC-SWAT: I uses this I bit different way. You can ignore it, I'll just leave this here as variant:
https://github.com/DC-SWAT/DreamShell/blob/master/firmware/isoldr/loader/include/main.h#L34

kernel/arch/dreamcast/kernel/cache.s
```diff
@@ -201,6 +207,7 @@ _dcache_purge_all:
 ! r4 is address for temporary buffer 32-byte aligned
 ! r5 is size of temporary buffer (8 KB or 16 KB)
 _dcache_purge_all_with_buffer:
+.align 2
```
DC-SWAT, line 210: This gives improvements for all functions?

andressbarajas: For the most part yes,  some are very minor ~1000ns

ljsebald: Shouldn't the alignment directives be before the symbols, not after?

kernel/arch/dreamcast/include/arch/cache.h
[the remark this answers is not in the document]
ljsebald: Those implementations from @DC-SWAT are definitely more correct. However, I'd suggest that if such macros are to be added, that they should probably be in `arch/memory.h` and should be more along the lines of being called `MEM_ADDR_P1` or names like that. Then comments can be added to each one to clarify where they'd be useful (i.e, saying that P1 is normally a cached address, P2 is an uncached address, etc).

I feel like `NONCACHED` and `CACHED` are too generic and have a distinct possibility of conflicting with user symbol/macro names.

kernel/arch/dreamcast/kernel/cache.s
andressbarajas: I dont think it matters but I can move them if that is the proper way to do it.  Im still an asm noob and Ive seen so many approaches.  Let me look it up right now.

andressbarajas: @ljsebald You're right.  I'll make the changes.

kernel/arch/dreamcast/kernel/cache.s
ljsebald, line 1: This whole file is going to need some formatting cleanup at some point... It doesn't necessarily have to be done in this PR, but it is a mish-mash of all kinds of formatting (tabs, spaces, sometimes both on the same line), which really needs to be cleaned up at some point...

Just seeing it in the PR view has drawn my attention to it, and it really does need to be fixed eventually. Like I said, no big deal right now with this PR. I'll eventually get around to it, sometime.

andressbarajas: I can take care of it.  Replace all tabs with spaces Im guessing?

kernel/arch/dreamcast/include/arch/cache.h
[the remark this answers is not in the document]
andressbarajas: I will remove these macros then and they can be in a future small PR.

DC-SWAT (approved):

kernel/arch/dreamcast/kernel/cache.s
ljsebald: Tabs should be replaced with 4 spaces (which it looks like you've already done by this point, so thank you for that). Also, the `.text` and `.align` directives should be indented like the `.globl` and various data type ones are, so as to make sure they stand out from labels that start with a dot.

Also, we should probably have a header comment at the top with copyright information and the like.... If you wouldn't mind adding that with at least yourself for the new stuff (and I can figure out what needs to be added for the history of the file and put it in as a suggestion later).

andressbarajas: Done

ljsebald (approved):

PR #343, opened 2023-11-07
Add sound effects example

andressbarajas:


examples/dreamcast/sound/sfx/main.c
```diff
@@ -0,0 +1,158 @@
+
```
ljsebald, line 1: Copyright header?

andressbarajas: Done

ljsebald (approved):

PR #350, opened 2023-11-10
Update romdisk examples to use new romdisk approach

andressbarajas:


examples/dreamcast/basic/exec/Makefile
```diff
@@ -26,7 +26,6 @@ $(SUBTARGET): $(SUBOBJS)
 	kos-cc -o sub.elf $(SUBOBJS)
 	$(KOS_OBJCOPY) -O binary sub.elf $(SUBTARGET)
 
-# You can safely remove the next two targets if you don't use a ROMDISK
```
ljsebald, line 29: Doesn't this one need `KOS_ROMDISK_DIR`? You probably still need to keep that romdisk.img has a dependency on the `$(SUBTARGET)`, of course.

ljsebald (changes requested):

examples/dreamcast/basic/exec/Makefile
andressbarajas: I didnt include KOS_ROMDISK_DIR to makefiles that have custom romdisk dependencies.  That includes this example (which seems broken btw because it doesnt load the ), bumpmap, and modifier volume with te. ture. I'll pull the latest changes and work on them.

ljsebald: Ah... Well, in that case, they'd still need the KOS_INIT_ROMDISK in the code, if they are to keep their old special target stuff.

andressbarajas: I got it all sorted out and pushed my changes.  The three examples I mentioned look to be working.

ljsebald (approved):

PR #352, opened 2023-11-11
 Fix accidental double free of stream and cleaning code

andressbarajas:


gyrovorbis (approved):
Looks good. Good catch.

PR #356, opened 2023-11-12
SPU Clean

andressbarajas:
Added these constants to spu.h and replaced redefined values across spu.c, snd_stream.c, and snd_iface.c:

#define SPU_RAM_BASE 0x00800000
#define SPU_RAM_UNCACHED_BASE (MEM_AREA_P2_BASE | SPU_RAM_BASE)

Replaced uint8, uint16, uint32 types.

Corrected comment that rounds to the nearest 4 for SQ transfer and reduced logic to calculate number of 32-bit words for other transfers:
https://godbolt.org/z/d5xjs4aWz

### DIFF
--- a/kernel/arch/dreamcast/hardware/spu.c
+++ b/kernel/arch/dreamcast/hardware/spu.c
@@ -62,7 +62,8 @@ void spu_memload(uintptr_t dst, void *src_void, size_t length) {
 
 void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     uint8_t *src = (uint8_t *)src_void;
-    int aligned_len, old;
+    int aligned_len;
+    g2_ctx_t ctx;
 
     /* Round up to the nearest multiple of 4 */
     if(length & 3) {
@@ -76,10 +77,13 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     /* Add in the SPU RAM base (cached area) */
     dst += SPU_RAM_BASE;
 
-    old = irq_disable();
-    do { } while(FIFO_STATUS & (FIFO_SH4 | FIFO_G2));
+    /* Make sure the FIFOs are empty */
+    ctx = g2_lock();
+    g2_fifo_wait();
+
     sq_cpy((void *)dst, src, aligned_len);
-    irq_restore(old);
+
+    g2_unlock(ctx);
 
     if(length > 0) {
         /* Make sure the destination is in a non-cached area */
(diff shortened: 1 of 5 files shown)


kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -60,30 +56,28 @@ void spu_memload(uint32 dst, void *src_void, int length) {
...
     /* Using SQs for all that is divisible by 32 */
     aligned_len = length & ~31;
     length &= 31;
 
+    /* Add in the SPU RAM base (cached area) */
+    dst += SPU_RAM_BASE;
+
     old = irq_disable();
-    do { } while(*(vuint32 *)0xa05f688c & (1 << 5)); /* FIFO_SH4 */
-    do { } while(*(vuint32 *)0xa05f688c & (1 << 4)); /* FIFO_G2 */
```
DC-SWAT, line 86: It must be replaced to g2_lock

andressbarajas: The new g2_lock function doesnt wait on FIFO_G2:

`static inline g2_ctx_t g2_lock(void) {
    g2_ctx_t ctx;

    ctx.irq_state = irq_disable();

    /* Suspend any G2 DMA */
    G2_DMA_SUSPEND_SPU = 1;
    G2_DMA_SUSPEND_BBA = 1;
    G2_DMA_SUSPEND_CH2 = 1;

    while(FIFO_STATUS & FIFO_SH4);

    return ctx;
}` Should that be updated too?

DC-SWAT: Just add g2_fifo_wait() after g2_lock()

DC-SWAT: In general, there are different cases. Somewhere we only need to wait for SH4, somewhere SH4 + G2, and somewhere SH4 + G2 + AICA.
In this case we need to wait SH4 + G2. In case with AICA registers, need wait AICA FIFO in additional. 
And also every 8 regs writes need wait AICA FIFO, as 8 writes to memory we wait G2 FIFO.

andressbarajas: DONE

kernel/arch/dreamcast/sound/snd_stream.c
```diff
@@ -391,6 +393,8 @@ snd_stream_hnd_t snd_stream_reinit(snd_stream_hnd_t hnd, snd_stream_callback_t c
 void snd_stream_destroy(snd_stream_hnd_t hnd) {
     filter_t *c, *n;
 
+    mutex_lock(&stream_mutex);
+
     CHECK_HND(hnd);
 
     if(!streams[hnd].initted)
```
DC-SWAT, line 400: Here is return without unlock

DC-SWAT (approved):

ljsebald (approved):

PR #357, opened 2023-11-12
ARM AICA Clean up types and spacing

andressbarajas:


kernel/arch/dreamcast/include/dc/sound/aica_comm.h
```diff
@@ -12,8 +12,8 @@
 #define __DC_SOUND_AICA_COMM_H
 
 #ifndef __ARCH_TYPES_H
-typedef unsigned long uint8;
-typedef unsigned long uint32;
+typedef unsigned char uint8_t;
+typedef unsigned long uint32_t;
 #endif
```
ljsebald, line 17: The types with _t aren't defined in `__ARCH_TYPES_H`, so this isn't really correct here...

I'd prefer if we kept these as the internal types, as this is low-level stuff that really shouldn't be using stdint stuff, IMO. That way we definitely control the types and can ensure they stay the same across the SH4/ARM divide.

In addition, we should not be defining stuff that is part of libc, so that's all the more reason to not do this...

andressbarajas: I understand.  I reverted those changes and left what I think should still be okay.  Let me know.

ljsebald (approved):

PR #358, opened 2023-11-12
Ensure getaddrinfo() handles IP adddress inputs without using DNS

has207:
Currently any input to getaddrinfo() gets resolved using DNS, including raw IP addresses. This is not ideal but tends to work when your DNS server is local, however it flat out fails when the DNS server is an external/public one and you're trying to connect to a local IP.

This change parses the hostname input string and fills relevant structures without making a DNS call when the input matches IPv4 or IPv6 address.

### DIFF
--- a/kernel/libc/koslib/getaddrinfo.c
+++ b/kernel/libc/koslib/getaddrinfo.c
@@ -706,12 +706,15 @@ int getaddrinfo(const char *nodename, const char *servname,
         return 0;
     }
 
-    // Try to handle input as IPv4 address
-    uint32_t ip4_addr;
-    if (inet_pton(AF_INET, nodename, &ip4_addr) > 0) {
-        ihints.ai_family = AF_INET;
-        *res = add_ipv4_ai(ip4_addr, port, &ihints, NULL);
-        return 0;
+    /* Try to handle input as an IPv4 address */
+    if(ihints.ai_family == AI_INET || ihints.ai_family == AF_UNSPEC) {
+        uint32_t ip4_addr;
+
+        if(inet_pton(AF_INET, nodename, &ip4_addr) > 0) {
+            ihints.ai_family = AF_INET;
+            *res = add_ipv4_ai(ip4_addr, port, &ihints, NULL);
+            return 0;
+        }
     }
 
     // Try to handle input as IPv6 address
--- a/kernel/libc/koslib/getaddrinfo.c
+++ b/kernel/libc/koslib/getaddrinfo.c
@@ -717,12 +717,15 @@ int getaddrinfo(const char *nodename, const char *servname,
         }
     }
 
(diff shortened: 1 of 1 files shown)


andressbarajas:
@Kazade Tagging you because your game has online play.

kernel/libc/koslib/getaddrinfo.c
```diff
@@ -706,6 +706,22 @@ int getaddrinfo(const char *nodename, const char *servname,
         return 0;
     }
 
+    // Try to handle input as IPv4 address
+    uint32_t ip4_addr;
+    if (inet_pton(AF_INET, nodename, &ip4_addr) > 0) {
+        ihints.ai_family = AF_INET;
+        *res = add_ipv4_ai(ip4_addr, port, &ihints, NULL);
+        return 0;
+    }
```
ljsebald, line 715:
```suggestion
    /* Try to handle input as an IPv4 address */
    if(ihints.ai_family == AI_INET || ihints.ai_family == AF_UNSPEC) {
        uint32_t ip4_addr;

        if(inet_pton(AF_INET, nodename, &ip4_addr) > 0) {
            ihints.ai_family = AF_INET;
            *res = add_ipv4_ai(ip4_addr, port, &ihints, NULL);
            return 0;
        }
    }
```

kernel/libc/koslib/getaddrinfo.c
```diff
@@ -706,6 +706,22 @@ int getaddrinfo(const char *nodename, const char *servname,
...
+
+    // Try to handle input as IPv6 address
+    struct in6_addr addr;
+    if (inet_pton(AF_INET6, nodename, &addr.s6_addr) > 0) {
+        ihints.ai_family = AF_INET6;
+        *res = add_ipv6_ai(&addr, port, &ihints, NULL);
+        return 0;
+    }
```
ljsebald, line 723:
```suggestion
    /* Try to handle input as an IPv6 address */
    if(ihints.ai_family == AF_INET6 || ihints.ai_family == AF_UNSPEC) {
        struct in6_addr addr;

        if(inet_pton(AF_INET6, nodename, &addr.s6_addr) > 0) {
            ihints.ai_family = AF_INET6;
            *res = add_ipv6_ai(&addr, port, &ihints, NULL);
            return 0;
        }
    }
```

ljsebald (changes requested):
Kernel code within KOS, including the network code should be written as C89/C90 (with a few exceptions like allowing for `<stdint.h>` types). Variables must be declared at the beginning of a scope, and comments should be done with `/* */`, not `//`.

In addition, it probably shouldn't be trying to parse an IPv4 address if the user specified IPv6 and vice-versa, I imagine.

has207:
Thanks for the feedback, suggestions merged!

ljsebald (approved):

PR #361, opened 2023-11-13
getsockopt(): add no-op handling for SO_ERROR

has207:
getsockopt() is supposed to return and clear errors when passed SO_ERROR option.
As it's not currently handled it always returns a "Protocol not available" error when called which signals to the calling code to kill the connection regardless of whether it's established or not.

This change adds a no-op handler for SO_ERROR so that the calls to getsockopt() return successfully.

It's not exactly correct, but given that we don't seem to have error info on the socket
structure I'm not sure it can be implemented easily. In any case, this at least makes it
possible for calling code to work normally.


ljsebald (approved):

ljsebald:
Thanks for the contribution.

There are definitely some loose ends in the network code, so if you see any more, please don't hesitate to let us know/PR fixes for them.

PR #364, opened 2023-11-15
Thread-safe audio fixes

DC-SWAT:
- Fixes for streams mutex
- Added a spinlock for sound memory

### DIFF
--- a/kernel/arch/dreamcast/sound/snd_mem.c
+++ b/kernel/arch/dreamcast/sound/snd_mem.c
@@ -1,7 +1,8 @@
 /* KallistiOS ##version##
 
    snd_mem.c
-   (C)2002 Megan Potter
+   Copyright (C) 2002 Megan Potter
+   Copyright (C) 2023 Ruslan Rostovtsev
 
  */
 
@@ -9,8 +10,10 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <string.h>
+#include <errno.h>
 #include <sys/queue.h>
 #include <dc/sound/sound.h>
+#include <arch/spinlock.h>
 
 /*
 
@@ -59,6 +62,8 @@ typedef struct snd_block_str {
 /* Our SPU RAM pool */
 static int initted = 0;
 static TAILQ_HEAD(snd_block_q, snd_block_str) pool = {0};
+static spinlock_t snd_mem_mutex = SPINLOCK_INITIALIZER;
+
 
 /* Reinitialize the pool with the given RAM base offset */
 int snd_mem_init(uint32 reserve) {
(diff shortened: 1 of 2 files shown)


gyrovorbis (approved):
Looks good! Just spent some time reviewing. Still want Crabby to see, but I didn't see anything wrong.

ljsebald:
A big problem with doing this is that it makes it completely impossible to do any of the functions you've outfitted with mutexes from within an interrupt context (mutex_lock and mutex_lock_timed do not work in an interrupt context -- only mutex_trylock is available in an interrupt context). I'm definitely not comfortable with that restriction on the memory allocation.

Also introducing mutex locking within the sound polling functionality could potentially lead to issues with dropped samples and the like... I think we need to at least think about this a little bit more before taking this approach.

DC-SWAT:
> A big problem with doing this is that it makes it completely impossible to do any of the functions you've outfitted with mutexes from within an interrupt context (mutex_lock and mutex_lock_timed do not work in an interrupt context -- only mutex_trylock is available in an interrupt context). I'm definitely not comfortable with that restriction on the memory allocation.

That's a good point, I didn't know about that. It can be remade, but I doubt that this will ever be called in interrupts; sound memory has a different use case.

> 
> Also introducing mutex locking within the sound polling functionality could potentially lead to issues with dropped samples and the like... I think we need to at least think about this a little bit more before taking this approach.

This is the main point, since it uses global resources such as buffer for PCM separation and the AICA DMA.
Separation buffer could be made for each stream, this would take up much more memory, but would be more thread-safe. It’s more complicated with the G2bus; this is only possible if you use different DMA channels or add a mutex for the AICA DMA channel.

ljsebald:
> > A big problem with doing this is that it makes it completely impossible to do any of the functions you've outfitted with mutexes from within an interrupt context (mutex_lock and mutex_lock_timed do not work in an interrupt context -- only mutex_trylock is available in an interrupt context). I'm definitely not comfortable with that restriction on the memory allocation.
> 
> That's a good point, I didn't know about that. I'll rewrite this, but I doubt that this will ever be called in interrupts; sound memory has a different use case.
I agree that it is unlikely. Thank you for reworking it though.

> > Also introducing mutex locking within the sound polling functionality could potentially lead to issues with dropped samples and the like... I think we need to at least think about this a little bit more before taking this approach.
> 
> This is the main point, since it uses global resources such as buffer for PCM separation and the AICA DMA. Separation buffer could be made for each stream, this would take up much more memory, but would be more thread-safe. It’s more complicated with the G2bus; this is only possible if you use different DMA channels or add a mutex for the AICA DMA channel.
Ok... I can see the point there then. Thanks for the explanation as to what you were going for. I hadn't thought about that at all...

ljsebald (approved):

PR #368, opened 2023-11-18
Implement SO_RCVBUF and SO_SNDBUF for setsockopt()

darcagn:
This PR implements the SO_RCVBUF and SO_SNDBUF options in setsockopt(). This allows the send and receive buffer sizes to be increased for a socket. Currently attempting to do so with setsockopt() results in a silent failure. 

Increasing the send and receive buffers from the default 8192 to the maximum 65535 results in a doubling in throughput for my application.

### DIFF
--- a/kernel/net/net_tcp.c
+++ b/kernel/net/net_tcp.c
@@ -1767,6 +1767,7 @@ static int net_tcp_setsockopt(net_socket_t *hnd, int level, int option_name,
                               const void *option_value, socklen_t option_len) {
     struct tcp_sock *sock;
     int tmp;
+    uint8_t *new_ptr;
 
     if(!option_value || !option_len) {
         errno = EFAULT;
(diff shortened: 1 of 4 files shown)


gyrovorbis (approved):
Looks great to me,

kernel/net/net_tcp.c
```diff
@@ -1808,6 +1809,48 @@ static int net_tcp_setsockopt(net_socket_t *hnd, int level, int option_name,
...
+                    sock->rcvbuf_sz = (*(uint32_t *)option_value > 65535) ?
+                                      65535 : *(uint32_t *)option_value;
+
+                    new_ptr = realloc(sock->data.rcvbuf, sock->rcvbuf_sz);
+                    if(!new_ptr) {
+                        sock->rcvbuf_sz = tmp;
+                        goto ret_nomem;
+                    }
+
+                    sock->data.rcvbuf = new_ptr;
```
ljsebald, line 1830: It would make more sense to structure this such that the `tmp` variable is where the option value is copied into and that you only update the `sock->rcvbuf_sz` when you update `sock->data.rcvbuf` down below.

Also this needs a check for a sensible minimum value as well. Linux apparently enforces a minimum of 256 here (per the manpage), so let's do the same.

kernel/net/net_tcp.c
```diff
@@ -1808,6 +1809,48 @@ static int net_tcp_setsockopt(net_socket_t *hnd, int level, int option_name,
...
+                    sock->sndbuf_sz = (*(uint32_t *)option_value > 65535) ?
+                                      65535 : *(uint32_t *)option_value;
+
+                    new_ptr = realloc(sock->data.sndbuf, sock->sndbuf_sz);
+                    if(!new_ptr) {
+                        sock->sndbuf_sz = tmp;
+                        goto ret_nomem;
+                    }
+
+                    sock->data.sndbuf = new_ptr;
```
ljsebald, line 1851: The same suggestion from up above with the receive buffer applies here too. The minimum Linux specifies for this value is 2048.

ljsebald (changes requested):

PR #370, opened 2023-11-18
Fix Typo in CMake Toolchain for NAOMI

cepawiel:


gyrovorbis (approved):
Yep, that's a f*ckup!

PR #372, opened 2023-11-18
Add <netinet/tcp.h> header

ljsebald:
- Add the `<netinet/tcp.h>` header called for in POSIX, with the required (single) macro defined in it.
- Add "support" for `TCP_NODELAY` in setsockopt() and getsockopt() -- Note, we don't implement Nagle's algorithm, so the options are effectively ignored.
- Clean up some formatting in net_tcp.c.


gyrovorbis (approved):
Yuuuuus more POSIX standards compliance!

PR #377, opened 2023-11-20
Fixed GD-ROM init for BIOS mod without CD inserted.

DC-SWAT:
- Fixed GD-ROM init for BIOS mod without CD inserted
- Added `cdrom_exec_cmd_timed()`
- Fixed `cdrom_get_status()`
- Fixed `gdc_abort_cmd()`
- Some cleanup.

### DIFF
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -10,11 +10,15 @@
  */
 #include <assert.h>
 
+#include <arch/timer.h>
+#include <arch/memory.h>
+
 #include <dc/cdrom.h>
 #include <dc/g1ata.h>
 
 #include <kos/thread.h>
 #include <kos/mutex.h>
+#include <kos/dbglog.h>
 
 /*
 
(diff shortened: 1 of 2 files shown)


gyrovorbis:
> * Added `cdrom_exec_cmd_timed()`
Damn, that's a good idea...

kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -282,32 +315,10 @@ int cdrom_reinit(void) {
...
-    while(timeout > 0) {
-        r = cdrom_exec_cmd(CMD_INIT, NULL);
-
-        if(r == 0) break;
-
-        if(r == ERR_NO_DISC || r == ERR_SYS) {
-            return r;
-        }
+    int r;
+    r = cdrom_exec_cmd_timed(CMD_INIT, NULL, 10000);
```
ljsebald, line 319: Just checking here, since it seems this is a bit of a behavior change... This has been adequately tested on the hardware, right?

If I understand the code correctly, the old path would resubmit the INIT command each pass through its loop if it didn't get a response back, whereas the new version only submits the command one time and then could wait up to 10 seconds...

DC-SWAT: This was written incorrectly.
The drive takes a long time to execute a command when there is no disk, rather than aborting commands with an error. Here it only works because the INIT command never returned an error, since the system calls were just used by the BIOS (in the case without a mod).
Here it is generally forbidden to call INIT without a syscall reset, which was done only by the BIOS, I just added it.
I tested both with and without disk on real hardware and emulators.

DC-SWAT: In any case, there is nothing to worry about if INIT fails here, since it will be performed again if something changes in the status when accessing the file system. Here it can be removed altogether to be honest.

DC-SWAT: As proof that it has never worked before - incorrect use of the abort command syscall :)

ljsebald: Ok, Thanks for the confirmation. :+1:

ljsebald (approved):

QuzarDC:
>     * Added `cdrom_exec_cmd_timed()`

Thank you. I originally added that comment about making a timed version many years ago when adding a bit of other functionality :)

PR #388, opened 2023-11-26
Fix GCC 13.2.0 Toolchain Compilation on Cygwin

ppxxcc:
Building dc-chain on Cygwin fails in gcc-13.2.0/gcc/config/sh/sh.cc, with the following error:

> build-gcc-sh-elf-13.2.0-pass1.log:../../gcc-13.2.0/gcc/system.h:150:33: エラー: ‘fputs_unlocked’ was not declared in this scope; did you mean ‘fputc_unlocked’?

In gcc-13.2.0/gcc/system.h, fputs() and similar functions are redefined to use their unlocked counterpart, like fputs_unlocked().

However, for whatever reason, this causes the build to fail on Cygwin.

As a workaround, I have added a patch, which will add a preprocessor conditional inside sh.cc to check for Cygwin and undefine the problematic symbols.

After the patch, the build process completes successfully.

### DIFF
--- a/utils/dc-chain/patches/gcc-13.2.0-kos.diff
+++ b/utils/dc-chain/patches/gcc-13.2.0-kos.diff
@@ -160,20 +160,3 @@ diff --color -ruN gcc-13.2.0/libstdc++-v3/configure gcc-13.2.0-kos/libstdc++-v3/
      mcf)	thread_header=config/i386/gthr-mcf.h ;;
  esac
  
-diff --color -ruN gcc-13.2.0/gcc/config/sh/sh.cc gcc-13.2.0-kos/gcc/config/sh/sh.cc
---- gcc-13.2.0/gcc/config/sh/sh.cc	2023-11-26 08:34:56.426962700 +0900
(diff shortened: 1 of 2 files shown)


AaronGlazer:
Thank you for the valuable contribution as always Shirobon. We are very excited for the future of Cygwin KallistiOS Sayga Drmcsat development.

darcagn (changes requested):
Thanks for the contribution! However, the `patches/gcc-13.2.0-kos.diff` file is for general KOS patches to GCC 13.2.0, for all platforms. For platform-specific patches, they are to be placed in e.g. `patches/<host triplet>/gcc-13.2.0.diff`, where `<host triplet>` is `x86_64-pc-cygwin` or whatever the triplet is on Cygwin (try running `gcc -dumpmachine` on the using the host Cygwin GCC).

ppxxcc:
Hi Darc, I removed the global patch and added a new platform specific patch to x86_64-pc_cygwin/gcc-13.2.0.diff
Please review.

darcagn (approved):

PR #393, opened 2023-12-01
Threaded ASIC IRQ handlers

pcercuei:
Update `irq_set_handler`, `irq_set_global_handler`, `asic_evt_set_handler`, `vblank_handler_add` to accept an extra argument, that will be passed along to the registered callback.

This is very useful to pass state structures to the IRQ handlers, which can then pass it down the functions.

This will later be used to pass structures between a first-level IRQ handler and second-level threaded IRQ handlers, when those are implemented.

### DIFF
--- a/kernel/arch/dreamcast/hardware/asic.c
+++ b/kernel/arch/dreamcast/hardware/asic.c
@@ -93,12 +93,15 @@
 
  */
 
+#include <stdbool.h>
+#include <stdlib.h>
 #include <string.h>
 #include <stdio.h>
 #include <assert.h>
 #include <arch/irq.h>
 #include <dc/asic.h>
 #include <arch/spinlock.h>
+#include <kos/genwait.h>
 
 /* XXX These based on g1ata.c and pvr.h and should be replaced by a standardized method */
 #define IN32(addr)         (* ( (volatile uint32_t *)(addr) ) )
@@ -114,6 +117,18 @@ typedef struct {
     asic_evt_handler hdl;
     void *data;
 } asic_evt_handler_entry_t;
+
+struct asic_thdata {
+    asic_evt_handler hdl;
+    uint32_t source;
+    kthread_t *thd;
+    int genwait_obj;
+    void *data;
+    volatile bool quit;
+    volatile bool pending;
+    void (*ack_and_mask)(uint16_t);
+    void (*unmask)(uint16_t);
+};
 
 /* Exception table -- this table matches each potential G2 event to a function
    pointer. If the pointer is null, then nothing happens. Otherwise, the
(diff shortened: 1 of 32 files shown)


pcercuei:
Rebased and updated to use the ugly void-casts for unused parameters.

kernel/arch/dreamcast/hardware/asic.c
```diff
@@ -110,26 +113,46 @@
 #define ASIC_EVT_REGS 3
 #define ASIC_EVT_REG_HNDS 32
 
+typedef struct {
+    asic_evt_handler hdl;
+    void *data;
+} asic_evt_handler_entry_t;
+
+struct asic_thdata {
```
QuzarDC, line 121: Is there a compelling reason to do it this way? In this way it's not possible to determine whether or not a given handler is threaded or not (and if we need to call its own free/remove) or not. By expanding the `asic_evt_handler_entry_t` to have some component (maybe flags or a pointer for asic_thdata) the management of the handlers could be more easily be managed. 

I presume some sort of change to this with my comment at the end. But am open to having missed why this would be better or that's really unnecessary.

kernel/arch/dreamcast/include/dc/asic.h
```diff
@@ -193,9 +195,17 @@ typedef void (*asic_evt_handler)(uint32_t code);
 
     \param  code            The ASIC event code to hook (see \ref asic_events).
     \param  handler         The function to call when the event happens.
+    \param  data            A user pointer that will be passed to the callback.
 
 */
-void asic_evt_set_handler(uint16_t code, asic_evt_handler handler);
+void asic_evt_set_handler(uint16_t code, asic_evt_handler handler, void *data);
+
```
QuzarDC, line 202: These two would need to get fleshed out dox.

QuzarDC (commented):
Looks great. Needs a bit more (the dox for the new functions). And I gave two other suggestions (unique thread names for the handlers + adjusting the data structure to detect type/collapse remove function).

kernel/arch/dreamcast/hardware/asic.c
```diff
@@ -230,3 +252,99 @@ void asic_init(void) {
...
+    //thd_set_prio(thdata->thd, 0);
+
+    asic_evt_set_handler(code, asic_thirq_dispatch, thdata);
+
+    irq_restore(flags);
+
+    return 0;
+}
+
+void asic_remove_threaded_handler(uint16_t code)
```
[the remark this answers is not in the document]
pcercuei: Your `asic_remove_handler` wouldn't know which handler is a threaded interrupt and which one is not, because it only has a `void *data` pointer...

kernel/arch/dreamcast/hardware/asic.c
pcercuei: My reason was keeping RAM usage low, but it's probably a better idea to add a field, yes.

kernel/arch/dreamcast/include/dc/asic.h
pcercuei: What do you mean? It's already documented.

pcercuei: Nevermind - just noticed you were talking about the two functions below this one.

pcercuei:
@QuzarDC I updated the PR with your suggestions and some more:
- It now sets the thread name and uses it in genwait_wait();
- Added documentation for the two new functions;
- I renamed `asic_remove_threaded_handler` to `asic_evt_remove_handler`;
- I renamed `asic_request_threaded_handler` to `asic_evt_request_threaded_handler`;
- `asic_evt_remove_handler` will work with both threaded and non-threaded handlers. I did not have to add a new field to `asic_evt_handler_entry_t`, because we know that threaded handlers will always have `.hnd = asic_thirq_dispatch`.

kernel/arch/dreamcast/include/dc/asic.h
QuzarDC: > Nevermind - just noticed you were talking about the two functions below this one.

Right, sorry, it made more sense in the review view where these are all inlined so the comment would have appeared directly in the spot where the dox go. I forgot they get truncated down in the normal views -_-;

kernel/arch/dreamcast/hardware/asic.c
QuzarDC: Checking against the dispatch function being installed makes fine sense instead of an additional field. If it ever becomes an issue to need to know elsewhere, a function could be exposed to make that same check.

kernel/arch/dreamcast/hardware/asic.c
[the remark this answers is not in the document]
QuzarDC: That was meant to be in the context of the other comment to add a way to distinguish them, but you figured that out by the end :) . My thinking was that for the handlers we want to make threaded the use cases we have seen have been that we want user-installable callbacks. It would make sense in those cases to use a regular handler by default ( which would likely be something like `{ ack_and_mask_func(); unmask_func(); }` ) then if a user requests a custom callback install the threaded version.

Overly long comment is just to document that reasoning for posterity.

QuzarDC (approved):

PR #413, opened 2023-12-16
Restore remaining GL examples for GLdc

darcagn:
This is a draft PR to track fixing and restoring the remaining non-working KOS GL examples for GLdc.


gyrovorbis:
You guys think this should be part of the `v2.1.0` milestone or not, now that we've decided to just keep the old KGL for these?

andressbarajas:
I dont think so.  I tried to get all the examples working last week.  I was able to get 5 or 6 more examples working but there are some KGL examples that GLdc currently does not currently support (issue with params values not being supoorted).  Keep in mind I am a OpenGL newb though.

PR #414, opened 2023-12-16
Testing and Example for libpthread Barrier Implementation

gyrovorbis:
Working on adding a new example/test to the pthread examples directory to fully flex the barrier implementation. 

- example directory + Makefile created
- Skeleton test application created, needs fleshing out


gyrovorbis:
This was originally supposed to be a test for KERNEL barriers, not pthread barriers... I also wound up making that test better than this one was anyway, so if we wind up needing a test for the pthread layer, I'd be better off starting a new PR based on the kthread barrier test branch.

PR #419, opened 2023-12-19
Refactor Multibuffered video modes

QuzarDC:
The purpose of this PR is to simplify and enhance the basic video mode settings primarily by restructuring the multibuffered mode. I'm not sure why the video driver was structured in this way to begin with but each multibuffer mode was simply a copy of its standard counterpart with a different set of positions for the frame buffer. This both made the file difficult to read and locked the modes to a static number of buffers.

Instead of simply splitting the vram into 4 2mb framebuffers, setting MB now automatically sets fb_count to match the maximum number of framebuffers that can fit in vram and only stores the size of a framebuffer. This shouldn't cause any incompatibility unless a user was explicitly attempting to read the relative position of the framebuffer via vid_mode->fb_base[x], which can now be done via vid_get_start(x) and any possible references to VID_MAX_FB which is now obsolete.

Other changes:

- Functions that take in a frame buffer as a parameter now all default to out-of-bounds meaning 'current'. Previously, the behavior of vid_flip would do 'current' if <0 (despite the docs stating '-1') and any above the fb_count would be truncated to a multiple of the fb_count, so with the default 4 if you passed in 6 it would flip to fb 2. This wasn't documented, and I can't imagine it would be relied upon.
- New convenience functions added. Though the nesting on some is a little much, they each served to isolate functionality that has a benefit to expose directly to the user.
- vid_disable/vid_enable - These are the method that vid_set_mode would use to prevent flashing while adjusting the PVR registers. Exposing them though gives the use the ability to blank the screen instantaneously without having to clear the framebuffer or flip to a blank one.
- vid_set_fb/vid_set_vram - These allow the user to independently manage the setting of the current displayed fb and the current convenience pointers. Previously the two could only be manipulated simultaneously by vid_set_start and vid_flip which each using its own ways to manage them. Now, if you can finish filling buffer 3 and move on to buffer 4 while still displaying buffer 0.
- Cleaned up all vid_mode data to be in dec rather than a mix of dec and hex.
- Cleaned up types and added an enum type for the pixel modes.
- vid_init now clears the entire vram rather than just the only or second framebuffer (in MB mode, after set mode vram_* is set to the second FB, so vid_clear would draw 0s to that one rather than the first).

I couldn't find any samples that used the multibuffer functionality so I went ahead and wrote one. The way it was set up makes it seem as though the changes shouldn't break basic usage and the example should work on both versions as well.


examples/dreamcast/video/multibuffer/multibuffer.c
```diff
@@ -0,0 +1,63 @@
...
+
+    multibuffer.c
+    Copyright (C) 2023 Donald Haase
+
+*/
+
+/*
+    This example is based off a combination of the libdream video mode and 
+    the bfont examples. It draws out four distinct framebuffers then rotates 
+    beteween them until stopped.
```
andressbarajas, line 11:
```suggestion
    between them until stopped.
```

QuzarDC:
Ok, so first time I had this up for review I realized that it would be best to provide more utility functions now to facilitate more flexible use of the multiple buffers. There were only two concerns I had of things that might make sense to change:
1) The previous implementation had an array of buffer addresses that could be accessed in the vid_mode. This could still be implemented as a dynamically allocated array, but it doesn't seem like that would provide much advantage at all over the function to look up the same address (vid_get_start)
2) Whether or not to rewrite internal functionality like screenshot, vid_clear, the fb_console to no longer use the vram_* convenience pointers. Some of these don't quite make much sense with the multibuffering (and never quite did) as they would be pointing to the FB you're drawing to as opposed to the one currently being displayed.

kernel/arch/dreamcast/include/dc/video.h
```diff
@@ -241,6 +227,16 @@ extern uint32 *vram_l;
...
 
+/** \brief   Set the VRAM convenience pointers.
+    \ingroup video_fb
+
+    This function sets the vram_s and vram_l pointers to specified offset
+    within VRAM. In multibuffered mode it allows manual management of them.
+
+    \param  base            The offset within VRAM to set the base to.
+*/
+void vid_set_vram(uint32_t base);
```
gyrovorbis, line 238:
```suggestion
void vid_set_vram(uintptr_t base);
```
Just talked about this one on Discord, wasn't sure what felt most idiomatic, but we agreed probably `uintptr_t` would actually be best here.

kernel/arch/dreamcast/include/dc/video.h
```diff
@@ -250,19 +246,43 @@ int vid_check_cable(void);
 
     \param  base            The offset within VRAM to set the base to.
 */
-void vid_set_start(uint32 base);
+void vid_set_start(uint32_t base);
```
gyrovorbis, line 249:
```suggestion
void vid_set_start(uintptr_t base);
```

kernel/arch/dreamcast/include/dc/video.h
```diff
@@ -250,19 +246,43 @@ int vid_check_cable(void);
...
+
+/** \brief   Get the VRAM base of a framebuffer.
+    \ingroup video_fb
+
+    This function gets the position of the specified framebuffer within VRAM.
+    Any invalid fb value will be treated as the current framebuffer.
+
+    \param  fb            The number of the framebuffer or -1 for current.
+*/
+uint32_t vid_get_start(int32_t fb);
```
gyrovorbis, line 259:
```suggestion
uintptr_t vid_get_start(ssize_t fb);
```

kernel/arch/dreamcast/include/dc/video.h
```diff
@@ -250,19 +246,43 @@ int vid_check_cable(void);
...
 
+    This function sets the displayed framebuffer to the specified buffer.
+    Unlike vid_set_fb, this does not point the vram pointers to the next
+    framebuffer, allowing for non-linear management of the FBs.
+
+    \param  fb              The framebuffer to display or any other value
+                                to display the next one.
+
+*/
+void vid_set_fb(int32_t fb);
```
gyrovorbis, line 272:
```suggestion
void vid_set_fb(ssize_t fb);
```

kernel/arch/dreamcast/include/dc/video.h
```diff
@@ -250,19 +246,43 @@ int vid_check_cable(void);
...
-    is masked against (vid_mode->fb_count - 1) in order to loop around.
+    allow for tearing-free framebuffer-direct drawing.
+
+    \param  fb              The framebuffer to display or any other value
+                                to display the next one.
 
-    \param  fb              The framebuffer to display (or -1 for the next one).
 */
-void vid_flip(int fb);
+void vid_flip(int32_t fb);
```
gyrovorbis, line 285:
```suggestion
void vid_flip(ssize_t fb);
```

kernel/arch/dreamcast/include/dc/video.h
```diff
@@ -279,14 +299,19 @@ void vid_flip(int fb);
     
     \return                 Old border color value (RGB888)
 */
-uint32 vid_border_color(int r, int g, int b);
+uint32_t vid_border_color(int r, int g, int b);
```
gyrovorbis, line 302:
```suggestion
uint32_t vid_border_color(uint8_t r, uint8_t g, uint8_t b);
```
I feel like we might as well fix this while we're at it... The only thing that makes sense for RGB888 is for each channel to be `uinht8_t`, which makes it easier on the user.

gyrovorbis (commented):
Also, I think for this:
```
#define CT_ANY          -1  /**< \brief Any cable type. Used only internally. */
#define CT_VGA          0   /**< \brief VGA Box */
#define CT_NONE         1   /**< \brief Nothing connected */
#define CT_RGB          2   /**< \brief RGB/SCART cable */
#define CT_COMPOSITE    3   /**< \brief Composite cable or RF switch */
```
which I know isn't your code, but while we're cleaning it up anyway, since it is enumerating independent values from an initial value, should really become:
```
typedef struct cable_type {
    CT_ANY = -1,
     CT_VGA,
     CT_NONE,
     CT_RGB,
      CT_COMPOSITE
 } cable_type_t;
 ```
 
Btw... the first thing I was thinking when seeing the (very welcome) API addition of `vid_enable()` and `vid_disable()` is that then I want a way to query for whether it's enabled... which then made me think it's more natural and orthogonal to perhaps just have:
`void vid_set_enabled(bool value);` for enabling and 
`bool vid_enabled(void);` for checking whether it's enabled?

Then it's more like a pair of get/set functions and is easier to toggle... Plus less API additions to have to add to the library exports and all that... Dunno, that's a personal preference, feel free to ignore if you disagree.

File: kernel/arch/dreamcast/include/dc/video.h (hunk 2 above)
Comment (QuzarDC) [reply kernel/arch/dreamcast/include/dc/video.h:236]: Does that make sense when it isn't a pointer? It's only an offset within the VRAM not an exact memory location.

File: kernel/arch/dreamcast/include/dc/video.h (hunk 4 above)
Comment (QuzarDC) [reply kernel/arch/dreamcast/include/dc/video.h:257]: This would mean also rewriting each instance of the fb index to be size_t too though, as well as fb_curr? My understanding was that size_t was used for things that could be a possible output of sizeof, which *can* be used as an index (for example buf[size_t var] after doing a sizeof(buf)) but that it's not for just any 'ol index value. As is, there's no array that you can plug that fb value into.

**Comment by QuzarDC** [comment]
> Also, I think for this:
> 
> ```
> #define CT_ANY          -1  /**< \brief Any cable type. Used only internally. */
> #define CT_VGA          0   /**< \brief VGA Box */
> #define CT_NONE         1   /**< \brief Nothing connected */
> #define CT_RGB          2   /**< \brief RGB/SCART cable */
> #define CT_COMPOSITE    3   /**< \brief Composite cable or RF switch */
> ```
> 
> which I know isn't your code, but while we're cleaning it up anyway, since it is enumerating independent values from an initial value, should really become:
> 
> ```
> typedef struct cable_type {
>     CT_ANY = -1,
>     CT_VGA,
>     CT_NONE,
>     CT_RGB,
>     CT_COMPOSITE
>  } cable_type_t;
> ```
> 
> Btw... the first thing I was thinking when seeing the (very welcome) API addition of `vid_enable()` and `vid_disable()` is that then I want a way to query for whether it's enabled... which then made me think it's more natural and orthogonal to perhaps just have: `void vid_set_enabled(bool value);` for enabling and `bool vid_enabled(void);` for checking whether it's enabled?
> 
> Then it's more like a pair of get/set functions and is easier to toggle... Plus less API additions to have to add to the library exports and all that... Dunno, that's a personal preference, feel free to ignore if you disagree.
> 
> EDIT: Oh yeah, plus I would say your API additions should make you an `\author` in the header, plus maybe update the `AUTHORS` file for 2024 too? lol.

The Cabletype defines aren't independent at all. They're an unfortunate mix of hardware values (everything but -1) and a convenience define (CT_ANY).

That's a great idea with the vid_set_enabled! I was thinking too MFC and ignored the much better possible design.

File: kernel/arch/dreamcast/include/dc/video.h (hunk 7 above)
Comment (QuzarDC) [reply kernel/arch/dreamcast/include/dc/video.h]: Good catch. Updated.

File: kernel/arch/dreamcast/include/dc/video.h (hunk 2 above)
Comment (QuzarDC) [reply kernel/arch/dreamcast/include/dc/video.h:237]: Read the discussion you mentioned on your discord. Made sense. I believe the most appropriate thing would be for these to be pvr_ptr_t. Unfortunately there's a bit of a snarl where vid_set_start is used by PVR files and passed a uint32, and there are dozens of functions which treat those values and others that interact with them as uint32.

I'll make sure to take note of this though as a cleanup to be done within the PVR code to universally use pvr_ptr_t for addresses in/pointers into the vram.

Comment (gyrovorbis) [reply kernel/arch/dreamcast/include/dc/video.h:237]: Sounds good to me, btw.

File: examples/dreamcast/video/multibuffer/multibuffer.c [hunk 8]
```diff
@@ -0,0 +1,60 @@
...
+    the bfont examples. It draws out four distinct framebuffers then rotates 
+    between them until stopped.
+
+ */
+
+#include <kos.h>
+
+int main(int argc, char **argv) {
+    unsigned short x, y, mb;
+
```
ljsebald, line 19:
```suggestion
```

examples/dreamcast/video/multibuffer/multibuffer.c
```diff
@@ -0,0 +1,60 @@
...
+    for(mb = 0; mb < vid_mode->fb_count; mb++) {
+
+        for(y = 0; y < 480; y++)
+            for(x = 0; x < 640; x++) {
+                int c = (x ^ y) & 0xff;
+                vram_s[y * 640 + x] = ( ((c >> 3) << 12)
+                                      | ((c >> 2) << 5)
+                                      | ((c >> 3) << (mb % 5))) & 0xffff;
+            }
+
```
ljsebald, line 41: I'd rather we had the curly braces on the outer for loop, even if they're technically not necessary... It's just more clear.

```suggestion
        for(y = 0; y < 480; y++) {
            for(x = 0; x < 640; x++) {
                int c = (x ^ y) & 0xff;
                vram_s[y * 640 + x] = ( ((c >> 3) << 12)
                                      | ((c >> 2) << 5)
                                      | ((c >> 3) << (mb % 5))) & 0xffff;
            }
        }

```

ljsebald (changes requested):

ljsebald (approved):

PR #420, opened 2023-12-19
timer_spin_sleep() and thread_sleep() Precision Enhancements

gyrovorbis:
I was working on a PR for enhancing precision everywhere on everything, but I realized there's so much that I might not be able to get it all done over Christmas break... so I'm going to do it in smaller pieces. Here's the next phase:

1. Timer API
    - deprecated `timer_spin_sleep()` and made it call into `timer_spin_sleep_ms()`
    - added `timer_spin_sleep_us()` and `timer_spin_sleep_ns()`
2. Thread API
    - deprecated `thd_sleep()` and made it call into `thd_sleep_ms()`
    - added declarations for `thd_sleep_us()` and `thd_sleep_ns()`
    - implementations of `thd_sleep_us()` and `thd_sleep_ns()` are currently just falling back to ms precision. `genwait_wait()` will need a precision enhancement as well for the threading back-end to hook into. That'll come at a later phase.


kernel/arch/dreamcast/kernel/timer.c
```diff
@@ -152,25 +152,36 @@ int timer_clear(int which) {
...
-void timer_spin_sleep(int ms) {
-    timer_prime(TMU1, 1000, 0);
+static void timer_spin_sleep_ticks(unsigned speed, int ticks) {
+    timer_prime(TMU1, speed, 0);
     timer_clear(TMU1);
     timer_start(TMU1);
 
-    while(ms > 0) {
+    while(ticks > 0) {
         while(!(TIMER16(tcrs[TMU1]) & UNF))
```
QuzarDC, line 163: Would it make sense to have this be based around a generic version of the newer timer_getticks that takes in the timer # as a param?

gyrovorbis: yeah, that might actually be a good idea.

pcercuei:
I would drop `timer_spin_sleep_ms()`.

Even one millisecond is a lot of time to busy-wait for. At that point it would be better to yield the thread. For reference, in the Linux kernel the threshold to sleep the thread vs. busy-waiting is 20 *micro* seconds.

PR #422, opened 2023-12-19
Logging Enhancements, fs.c Cleanup

gyrovorbis:
While researching for why `stdout` has been so slow lately and why `stdout`/`stderr` redirection was broken, I saw a few missing API calls in the logger APIs, a few missed optimizations, and some clean-up that needed to be done. 

A. dbglog and dbgio printf-style interfaces with `...` variadic arguments have all had equivalent `va_list`-based functions added to them. Why? 
    1) Functions with variadic argument lists cannot chain into each other. This means when dealing with `printf()`-style formatters, the very first function must process the variadic args into a temporary buffer before passing it on to the next function which will attempt the same thing. It's unnecessary overhead for intermediate format processing.
    2) Many languages can bind to C APIs with `va_list`-style additional arguments, but not also `...` variadic arguments.

Because of these two points, I'd like to propose every variadic argument list function within the codebase have an equivalent `va_list` function.
    
B. dbglog and dbgio interfaces have had their internal static buffers for temporary string storage removed, along with their synchronization primitives and are instead doing string processing within a stack-local temporary variable. This has the added benefit of not wasting 2KB consistently at runtime plus also no longer needing a lock for reentrancy. 

C. I was in `fs.c` doing some work and noticed how bad a bunch of the styling and formatting was. Very inconsistent with the rest of the codebase. I decided that I might as well clean it up.


kernel/fs/fs.c
```diff
@@ -36,36 +36,34 @@ something like this:
...
 #include <kos/dbgio.h>
 
 /* File handle structure; this is an entirely internal structure so it does
    not go in a header file. */
 typedef struct fs_hnd {
-    vfs_handler_t   *handler;   /* Handler */
-    void *      hnd;        /* Handler-internal */
-    int     refcnt;     /* Reference count */
+    vfs_handler_t *handler;   /* Handler */
+    void *        hnd;        /* Handler-internal */
```
andressbarajas, line 46:
```suggestion
    void         *hnd;        /* Handler-internal */
```

kernel/libc/newlib/newlib_isatty.c
```diff
@@ -20,6 +20,7 @@ int isatty(int fd) {
 
 int _isatty_r(struct _reent *reent, int fd) {
     (void)reent;
+    (void)fd;
```
QuzarDC, line 23: Why is this needed if fd is used in the function?

gyrovorbis: Good point. I had modified the function to take out the following check temporarily, added that to suppress the warning, then put the check back in and forgot to remove this. lol.

include/kos/dbgio.h
```diff
@@ -1,7 +1,9 @@
 /* KallistiOS ##version##
 
    kos/include/dbgio.h
```
andressbarajas, line 3:
```suggestion
   include/kos/dbgio.h
```

kernel/arch/dreamcast/fs/fs_dclsocket.c
```diff
@@ -634,7 +637,7 @@ static int dcls_fake_shutdown(void) {
     return 0;
 }
 
-static int dcls_writebuf(const uint8 *buf, int len, int xlat) {
+static int dcls_writebuf(const uint8_t *buf, size_t len, int xlat) {
```
QuzarDC, line 640: Will this require casting len when passing to htonl later on?

kernel/arch/dreamcast/kernel/panic.c
```diff
@@ -12,6 +13,6 @@
    can recover, call this. This is a pretty standard tactic from *nixy
    kernels which ought to be avoided if at all possible. */
 void arch_panic(const char *msg) {
-    printf("kernel panic: %s\r\n", msg);
+    fprintf(stderr, "kernel panic: %s\r\n", msg);
```
QuzarDC, line 16: We don't use stderr anywhere else in the kernel (except in malloc where it wasn't rewritten from the desktop version). Pretty sure if this should be anything it would be a dbglog, but it would be with a level to ensure it always shows up.

kernel/debug/dbgio.c
```diff
@@ -151,39 +152,34 @@ int dbgio_write_buffer_xlat(const uint8 *data, int len) {
...
 }
 
-// Not re-entrant
-static char printf_buf[1024];
-static spinlock_t lock = SPINLOCK_INITIALIZER;
+int dbgio_vprintf(const char *fmt, va_list *var_args) { 
+    char buffer[1024];
+    int i = vsnprintf(buffer, sizeof(buffer), fmt, *var_args);
+
+    if(i >= (int)sizeof(buffer))
```
QuzarDC, line 165: Shouldn't this trigger a growth of the buffer? At least it's better than before where it won't just overflow print_buf. Thinking something like:

    char *overflow = &buffer;
    int i = vsnprintf( all that );
    
    if(i >= sizeof) {
        overflow = malloc(i + 1);
        vsnprintf (overflow, i, fmt, *var_args);
    }
    dbgio_write(overflow)
    
    if(overflow != &buffer)
       free(overflow);
  

And then also should be error checking vsnprintf for possible negative return for encoding issues.

QuzarDC (commented):
I know it's still in draft, but figured I'd give some review. Nothing too big.

QuzarDC:
From what I can tell the changes here are all generally moot at this point. We had substantial updates that have allowed stdio to work better, and dbglog no longer chains into dbgio in the way that this would benefit from. The majority of the rest of it was cleanups that have all been done in the meantime.

PR #423, opened 2023-12-19
Increased Resolution of C's clock() to Microseconds

gyrovorbis:
- Added #define to kos/time.h to provide newlib with an initial value for `CLOCKS_PER_SEC`
- Modified newlib_times.c to use the microsecond timer, which is what backs the `clock()` C stdlib function
- Updated CHANGELOG

EDIT: Just wanted to mention why this change is important. With the newfound `TMU` precision, we can increase to micro and nanosecond to timing; however, as noted here: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/time.h.html

` The value of CLOCKS_PER_SEC shall be 1 million on XSI-conformant systems. `

I figured we might as well be XSI-conformant and bump this bad boy up to microsecond resolution...

### DIFF
--- a/doc/CHANGELOG
+++ b/doc/CHANGELOG
@@ -212,7 +212,7 @@ KallistiOS version 2.1.0 -----------------------------------------------
 - DC  Added basic example for micropython KOS port [Aaron Glazer == AG]
 - DC  Improved performance of IRQ context save / restore [PC]
 - DC  Increased resolution of TMU timers + date/time functions [FG && PC]
-- DC  Increased resolution of C's clock() and CLOCKS_PERS_SEC to microseconds [FG]
+- *** Increased resolution of clock() and CLOCKS_PER_SEC to microseconds [FG]
 
 KallistiOS version 2.0.0 -----------------------------------------------
 - DC  Broadband Adapter driver fixes [Megan Potter == MP]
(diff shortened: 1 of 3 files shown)


andrewapperley (approved):
Looks good to me! Easy enough change.

doc/CHANGELOG
```diff
@@ -212,6 +212,7 @@ KallistiOS version 2.1.0 -----------------------------------------------
 - DC  Added basic example for micropython KOS port [Aaron Glazer == AG]
 - DC  Improved performance of IRQ context save / restore [PC]
 - DC  Increased resolution of TMU timers + date/time functions [FG && PC]
+- DC  Increased resolution of C's clock() and CLOCKS_PERS_SEC to microseconds [FG]
```
ljsebald, line 215:
```suggestion
- DC  Increased resolution of clock() and CLOCKS_PER_SEC to microseconds [FG]
```

ljsebald (changes requested):
See above.

gyrovorbis:
> Merging is blocked
> Merging can be performed automatically once the requested changes are addressed.

Oh yeah, good point. Addressed.

gyrovorbis:
> See above.

Actually, based on that comment, is the above entry in the CHANGELOG not also incorrect? TMU and stdlib crap would also not be DC-specific, would it?

ljsebald:
> > See above.
> 
> Actually, based on that comment, is the above entry in the CHANGELOG not also incorrect? TMU and stdlib crap would also not be DC-specific, would it?

TMU is certainly DC-specific. This change changes nothing in the arch/dreamcast directory and only changes things in the supposedly non-arch-specific directories, hence marking it not as DC only.

That said, the distinction is kinda moot, since there aren't any other arches anyway.

gyrovorbis:
OOOOOH. I keep getting the subarch confused with the architecture. In my mind, the `TMU` stuff was "architecture-independent" for being also a change that would've influenced NAOMI. Now it all makes sense. :)

ljsebald (approved):

PR #425, opened 2023-12-20
Separate Performance Counter Driver

gyrovorbis:
This PR is to separate the performance counter driver and its API from the timer driver and API. We have been discussing this for awhile now, and we agree that it was a mistake to stick the two separate peripheral drivers into a single file when they are separate. The only reason for this was because we thought we would be using the performance counters as timers before we found out that they were unable to keep time correctly when the processor goes to sleep.

This PR is the first of a big endeavor to build a profiler and useful higher-level user-facing API around the low-level performance counter driver. More will be following.

Anyway, what this PR entails:
- Created `perfctr.h` and `perfctr.c` for the new driver
- Went back to actually implement the `timer_ns_gettime64()` function which was previously clashing with the existing perf counter version (which didn't work as intended), since we can now do it with `TMU2`. 
- Cleaned up and redid the typing of the whole API. There was just no reason to make everything a `#define`, when every set of data was is a zero-starting sequential value which the user will never be masking or bitbanging directly. Instead, by using enumerations, we've made the API strongly typed/type-safe and better organized, without actually changing value names
- Added a way to query whether the perf counter timer is enabled or not 
- Fleshed out Doxygen more 
- Commented bit fields in the driver 

EDIT: Yes, this is technically a teensy API breaking change, because now the symbols from `perfctr.h` are not auto-included within `timer.h`... but I've personally spoken with a LOT of people, and the only people using this stuff out there in the wild is @DC-SWAT along with @andressbarajas, who I will make sure know about the change. Worst case, if this is unacceptable, we auto-include `perfctr.h` within `timer.h`.


kernel/arch/dreamcast/include/arch/perfctr.h
```diff
@@ -0,0 +1,271 @@
...
+    use perf_cntr_timer_disable() or perf_cntr_start() to reconfigure it for
+    something else. When disabled, the timer will simply fall through to use
+    timer_ns_gettime64() from the timer driver, decreasing the resolution of
+    each tick to 80ns.
+
+    \warning
+    The performance counter timer is only counting *active* CPU cycles. This
+    means that when KOS's thread scheduler uses the "sleep" instruction, 
+    putting the CPU to sleep, these counters cease to record elapsed time. 
+    Because of this, they should only be used to measure small delats that
```
andressbarajas, line 209: Delats?

kernel/arch/dreamcast/include/arch/perfctr.h
```diff
@@ -0,0 +1,271 @@
...
+    to using the timer_ns_gettime64(), which decreases the resolution of each
+    tick to 80ns.
+
+    \retval true    The nanosecond timer is configured and running
+    \retval false   The nanosecond timer is not configured and/or isn't
+                    running
+*/
+bool perf_cntr_timer_enabled(void);
+
+/** \brief  Get the current uptime of the system (in nanoseconds).
```
andressbarajas, line 257: Since we found out that the sleep instruction messes with the timing wouldn't the documentation on the function be incorrect?

andressbarajas: Maybe add a warning to this function that it doesnt include the sleep time?

gyrovorbis: SHIT. I did that in the topic, but you're right, it needs to be done there as well.

kernel/arch/dreamcast/include/arch/perfctr.h
```diff
@@ -0,0 +1,281 @@
...
+
+    \retval true    The nanosecond timer is configured and running
+    \retval false   The nanosecond timer is not configured and/or isn't
+                    running
+*/
+bool perf_cntr_timer_enabled(void);
+
+/** \brief  Gets elapsed CPU time (in nanoseconds)
+
+    This function retrieves the number total amount of \a active CPU time since
```
andressbarajas, line 263:
```suggestion
    This function retrieves the total amount of \a active CPU time since
```

kernel/arch/dreamcast/include/arch/perfctr.h
ljsebald, line 1: I feel that this should probably be in the `kernel/arch/dreamcast/include/dc` directory, rather than the `kernel/arch/dreamcast/include/arch` one. The `arch` directory is more for common functionality that one might expect all ports to have, where the `dc` one is for Dreamcast-specific hardware.

ljsebald (changes requested):

kernel/arch/dreamcast/include/dc/perfctr.h
```diff
@@ -0,0 +1,278 @@
+/* KallistiOS ##version##
+
+   arch/dreamcast/include/perfctr.h
+   Copyright (C) 2023 Andy Barajas
+   Copyright (C) 2023 Falco Girgis
+   
+*/
+
+/** \file    arch/perfctr.h
```
andressbarajas, line 9:
```suggestion
/** \file    dc/perfctr.h
```

kernel/arch/dreamcast/include/dc/perfctr.h
```diff
@@ -0,0 +1,278 @@
...
+    This file contains the low-level driver for interacting with and
+    utilizing the SH4's two Performance Counters, which are primarily
+    used for profiling and performance tuning. 
+
+    \author MoopTheHedgehog
+    \author Andy Barajas
+    \author Falco Girgis
+*/
+
+#ifndef __ARCH_PERFCTR_H
```
andressbarajas, line 22:
```suggestion
#ifndef __DC_PERFCTR_H
```

kernel/arch/dreamcast/include/dc/perfctr.h
```diff
@@ -0,0 +1,278 @@
...
+#define __ARCH_PERFCTR_H
```
andressbarajas, line 23:
```suggestion
#define __DC_PERFCTR_H
```

kernel/arch/dreamcast/include/dc/perfctr.h
```diff
@@ -0,0 +1,278 @@
...
+*/
+uint64_t perf_cntr_timer_ns(void);
+
+/** @} */
+
+/** @} */
+
+__END_DECLS
+
+#endif /* __ARCH_PERFCTR_H */
```
andressbarajas, line 278:
```suggestion
#endif /* __DC_PERFCTR_H */
```

kernel/arch/dreamcast/include/dc/perfctr.h
```diff
@@ -0,0 +1,278 @@
+/* KallistiOS ##version##
+
+   arch/dreamcast/include/perfctr.h
```
andressbarajas, line 3:
```suggestion
   arch/dreamcast/include/dc/perfctr.h
```

ljsebald (approved):

PR #427, opened 2023-12-21
Added CLOCK_PROCESS_CPUTIME_ID + clock_getcpuclockid()

gyrovorbis:
- Removed some unecessary defines from time.h for different clock types, apparently defining the correct POSIX support makes Newlib do it for you.
- Removed support for `CLOCK_THREAD_CPUTIME_ID`, because it was a lie. We don't support that.
- Made `CLOCK_PROCESS_CPUTIME_ID` actually measure real active CPU time now, by using the performance counters (which don't run during sleep), when they're configured properly
- added `clock_getcpuclockid()`, which gets the `CLOCK_PROCESS_CPUTIME` clock ID for the given process ID. Yes, we're single-process, but the standard actually states it should then work with just pid 0 (current) and fail for any other pid.

NOTE: This PR depends on https://github.com/KallistiOS/KallistiOS/pull/425. So I'll take it out of draft once it's merged.


andressbarajas:
@gyrovorbis This can be taken out of draft now that #425 was merged

gyrovorbis:
> @gyrovorbis This can be taken out of draft now that #425 was merged

Good call. On it!

kernel/libc/posix/clock_gettime.c
```diff
@@ -1,24 +1,43 @@
...
+#include <time.h>
 #include <errno.h>
+#include <assert.h>
+#include <stdlib.h>
+
+int clock_getcpuclockid(pid_t pid, clockid_t *clock_id) {
+    /* pid of 0 means the current process,
+       and we only support a single process. */
+    if(pid)
+        return ENOSYS;
```
QuzarDC, line 21: Would it be more appropriate for this to be ESRCH? Based on this: https://pubs.opengroup.org/onlinepubs/009696899/functions/clock_getcpuclockid.html it's what should be returned when you can't find the process passed as pid.

QuzarDC:
Was seeing if I can review this, and was wondering: what's the purpose of using the perfcounter clock for a single process if it's a single process anyways? Is the idea that the user would be manipulating the runtime of the perfcounter as desired in the interim to mimic the normal usage of the functionality? 

I haven't been following much the perfcounter stuff so maybe with an understanding of it, the answer is obvious. But if that's what it is, perhaps some notes in the comments make sense to convey this.

gyrovorbis:
> Was seeing if I can review this, and was wondering: what's the purpose of using the perfcounter clock for a single process if it's a single process anyways? Is the idea that the user would be manipulating the runtime of the perfcounter as desired in the interim to mimic the normal usage of the functionality?
> 
> I haven't been following much the perfcounter stuff so maybe with an understanding of it, the answer is obvious. But if that's what it is, perhaps some notes in the comments make sense to convey this.

Yeah, it's kind of a weird thing you'd have to have either read the perf counter documentation to understand or have to have followed along with this... crazy Saga to really know...

So I *did* try to make it clear that the perfcounter timer had this... weird little quirk in how it works that makes it totally ill-suited to be a regular monotonic timer like what we've been doing with the `TMU` driver, but that actually makes it work for a `CPU-time counter`.... It does not count anything or increment at all when the `SLEEP` instruction is issued... This makes it worthless for maintining actual wall or monotonic time, but *does* make it useful for measuring actual time spent within our overall "single KOS process" for measuring actual useful/active execution time, which is what `CLOCK_PROCESS_CPU_TIME` is supposed to be representing... So basically by dumb luck, that's what our performance counter is actually monitoring.

kernel/libc/posix/clock_gettime.c
gyrovorbis: Good call!

gyrovorbis: Shit. Okay. Now I remember what the deal was with this... So I was returning `ENOSYS` rather than `ESRCH`, because it looks as though we're returning the thread ID from `getpid()` or whatever standard POSIX process ID function or whatever... So it would be kind of weird to say we couldn't find the pid when they could potentially pass it straight from that function into here... Thought the "No support" error made more sense given that context? Maybe @ljsebald has an opinion on this one? I can see your point too...

ljsebald: I agree with @QuzarDC that this should be `ESRCH`.

gyrovorbis: Thank you, both! That's what I was waiting to hear. On it!

kernel/libc/posix/clock_gettime.c
```diff
@@ -1,24 +1,43 @@
...
+    return 0;
+}
 
 int clock_getres(clockid_t clk_id, struct timespec *ts) {
     switch(clk_id) {
         case CLOCK_REALTIME:
         case CLOCK_MONOTONIC:
         case CLOCK_PROCESS_CPUTIME_ID:
-        case CLOCK_THREAD_CPUTIME_ID:
+        /* case CLOCK_THREAD_CPUTIME_ID: */
```
ljsebald, line 35:
```suggestion
```

ljsebald (changes requested):
One tiny cleanup....

ljsebald (approved):

Issue #264, opened 2023-08-09
libdream/800x608 example gives error "invalid generic mode 1002"


darcagn:
Tested using GCC 9.3.0 and 13.2.0

Example prints error `vid_set_mode: invalid generic mode 1002`.

This is output:
![Screenshot_20230809_141201](https://github.com/KallistiOS/KallistiOS/assets/5105103/a06d6d01-6542-40ab-829b-1277776e7ad0)

`kernel/arch/dreamcast/hardware/video.c` says this video mode is broken:
![Screenshot_20230809_141500](https://github.com/KallistiOS/KallistiOS/assets/5105103/34fe449a-3e2c-43e4-8c85-a37d74a20838)

![image](https://github.com/KallistiOS/KallistiOS/assets/5105103/cf8ae659-00f2-4c99-b5fb-900b22e91248)

So perhaps it should not be a demo?

ljsebald:
The example itself has a comment about things being broken potentially too, so...
```c
/* This sample program shows off 800x608, the mythical "unsupported" video
   mode that Sega will probably complain doesn't work (like they did for
   our 320x240 modes earlier on ;-). This will probably only work on a VGA
   monitor and I highly recommend that you don't try it on a non-multisync.
   Still needs some tweaking.

   In here I also demonstrate how to use the bfont BIOS font routines, and
   how to initialize only selected parts of KOS (resulting in a much
   smaller and less polluted binary).

 */
```

gek169:
Let this issue be accursed

andressbarajas:
Might have no hope of fixing it but we can look at dreamhal video modes and replace them?

https://github.com/sega-dreamcast/dreamhal/blob/3620f5b96af2668d8c065bf78abf219224d236b8/startup/startup_support.c#L689

gyrovorbis:
Can someone make sure we also add a `printf()` to the example that says it's currently not supported properly or something, just so in `v1.2.0`, it doesn't look as broken to the users? Not saying we need to add support for this release, though.

gyrovorbis:
Closing this issue, since we've removed it for now... If we want to track trying to implement it correctly or fix it and bring that back, we can open another issue for it. I think the scope of what this one was capturing is now complete.

PR #429, opened 2023-12-21
Added Disclaimer to 800x608 Resolution Example

gyrovorbis:
- Because users, including our own developers, are continually being confused by the garbled output after not fully reading the example's comment. :)

Related Issue: https://github.com/KallistiOS/KallistiOS/issues/264


ljsebald (approved):
Sensible to have said warning.

PR #431, opened 2023-12-22
Added Toolchain Option to Enable Newlib's Multibyte/Wide Character Support

gyrovorbis:
Someone actually asked about this today... maybe they were joking, but it doesn't seem that unreasonable. I've really been considering how I'd be handling localization lately on a game, and this could be a potential avenue. Might as well give them the option, because I tested, and it works fine!

- Added check for variable and application of flag to Newlib in init.mk
- Added a (default disabled) flag for enabling it with a description in config.mk.stable.sample

### DIFF
--- a/utils/dc-chain/config/config.mk.10.5.0.sample
+++ b/utils/dc-chain/config/config.mk.10.5.0.sample
@@ -138,6 +138,13 @@ thread_model=kos
 # ptrdiff_t, intmax_t, and sized integral types.
 newlib_c99_formats=1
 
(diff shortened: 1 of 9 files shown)


ljsebald:
The changes should probably be propagated to the other configuration samples too.

darcagn (approved):

darcagn:
I updated the other configurations. I can't merge now since I'm the last pusher, so feel free to merge this @gyrovorbis or @ljsebald.

ljsebald (approved):

PR #433, opened 2023-12-23
Properly clean dc-chain builds from SH4 Pass1 & GDB

cepawiel:
Due to the split of Pass1 & Pass2 build folders the clean_up helper needed to be added to Pass1 as `build-gcc-sh4-*-pass1` wasn't being cleaned up when `erase=1` was specified in the config. This change also allowed the removal of an "hack" for removing the ARM Pass1 build folder. 

Also added the clean_up helper to the gdb install step to similarly delete the build folder when `erase=1`.


sizious:
That's great, I noticied that in the past but you were faster than me to submit something 👍

darcagn (approved):

PR #435, opened 2023-12-24
Add some gldc examples

andressbarajas:
Fixed these examples for gldc (from kgl) by using a working pvr-texture.c file


examples/dreamcast/gldc/basic/gl/pvr-texture.c
ljsebald, line 1: I feel like it might be a good idea to perhaps mainline this into gldc at some point. @Kazade thoughts?

Kazade: The pvr-texture file? Do you mean so people can use it from gldc?

andressbarajas: Yep, I think he is referencing the pvr-texture.c file.  A lot of the gldc examples in this PR depend on it so its source is copied many times over.  You also already have it in one of your samples in your gitlab repo.

ljsebald: Yes, that's exactly what I'm referring to. Would be a very nice thing to add to the KOS-specific stuff (like the rest of the stuff in glkos.h).

Kazade: No texture loading stuff doesn't belong in libGL. The glkos.h header is for GL extensions and context creation (and if we had a libEGL we wouldn't even have the context creation stuff there).

I think it would be better if pvr-texture was a library in kos-ports that people can use it they want, many people also use .dtex (texconv) format so perhaps a libDCTexture or something could include loaders for both.

andressbarajas: Sounds like a great idea.  I'll try and tackle it unless someone else wants to do it.

andressbarajas: @Kazade Sorta related, do you think LibGL could have a function that takes a pointer to texture data already residing in the PVR?  Use case is decoding YUV420 video frames with pl_mpeg and using the PVR to convert them to Dreamcasts YUV422 format which means they are already in texture memory.  Maybe like this:

```c
GLAPI void APIENTRY glTexMemoryImage2DARB(GLenum target,
        GLenum internalformat,
        GLsizei width,
        GLsizei height,
        GLsizei imageSize,
        const GLvoid *data);
```

So we can use libGL to render video frame and maybe other stuff on top (Press Start).  I know libGL currently doesnt support Dreamcasts YUV422 but that shouldn't be too hard to add looking at the code.  Just wanted your opinion before I start a PR for libGL.

Kazade: I'd need to understand better the mechanism of using the PVR to decode stuff to work out what would be the best approach. For example, maybe glTexImage could do that with a different source format parameter to indicate the source data is YUV420. Could you file an issue in GLdc? and then I'll try to find time to look at it at some point.

ljsebald (approved):

Issue #441, opened 2023-12-28
Create example(s) to demonstrate all video modes


QuzarDC:
KOS currently supports a number of video modes that have no examples that demonstrate their use. Among them 256x256 and 720x480. With the long-time misconceptions around 800x608, (and the newer misconceptions around Moop's HD modes) it would be useful to ensure there are examples that can demonstrate all of our video modes (especially the ones above 640x480).

I may take this up myself, but wanted to make sure to put in an issue to record the gap. To my mind the examples should also clearly be able to demonstrate if the mode is not working correctly, so perhaps rendering reference test patterns rather than just an xor pattern or a texture that gets stretched.

gyrovorbis:
@QuzarDC didn't this actually just get checked in as part of your video mode refactor endeavor?

QuzarDC:
No, that was merely a demonstration of multi-buffering since we had no example that used it.

PR #442, opened 2023-12-28
Remove non-functional 800x608 mode

QuzarDC:
As title. Addresses #264 and the goal of having all examples working for the 2.1.0 milestone. Some notes:

- Going back to the libdream sources and configuring the mode to work as libdream had it simply didn't work right. The notes and in the import of examples to KOS indicate that it was always 'experimental'. When set my monitor understood it as 800x600 but the actual screen space presented was somewhere around 600x576.
- Looking through DreamHAL's implementation of 'HD modes' ( https://github.com/sega-dreamcast/dreamhal/blob/master/inc/startup_support.h#L222 and https://github.com/sega-dreamcast/dreamhal/blob/master/startup/startup_support.c#L609 ) seems to implicitly explain how an 800x608 mode with a matching framebuffer is not possible. While it's not immediately apparent, each of the higher resolution modes defined require having smaller framebuffers and using timings to have modern monitors interpret the output as being in a higher resolution (so the monitor would see 800x600 but in software a 544x608 framebuffer would be used.
- There shouldn't really be any concerns for the historical preservation of the libdream examples. We have already moved a number of them out of the same folder (for instance hello and ghettoplay). The intent was to provide regression against libdream and demonstrate how to move from libdream to KOS. Not only is that no longer an issue, but since the example never worked in KOS it was always just misleading. On the same lines I've opened #441 in order to add better examples.


ljsebald:
Perhaps someday in the future we'll have to reevaluate the video subsystem and look into the modes in dreamhal (with non-matching framebuffer sizes), alas that day is not today.

ljsebald (approved):
Normally I wouldn't so quickly approve something that removes functionality... But removing functionality that has been explicitly tagged as broken forever is a different story. As I said before, perhaps we'll eventually revisit the video modes situation, but that is for another day.

PR #454, opened 2024-01-08
Fixed Build Issue with -Og for inode.c

gyrovorbis:
- When using debug optimization level, inode.c:1270 was issuing a warning about "dent" being used without being initialized. This warning was treated as an error.
- Initialized "dent" to NULL at the beginning of the function.

Why does this only happen with `-Og`? ...I have no idea.

Here's the actual error:
```
make[2]: Entering directory '/opt/toolchains/dc/kos/addons/libkosext2fs'
kos-cc  -c ext2fs.c -o ext2fs.o
kos-cc  -c bitops.c -o bitops.o
kos-cc  -c block.c -o block.o
kos-cc  -c inode.c -o inode.o
inode.c: In function ‘ext2_inode_by_path’:
inode.c:1270:22: error: ‘dent’ may be used uninitialized [-Werror=maybe-uninitialized]
 1270 |     *inode_num = dent->inode;
      |                  ~~~~^~~~~~~
inode.c:1045:20: note: ‘dent’ was declared here
 1045 |     ext2_dirent_t *dent;
      |                    ^~~~
cc1: all warnings being treated as errors
make[2]: *** [/opt/toolchains/dc/kos/Makefile.rules:13: inode.o] Error 1
```

### DIFF
--- a/addons/libkosext2fs/inode.c
+++ b/addons/libkosext2fs/inode.c
@@ -1042,7 +1042,7 @@ int ext2_inode_by_path(ext2_fs_t *fs, const char *path, ext2_inode_t **rv,
     int blocks, i, block_size;
     uint8_t *buf;
     uint32_t *ib;
-    ext2_dirent_t *dent;
+    ext2_dirent_t *dent = NULL;
     int err = 0;
     size_t tmp_sz;
     char *symbuf;


ljsebald:
I mean... technically this won't fix anything, and if we somehow got to where it is complaining that it is being used uninitialized, it'd do a `NULL` dereference. This probably should be fixed with not only doing that, but also adding a check at the bottom for if `dent` is `NULL`, and exiting accordingly with an appropriate error (perhaps `ENOENT`?).

Mind you, I'd imagine that it isn't actually possible to get to that point with `dent` not set (or it would be triggering that warning on more than just `-Og`), but...

gyrovorbis:
> I mean... technically this won't fix anything, and if we somehow got to where it is complaining that it is being used uninitialized, it'd do a `NULL` dereference. This probably should be fixed with not only doing that, but also adding a check at the bottom for if `dent` is `NULL`, and exiting accordingly with an appropriate error (perhaps `ENOENT`?).
> 
> Mind you, I'd imagine that it isn't actually possible to get to that point with `dent` not set (or it would be triggering that warning on more than just `-Og`), but...

Yeah, I was actually thinking that it was probably some GCC bug (there are a few other places where warnings only appear at certain optimization levels that are just wrong)... but I have to admit that I did not sit there and trace this through to ensure that it's just a false alarm.. I can go ahead and do that.

gyrovorbis:
Alrighty, after carefully studying the code, yes, it does appear that for whatever reason, at this particular optimization level, GCC is full of it... I cannot see how control flow could allow for execution to reach that point without `dent` being set to a valid pointer.

ljsebald (approved):

PR #456, opened 2024-01-09
Implemented Newlib try_lock() Functions

gyrovorbis:
So this one really isn't going to do a whole lot, considering that the functions I've implemented clearly weren't being used, or they'd be aborting, PLUS I did a search through the Newlib repository and couldn't find anywhere where they were actually used; however, I did want to go ahead and implement these, since they are part of what Newlib expects to be there, and there doesn't seem to be any real reason not to have them now that we have `spinlock_trylock()`. 

Oh, also, our function signatures on those two were incorrect. They should return an `int`. 

- `__newlib_lock_try_acquire()` and `__newlib_lock_try_acquire_recursive()` were stubbed out to assert and abort due to not being implemented
- Implemented both using the new `spinlock_trylock()` API
- Also had to fix their prototypes, since they both should be returning an integer for whether the lock was successfully obtained

### DIFF
--- a/kernel/libc/newlib/lock_common.c
+++ b/kernel/libc/newlib/lock_common.c
@@ -45,17 +45,7 @@ void __newlib_lock_close_recursive(__newlib_recursive_lock_t *lock) {
 }
 
 void __newlib_lock_acquire_recursive(__newlib_recursive_lock_t *lock) {
-    int old;
-    int iscur;
-
-    // Check to see if we already own it. If so, everything is clear
-    // to incr nest. Otherwise, we can safely go on to do a normal
-    // spinlock wait.
-    old = irq_disable();
-    iscur = lock->owner == thd_current;
-    irq_restore(old);
-
-    if(iscur) {
+    if(lock->owner == thd_get_current()) {
         lock->nest++;
         return;
     }
@@ -64,7 +54,7 @@ void __newlib_lock_acquire_recursive(__newlib_recursive_lock_t *lock) {
     spinlock_lock(&lock->lock);
 
     // We own it now, so it's safe to init the rest of this.
-    lock->owner = thd_current;
+    lock->owner = thd_get_current();
     lock->nest = 1;
 }
 
--- a/kernel/libc/newlib/lock_common.c
+++ b/kernel/libc/newlib/lock_common.c
@@ -1,8 +1,9 @@
 /* KallistiOS ##version##
 
    lock_common.c
-   Copyright (C)2004 Megan Potter
-
+   Copyright (C) 2004 Megan Potter
+   Copyright (C) 2024 Falco Girgis
+   Copyright (C) 2025 Eric Fradella
 */
 
 #include <assert.h>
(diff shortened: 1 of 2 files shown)


gyrovorbis:
Moving back to draft temporarily since zcrc pointed out the race conditionz!

darcagn:
Removed merge commits and rebased this commit on top of current `master`.

This looks OK to me?

Paul said in Discord on January 9, 2024:
> Your `lock->nest++` isn't atomic, and you perform it without holding the spinlock, so two concurrent calls can race.

But it _is_ performed while holding the lock. If `iscur` is true or `spinlock_trylock(&lock->lock)` is true, we are the holding the lock, so where's the race condition?

Either way, if I'm wrong, this is a simple PR so let's take it out of draft and knock it out.

darcagn (dismissed):

pcercuei:
> But it _is_ performed while holding the lock. If `iscur` is true or `spinlock_trylock(&lock->lock)` is true, we are the holding the lock, so where's the race condition?

`iscur` is only valid as long as IRQs are disabled. But it is checked after the IRQs are re-enabled, so it races.

kernel/libc/newlib/lock_common.c
```diff
@@ -68,12 +67,31 @@ void __newlib_lock_acquire_recursive(__newlib_recursive_lock_t * lock) {
...
-    (void)lock;
-    assert_msg(0, "We don't support try_acquire");
+/* Similar to __newlib_lock_acquire_recursive(), except that it can
+   fail to obtain the lock. */
+int __newlib_lock_try_acquire_recursive(__newlib_recursive_lock_t *lock) {
+    int old;
+    int iscur;
+
+    old = irq_disable();
+    iscur = lock->owner == thd_current;
```
QuzarDC, line 77: `thd_get_current()`

kernel/libc/newlib/lock_common.c
```diff
@@ -68,12 +67,31 @@ void __newlib_lock_acquire_recursive(__newlib_recursive_lock_t * lock) {
...
+    irq_restore(old);
+
+    if(iscur) {
+        lock->nest++;
+        return 1;
+    }
+
+    if(spinlock_trylock(&lock->lock)) {
+        lock->owner = thd_current;
```
QuzarDC, line 86: `thd_get_current()`

We should use the getter rather than pulling it directly.

darcagn:
> > But it _is_ performed while holding the lock. If `iscur` is true or `spinlock_trylock(&lock->lock)` is true, we are the holding the lock, so where's the race condition?
> 
> `iscur` is only valid as long as IRQs are disabled. But it is checked after the IRQs are re-enabled, so it races.

How do we lose the lock from the time IRQs are disabled to when `iscur` is checked?

kernel/libc/newlib/lock_common.c
```diff
@@ -68,12 +67,31 @@ void __newlib_lock_acquire_recursive(__newlib_recursive_lock_t * lock) {
...
+    int iscur;
+
+    old = irq_disable();
+    iscur = lock->owner == thd_current;
+    irq_restore(old);
+
+    if(iscur) {
+        lock->nest++;
+        return 1;
+    }
```
QuzarDC, line 83: This'll race between the restore of the irq and the testing of `iscur`.

I think it could be solved by having something like:

```
{
    irq_disable_scoped();
    if(lock->owner == thd_get_current()) {
        lock->nest++;
        return 1;
    }
}

QuzarDC:
> > > But it _is_ performed while holding the lock. If `iscur` is true or `spinlock_trylock(&lock->lock)` is true, we are the holding the lock, so where's the race condition?
> > 
> > 
> > `iscur` is only valid as long as IRQs are disabled. But it is checked after the IRQs are re-enabled, so it races.
> 
> How do we lose the lock from the time IRQs are enabled to when `iscur` is checked?

Threading timer fires, and another thread tries, and succeeds, to take the lock before we switch back.

gyrovorbis:
Yes plz, don’t think that existed when I write this PR.

kernel/libc/newlib/lock_common.c
gyrovorbis: Good catch.

pcercuei:
Actually I think @darcagn is right and there is no race. If `iscur` is true the current thread holds the lock, which will still be the case after the unlock - because the only way it wouldn't be true is if another thread unlocks it while not holding it, which would be a fault anyway.

With that said, there is no reason to disable/enable IRQs there.

darcagn:
> > > > But it _is_ performed while holding the lock. If `iscur` is true or `spinlock_trylock(&lock->lock)` is true, we are the holding the lock, so where's the race condition?
> > > 
> > > 
> > > `iscur` is only valid as long as IRQs are disabled. But it is checked after the IRQs are re-enabled, so it races.
> > 
> > 
> > How do we lose the lock from the time IRQs are enabled to when `iscur` is checked?
> 
> Threading timer fires, and another thread tries, and succeeds, to take the lock before we switch back.

But we're not trying to obtain the lock, we already have the lock?

kernel/libc/newlib/lock_common.c
QuzarDC: Entirely wrong. Whoops! The irq disabling and `iscur` aren't needed at all. If we are owner, then it won't be changing, and if we aren't owner, we won't become owner while in this function.

QuzarDC:
> > > > > But it _is_ performed while holding the lock. If `iscur` is true or `spinlock_trylock(&lock->lock)` is true, we are the holding the lock, so where's the race condition?
> > > > 
> > > > 
> > > > `iscur` is only valid as long as IRQs are disabled. But it is checked after the IRQs are re-enabled, so it races.
> > > 
> > > 
> > > How do we lose the lock from the time IRQs are enabled to when `iscur` is checked?
> > 
> > 
> > Threading timer fires, and another thread tries, and succeeds, to take the lock before we switch back.
> 
> But we're not trying to obtain the lock, we already have the lock?

You're absolutely right. And it is that the value can change out from under us, but only from something that's not the current thread to a different thing that's not the current thread.

darcagn:
- Removed the irq_disable()/irq_enable() as it turns out they weren't necessary at all.
- Did the same for __newlib_lock_acquire_recursive().
- Changed both functions to use thd_get_current().

QuzarDC (approved):

PR #457, opened 2024-01-10
Fixed Newlib's getpid() and Implemented kill()

gyrovorbis:
- Our implementation of getpid() for newlib was sketchily returning the current thread ID as the process ID
- We're now using a signle constnant value, KOS_PID, as the single process ID for a KOS executable, which is returned by getpid()
- kill() is now implemented to kill the current process if the correct PID is passed or a negative value which is meant to kill a related group
- `clock_getcpuclockid()` updated to support the new default `KOS_PID`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,7 +43,7 @@ Joe Fenton: 2016
 Stefan Galowicz: 2016, 2017
 Luke Benstead: 2020, 2021, 2022, 2023
 Eric Fradella: 2023, 2024
-Falco Girgis: 2023
+Falco Girgis: 2023, 2024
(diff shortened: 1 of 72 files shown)


ljsebald (commented):
If we're going to be implementing kill, then we really should be implementing signal handling as well, which I don't think is a can of worms we want to open just now...

gyrovorbis:
> If we're going to be implementing kill, then we really should be implementing signal handling as well, which I don't think is a can of worms we want to open just now...

Sounds good. As discussed in Discord, I removed that aspect from the PR, leaving just the changes to `getpid()`.

ljsebald (approved):

PR #459, opened 2024-01-11
Hello Example Cleanup and Update

gyrovorbis:
This one is pretty opinionated, so please everyone feel free to totally disagree whether this should be done or not... but I'm of the opinion this example should be updated for a few reasons:
1) It requires a romdisk which is neither actually used nor is it even described or introduced to the user, giving the impression that it is a requirement and that we have some unnecessary complexity in our build system that they need to duplicate for any project.
2) It explicitly uses `KOS_INIT_FLAGS()`.... Yes, I know, this is weird, but I actually think the vanilla "hello world" example would be better suited showing off how you can compile a 100% standard, vanilla C file using just the stdlib and a regular hello world that will build anywhere and run it on Dreamcast just fine without pulling in a single KOS-specific header file or doing anything platform-dependent. I believe there's actually real value in the fact we can do that.
3) I thought we could improve some of the comments and descriptions a little better than what we have. 

- Removed totally unneeded romdisk which wasn't even used
- Added more comments and descriptions to the Makefile for each rule
- Commented out explicit `KOS_INIT_FLAGS()` to show how defaults can be used without including a single KOS header
- Added more details about how `KOS_INIT_FLAGS()` work.

### DIFF
--- a/examples/dreamcast/hello/hello.c
+++ b/examples/dreamcast/hello/hello.c
@@ -2,25 +2,38 @@
 
    hello.c
    Copyright (C) 2001 Megan Potter
+   Copyright (C) 2024 Falco Girgis
 */
 
-#include <kos.h>
+#include <stdio.h>
 
-/* These macros tell KOS how to initialize itself. All of this initialization
-   happens before main() gets called, and the shutdown happens afterwards. So
-   you need to set any flags you want here. Here are some possibilities:
+/* #include <kos.h>
+   KOS_INIT_FLAGS(INIT_DEFAULT);
+*/
+
+/* KOS_INIT_FLAGS() tells KOS how to initialize itself and which drivers to
+   pull into your executable. All of this initialization happens before main()
+   gets called, and the shutdown happens afterwards.
+
+   Here are some possible flags to pass to KOS_INIT_FLAGS():
 
    INIT_NONE        -- don't do any auto init
    INIT_IRQ         -- Enable IRQs
    INIT_NET         -- Enable networking (including sockets)
    INIT_MALLOCSTATS -- Enable a call to malloc_stats() right before shutdown
 
-   You can OR any or all of those together. If you want to start out with
-   the current KOS defaults, use INIT_DEFAULT (or leave it out entirely). */
-KOS_INIT_FLAGS(INIT_DEFAULT | INIT_MALLOCSTATS);
+   Refer to kos/init.h and arch/init_flags.h for the full list of flags.
+   You can OR any or all of these together.
+
+   If you want to start out with the current KOS defaults, use INIT_DEFAULT.
+   You may now also omit KOS_INIT_FLAGS() altogether, which will also use
+   INIT_DEFAULT automatically, as we're doing here. This can be useful if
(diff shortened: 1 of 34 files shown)


examples/dreamcast/hello/hello.c
```diff
@@ -2,25 +2,38 @@
...
-   the current KOS defaults, use INIT_DEFAULT (or leave it out entirely). */
-KOS_INIT_FLAGS(INIT_DEFAULT | INIT_MALLOCSTATS);
+   Refer to kos/init.h and arch/init_flags.h for the full list of flags.
+   You can OR any or all of these together.
+
+   If you want to start out with the current KOS defaults, use INIT_DEFAULT.
+   You may now also omit KOS_INIT_FLAGS() altogether, which will also use
+   INIT_DEFAULT automatically, as we're doing here. This can be useful if
+   you're lazy or if you're building regular C or C++ code and don't want to
+   introduce KOS-specific DC code into the file.
```
ljsebald, line 32:
```suggestion
   If you wish to use the default initialization settings, specify
   INIT_DEFAULT in your KOS_INIT_FLAGS(). You may also omit KOS_INIT_FLAGS()
   from your program (as we are doing here since the above code is commented
   out), which will automatically use the default initialization settings.
```

I don't think it's necessary to have the last sentence at all, and I've reworded what you have a little bit (since it isn't like the default flags being implied without a `KOS_INIT_FLAGS` is new or anything.

Spacing might need to be adjusted here, but it should be good enough.

examples/dreamcast/hello/hello.c
```diff
@@ -2,25 +2,38 @@
...
+/* KOS_INIT_FLAGS() tells KOS how to initialize itself and which drivers to
+   pull into your executable. All of this initialization happens before main()
+   gets called, and the shutdown happens afterwards.
+
+   Here are some possible flags to pass to KOS_INIT_FLAGS():
 
    INIT_NONE        -- don't do any auto init
    INIT_IRQ         -- Enable IRQs
    INIT_NET         -- Enable networking (including sockets)
    INIT_MALLOCSTATS -- Enable a call to malloc_stats() right before shutdown
```
ljsebald, line 23:
```suggestion
   INIT_DEFAULT     -- do a normal init with the most common settings
   INIT_IRQ         -- Enable IRQs (implied by INIT_DEFAULT)
   INIT_NET         -- Enable networking (including sockets)
   INIT_MALLOCSTATS -- Enable a call to malloc_stats() right before shutdown
   INIT_NONE        -- don't do any auto init (you must manually init hardware)
```

ljsebald (changes requested):
A few changes here...

I might also suggest writing about romdisks in the comments here, showing how one can be set up. Also, adding the romdisk stuff back into the Makefile but commenting it out.

gyrovorbis:
> A few changes here...
> 
> I might also suggest writing about romdisks in the comments here, showing how one can be set up. Also, adding the romdisk stuff back into the Makefile but commenting it out.

Oh yeah, probably a good idea, since I guess part of the original intention was for this to also serve as a skeleton/template project... Done!

ljsebald (approved):

PR #460, opened 2024-01-11
Add environ variable for SH4 floating point precision configuration

darcagn:
This PR adjusts `environ_dreamcast.sh` so that the SH4 FP precision mode can be overridden based on the environment variable `KOS_SH4_PRECISION`. If the setting is omitted from the user's `environ.sh`, the setting defaults to `-m4-single-only`. The setting is added towards the bottom of the `environ.sh.sample`, but it is noted that only `-m4-single-only` is supported and `-m4`/`-m4-single` are experimental.

### DIFF
--- a/doc/environ.sh.sample
+++ b/doc/environ.sh.sample
@@ -2,22 +2,22 @@
 #
 # This is a sample script for configuring and customizing your
 # KOS build environment. Modify it to suit your setup. Several
-# settings may be enabled optionally or are provided with 
+# settings may be enabled optionally or are provided with
 # alternative values.
 #
-# This script is typically sourced in your current shell environment 
+# This script is typically sourced in your current shell environment
 # (probably by .bashrc, .bash_profile, or something similar), so that
 # the KOS environment is set up automatically for each shell session.
 #
 
-# Build Architecture 
+# Build Architecture
 #
 # Set the major architecture you'll be building for.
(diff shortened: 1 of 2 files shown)


doc/environ.sh.sample
```diff
@@ -107,6 +107,7 @@ export KOS_CFLAGS=""
 export KOS_CPPFLAGS=""
 export KOS_LDFLAGS=""
 export KOS_AFLAGS=""
+export KOS_SH4_PRECISION=""
```
gyrovorbis, line 110: Are you doing this here so that they can comment out the one below and have everything still work or? Because the one below is just overriding this empty default? The others have empty defaults because the flags get appended to these variables rather than having them just get set outright.

darcagn: I went back and forth on whether I should include this line, and my justification was that I thought one of the other ones also didn't add to itself, but looking at the other ones, they all do, so it's probably cleaner to remove it.

ljsebald:
I would like a more stern warning about how things will likely break if the precision is changed from the default. As of right now, a casual reader who doesn't know any better might assume from the wording that enabling double precision mode would have no negative consequences (just that it is experimental), which is definitely not the case.

Also, it should probably mention (although this should be obvious) that the toolchain must be compiled with support for the other precision modes for that flag to work at all.

ljsebald (approved):

PR #466, opened 2024-01-19
Update Dockerfile for repo as its pointing to a dead repo

futurepr0n:
The old repo no longer exists, so I updated it point to the current https://github.com/KallistiOS/KallistiOS/


futurepr0n:
accidentally thought it was a dead repo, just a repo outage, ignore pull request

PR #469, opened 2024-01-21
1-20-24 - Fixed depth test issues and added color buffer clear to act…

OniEnzeru:
Fixed depth test issues in GLDC example and added a line to clear the depth and color buffers to allow the background clear color to actually be set.


gyrovorbis:
@Kazade Yo, is this going to be something that needs to be fixed in the actual GLdc repo as well? Wasn't this example just copied from your repo, or? Wasn't sure where it came from.

ljsebald (approved):

Kazade:
@gyrovorbis I'm pretty sure this was an original kgl example - don't think it's from GLdc

PR #471, opened 2024-01-27
Increased Timing Resolution of pvr_stats_t

gyrovorbis:
- Now that the TMU driver has been upgraded, this can all be bumped up from millisecond to nanosecond resolution for free


ljsebald (changes requested):
I don't necessarily mind doing this, but it probably should also involve a bit more of a notice (in the release notes or something like that)...

Also, are there any examples that use the data that had its resolution changed here that need to be updated?

gyrovorbis:
> I don't necessarily mind doing this, but it probably should also involve a bit more of a notice (in the release notes or something like that)...
> 
> Also, are there any examples that use the data that had its resolution changed here that need to be updated?

Okay, I added a note to the CHANGELOG for it, and did a search through the examples for everything using `pvr_get_stats()` to ensure that nothing should be broken from this change. 95% of them are only using the framerate or are simply printing "frame time" without any associated units, so they should be fine; however, `serpent_dma` does have some fancy frame-time visualizer whose sizes needed to be adjusted for the new timing resolution... good call.

QuzarDC:
So, took a sec to review this and I think there's a bit of a need to clean up/adjust the types here. A u64 time is retrieved which then has the stored time, an int, subtracted from it. That's then cast to a u32 and stored in an int.

Aside from that though, there may need to be checks for rollovers in the timer since it's now 1000000x more precise. Before the max was in the scale of weeks where it just wouldn't really ever matter, but now it'd be in minutes.

gyrovorbis:
> So, took a sec to review this and I think there's a bit of a need to clean up/adjust the types here. A u64 time is retrieved which then has the stored time, an int, subtracted from it. That's then cast to a u32 and stored in an int.
> 
> Aside from that though, there may need to be checks for rollovers in the timer since it's now 1000000x more precise. Before the max was in the scale of weeks where it just wouldn't really ever matter, but now it'd be in minutes.

You're exactly right. I hated the fact we were storing timestamps as a signed integer to start with, but yes, now it's what, just over 4 seconds before an UNSIGNED 32-bit int would roll over? Honestly I feel like the best and most correct solution would be to make these fields `uint64_t`, as are the times that it's querying for... It's 16 more bytes but there's only ever one of these instantiated... I feel like it's worth it for the profiling capabilities the increased precision gives and to not have to add a bunch of ugly/slow logic to attempt to do graceful rollover...

gyrovorbis:
Alrighty, guys, ready for another review. She's bumped up to nanosecond resolution, there won't be any overflowing, I've reorganized the structure members for optimal packing, and I kept us from unnecessarily querying for the current timestamp for one PVR event type that wasn't actually using it.

I just uploaded a screen capture to my Twitter/X here: https://x.com/falco_girgis/status/1772484715032850753?s=20

ljsebald (approved):

PR #476, opened 2024-01-29
Small Fixes from DCPlaya's KOS branch

QuzarDC:
These are three small fixes that are derived from DCPlaya's branch of KOS as found here: https://github.com/zig/kos-dcplaya/commit/c7b02b372e8da65e3cd83016f15840841d76754a

- Prevent the loading of multiple maple drivers for the same function. This would help prevent possible issues when loading them individually/manually or if someone brews up alternative drivers.
- Prevent doublt init or unitted shutdown of fs_ramdisk. Should be helpful if loading manually, and has it match other FSes which have similar protections.
- Shortcut boosting the currently running thread. If thd_schedule_next is called on thd_current, it can just return, it's work done.

### DIFF
--- a/kernel/fs/fs_ramdisk.c
+++ b/kernel/fs/fs_ramdisk.c
@@ -891,7 +891,7 @@ int fs_ramdisk_shutdown(void) {
     rd_file_t *f1, *f2;
 
     /* Test if initted */
-    if(rootdir === NULL)
+    if(rootdir == NULL)
         return -1;
 
     /* For now assume there's only the root dir, since mkdir and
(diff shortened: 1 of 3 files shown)


kernel/fs/fs_ramdisk.c
```diff
@@ -885,6 +889,11 @@ int fs_ramdisk_init(void) {
 /* De-init the file system */
 int fs_ramdisk_shutdown(void) {
     rd_file_t *f1, *f2;
+
+    /* Test if initted */
+    if(rootdir === NULL)
```
ljsebald, line 894:
```suggestion
    if(rootdir == NULL)
```

ljsebald (changes requested):

ljsebald (approved):

PR #480, opened 2024-01-31
Quick cleanup for 2D_tex_quad example in GLdc

OniEnzeru:
When recompiling KOS I noticed that dc-tool in environ.sh already has a -x making the one in my makefile redundant.  Also cleaned up an unused variable warning in main.cpp.


gyrovorbis (approved):
Oh snap, I didn't even notice the new warnings there for unused variables... You got away with this one! jk jk ;)

PR #484, opened 2024-02-04
Fix cmake function kos_add_romdisk not getting KOS_BASE from ENV

maishuji:
This PR is about fixing the path given in the cmake function kos_add_romdisk.
Indeed, we use ${KOS_BASE}  whereas we should take it from $ENV{KOS_BASE} (as it is done  line 54 in the same file for exmaple). The result of this bug was a failure when generating romdisk, as it didn't find the library romdiskbase

### DIFF
--- a/utils/cmake/dreamcast.cmake
+++ b/utils/cmake/dreamcast.cmake
@@ -60,7 +60,7 @@ function(kos_add_romdisk target romdiskPath)
     add_custom_command(
         OUTPUT  ${obj}
         DEPENDS ${obj_tmp}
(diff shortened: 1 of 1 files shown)


gyrovorbis:
He's right. I have no idea how this worked for me, but oh well! Thank you!

gyrovorbis (approved):

gyrovorbis:
Still in the hotel room in Brussels at FOSDEM without my main dev machine, so I can't update the SourceForge repo after merging... I'll approve then leave it for someone else to merge.

PR #491, opened 2024-03-03
Update configs with GDB 14.1 to 14.2

darcagn:
GDB 14.2 was released today:
https://sourceware.org/pipermail/gdb-announce/2024/000138.html

This PR updates all `dc-chain` configs with GDB 14.1 to 14.2.

This is in draft until I can actually confirm things build properly.


darcagn:
Finally got around to building it, it builds, so this is coming out of draft.

ljsebald (approved):

PR #499, opened 2024-03-25
Avoid flexible array member in struct dirent

dressupgeekout:
GCC doesn't like having the dirent member be in the middle of DIR because struct dirent's `d_name` is a flexible array. This could be fixed by swapping the placement of DIR::d_ent and DIR::d_name, but I think a better solution is to have the size of struct dirent's `d_name` be exactly NAME_MAX bytes so that it matches DIR's `d_name`.

This change allows me to work past an error such as:

```
In file included from /opt/toolchains/dc/sh-elf/sh-elf/include/dirent.h:39,                                                                                                                                                           
                 from dreamcast_main.cpp:9:                                                                                                                                                                                           
/opt/toolchains/dc/kos/include/sys/dirent.h:69:15: error: flexible array member ‘dirent::d_name’ not at end of ‘struct<unnamed>’                                                                                                      
   69 |     char      d_name[];           /**< \brief Filename */                                                                                                                                                                     
      |               ^~~~~~                                                                                                                                                                                                          
/opt/toolchains/dc/kos/include/sys/dirent.h:85:21: note: next member ‘char <unnamed struct>::d_name [256]’ declared here
   85 |     char            d_name[NAME_MAX]; /**< \brief Filename */                                              
      |                     ^~~~~~                       
/opt/toolchains/dc/kos/include/sys/dirent.h:82:16: note: in the definition of ‘struct<unnamed>’
   82 | typedef struct {                                 
      |                ^                                 
```

I've tested with a small program which uses opendir() and readdir() and it shows me reasonable results (file system entries `/vmu`, `/cd`, etc.)

### DIFF
--- a/include/sys/dirent.h
+++ b/include/sys/dirent.h
@@ -66,7 +66,7 @@ struct dirent {
     off_t     d_off;              /**< \brief File offset */
     uint16_t  d_reclen;           /**< \brief Record length */
     uint8_t   d_type;             /**< \brief File type */
-    char      d_name[];           /**< \brief Filename */
+    char      d_name[0];          /**< \brief Filename */
 };
 
 /** \brief  Type representing a directory stream.


gyrovorbis:
Well, the problem with this solution is that you're going to be wasting RAM. The flexible array member allows us to over-allocate *exactly* the number of bytes needed to store the name string and no more.

You're getting that error in C++, because i'm presuming you're compiling with `-std=c++xx` rather than `-std=gnuxx`, which should allow C++ to handle flexible arrays fine.

gyrovorbis:
Ugh. I'm the one who did this... Is this something that was working in C, but I didn't test in C++?

The crappy thing is that the flexible array member on `dirent` allowed us to over-allocate *exactly* only the number of bytes needed to store the name, so we were conserving RAM.

Then secondly, the `DIR::d_name[NAME_MAX]` field is there so that the flexible array member *above it*from the `dirent` structure would simply spill over and have exactly enough static storage allocated to hold the maximum name size, as previously suggested.

If we do take this change, that field, `DIR::d_name` should be removed, since it becomes redundant.

gyrovorbis:
Whelp, I've reproduced it here... 100% a C++ only issue... Works fine in C... goddamnit... 

Is it some aliasing rule where the FAM can't alias over the following members in `DIR` or something C++? ...Interesting... Playing with it...

include/sys/dirent.h
```diff
@@ -66,7 +66,7 @@ struct dirent {
     off_t     d_off;              /**< \brief File offset */
     uint16_t  d_reclen;           /**< \brief Record length */
     uint8_t   d_type;             /**< \brief File type */
-    char      d_name[];           /**< \brief Filename */
+    char      d_name[NAME_MAX];   /**< \brief Filename */
```
gyrovorbis, line 69:
```suggestion
    char      d_name[0];   /**< \brief Filename */
```
Yes, I know, WTF.... This is a GNU extension from before FAMs being around which does not seem to freak the compiler out when its structure is used within another structure without being the last member in C++... Can you test that this works for you? I just tested locally with `scandir()`, but I'd feel better if you did another test. :)

gyrovorbis, line 69:
```suggestion
    char      d_name[0];          /**< \brief Filename */
```

gyrovorbis (commented):
Can you test if changing it to be size 0 fixes the issue you're seeing as well? It's an old GNU extension that compiles and seems to work over here for both C and C++. 

(no, `dirent::d_name` will not actually have a size of 0. It will be a NULL-terminated C string, because it should always get allocated with additional storage for the string following the struct.)

dressupgeekout:
Yes this affects only C++.  I'm using `sh-elf-g++ (GCC) 13.2.0`

include/sys/dirent.h
dressupgeekout: Yeah I can confirm this little trick works with GCC 13.2.0!

gyrovorbis: Awesome. I'm making a note to do this in the future when I need FAM behavior.... Sorry about breaking C++!

dressupgeekout:
Oh whoops, don't merge yet, I messed up.

dressupgeekout:
OK try this!

QuzarDC (approved):

PR #501, opened 2024-04-04
cmake: Simplify toolchain script

pcercuei:
Instead of duplicating all the CFLAGS and LDFLAGS that should be set by the user in the environ*.sh scripts, use the KOS wrappers as the compilers, which makes sure that the user flags are used by default.

### DIFF
--- a/utils/cmake/dreamcast.cmake
+++ b/utils/cmake/dreamcast.cmake
@@ -1,4 +1,22 @@
+# Auxiliary CMake Utility Functions
+#   Copyright (C) 2023 Colton Pawielski
+#   Copyright (C) 2024 Falco Girgis
+#
+# This file implements utilities for the following additional functionality
+# which exists in the KOS Make build system:
+#   1) linking to existing binaries
+#   2) adding a romdisk
+#
+# NOTE: When using the KOS CMake toolchain file, you do not need to include
+#       thils file directly!
+
 cmake_minimum_required(VERSION 3.23)
+
+#### Set Configuration Variables From Environment ####
+if(NOT DEFINED ENV{KOS_BASE}
+   OR NOT DEFINED ENV{KOS_CC_BASE})
+    message(FATAL_ERROR "KallistiOS environment variables not found")
+endif()
 
 ### Helper Function for Bin2Object ###
 function(kos_bin2o inFile symbol)
@@ -18,8 +36,9 @@ function(kos_bin2o inFile symbol)
 endfunction()
 
 function(kos_add_binary target inFile symbol)
+    file(REAL_PATH "${inFile}" inFile)
     set(outFile ${CMAKE_CURRENT_BINARY_DIR}/${symbol}.o)
-    kos_bin2o(${CMAKE_CURRENT_SOURCE_DIR}/${inFile} ${symbol} ${outFile})
+    kos_bin2o(${inFile} ${symbol} ${outFile})
     target_sources(${target} PRIVATE ${outFile})
 endfunction()
 
@@ -32,7 +51,7 @@ function(kos_add_romdisk target romdiskPath)
         set(romdiskName ${ARGN})
     endif()
 
-    set(romdiskPath ${CMAKE_CURRENT_SOURCE_DIR}/${romdiskPath})
+    file(REAL_PATH "${romdiskPath}" romdiskPath)
 
     set(obj     ${CMAKE_CURRENT_BINARY_DIR}/${romdiskName}.o)
     set(obj_tmp ${CMAKE_CURRENT_BINARY_DIR}/${romdiskName}_tmp.o)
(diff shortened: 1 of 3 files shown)


gyrovorbis:
OKAY. I tested the crap out of it with my `libGimbal` CMake project from hell, both from the command-line and from `Qt Creator`, and it seems as though using the GNU wrappers will actually work for us now. 

I did have to change a few things in the `dreamcast.cmake` utility file. One was a direct change, one was a small break...
1) Using `kos-cc` as our compiler breaks the custom build rule for generating the romdisk... but I think I am reasonably using the default compiler given that we know the path to their toolchain
2) Out-of-source includes for either bin2o or romdisks were not working at all and were failing. Had to do some hacky crap with `file(REAL_PATH)` to resolve the paths properly in both scenarios
3) Got rid of the fast-math function and clutter it introduced, since now that's perfectly documented and is configured from within the environment variables.
4) Added a header to dreamcast.cmake to clarify that it doesn't need to be manually included if the toolchain file is being used
5) We were defining `__DREAMCAST__` within the CMake toolchain, which I think is the most idiomatic architecture-specific define... which broke my build... So I propose adding it back to the environment directly...

utils/cmake/dreamcast.cmake
```diff
@@ -60,40 +79,10 @@ function(kos_add_romdisk target romdiskPath)
     add_custom_command(
         OUTPUT  ${obj}
         DEPENDS ${obj_tmp}
-        COMMAND ${CMAKE_C_COMPILER} -o ${obj} -r ${obj_tmp} -L$ENV{KOS_BASE}/lib/dreamcast -Wl,--whole-archive -lromdiskbase
+        COMMAND ${KOS_CC_BASE}/bin/sh-elf-gcc -o ${obj} -r ${obj_tmp} -L$ENV{KOS_BASE}/lib/dreamcast -Wl,--whole-archive -lromdiskbase
```
pcercuei, line 82: I think you can use `${KOS_BASE}`, it should have been set by the main toolchain script.

gyrovorbis: But that's just the KOS repo root path. We need the path to the toolchain folder, which is separate and usually one level up.

utils/cmake/dreamcast.cmake
```diff
@@ -1,5 +1,23 @@
...
+#   Copyright (C) 2023 Colton Pawielski
+#   Copyright (C) 2024 Falco Girgis
+#
+# This file implements utilities for the following additional functionality
+# which exists in the KOS Make build system:
+#   1) linking to existing binaries
+#   2) adding a romdisk
+#
+# NOTE: When using the KOS CMake toolchain file, you do not need to include
+#       thils file directly!
```
maishuji, line 11: typo

gyrovorbis: Oh shit, thank you for that good catch!

gyrovorbis:
Yo, @maishuji, as one of our devs who I know is using at least the CMake utility include, is there any way you can test this PR to make sure that it works fine with your environment? 

I did the same and got it all working, but I did have to do a bunch of tweaking... i do think this is the direction we want to take the CMake stuff in, though, to keep it synchronized with the rest of the KOS environment.

utils/cmake/dreamcast.cmake
```diff
@@ -1,5 +1,23 @@
...
+#
+# NOTE: When using the KOS CMake toolchain file, you do not need to include
+#       this file directly!
+
 cmake_minimum_required(VERSION 3.23)
 
+#### Set Configuration Variables From Environment ####
+if(NOT DEFINED ENV{KOS_BASE}
+   OR NOT DEFINED ENV{KOS_CC_BASE})
+    message(FATAL_ERROR "KallistiOS environment variables not found")
```
maishuji, line 18: Here we test the env variable for KOS_CC_BASE, but below at line 82 its using the non-env variable ( which can be not set if we don't pass by the toolchain.cmake.  Is the case for example in my project as I was using a custom toolchain.cmake)

utils/cmake/dreamcast.cmake
maishuji: I think we should use either $ENV{KOS...} or ${KOS...}, but not mixed them. From my understanding, as the cmake variables are defined in the toolchain.cmake and taken from env variable, I suggest we replace $ENV{KOS_BASE} by ${KOS_BASE} here, and modify the test line 16 to check the cmake variables instead of the env variables, wdyt?

gyrovorbis: @maishuji Sorry for the late response. Had a lot going on... but yes, you're right! On it!

gyrovorbis: Alrighty, should be done.

gyrovorbis:
@maishuji Can you check out the latest changes? I've basically made it so that we no longer directly reference environment variables within the `dreamcast.cmake` utility file and instead use `KOS_BASE` and `KOS_CC_BASE` which come from:

1) Inherited from `dreamcast.toolchain.cmake` including the file for you
2) Set externally via command-line or elsewhere
3) The environment variables

I think this is the most versatile solution, so that the file works as intended either automatically getting included by the toolchain OR getting included standalone directly.

Thank you for your input, btw! :)

maishuji (approved):
I tested the last changes on my machine, it works properly on my side.
For me, everything seems OK

gyrovorbis:
> I tested the last changes on my machine, it works properly on my side. For me, everything seems OK

Thanks so much, my friend! I really appreciate having another set of eyes on this! Your input was great.

gyrovorbis (approved):
Alrighty, we've all tested this bad boy, and it seems to work across the board and simplify the CMake toolchain...

gyrovorbis:
Oh.. damnit... I can't actually approve the changes since I participated in the coedz... lol...

QuzarDC (approved):
Tested with building the cmake kos-ports and they all worked fine.

PR #502, opened 2024-04-06
change g2_ctx_t->irq_state type from `uint32_t` to `int`

kapodamy:
Suppress warnings generated by **-Wsign-conversion**, both `irq_disable()` and `irq_restore()` uses `int` type.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/g2bus.h
+++ b/kernel/arch/dreamcast/include/dc/g2bus.h
@@ -140,7 +140,7 @@ void g2_dma_shutdown(void);
     is used in with g2_lock() and g2_unlock().
 */
 typedef struct { 
-    uint32_t irq_state;    /** \brief IRQ state when entering a G2 critical block */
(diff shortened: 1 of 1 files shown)


gyrovorbis:
> Suppress warnings generated by **-Wsign-conversion**, both `irq_disable()` and `irq_restore()` uses `int` type.

Grrrrr I *haaaate* the fact that we overuse `int` for everything ever, and especially for something that's supposed to be a mask... But I guess there's not much we can do, since we don't want to break the whole IRQ API... 

By the way, is this warning enabled by default or what? I haven't noticed that warning, and I'm usually pretty good about spotting them. Thank you! :)

gyrovorbis (approved):
Looks good to me, but leaving it for someone else to merge, just to be sure the authors of this API don't disagree.

kapodamy:
this happen because `g2_lock() `and `g2_unlock()` are inline functions

PR #504, opened 2024-04-10
Minor Floating-Point Precision Fixes

gyrovorbis:
~~1) `cdefs.h`
    - added a new function attribute, `__sh4_single_no_inline`, which
      conditionally only declares a function non-inlined when using
      `-m4-single`.~~
2) `vec3f.h`
    - explicitly truncated two double-precision fp constants to
      single-precision floats

### DIFF
--- a/include/kos/cdefs.h
+++ b/include/kos/cdefs.h
@@ -164,10 +164,20 @@
 #define __no_inline __attribute__((__noinline__))
 #endif
 
-#if __SH4_SINGLE_ONLY__
-#   define __sh4_single_no_inline
-#else
-#   define __sh4_single_no_inline __no_inline
+#ifndef __sh4_single_only_inline
+#   if __SH4_SINGLE_ONLY__
+#       define __sh4_single_only_inline inline static
+#   else
+#       define __sh4_single_only_inline __no_inline
+#   endif
+#endif
+
+#ifndef __sh4_single_only_always_inline
+#   if __SH4_SINGLE_ONLY__
+#       define __sh4_single_only_always_inline __always_inline
+#   else
+#       define __sh4_single_only_always_inline __no_inline
+#   endif
 #endif
 
 /* GCC macros for special cases */
--- a/include/kos/cdefs.h
+++ b/include/kos/cdefs.h
@@ -164,22 +164,6 @@
 #define __no_inline __attribute__((__noinline__))
 #endif
 
-#ifndef __sh4_single_only_inline
-#   if __SH4_SINGLE_ONLY__
-#       define __sh4_single_only_inline inline static
-#   else
-#       define __sh4_single_only_inline __no_inline
-#   endif
-#endif
-
-#ifndef __sh4_single_only_always_inline
-#   if __SH4_SINGLE_ONLY__
-#       define __sh4_single_only_always_inline __always_inline
-#   else
-#       define __sh4_single_only_always_inline __no_inline
-#   endif
-#endif
-
 /* GCC macros for special cases */
 /* #if __GNUC__ ==  */
 
--- a/include/kos/cdefs.h
+++ b/include/kos/cdefs.h
@@ -164,6 +164,12 @@
 #define __no_inline __attribute__((__noinline__))
 #endif
 
+#if __SH4_SINGLE_ONLY__
+#   define __sh4_single_no_inline
+#else
+#   define __sh4_single_no_inline __no_inline
+#endif
+
 /* GCC macros for special cases */
 /* #if __GNUC__ ==  */
 
--- a/kernel/arch/dreamcast/include/dc/vec3f.h
+++ b/kernel/arch/dreamcast/include/dc/vec3f.h
@@ -33,8 +33,8 @@ typedef struct vec3f {
 } vec3f_t;
 
 /** \cond */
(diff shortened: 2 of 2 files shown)


pcercuei:
What's with the `__sh4_single_no_inline`? You define it here (and it's completely unrelated to `-m4-single`) and you're not even suing it anywhere?

gyrovorbis:
> What's with the `__sh4_single_no_inline`? You define it here (and it's completely unrelated to `-m4-single`) and you're not even suing it anywhere?

Oh yeah, I forgot how weird that one must look without it being used anywhere... 

I'm almost positive this is needed for some of the stuff going on in GLdc for proper `-m4-single` support. Basically we cannot allow functions which assume single-point precision upon entry to get inlined within other functions, as the calling function may have temporarily changed the FP precision/environment around the inline function.

This macro was a way to allow for regular inlining with `-m4-single-only` but to no longer inline for `-m4-single`, so that the FP environment will always be single-precision upon function entry.

gyrovorbis:
Okay, I've reversed the naming scheme on the macro and have added another one... Does the name make more sense now?

Basically you'd use it like:
`__sh4_single_only_always_inline void glDCLoadMatrix(Matrix4 *matrix) {
    __asm__(" Do some fast matrix loading assembly crap that
                      swaps to double precision FPU mode here to load
                      matrix elements faster!");
}`

The previous function would REQUIRE that the FP state be defaulted to single-precision upon function entry, which is always assured with `-m4-single-only`, but in the case of `-m4-single`, I believe this function could get inlined within another function context which has changed the FP mode to double... It's only required to be changed back to single for function-call boundaries, and I don't believe there is one in the case of an inline function...

So TL;DR, we need a mechanism that simultaneously allows or forces a function to be inlined with `-m4-single-only` but prevents it from being inlined with `-m4-single`.

gyrovorbis:
Alrighty, nevermind. I have decided against the conditional inline macros... probably  not a good idea to "sometimes" inline a function based on what FP configuration we're building as... 

This PR is ready. The scope of it is now a tiny bugfix. :)

QuzarDC (approved):

olehermanse:
@gyrovorbis sorry if you find this noisy since I'm not a contributor to the project, and I'm lacking a lot of context, but seeing the commits made me curious - what does changing these constants from double to single precision achieve / what is the bug being fixed here?

gyrovorbis:
> @gyrovorbis sorry if you find this noisy since I'm not a contributor to the project, and I'm lacking a lot of context, but seeing the commits made me curious - what does changing these constants from double to single precision achieve / what is the bug being fixed here?

Not noisy at all, my friend, you don't have to be a contributor to participate! That's a DAMN good question that is... pretty deep.

Okay, historically Dreamcast binaries have always been built with the `-m4-single-only` flag (including in official Sega libraries) which FORCES the compiler to truncate all double-precision floating point values into single-precision values. KOS has been built with and only intended to be used with this flag since its inception, all the way up until about a year ago.

Why in god's name would you do this? Especially when the SH4 actually does have support for double-precision floats? Performance is one of the main arguments I've heard. In order to do double-precision FPU operations, the FP precision mode must be swapped via the `FSCHG` instruction. Then again, if you want to swap back to single-precision, you have to issue another `FSCHG` instruction again to change the FP mode back. There is an argument that, sure, you can increase performance of code that relies on double-precision floats by simply forcing them to be single-precision, so that you will never have to swap between the SH4's FP mode by issuing the `FSCHG` instruction.

Fast forward out of the dark ages and into 2024, and we have run into many scenarios where we need true full double-precision floating point values... Take the C function `difftime()` which is finding the difference between two 64-bit integers and is returning that value as a `double`. If we were to truncate that value to a single-precision float, then almost a full minute would have to elapse before the `difftime()` truncated result would show any change in time... making it totally worthless for handling frame-deltas. We've literally run into this exact problem with Objective-C's Foundation stdlib doing this for NSTimer as well as Raylib doing this for frame-based deltatime.

This year, several of us have decided to start actually going out of our way to build the toolchains with the `-m4-single` flag supported and tested with KOS (this means *default* to single-precision upon function entry, but allow for temporarily swapping to double-precision as needed). There has been a LOT of work going through to ensure that KOS does not actually suffer any negative performance repercussions from this... and part of that is ensuring that random FP constants like the ones in this PR are actually single-precision floats, rather than double (hence the `f` suffix).

These constants are actually used in very critical, fast-path, assembly-optimized routines for degree/radian conversions. The last thing we want is some dumb little FP constant used in this path to force the processor to have to swap FP precision modes, do a calculation on a double, then swap back. We're wasting time on the extra instructions, and then there's added register pressure when in double FP mode, since you get 1/2 as many double-sized registers as you otherwise would...

Anyway, I hope that kind of explained why this is such a big deal... There's a lot more that has to go into this as well, lots of nuances, because we have inline assembly that will ALWAYS ASSUME that the function is entered in single-precision FP mode... but what happens if this function gets inlined within a context where the FPU has swapped to double-precision? It would blow up... So we also cannot be inlining any functions with inline assembly that makes any sort of assumption about the FP precision mode upon entry...

olehermanse:
@gyrovorbis Thanks for the thorough explanation. That's great, and not what I'd expect 😅

PR #505, opened 2024-04-10
Worker threads

pcercuei:
Worker threads are threads that are associated to a given work function.
When notified (through thd_worker_notify()), the worker thread will call
its work function.

This is useful for implementing asynchronous function calls. It is
especially useful to implement threaded interrupt handlers.


kernel/thread/worker.c
```diff
@@ -0,0 +1,93 @@
...
+
+struct thd_worker {
+    kthread_t *thd;
+    void (*routine)(void *);
+    void *data;
+    bool pending;
+    bool quit;
+};
+
+static void * thd_worker_thread(void *d)
```
andressbarajas, line 20:
```suggestion
static void * thd_worker_thread(void *d) {
```

kernel/thread/worker.c
```diff
@@ -0,0 +1,93 @@
...
+{
```
andressbarajas, line 21:
```suggestion
```

kernel/thread/worker.c
```diff
@@ -0,0 +1,93 @@
...
+
+    /* (ab)use the rv pointer to reference our worker struct */
+    worker->thd->rv = worker;
+
+    irq_restore(flags);
+
+    return worker->thd;
+}
+
+void thd_worker_wakeup(kthread_t *thd)
```
andressbarajas, line 73:
```suggestion
void thd_worker_wakeup(kthread_t *thd) {
```

kernel/thread/worker.c
```diff
@@ -0,0 +1,93 @@
...
+{
```
andressbarajas, line 74:
```suggestion
```

kernel/thread/worker.c
```diff
@@ -0,0 +1,93 @@
...
+    struct thd_worker *worker = thd->rv;
+    uint32_t flags = irq_disable();
+
+    worker->pending = true;
+    genwait_wake_one(worker);
+
+    irq_restore(flags);
+}
+
+void thd_worker_destroy(kthread_t *thd)
```
andressbarajas, line 84:
```suggestion
void thd_worker_destroy(kthread_t *thd) {
```

kernel/thread/worker.c
```diff
@@ -0,0 +1,93 @@
...
+{
```
andressbarajas, line 85:
```suggestion
```

andressbarajas:
Dont forget to include your copyright under "Copyright (C) 2009, 2010, 2016 Lawrence Sebald"
in the thread.h file

gyrovorbis:
> Dont forget to include your copyright under "Copyright (C) 2009, 2010, 2016 Lawrence Sebald" in the thread.h file and maybe edit the CHANGELOG if you want

lol forreal. I had done a decent amount of work and forgot to include mine on `thread.h` as did @cepawiel... oh well, doing so with the PR for CPU time... but yeah, especially if you like contributed something to the public API. Also definitely think this is a big enough new feature addition to warrant a CHANGELOG update too.

pcercuei:
@andressbarajas thanks for the review, I updated the commit.

The API now also uses a `kthread_worker_t`, which is opaque on purpose (but I can change that with enough social pressure); and I added a `kthread_worker_get_thread()` API function to retrieve the underlying `kthread_t`.

This should prevent any mistake when using the API (e.g. passing regular threads to `thd_worker_wakeup()`). It also means we don't need the "rv" hack anymore.

gyrovorbis:
The only thing I really have to add is that I would personally prefer to see the new API stuff in its own `worker.h` or something to go with `worker.c`, cuz I feel like all of the threading and primitives are split into their own respective files surrounding each unique type...

...but I'm not picky if you guys disagree. I think this looks great and can't wait to start making stuff use it. lol.

pcercuei:
Updated, the API is now in its own `<kos/worker_thread.h>` header.

kernel/thread/worker.c
```diff
@@ -0,0 +1,89 @@
...
+            break;
+
+        worker->pending = false;
+        worker->routine(worker->data);
+    }
+
+    return NULL;
+}
+
+kthread_worker_t * thd_worker_create(void (*routine)(void *), void *param) {
```
QuzarDC, line 43: I'm not sure how this should get factored out, but I think it would make a lot of sense to have a version of this as well (perhaps `thd_worker_create_ex`) that would take in a `kthread_attr_t` to define detailed params. Since the worker is only run on demand I could see though it makes sense if the intent was that all those could be manipulated after the worker is created.

kernel/thread/worker.c
```diff
@@ -0,0 +1,89 @@
...
+    irq_restore(flags);
+
+    return worker;
+}
+
+void thd_worker_wakeup(kthread_worker_t *worker) {
+    uint32_t flags = irq_disable();
+
+    worker->pending = true;
+    genwait_wake_one(worker);
```
QuzarDC, line 74: What happens if the worker thread just hasn't run yet in order to sleep on itself? The way I read it the genwait_wake_one will fail, but now the pending will be set to true so whenever the thread does run for the first time it won't sleep but just run through? Makes sense, but just wanted to verify my reading.

That led me to the issue in that calling `thd_worker_wakeup` X times does not guarantee that the worker routine ran X times. I'm not sure if that is intentional, but it would certainly seem to me to limit the usefulness. 

Changing `pending` to an int rather than a bool could fix that and generally improve the usability of the system, but otherwise I'd think that this wakeup would need to at least be able to signal to the caller of it if the worker was already awake (to know that calling the function would have no effect).

kernel/thread/worker.c
```diff
@@ -0,0 +1,89 @@
...
+void thd_worker_destroy(kthread_worker_t *worker) {
+    worker->quit = true;
+    genwait_wake_one(worker);
+
+    thd_join(worker->thd, NULL);
+    free(worker);
+}
+
+kthread_t * thd_worker_get_thread(kthread_worker_t *worker) {
+    return worker->thd;
```
QuzarDC, line 88: Adding the comment here but generally, for each of the functions there should be NULL checks to prevent deref.

kernel/thread/worker.c
```diff
@@ -0,0 +1,89 @@
...
+
+kthread_worker_t * thd_worker_create(void (*routine)(void *), void *param) {
+    kthread_worker_t *worker;
+    uint32_t flags;
+
+    worker = malloc(sizeof(*worker));
+    if (!worker)
+        return NULL;
+
+    worker->data = param;
```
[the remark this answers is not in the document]
pcercuei: I guess I couldn't decide.

kernel/thread/worker.c
pcercuei: > What happens if the worker thread just hasn't run yet in order to sleep on itself? The way I read it the genwait_wake_one will fail, but now the pending will be set to true so whenever the thread does run for the first time it won't sleep but just run through? Makes sense, but just wanted to verify my reading.

Correct.

The "pending" is used when `thd_worker_wakeup` is called while the worker thread is running; it tells it that more work was added while the thread was running, so the worker should restart when it's done with the current load.
 
> That led me to the issue in that calling `thd_worker_wakeup` X times does not guarantee that the worker routine ran X times. I'm not sure if that is intentional, but it would certainly seem to me to limit the usefulness.

It is not an issue, it is by design. The wakeup function basically says "there is work to do", it doesn't actually give your worker function any work load - this is completely up to you how you implement this.

See it for instance in the context of a IRQ handler: when the worker function start, it would read the "pending IRQ" register, cache it into a variable, ACK the interrupts, and then for each bit set, handle the corresponding interrupt. It wouldn't handle just one single interrupt then return, but handle all the ones that were pending. And if an IRQ fires while the worker thread is running, `thd_worker_wakeup` would be called, which would cause the worker thread to restart the IRQs processing after it's done with the current batch.

What you are asking for is trivial to implement on top of the existing API, by the way; you'd just need to implement a FIFO for workloads. Wakeup the worker when something was added, and the worker thread processes the jobs from the FIFO until it's empty.

I can totally add an API for that, but my point is, the current one is fine as-is.

> 
> Changing `pending` to an int rather than a bool could fix that and generally improve the usability of the system, but otherwise I'd think that this wakeup would need to at least be able to signal to the caller of it if the worker was already awake (to know that calling the function would have no effect).

I don't see why the caller needs to know anything about that.

kernel/thread/worker.c
pcercuei: I adhere to the school of thought that you (the program/game developer) is responsible for making sure that you don't pass garbage to API functions.

kernel/thread/worker.c
pcercuei: I can do that, if you think that's needed. Currently the attributes can be added after the worker is created, yes.

pcercuei:
Rebased to latest master. I fixed the "param/data" variable mismatch, and added a new optional "jobs" API to work alongside the worker threads.

kernel/thread/worker.c
gyrovorbis: Ah, that would be a good idea, actually. For consistency and orthogonality and shit.

kernel/thread/worker.c
gyrovorbis: I guess if @QuzarDC wants that, at least do it as an `assert()`, so everyone doesn't have to pay for it? We exposed `NDEBUG` in the `environ.sh.sample` now, and I'm building with it...

gyrovorbis:
Great work on the Doxygen, btw...

kernel/thread/worker.c
pcercuei: @gyrovorbis it's done in the new version.

kernel/thread/worker.c
pcercuei: Done.

gyrovorbis (approved):
Other than these couple trivial formatting changes, I'm good with this PR. I'll approve but let @QuzarDC merge it.

include/kos/worker_thread.h
```diff
@@ -0,0 +1,156 @@
...
+                            Passing NULL will initialize all attributes to their
+                            default values.
+    \param  routine         The function to call in the worker thread.
+    \param  data            A parameter to pass to the function called.
+
+    \return                 The new worker thread on success, NULL on failure.
+
+    \sa thd_worker_destroy, thd_worker_wakeup
+*/
+kthread_worker_t * thd_worker_create_ex(const kthread_attr_t *attr,
```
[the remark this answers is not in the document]
pcercuei: You guys sure have a strange coding style.

pcercuei:
@gyrovorbis updated.

gyrovorbis (dismissed):

QuzarDC (approved):

PR #506, opened 2024-04-10
Per-thread CPU time implementation

gyrovorbis:
This PR is finally introducing the elusive PER-THREAD CPU time! The last major POSIX clock type we didn't support. 

- Added two `uint64_t` values to the kernel thread type to allow us to maintain CPU time.
- Maintaining and updating threads' CPU time upon context switches
- Printing out time and % of whole when doing `thd_pslist()` to print
- Also accidentally snuck in a bunch of general clean-up
- updated `clock_gettime()` to support `CLOCK_THREAD_CPUTIME_ID`.
- enabled `_POSIX_CLOCK_CPUTIME` in `time.h`.


include/kos/thread.h
```diff
@@ -189,7 +231,7 @@ typedef __attribute__((aligned(32))) struct kthread {
     uint32_t *stack;
 
     /** \brief  Size of the thread's stack, in bytes. */
-    uint32_t stack_size;
+    size_t stack_size;
```
cepawiel, line 234: Did you mean `usize_t` here? Or is there a reason to have it signed?

include/kos/thread.h
```diff
@@ -251,10 +267,10 @@ typedef __attribute__((aligned(32))) struct kthread {
 */
 typedef struct kthread_attr {
     /** \brief  1 for a detached thread. */
-    int create_detached;
+    bool create_detached;
 
     /** \brief  Set the size of the stack to be created. */
-    uint32_t stack_size;
+    size_t stack_size;
```
cepawiel, line 273: Same as other `stack_size`?

kernel/thread/thread.c
```diff
@@ -66,13 +68,13 @@ static struct ktqueue run_queue;
...
 /* Thread mode: uninitialized or pre-emptive. */
-static int thd_mode = THD_MODE_NONE;
+static kthread_mode_t thd_mode = THD_MODE_NONE;
 
 /* Reaper semaphore. Counts the number of threads waiting to be reaped. */
 static semaphore_t thd_reap_sem;
 
 /* Number of threads active in the system. */
-static uint32_t thd_count = 0;
+static size_t thd_count = 0;
```
cepawiel, line 77: This one too!

include/kos/thread.h
gyrovorbis: `size_t` is supposed to be unsigned? I thought `ssize_t` was the (nonstandard) signed equivalent?

darcagn: `size_t` is unsigned, I don't think `usize_t` is a standard thing?

cepawiel: Woops my bad! Think I got mixed it up with Rust types  :sweat_smile:

kernel/thread/thread.c
```diff
@@ -773,7 +806,7 @@ static void thd_timer_hnd(irq_context_t *context) {
 /* Thread blocking based sleeping; this is the preferred way to
    sleep because it eases the load on the system for the other
    threads. */
-void thd_sleep(int ms) {
+void thd_sleep(unsigned ms) {
```
andressbarajas, line 809:
```suggestion
void thd_sleep(unsigned int ms) {
```

gyrovorbis: Whoops. Dun let my own coding style slink into KOS...

kernel/libc/posix/clock_gettime.c
```diff
@@ -86,8 +87,21 @@ int clock_gettime(clockid_t clk_id, struct timespec *ts) {
...
             return 0;
 
-        /* Fail out for any other unsupported CPU type */
-        /* case CLOCK_THREAD_CPUTIME_ID: */
+        /* Use the kthread-specific CPU time counters */
+        case CLOCK_THREAD_CPUTIME_ID:
+            /* Check whether they are configured properly
+               as an interval timer. */
+            if(!perf_cntr_timer_enabled()) {
+                errno = EINVAL;
```
QuzarDC, line 95: Just a minor concern. I see that EINVAL is used for the PROCESS CPUTIME ID as well for if perf counters are disabled, but that doesn't directly map to the standard which is that it's used for 'Doesn't specify a known clock'. It would be useful if there was some way to flag the user that the reason they're getting this error is due to a misconfiguration.

kernel/thread/thread.c
```diff
@@ -122,22 +123,35 @@ int thd_each(int (*cb)(kthread_t *thd, void *user_data), void *data) {
 int thd_pslist(int (*pf)(const char *fmt, ...)) {
     kthread_t *cur;
 
+    const uint64_t ns_time = perf_cntr_timer_ns();
```
QuzarDC, line 126: Unless I missed something this should probably move to inside the `if(perf_cntr_timer_enabled())` check right? Because the value is unused if we don't have them enabled, so we'd be polling the fallback regular timer for no reason.

kernel/thread/thread.c
```diff
@@ -613,6 +627,20 @@ int thd_set_prio(kthread_t *thd, prio_t prio) {
...
 /* Scheduling routines */
 
+static void thd_update_cpu_time(kthread_t *thd) {
+    if(perf_cntr_timer_enabled()) {
+
+        const uint64_t ns = perf_cntr_timer_ns();
+
+        if(thd_current) {
+            thd_current->cpu_time.total +=
+                ns - thd_current->cpu_time.scheduled;
```
QuzarDC, line 637: Should there be some check for the scheduled being 0/uninitialized? It would seem that since perf_cntr_timer_enabled is being treated as run-time variable and not tied to some mode that you'd get really weird garbage for all threads if you enabled it on-the-fly. 

Or maybe it's not meant to really be runtime settable and it's meant more to be something that is static during the course of execution?

kernel/thread/thread.c
```diff
@@ -1055,6 +1100,19 @@ int thd_init(void) {
...
+        /* Get initial total process CPU time */
+        const uint64_t ns = perf_cntr_timer_ns();
+        /* Set it as our scheduled timestamp */
+        kern->cpu_time.scheduled = ns;
+        /* Also set it as our total CPU time, so that the main
+           thread is credited with consuming this CPU time, otherwise
+           it will be lost and thread totals will not equal process
+           total. */
+        kern->cpu_time.total = ns;
+    }
```
QuzarDC, line 1114: Should these be 0'd in the negative case? See the earlier comment about it being really runtime variable or not.

QuzarDC (commented):
Just some questions about minor bits. Wanted to mostly understand more the expectations of perf counters being enabled or not.

Otherwise looks good.

include/kos/thread.h
```diff
@@ -269,18 +295,19 @@ typedef struct kthread_attr {
...
     The threading system will always be in one of the following modes. This
     represents either pre-emptive scheduling or an un-initialized state.
-
-    @{
 */
-#define THD_MODE_NONE       -1  /**< \brief Threads not running */
-#define THD_MODE_COOP       0   /**< \brief Cooperative mode \deprecated */
-#define THD_MODE_PREEMPT    1   /**< \brief Preemptive threading mode */
-/** @} */
+typedef enum kthread_mode {
```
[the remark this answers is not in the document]
gyrovorbis: Because I don't want to make this entire file pretty, idiomatic C and then just randomly leave this one deprecated piece using a completely different coding style... Seemed very inconsistent, and at least if you're using a deprecated API, you're doing it with more strongly typed enumerations.

Also enumerations are far more useful in terms of documentation and doxygen searching and grouping... Not that anyone should be searching for this since it's deprecated... lol.

include/kos/thread.h
```diff
@@ -690,26 +743,12 @@ int thd_pslist(int (*pf)(const char *fmt, ...));
...
-
-    This is normally done for you by default when KOS starts. This will also
-    initialize all the various synchronization primitives.
-
-    \retval -1              If threads are already initialized.
-    \retval 0               On success.
-
-    \sa thd_shutdown
-*/
+/** \cond Handled internally by KOS */
```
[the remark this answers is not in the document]
gyrovorbis: Yeah, that's what `\cond` and `\endcond` which I've added do... but I guess what you'd like to see is both the conditional stuff plus the existing doxygen comments...

kernel/libc/posix/clock_gettime.c
gyrovorbis: Do you see any POSIX-complaint way to signal that? This seemed like the most POSIX-y thing to do, but you have a point... Was just going to make it super clear in the Doxygen.

kernel/thread/thread.c
gyrovorbis: It was originally intended to be static throughout the course of execution, like how you can't actually modify the CPU clock as you can the REALTIME clock on Linux... so I was operating under the assumption that if you had chosen to reconfigure or claim that first perf counter that KOS uses for this timing stuff, that it was understood that timing would be off and you would no longer be able to measure per-thread or per-process CPU time...

With this in mind, do you think I'm better off hard zeroing it out here if they did mess with that performance counter? Or leave it as its last valid value, which I think is what this should do?

kernel/thread/thread.c
gyrovorbis: Responded.... Good question... should they? What do you think?

kernel/thread/thread.c
QuzarDC: Ah ok, that makes perfect sense.  I figured I was just missing why but there was likely a good one :)

kernel/libc/posix/clock_gettime.c
QuzarDC: Nope! I looked around and couldn't find one myself. That's why I left it as 'minor concern' cause was pretty sure that is just the way to do it and hopefully it doesn't cause any issue.

include/kos/thread.h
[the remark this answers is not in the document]
QuzarDC: Well, I don't know how it would work well. It makes sense not to clutter up doxygen with stuff that end users should neither know about nor have visibility into (not part of the 'public' API as much as we have such a thing). At the same time we'd want to have our own internal nice documentation for it to easily grok things.

For this though it's super pointless since the documentation is just `thd_init inits threads 0 is good -1 is bad`, but would be something to keep in mind for any internals that can actually benefit from docs.

kernel/thread/thread.c
QuzarDC: Dug deeper. On thread creation we do a memset 0, which is why this was working fine for new threads. So it's already 0 initted as a default.

kernel/thread/thread.c
QuzarDC: I think the thing that I missed in my thinking on it was that perf counters are by default enabled and that happens directly before threads are initted. So I was a bit concerned about something like the default thread having an uninitialized value for cpu_time.scheduled/total. But since new threads are 0-initted that would only be an issue when someone has disabled perf counters, created new threads, then reenabled perf counters. I think that'd be on the user, and am pretty sure this would be safe even if `(ns-scheduled) <0` it'd just give garbage that they can deal with.

What's the reason though for not just allowing the perfcounters to use the fallback value out of timer_ns_gettime64 if they happen to be disabled? I'm guessing it has something to do with the differences between those systems (counting nops?) but it would seem like that it might be preferable to get slightly incorrect info from them than the wildly incorrect from stopping the counting? If that could work as the backup then you could do away with all the checks for the perf counters being enabled.

kernel/thread/thread.c
```diff
@@ -613,6 +627,20 @@ int thd_set_prio(kthread_t *thd, prio_t prio) {
 /*****************************************************************************/
 /* Scheduling routines */
 
+static void thd_update_cpu_time(kthread_t *thd) {
+    if(perf_cntr_timer_enabled()) {
+
+        const uint64_t ns = perf_cntr_timer_ns();
+
+        if(thd_current) {
```
QuzarDC, line 635: It just occurred to me. Is this intended to be just checking thd_current for NULL? Because that's basically an assertable condition and we don't check for it anywhere else. Technically we set it to NULL during initialization, so we have a check for it during thread creation just to have default pwds for kern/reaper/idle but after init it should always be set.

kernel/thread/thread.c
gyrovorbis: Yeah, actually, everything you've said is correct here.

The reason I wasn't just falling back to `timer_ns_gettime64()` was simply because of fact the perf counters in this mode were not counting `SLEEP` instructions... but... I guess this may still be preferable to just garbage data... good point.

kernel/thread/thread.c
```diff
@@ -931,17 +964,29 @@ struct _reent *thd_get_reent(kthread_t *thd) {
...
 }
 
+uint64_t thd_get_cpu_time(kthread_t *thd) {
+    /* Check whether we should force an update immediately for accuracy. */
+    if(thd == thd_get_current() && perf_cntr_timer_enabled()) {
+        const uint64_t ns = perf_cntr_timer_ns();
+
+        thd->cpu_time.total += ns - thd->cpu_time.scheduled;
+        thd->cpu_time.scheduled = ns;
+    }
```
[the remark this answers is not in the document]
gyrovorbis: DOH. Yeah, seriously!

gyrovorbis:
@QuzarDC The main thing to note with these changes that I just made:

I had already made `perf_cntr_timer_ns()` gracefully fall back to the lower resolution `timer_ns_gettime64()` when the perf counter timer isn't enabled... So all of that conditional logic was totally unnecessary.

kernel/thread/thread.c
```diff
@@ -1059,6 +1091,8 @@ int thd_init(void) {
     thd_current = kern;
     irq_set_context(&kern->context);
 
+    thd_update_cpu_time(thd_current);
```
QuzarDC, line 1094: Good catch!

kernel/thread/thread.c
```diff
@@ -1004,6 +1035,7 @@ int kthread_key_delete(kthread_key_t key) {
 
 /* Init */
 int thd_init(void) {
+    uint64_t ns;
```
QuzarDC, line 1038:
```suggestion
```
Don't need this anymore.

include/kos/thread.h
```diff
@@ -690,27 +743,34 @@ int thd_pslist(int (*pf)(const char *fmt, ...));
...
+/** \brief  Initialize the threading system.
+    
     This is normally done for you by default when KOS starts. This will also
     initialize all the various synchronization primitives.
-
+    
     \retval -1              If threads are already initialized.
     \retval 0               On success.
-
+
```
QuzarDC, line 755:
```suggestion

```

include/kos/thread.h
```diff
@@ -690,27 +743,34 @@ int thd_pslist(int (*pf)(const char *fmt, ...));
...
 
-/** \brief   Initialize the threading system.
+/** \cond INTERNAL */
 
+/** \brief  Initialize the threading system.
+    
     This is normally done for you by default when KOS starts. This will also
     initialize all the various synchronization primitives.
-
+
```
QuzarDC, line 752:
```suggestion

```

include/kos/thread.h
```diff
@@ -572,20 +604,41 @@ int *thd_get_errno(kthread_t *thd);
 */
 struct _reent *thd_get_reent(kthread_t *thd);
 
+
+/** \brief       Retrieves the thread's elapsed CPU time
+    \relatesalso kthread_t
+
+    Returns the amount of active CPU time the thread has consume in
```
QuzarDC, line 611:
```suggestion
    Returns the amount of active CPU time the thread has consumed in
```

QuzarDC (changes requested):
Just a few tiny things that crept in. Tried to make them all 'suggestions' to speed it along.

QuzarDC:
> @QuzarDC The main thing to note with these changes that I just made:
> 
> I had already made `perf_cntr_timer_ns()` gracefully fall back to the lower resolution `timer_ns_gettime64()` when the perf counter timer isn't enabled... So all of that conditional logic was totally unnecessary.

That's exactly why I was suggesting it :)

QuzarDC (approved):

PR #507, opened 2024-04-11
Fix erroneous timeout math and change implementation to CLOCK_MONOTONIC

darcagn:
* Switches our pthreads implementation in `libpthread` branch to use `CLOCK_MONOTONIC` instead of `CLOCK_REALTIME`.
* Currently the pthread implementation erroneously calculates timeout `tmo` by calculating the difference in seconds and the difference in milliseconds, and then adding seconds and milliseconds together. This PR fixes the issue by converting seconds to milliseconds before adding to the difference in milliseconds.

### DIFF
--- a/addons/libpthread/pthread_cond_timedwait.c
+++ b/addons/libpthread/pthread_cond_timedwait.c
@@ -2,6 +2,7 @@
 
    pthread_cond_timedwait.c
    Copyright (C) 2023 Lawrence Sebald
+   Copyright (C) 2024 Eric Fradella
 
 */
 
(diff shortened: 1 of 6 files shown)


gyrovorbis:
If this PR also fixes the condition variable timing issue with Rust: https://github.com/KallistiOS/KallistiOS/pull/513, then I say we pull that out of this PR and just use this PR for the timing calculation fixes.

OH. But make sure that the other `timedwait()` functions aren't suffering from the same issue... I know there's at least `mutex_timedlock()`, then a couple for `rwlock_t`.

darcagn:
Closing this PR at the moment as @gyrovorbis will incorporate my changes and his changes into a single branch and PR.

PR #509, opened 2024-04-14
init: Add KOS_NO_SHUTDOWN flag to disable shutdown

pcercuei:
Serious homebrew games and apps generally don't have any way to return to the calling application; exiting them means rebooting the console.

Those applications can now shave off a few kilobytes of RAM by using the KOS_NO_SHUTDOWN init flag.


gyrovorbis (approved):
Awesome. Didn't even think about the savings for GC-ing the "shutdown" code!

PR #511, opened 2024-04-18
fs: Remove duplicated fcntl() function

pcercuei:
Newlib already provides fcntl(), which will end up calling _fcntl_r() provided by KOS. This previously did not cause problems as the linker would pick the first symbol found (which means the link order mattered).

Rust however will complain that the symbol is duplicated. Address this by removing the KOS' fcntl(), which should not have any side effects.

### DIFF
--- a/kernel/fs/fs.c
+++ b/kernel/fs/fs.c
@@ -696,7 +696,7 @@ int fs_rmdir(const char * fn) {
     }
 }
 
-int fs_vfcntl(file_t fd, int cmd, va_list ap) {
+static int fs_vfcntl(file_t fd, int cmd, va_list ap) {
     fs_hnd_t *h = fs_map_hnd(fd);
     int rv;
 
--- a/kernel/libc/newlib/newlib_fcntl.c
+++ b/kernel/libc/newlib/newlib_fcntl.c
@@ -5,25 +5,13 @@
 
 */
 
-#include <sys/reent.h>
 #include <sys/fcntl.h>
-#include <stdarg.h>
 
 #include <kos/fs.h>
+
+struct _reent;
 
 int _fcntl_r(struct _reent *reent, int fd, int cmd, int arg) {
     (void)reent;
     return fs_fcntl(fd, cmd, arg);
-}
-
-extern int fs_vfcntl(file_t fd, int cmd, va_list ap);
-
-int fcntl(int fd, int cmd, ...) {
-    va_list ap;
-    int rv;
-
-    va_start(ap, cmd);
-    rv = fs_vfcntl(fd, cmd, ap);
-    va_end(ap);
-    return rv;
 }


darcagn:
Thanks for this! However, I'm not sure if we should've removed it on the KOS side instead of the Newlib side. The `fcntl()` in Newlib has the function signature `fcntl(int fd, int flag, int arg)` (see [newlib repo](https://sourceware.org/git/?p=newlib-cygwin.git;a=tree;f=newlib/libc/syscalls;h=e33f581fbbecf490086f99e2c0bec72c2de3b9cc;hb=26f7004bf73c421c3fd5e5a6ccf470d05337b435)) whereas in KOS it's `fcntl(int fd, int cmd, ...)`. The KOS signature with variadic arguments matches the existing bindings in Rust's `libc` bindings crate.

gyrovorbis:
If we do this, are we going to lose the ability to call `fcntl()` with truly variadic args? Since we're only ever implementing the scenario when an integer is passed to from Newlib, aren't we losing functionality technically here by doing this? There's no hook now for handling the truly variadic case?

pcercuei:
@gyrovorbis In which particular case do you need to call `fcntl()` with variadic args? ;)

If the case ever happens, we can fix it in Newlib. The important thing is that it won't break the ABI.

gyrovorbis:
> @gyrovorbis In which particular case do you need to call `fcntl()` with variadic args? ;)

Durrr... You got me... I have zero use-cases for that currently... Sounds like a plan.

gyrovorbis (approved):
Okay, so I kind of agonized about this for awhile, then I realized... I think @pcercuei is right, and we're going to be fine with removing the KOS-side definition.

APPARENTLY, according to the POSIX spec, `fcntl()` isn't *truly* fully variadic anyway. It simply takes an optional third argument which may or may not be supplied... 

Now as for what that third argument is:
1) From newlib it's always an `int`
2) I went back through every KOS VFS driver, and all of the ones that did use the optoinal third argument did so as a `long`, which is the same size as an `int`...

So in terms of what we have going on, I cannot see how this could possibly break... The only thing I *guess* could cause issues is if we did something weird like added support for a custom command type that takes a float in some hypothetical file system... then we used it with like `-m4-single`.... In that case, the arg would be promoted to a `double`, which is no longer the same size as an `int` and is no longer getting passed through the same registers... But this will probably never ever happen... lol.

So anyway, I'm approving this.

pcercuei:
> The only thing I _guess_ could cause issues is if we did something weird like added support for a custom command type that takes a float in some hypothetical file system... then we used it with like `-m4-single`.... In that case, the arg would be promoted to a `double`, which is no longer the same size as an `int`

Why would it be promoted as a `double`? That's not what `-m4-single` (or `-m4` for that matter) does.

PR #513, opened 2024-04-20
Added support for pthread_condattr_setclock()

gyrovorbis:
- Added public and private unions for `pthread_condattr_t`.
- Added attribute pointer to `pthread_cond_t`
- Initialized `pthread_condattr_t::clock_id` to `CLOCK_REALTIME` by default
- Implemented `pthread_condattr_setclock()` with support for `CLOCK_MONOTONIC` and `CLOCK_REALTIME`
- Warezed darc's timing fixes for `pthread_cond_timedwait()`
- Made `pthread_cond_timedwait()` respect the requested clock ID
- Added all of @darcagn's other fixes for `mutex` and `rwlock` timed functions

@darcagn can you verify if this solves the same problem that changing the clock default to `CLOCK_MONOTONIC` did for you?


addons/libpthread/pthread-internal.h
```diff
@@ -39,7 +49,10 @@ typedef union pthread_mutex_t {
 } pthread_mutex_t;
 
 typedef union pthread_cond_t {
-    condvar_t cond;
+    struct {
+        condvar_t cond;
+        const pthread_condattr_t *attr;
```
ljsebald, line 54: Storing a pointer like this is almost certainly not the correct way to do this.

If the attribute is allocated on the stack and goes out of scope before the condvar does, this will probably break. I suspect that that is not exactly an uncommon case to have happen (i.e, some function sets up/inits things for a statically allocated condvar in .bss/.data, but the attr is made on the stack in that function since it shouldn't need to be around after the condvar is created).

ljsebald: You probably just want to copy the clockid into the structure (or copy the whole condattr into it if you want to be really safe).

ljsebald: Also, since this will probably change structure a bit of other parts of this, I'll hold off on reviewing the rest of the PR until this is resolved.

gyrovorbis:
@ljsebald Alrighty, fixed the stupid and bumped up the `pthread_condattr_t`'s default size to `16`. Ready for review (again). :)

ljsebald (approved):

PR #516, opened 2024-04-20
init: Add init flags to enable/disable DCLOAD, CDROM

pcercuei:
Add a `INIT_CDROM` init flag, enabled by default, which can control whether the CD-ROM will be initialized and shut down automatically by KallistiOS.

Use the existing `INIT_NO_DCLOAD` flag to garbage-collect dcload code in case that flag is present.


gyrovorbis (approved):
Awesome! Curious if you checked how many KB this frees up with all of these flags disabled?

pcercuei:
Something about ~60 KiB, I think.

PR #517, opened 2024-04-20
maple: Fix race between IRQ and threads

pcercuei:
The IRQ handler and threads would read-modify-write the state->queue_len field as if races didn't exist.

Address this issue by using atomics instead.

Note that the code in the IRQ handler is using atomic routines as well just to be future-proof, as it may very well be called from a threaded interrupt handler in the future.

This has the added benefit that it also tells GCC that the 'queue_len' may be updated outside the visible scope, and therefore its value may change at any moment. This fixes the code not working properly when building with LTO.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/maple/keyboard.h
+++ b/kernel/arch/dreamcast/include/dc/maple/keyboard.h
@@ -28,6 +28,8 @@ __BEGIN_DECLS
 #include <arch/types.h>
 #include <dc/maple.h>
 
+#include <stdatomic.h>
+
 /** \defgroup kbd   Keyboard
     \brief          Driver for the Dreamcast's Keyboard Input Device
     \ingroup        peripherals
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/include/dc/maple/keyboard.h
```diff
@@ -297,7 +299,7 @@ typedef struct kbd_state {
     uint32 key_queue[KBD_QUEUE_SIZE];
     int queue_tail;                     /**< \brief Key queue tail. */
     int queue_head;                     /**< \brief Key queue head. */
-    int queue_len;                      /**< \brief Current length of queue. */
+    atomic_int queue_len;               /**< \brief Current length of queue. */
```
gyrovorbis, line 302: Shouldn't this really be atomic_size_t?

gyrovorbis, line 302:
```suggestion
    atomic_size_t queue_len;               /**< \brief Current length of queue. */
```

pcercuei: It was an `int`, not a `size_t`. If I change it to a `atomic_size_t` then I'm introducing a change that's completely unrelated to the bug being fixed.

gyrovorbis: You're right. I didn't even realize the damn thing started out as an `int`... Oops.

gyrovorbis (approved):
Looks good. Thanks!

PR #518, opened 2024-04-21
Updated dc-chain for DLang +  BETA m4-single support

gyrovorbis:
- Added an option for D to the pass2 langauges list, with a note that dependencies must already have been installed.
- Added `--disable-libphobos` flag to the pass2 script to prevent it from being built with D (which doesn't support it), but it's also gracefully ignored without D.
- Added `-m4-single` to the default list of FP configurations to build WHILE STILL KEEPING a strict warning about early beta support for it.

@darcagn If you've reviewed this, can you help me port these changes to your other configs? Or at very least the new GCC14 config, I guess.

### DIFF
--- a/utils/dc-chain/config/config.mk.13.2.1-dev.sample
+++ b/utils/dc-chain/config/config.mk.13.2.1-dev.sample
@@ -114,24 +114,27 @@ install_mode=install-strip
 # detected on some OS.
 makejobs=-j2
 
-# Languages (c|c++|objc|obj-c++)
-# Set the languages to build for pass 2 of building gcc for sh-elf. The default
-# here is to build C, C++, Objective C, and Objective C++. You may want to take
-# out the latter two if you're not worried about them and/or you're short on
-# hard drive space.
+# Languages (c|c++|objc|obj-c++|d)
+# Set the languages to build for sh-elf gcc compilation pass 2. The default is
+# to build everything used by all KOS examples; however, only C is actually
+# required for KOS itself. D may also be enabled optionally on POSIX platforms
+# which have its external dependencies installed.
 pass2_languages=c,c++,objc,obj-c++
 
-# Floating point precision support (m4|m4-single|m4-single-only)
+# Floating point precision support (m4-single-only|m4-single|m4)
 # Build support for various SH4 floating-point operation ABIs. KallistiOS only
-# officially supports single-precision-only mode. Add m4 (double precision) or
-# m4-single (single precision) to build experimental support for those ABIs.
-precision_modes=m4-single-only
-#precision_modes=m4,m4-single,m4-single-only
+# officially supports single-precision-only mode (m4-single-only); however,
+# experimental support for single-precision-default mode (m4-single) has just
+# been added to allow for the use of full 64-bit doubles. You may also add m4
+# (double-precision-default) which hasn't been tested, or you can simply stick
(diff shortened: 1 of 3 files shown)


ljsebald:
> Why not switch completely to `-m4-single`? YOLO :)

Absolutely not. I will not allow that switch at this point.

gyrovorbis:
> > Why not switch completely to `-m4-single`? YOLO :)
> 
> 
> 
> Absolutely not. I will not allow that switch at this point.

Omg I was literally about to respond with "...because Crabby will have our heads!" Lmao.

I did want to make it something that gets built by default, though, because I'm seeing people wanting it for raylib (does the double time step shenanigans that objective-c did), PLUS -m4-single-only is literally broken with C++23 (compiler ICE, already reported). 

Just seeing the same pattern of "darn I didn't know I needed to enable that in the config file!" stuff I was seeing with people missing the c99 formatters awhile back.

darcagn:
IMO I'm in support of having support for it built by default. Some things require `-m4-single-only` to work and some things require `-m4-single` so it seems reasonable now to just build both.

pcercuei:
@darc what requires `-m4-single-only`?

ljsebald:
> @darc what requires `-m4-single-only`?

KOS.

utils/dc-chain/config/config.mk.13.2.1-dev.sample
```diff
@@ -114,24 +114,27 @@ install_mode=install-strip
...
-# m4-single (single precision) to build experimental support for those ABIs.
-precision_modes=m4-single-only
-#precision_modes=m4,m4-single,m4-single-only
-
-# Default floating point mode (m4|m4-single|m4-single-only)
+# officially supports single-precision-only mode (m4-single-only); however,
+# experimental support for single-precision-default mode (m4-single) has just
+# been added to allow for the use of full 64-bit doubles. You may also add m4
+# (double-precision-default) which hasn't been tested, or you can simply stick
+# with m4-single-only to save disc space.
```
ljsebald, line 130:
```suggestion
# with m4-single-only to save disk space.
```

ljsebald (changes requested):
Spelling correction.

ljsebald (approved):
Approved, although gonna let @darcagn deal with merging and such (since this is really his domain).

pcercuei:
> > @darc what requires `-m4-single-only`?
> 
> KOS.

What in KOS requires `-m4-single-only`?

darcagn:
> @darc what requires `-m4-single-only`?

As I understand it there are some kos-ports that need adjusting but I don't know which ones exactly. GLdc was one of them but is fixed now, there are likely others. This needs testing. 

As far as KOS itself goes there are no *known* bugs or incompatibilities, but testing is only to the extent that a handful of people have been using it in their normal course of use lately. It's certainly not a complete coverage for testing. 

I want to switch to `-m4-single` by default too but we really should make a release before we introduce any more stuff that would break third party code. Then IMO we can run wild. In the mean time since there's more and more stuff relying on it (Raylib, Rust, Foundation, Fortran, etc.) it makes sense to build support and offer the option. With this PR it's trivial now.

darcagn:
Tested and `--disable-libphobos` is still gracefully ignored back in 4.7.4, so I'll go ahead and merge this.

PR #519, opened 2024-04-21
Fixed Release Build Warnings from Defining NDEBUG.

gyrovorbis:
- All of them were for unused variables. They were being used within `assert()` statements that have now been compiled out.
- Added `(void)` expressions around the variables to get GCC to shut up.
- Also fixed all Doxygen warnings while I was at it.

### DIFF
--- a/include/sys/dirent.h
+++ b/include/sys/dirent.h
@@ -62,11 +62,11 @@ __BEGIN_DECLS
     \headerfile sys/dirent.h
  */
 struct dirent {
-    int       d_ino;              /**< \brief File unique identifier. */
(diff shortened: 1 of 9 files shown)


pcercuei:
Why don't we start using `-Wno-unused-parameter`? Seriously. These `(void)variable;` are getting ridiculous.

gyrovorbis:
> Why don't we start using `-Wno-unused-parameter`? Seriously. These `(void)variable;` are getting ridiculous.

....I'm going to have to second this. It is really annoying. I think you can even leave off the parameter names like you can in C++ in C23 now... lol.

ljsebald (approved):

PR #520, opened 2024-04-21
Dynamically Modify Scheduler Frequency HZ at Runtime

gyrovorbis:
For the price of a single static `unsigned` variable, we can now allow the user to dynamically alter the scheduling timer interrupt frequency. This could be useful for scenarios when you might have different threading needs between cutscenes, regular gameplay, media playback, loadscreens, etc. 

- Defaults to the same `HZ` define we've always used
- `thd_set_hz()`: sets the current scheduler frequency (1-1000 hz)
- `thd_get_hz()`: gets the current scheduler frequency
- pulled in @pcercuei's `libudiv` to ensure that the division when calculating the timestep for scheduling the next wakeup is done as efficiently as possible, plus there are other places in KOS which could leverage it as well.


kernel/thread/thread.c
```diff
@@ -940,6 +943,19 @@ int thd_get_mode(void) {
     return thd_mode;
 }
 
+unsigned thd_get_hz(void) {
+    return thd_hz;
+}
+
+int thd_set_hz(unsigned hertz) {
+    if(!hertz || hertz >= 1000)
```
andressbarajas, line 951:
```suggestion
    if(!hertz || hertz > 1000)
```
In the documentation you said 1-1000 are valid values

gyrovorbis: Shit. You're right. I even fixed that one locally and forgot to commit. lol.

gyrovorbis:
Taking this bad boy back to draft, since zcrc mentioned I should be using libudiv to do the `1000 / thd_hz` calculations much more efficiently.

include/libudiv/udiv.h
```diff
@@ -0,0 +1,124 @@
...
+   Copyright (C) 2024 Paul Cercueil <paul@crapouillou.net>
+*/
+
+/** \file    libudiv/udiv.h
+    \brief   Fast uint32_t division routines
+    \ingroup libudiv
+
+    This file was originally from `libudiv`.
+
+    \author     Paul Cercueil
```
gyrovorbis, line 15:
```suggestion
    \author Paul Cercueil
```

include/libudiv/udiv.h
```diff
@@ -0,0 +1,124 @@
...
+#define __LIBUDIV_H__
+
+#include <kos/cdefs.h>
+
+/** \defgroup   libudiv     libudiv
+    \brief      Optimized unsigned 32-bit division routines
+    \ingroup    math_general
+
+    libudiv is an optimized math library that provides routines for efficiently
+    dividing an unsigned 32-bit variable by a known unsigned 32-bit constant.
```
pcercuei, line 28: libudiv is a small library that provides optimized unsigned 32-bit division routines, to be used when the divisor is constant or seldomly updated.

include/libudiv/udiv.h
```diff
@@ -0,0 +1,124 @@
...
+
+    @{
+*/
+
+/** \brief Divider for a uint32_t
+
+    This structure represents a cached, optimized form of a constant uint32_t
+    divisor value.
```
pcercuei, line 36: This structure represents a pre-processed form of an unsigned 32-bit divisor.

include/libudiv/udiv.h
```diff
@@ -0,0 +1,124 @@
...
+    unsigned int p = UDIV_P(div);
+    unsigned int m = UDIV_M(div, p);
+
+    return (udiv_t){ .p = p, .m = m, };
+}
+/** \endcond */
+
+/** \brief Returns a udiv_t from a constant uint32_t
+
+    Used to convert a constant uint32_t value into an optimized udiv_t form.
```
pcercuei, line 63: Convert a unsigned 32-bit divisor value into a pre-processed udiv_t form.

QuzarDC:
If going through all the trouble with udiv for this, why not just create a companion to HZ which stores the actual scheduling time (1000/HZ)? The udiv system is effectively storing the processed dividing info which is more than if we just stored the result since it never changes unless we're calling the set_divider again. That would be more efficient and use less memory.

Aside from it just not making much sense for this use case, I'd still advocate against bundling the two things together. If we did go with something like a udiv for these fast operations, the best way in my opinion would be to add it everywhere in the codebase that it would help at once. That way it's not an isolated 'what's this doing?' and is just the way we optimize some of these time-critical bits. 

Though for that it might make sense to have a coherent designation of time critical functions or areas of the code so that optimization can be prioritized over simplicity there.

One last little bit. Maybe change the comments around HZ to indicate at least that it's merely the default (or we do away with it as a define entirely, as you've already basically taken care of).

pcercuei:
Yeah caching the division would be smarter :)

gyrovorbis:
Right, that's a good point. But lets think about why that is for a sec and when employing libudiv would be the smarter option... It's cacheable just because the numerator happens to be a compile-time constant, while if it weren't, this would be the way we have to go, correct?

pcercuei:
@gyrovorbis it's cacheable because you can save the result of `1000 / div` into a static variable, instead of saving just the divisor (as udiv_t or not) and doing the division every single time.

QuzarDC:
> Right, that's a good point. But lets think about why that is for a sec and when employing libudiv would be the smarter option... It's cacheable just because the numerator happens to be a compile-time constant, while if it weren't, this would be the way we have to go, correct?

From reading through it, and the blog its based on, it's just the exact other case that it's best at. Where the denominator is static or changes rarely and the numerator is variable. It relies on preprocessing the denominator on update, then having a faster way to work with an arbitrary numerator. But even then, what are the savings overall for this? Is it just in a matter of a few cycles per division?

I struggle to see when or where in our codebase the value of this kind of clever dividing would be so beneficial so as to outweigh how non-standard it is. Maybe though I just don't know the scope of how much faster it is, or am forgetting some places where we do a lot of heavy dividing of uint32s frequently.

gyrovorbis:
Alrighty, I've removed libudiv, but I've kept it locally with all of the Doxygen work and everything in case we do see other places in the codebase that can benefit from it later on. Should be good to go.

include/kos/thread.h
```diff
@@ -598,6 +598,29 @@ int thd_set_mode(int mode) __deprecated;
 */
 int thd_get_mode(void) __deprecated;
 
+/** \brief   Set the scheduler's frequency.
+
+    Sets the frequency of the scheduler interrupts in hertz.
+
+    \param hertz    The new frequency in hertz (2-1000)
```
andressbarajas, line 605:
```suggestion
    \param hertz    The new frequency in hertz (1-1000)
```

andressbarajas: We can now have the value be one again

gyrovorbis: Crap. Good call!

andressbarajas:
You can probably add your copyright to the top of thread.h and thread.c. You did some TLS stuff in there too right?  More reason to add your copyright

gyrovorbis:
> You can probably add your copyright to the top of thread.h and thread.c. You did some TLS stuff in there too right? More reason to add your copyright

OMG I FREAKING FORGOT AGAIN.

One of my biggest regrets working on KOS was that after all of the TLS work (one of the hardest things I ever did), I didn't bother putting a copyright on the sacred `thread.h` or `thread.c`... So I just added it for 2023 as well.

...but at the same time, @cepawiel was absolutely vital in helping get TLS up and going as well, and he did personally modify both files along with me. I really think he needs to be on them as well.

doc/CHANGELOG
```diff
@@ -221,6 +221,7 @@ KallistiOS version 2.1.0 -----------------------------------------------
 - DC  Added new driver for the SH4's UBC + high-level breakpoint API [FG]
 - DC  Add support for French AZERTY keyboards [PC]
 - DC  Increased the resolution of pvr_stats_t from milli to nanoseconds [FG]
+- *** Added support for modifying scheduler frequency at runtime [FG && PC]
```
ljsebald, line 224: Since libudev was removed, is the attribution here still correct, or should it just be you, @gyrovorbis?.

include/kos/thread.h
```diff
@@ -598,6 +600,29 @@ int thd_set_mode(int mode) __deprecated;
...
+    Sets the frequency of the scheduler interrupts in hertz.
+
+    \param hertz    The new frequency in hertz (1-1000)
+
+    \retval 0       The frequency was updated successfully.
+    \retval -1      \p hertz is invalid.
+
+    \sa thd_get_hz(), HZ
+*/
+int thd_set_hz(unsigned hertz);
```
ljsebald, line 614:
```suggestion
int thd_set_hz(unsigned int hertz);
```

kernel/thread/thread.c
```diff
@@ -940,6 +945,19 @@ int thd_get_mode(void) {
     return thd_mode;
 }
 
+unsigned thd_get_hz(void) {
+    return 1000 / thd_sched_ms;
+}
+
+int thd_set_hz(unsigned hertz) {
```
ljsebald, line 952:
```suggestion
int thd_set_hz(unsigned int hertz) {
```

kernel/thread/thread.c
```diff
@@ -50,6 +52,9 @@ static inline size_t align_to(size_t address, size_t alignment) {
 /*****************************************************************************/
 /* Thread scheduler data */
 
+/* Scheduler timer interrupt frequency (Hertz) */
+static unsigned thd_sched_ms = 1000 / HZ;
```
ljsebald, line 56:
```suggestion
static unsigned int thd_sched_ms = 1000 / HZ;
```

ljsebald (changes requested):
Sorry... I really like to be more explicit about unsigned types than just `unsigned`.

doc/CHANGELOG
gyrovorbis: Updating.

ljsebald (approved):

PR #526, opened 2024-04-27
Update GCC configurations and patches for new separate GCC 15 and 14 branches, D support, and building m4-single support

darcagn:
- In anticipation of GCC 14 release in the next week or so, GCC 14 now has its own branch. So our master branch configuration is now updated to reflect being GCC 15.0.0, while the GCC 14 development branch is added in our configuration.
- Reverting changes made to patches in #500. This issue was fixed upstream and no longer applies.
- Adjusted all configuration files to show D language support as an option and add `-m4-single` support building by default, see #518. 
- Patch offsets adjusted to cleanly patch for all 3 dev configurations.


QuzarDC (approved):
Looks good to me. Tested with the new gcc 15 and it all built fine (had to update my libmpc though).

PR #528, opened 2024-04-27
Makefile.rules: Add rule to create binary executables

pcercuei:
Add a rule to create (unscrambled) .bin files from .elf files.


gyrovorbis:
Nice! i wonder if there should be a target or utility function added to the CMake stuff to do the same thing there...

ljsebald (approved):

PR #530, opened 2024-04-28
Patch newlib headers to expose lstat

QuzarDC:
Should resolve https://github.com/KallistiOS/kos-ports/issues/54 issue with libjimtcl and the missing lstat.

For whatever reason, I was getting warnings if I didn't add the extra newline at the end of the first patch file. It takes me quite a while to rebuild my toolchain, so I only actually tested the 4.4.0 and figured that perhaps others could test further back (@darcagn ?) The nature of the patch should mean that it'd be valid all the way back.


darcagn:
Fixed an issue in your 4.4.0 patch, and built toolchains with the 4 Newlibs, all of which compile `libjimtcl` successfully. This can be merged now @QuzarDC.

darcagn (approved):

PR #533, opened 2024-04-30
dc-chain: Improvements in functionality and documentation

darcagn:
- GCC 4.7.4 is removed.
- GCC 14.1.0 is added.
- Software package versions and download types are now in individual "profile" `profiles/profile.[name].mk` files, while the actual toolchain customization options are consolidated in a single common `Makefile.cfg` file that is already provided for the user. The version profile can be selected in the `Makefile.cfg` file; the stable toolchain is chosen for the user by default. This should make things simpler for users and maintainers alike.
- Package download types are stored in `scripts/profile-defaults.mk` but can be overridden in individual profile files.
- In `Makefile.cfg` settings are categorized and formatted to make it easier to scan through for options.
- In `Makefile.cfg` many settings' descriptions are revised or elaborated to be more understandable for users not familiar with the toolchain setup.
- In the configuration, SH and ARM toolchain path options are separated. Users can now specify the entire path including the trailing `/sh-elf` and `/arm-eabi`. This now matches the setting in `environ.sh` more closely and makes things more flexible when building multiple alternate toolchains with their own custom paths.
- The `makejobs` option now takes the number of jobs directly, instead of with `-j` added (e.g., now one would specify `2` instead of `-j2`). 
- The toolchain debug symbol option was simplified: `toolchain_debug` is set to 1 to keep symbols, or 0 to strip symbols. Previously this was called `install_mode` and needed to be set to `install` or `install-strip`. 
- Programming language options are all separated into their own options now with more detailed descriptions.
- `scripts/banner-variables.mk` was eliminated and its contents just added to `scripts/banner.mk`.
- `make clean` commands were revised so that errors do not fill the console when cleaning files that have already been deleted or don't exist.
- The ARM toolchain is no longer built by default. `make build-arm` must be run separately to build the ARM toolchain.
- Toolchain profiles added for GCC forks for GCCRS and libgccjit development.
- New target behavior:
  - `make` or `make default` installs SH toolchain and GDB.
  - `make all` builds SH toolchain, ARM toolchain, and GDB.
  - `make build-sh4` builds SH toolchain only.
  - `make build-arm` builds ARM toolchain only.
  - `make gdb` builds GDB only.
- Updated Dockerfile to adjust for new changes. Added ability to pass `makejobs` to Dockerfile.
- Added Markdown links throughout the documentation to improve browsability.
- Converted `changelog.txt` to `CHANGELOG.md`.
- Removed the Stalin/Jim Ursetto messages from the top of every file. These are not original and were added by SiZiOUS in that manner when he broke out the scripts into many files, and many of those files do not contain anything they wrote. This tool has grown much larger beyond what they initially created, so their contributions and the contributions of others are listed in the `CONTRIBUTORS.md` and `CHANGELOG.md` files.
- Removed `kos-patch` target. This allowed us to include patches to KOS itself in dc-chain but it's currently unused and I think never used. It currently creates a stamp file outside of the repo tree, which I think is poor behavior, so I removed the feature outright instead of fixing it.
- Fixed issue where the `extra_configure_args` would be sent to GCC `./configure` multiple times. `extra_configure_args` should not have been used to pass optional arguments from `init.mk`, so `extra_configure_args` is renamed to `cpu_configure_args` to clarify intent and a separate `gcc_extra_configure_args` variable was created for the new purpose.


gyrovorbis:
Whelp, I've officially built GCC14.0.1 with C, C++, ObjC, ObjC++, D, Ada, Rust, and libgccjit along with a bunch of the more advanced configuration flags, and everything worked great for me with this new model.

darcagn:
I'm marking the PR ready for review now! There are still some documentation adjustments I'd like to make, but they mostly involve checking and updating outdated instructions. I'll work on that in a follow-up PR both so I can get feedback after this one and so that they will match up with documentation updates outside of `dc-chain` that are coming soon too.

gyrovorbis (approved):
This looks good. I cannot see anything wrong with it, but it's also a massive review with a lot of changes... I'm about to start rebuilding the toolchain right now and testing locally, but in case I fall asleep or die before it finishes and I merge, I did look at these changes, and I approve.

PR #535, opened 2024-05-01
Implement Standard C Signals

gyrovorbis:
Adding support for standard C signal types and API: `signal()`, `kill()`, `raise()`.


QuzarDC:
Some general notes:
'Disabling' signals *just* disables the irq handlers for automatically uptaking exception signals right? Cause it makes it seem as though they end up just unimplemented entirely otherwise.

This could/should take into account the exception handler that already is in dcload as well as what we set up for gdb which hooks into the same irqs.

The newlib patch could likely be updated to not have to patch around the stack protector anymore. There's also probably other functionality that could work if aware that we have signal.

PR #537, opened 2024-05-02
Scheduling Precision Upgrade

gyrovorbis:
FINALLY attempting to complete the remainder of the precision upgrades we an now enjoy from the new TMU driver all the way down to the scheduler level... WIP!

- timer_spin_sleep is a macro that resolves to timer_spin_sleep_ms().
- timer_spin_sleep_us() and timer_spin_sleep_ns() have been added.
- timer_spin_sleep_xxx() no longer even directly consume a TMU unit.
  Instead, they simply spin in a loop comparing a time delta against
timer_xx_gettime64() values.
- Lots of Doxygen update/cleanup.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/timer.h
+++ b/kernel/arch/dreamcast/include/arch/timer.h
@@ -48,7 +48,7 @@ __BEGIN_DECLS
     of 80ns ticks.
 
     \warning
-    Under normal circumstances, all 3 TMU channels are reserved by KOS for
+    Under normal circumstances, 2 TMU channels are reserved by KOS for
     various OS-related purposes. If you need a free general-purpose interval
     timer, consider using the Watchdog Timer.
 
@@ -80,9 +80,7 @@ typedef enum timer_channel {
 
     /** \brief  SH4 Timer Channel 1.
 
-        \warning
-        This timer channel is used for the timer_spin_sleep() function, which
-        also backs the kthread, C, C++, and POSIX sleep functions.
+        This timer is currently free to use by an application or driver.
     */
     TMU1 = 1,
 
@@ -350,15 +348,11 @@ uint64_t timer_ns_gettime64(void);
 
 /** @} */
 
-/** \defgroup tmu_sleep     Sleeping
-    \brief                  Low-level thread sleeping
+/** \defgroup tmu_spin_wait Spin Waiting
+    \brief                  Low-level spin waiting
 
     This API provides the low-level functionality used to implement thread
-    sleeping, used by the KOS, C, C++, and POSIX threading APIs.
-
-    \warning
-    This API and its underlying functionality are using \ref TMU1, so any
-    direct manipulation of it will interfere with the API's proper functioning.
+    spin waiting/sleeping, used by the KOS, C, C++, and POSIX threading APIs.
 
     @{
 */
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/include/arch/timer.h
```diff
@@ -228,150 +232,202 @@ int timer_ints_enabled(int channel);
...
-    \ingroup tmu_uptime
 
     This function retrieves the number of seconds and milliseconds since KOS was
     started.
 
-    \param  secs            A pointer to store the number of seconds since boot
-                            into.
+    \note
+    To get the total number of milliseconds since boot, calculate
+    (*secs * 1000) + *msecs, or uses the timmer_ms_gettime64() function.
```
andressbarajas, line 265:
```suggestion
    (*secs * 1000) + *msecs, or use the timer_ms_gettime64() function.
```

pcercuei:
I tend to disagree with the implementation and the claim that it "complete the remainder of the precision upgrades":
- It does not address the elephant in the room, the fact that `thd_sleep(1)` will sleep for a minimum of 10 milliseconds;
- `timer_spin_sleep_ms()` shouldn't really exist IMHO, and `timer_spin_sleep()` should just resolve to `thd_sleep()` instead.

gyrovorbis:
> I tend to disagree with the implementation and the claim that it "complete the remainder of the precision upgrades":
> 
> * It does not address the elephant in the room, the fact that `thd_sleep(1)` will sleep for a minimum of 10 milliseconds;
> * `timer_spin_sleep_ms()` shouldn't really exist IMHO, and `timer_spin_sleep()` should just resolve to `thd_sleep()` instead.

1) Yeah, that's a fair point... I wasn't even fully aware of that fact when I first set out to start making these changes... but I guess the *scheduler* is what is updating `genwait()` with the delta time, NOT the other way around, so the best-case latency for a sleep is still going to be the remaining time before a the scheduler interrupt gets called... Basically the logic for the scheduler rescheduling itself should be the MIN between the scheduling interval or the next timeout from the `genwait()` queue...

2) Oh... yeah... That wasn't on my radar, because I was worried that it would be considered an API break to map `timer_spin_sleep()` to an actual thread sleep, but now that I think about it, IIRC if you attempt to sleep in an interrupt it basically does a spin sleep anyway, so this should actually still work in all scenarios.... So maybe this should be how we implement that call...

kernel/libc/koslib/nanosleep.c
```diff
@@ -23,24 +26,21 @@ int nanosleep(const struct timespec *rqtp, struct timespec *rmtp) {
...
-
-    /* We need to sleep for *at least* how long is specified, so if they've
-       given us a non-whole number of milliseconds, then add one to the time. */
-    if(rqtp->tv_nsec % 1000000)
-        ++ms;
+    ns = rqtp->tv_sec * 1000000000 + rqtp->tv_nsec;
 
     /* Make sure they gave us something valid. */
-    if(ms < 0) {
+    if(!(ns > 0)) {
```
andressbarajas, line 32: Does that read better than ns == 0?

gyrovorbis: No, I had a total moron moment. ;)

PR #538, opened 2024-05-02
examples/dreamcast/makefile rework

andressbarajas:
Instead of stopping the compilation process of all the examples when the first error occurs, save the error in a log and continue compiling.  At the end of compilation, report any errors.

Here is an example of output when it fails for one directory (including newline):

----------------------
Errors occurred during build:
Failed to build in 2ndmix


examples/dreamcast/Makefile
```diff
@@ -12,11 +13,36 @@ ifdef KOS_CCPLUS
...
-	for i in $(DIRS); do $(KOS_MAKE) -C $$i || exit -1; done
+ERROR_LOG = build_errors.log
 
-clean:
-	for i in $(DIRS); do $(KOS_MAKE) -C $$i clean || exit -1; done
+.PHONY: all clean dist show_errors clean_up
 
-dist:
-	for i in $(DIRS); do $(KOS_MAKE) -C $$i dist || exit -1; done
+all: show_errors clean_up
```
pcercuei, line 20: This looks wrong. `make -j2` will then run `make -C $$i clean` and `make -C $$i` in parallel.

gyrovorbis:
This is great... I have no idea wtf the mechanism is behind how the `build_all.sh` util script works in kos-ports, but it's gonna need this treatment too if someone's looking for something to do... really annoying how all kos-ports cease to build upon one single port failing.

andressbarajas:
After playing around with it a bit more, this still needs a lot more work.  The current changes only give the parent directory that failed.  For example directories with 1 level (2ndmix) that works fine but for deeply nested examples like gldc it would only output:

```
Failed to build in gldc
```

which is not too helpful.  I think Im gonna end up touching a lot of example makefiles for this one but I wont have the time to do that till next Tuesday.

I think it might be easier to do the kos-ports first.

andressbarajas:
I think I got the kos-ports version of this done.  I tested with my mruby dependency :^p:

https://github.com/KallistiOS/kos-ports/pull/65

examples/dreamcast/Makefile
andressbarajas: Thanks!  I took care of it.

andressbarajas:
Okay Round 2.  I updated the build algorithm in example/dreamcast/Makefile so it only runs make on child directories that have a makefile in it.  This is done so I can grab the path of examples that fail to build and log them to be outputted later.  Example of the output:

---------------------------------------------
2 Error(s) occurred during the build process:
2ndmix: Build failed with return code 2
kgl/demos/mipmap: Build failed with return code 2

QuzarDC (approved):

PR #539, opened 2024-05-03
pvr.h Refactor + Cleanup

gyrovorbis:
Guys, I'm sorry to say, but I have long thought `pvr.h` is probably the single worst source code file in all of KOS for a variety of offenses:
- It's too goddamn big and bloated, it needs to be broken down into multiple files. This makes it extremely hard to navigate through the file or to tell what is relevant when you're just trying to render a scene with the API.
- It is extremely bad about mixing driver implementation details and actual public API. The learning curve on this API is already extremely high, now factor in having to wade through register/memory maps and macros for internal command layouts when you're just trying to figure out how to render a polygon.
- It's very unidiomatic, newbish C: *everything* is an integer. Whether it's a bit mask, an unsigned type, something that is better suited as an enum, or something that's used as a bool, we're doing the pathological worst-case thing and are just making everything ever an int.

How much of the PVR API's learning curve is artificial due to the file just being a monstrosity? Because I honestly think the API itself is quite well designed and intuitive.

I am *positive* that I can clean this up in the same manner that I've been cleaning up other APIs/drivers as I have been touching them and break the thing into logical components. I'm also positive that I can do this all 100% without breaking the API, which I'm making a hard requirement, so don't worry! I will be testing that this still compiles against every PVR-using application I can get my hands on.

List of things this PR entails so far:

1) Dividing pvr.h into a series of individual subheaders within /dc/pvr
  which then get included within the main pvr.h header.
2) Cleaning up the styling and C idiomacy of the entire API
    a. boolean types where applicable
    b. stdint types where applicable
    c. enumeration where applicable
    d. typedefs where applicable
    e. const correctness
    f. struct tag names for all typedefed structs
3) Optimized pvr_poly_cxt_t structure for less memory (196 -> 73 bytes)
    a. packing all related enumerated types
    b. packing outer struct
    c. packing inner anonymous structs
    e. using booleans instead of ints for simple toggle fields
4) Added a todo list
5) Expanded upon documentation
    - Added more detailed descriptions on what primitive contexts are
      and how to use them
        * Added a warning about performance loss of recompiling every frame 
     - Removed some legacy documentation that wasn't relevant anymore 
6) Configured Doxygen so that the first line of comments above a symbol is assumed to be the \brief line without needing the tag to save
space
7) Created pvr_legacy.h file
    - where all macros which are doing nothing but reinventing boolean
      1/0 value types can go to get deprecated and die
    - we're now just making those fields bool


kernel/arch/dreamcast/include/dc/pvr/pvr_context.h
```diff
@@ -0,0 +1,635 @@
...
+*/
+void pvr_poly_cxt_txr_mod(pvr_poly_cxt_t *dst, pvr_list_t list,
+                          pvr_txr_fmt_t textureformat, size_t tw, size_t th,
+                          pvr_ptr_t textureaddr, pvr_filter_t filtering,
+                          pvr_txr_fmt_t textureformat2, size_t tw2, size_t th2,
+                          pvr_ptr_t textureaddr2, pvr_filter_t filtering2);
+
+__END_DECLS
+
+#endif
```
andressbarajas, line 635:
```suggestion
#endif

```

kernel/arch/dreamcast/include/dc/pvr/pvr_dma.h
```diff
@@ -0,0 +1,259 @@
...
+    \param  type            The type of SQ/DMA transfer to do (see list of modes).
+    \return                 The original value of dest.
+
+    \sa pvr_sq_set16()
+*/
+void *pvr_sq_set32(void *dest, uint32_t c, size_t bytes, pvr_dma_t type);
+
+__END_DECLS
+
+#endif
```
andressbarajas, line 259:
```suggestion
#endif

```

kernel/arch/dreamcast/include/dc/pvr/pvr_fog.h
```diff
@@ -0,0 +1,122 @@
...
+    have 129 entries, where the 0th entry is farthest from the eye and the last
+    entry is nearest. Higher values = heavier fog.
+
+    \param  tbl1            The table of fog values to set
+*/
+void pvr_fog_table_custom(float tbl1[]);
+
+__END_DECLS
+
+#endif
```
andressbarajas, line 122:
```suggestion
#endif

```

kernel/arch/dreamcast/include/dc/pvr/pvr_mem.h
```diff
@@ -0,0 +1,110 @@
...
+    \ingroup pvr_mem_mgmt
+
+    This prints out statistics like what malloc_stats() provides. Also, if
+    KM_DBG is enabled in pvr_mem.c, it prints the list of allocated blocks.
+*/
+void pvr_mem_stats(void);
+
+__END_DECLS
+
+#endif
```
andressbarajas, line 110:
```suggestion
#endif

```

kernel/arch/dreamcast/include/dc/pvr/pvr_misc.h
```diff
@@ -0,0 +1,184 @@
...
+    All of the color values should be between 0 and 1.
+
+    \param  a               Alpha value
+    \param  r               Red value
+    \param  g               Green value
+    \param  b               Blue value
+    \return                 The packed color value
+*/
+#define PVR_PACK_COLOR(a, r, g, b) ( \
+                                     ( ((uint8)( a * 255 ) ) << 24 ) | \
```
andressbarajas, line 44:
```suggestion
                                     ( ((uint8_t)( a * 255 ) ) << 24 ) | \
```

kernel/arch/dreamcast/include/dc/pvr/pvr_misc.h
```diff
@@ -0,0 +1,184 @@
...
+                                     ( ((uint8)( r * 255 ) ) << 16 ) | \
```
andressbarajas, line 45:
```suggestion
                                     ( ((uint8_t)( r * 255 ) ) << 16 ) | \
```

kernel/arch/dreamcast/include/dc/pvr/pvr_misc.h
```diff
@@ -0,0 +1,184 @@
...
+                                     ( ((uint8)( g * 255 ) ) << 8 ) | \
```
andressbarajas, line 46:
```suggestion
                                     ( ((uint8_t)( g * 255 ) ) << 8 ) | \
```

kernel/arch/dreamcast/include/dc/pvr/pvr_misc.h
```diff
@@ -0,0 +1,184 @@
...
+                                     ( ((uint8)( b * 255 ) ) << 0 ) )
```
andressbarajas, line 47:
```suggestion
                                     ( ((uint8_t)( b * 255 ) ) << 0 ) )
```

kernel/arch/dreamcast/hardware/pvr/pvr_prim.c
```diff
@@ -321,7 +321,7 @@ void pvr_sprite_cxt_txr(pvr_sprite_cxt_t *dst, pvr_list_t list,
     dst->txr.format = textureformat;
 }
 
-void pvr_sprite_compile(pvr_sprite_hdr_t *dst, pvr_sprite_cxt_t *src) {
+void pvr_sprite_compile(pvr_sprite_hdr_t *dst, const pvr_sprite_cxt_t *src) {
     int u, v;
     uint32 txr_base;
```
andressbarajas, line 326:
```suggestion
    uint32_t txr_base;
```

kernel/arch/dreamcast/hardware/pvr/pvr_prim.c
```diff
@@ -457,7 +457,7 @@ void pvr_mod_compile(pvr_mod_hdr_t *dst, pvr_list_t list, uint32 mode,
 
 /* Compile a polygon context into a polygon header that is affected by
    modifier volumes */
-void pvr_poly_mod_compile(pvr_poly_mod_hdr_t *dst, pvr_poly_cxt_t *src) {
+void pvr_poly_mod_compile(pvr_poly_mod_hdr_t *dst, const pvr_poly_cxt_t *src) {
     int u, v;
     uint32  txr_base;
```
andressbarajas, line 462:
```suggestion
    uint32_t  txr_base;
```

kernel/arch/dreamcast/hardware/pvr/pvr_scene.c
```diff
@@ -70,7 +70,7 @@ void * pvr_vertbuf_tail(pvr_list_t list) {
     return bufbase + pvr_state.dma_buffers[pvr_state.ram_target].ptr[list];
 }
 
-void pvr_vertbuf_written(pvr_list_t list, uint32 amt) {
+void pvr_vertbuf_written(pvr_list_t list, size_t amt) {
     uint32 val;
```
andressbarajas, line 74:
```suggestion
    uint32_t val;
```

kernel/arch/dreamcast/include/dc/pvr/pvr_texture.h
```diff
@@ -0,0 +1,136 @@
...
+                            \ref PVR_TXRLOAD_FMT_TWIDDLED) and
+                            \ref PVR_TXRLOAD_INVERT_Y in the flags.
+    \note                   DMA and Store Queue based loading is not available
+                            from this function if it twiddles the texture while
+                            loading.
+*/
+void pvr_txr_load_kimg(const kos_img_t *img, pvr_ptr_t dst, pvr_txr_load_t flags);
+
+__END_DECLS
+#endif
```
andressbarajas, line 136:
```suggestion
#endif

```

kernel/arch/dreamcast/include/dc/pvr/pvr_palette.h
```diff
@@ -0,0 +1,93 @@
...
+    \param  idx             The index to set to (0-1023)
+    \param  value           The color value to set in that palette entry
+*/
+static inline void pvr_set_pal_entry(size_t idx, uint32_t value) {
+    PVR_SET(PVR_PALETTE_TABLE_BASE + 4 * idx, value);
+}
+
+__END_DECLS 
+
+#endif
```
andressbarajas, line 93:
```suggestion
#endif

```

kernel/arch/dreamcast/include/dc/pvr/pvr_prim.h
```diff
@@ -0,0 +1,399 @@
...
+    uint32_t d4;                 /**< \brief Dummy value */
+    uint32_t d5;                 /**< \brief Dummy value */
+    uint32_t d6;                 /**< \brief Dummy value */
+} pvr_modifier_vol_t;
+
+/** @} */
+
+__END_DECLS
+
+#endif
```
andressbarajas, line 399:
```suggestion
#endif

```

kernel/arch/dreamcast/include/dc/pvr/pvr_legacy.h
```diff
@@ -0,0 +1,147 @@
...
+    \ingroup                        pvr_ctx_modvol
+    @{
+*/
+#define PVR_MODIFIER_CHEAP_SHADOW   0
+#define PVR_MODIFIER_NORMAL         1
+/** @} */
+
+__END_DECLS
+
+#endif
```
andressbarajas, line 147:
```suggestion
#endif

```

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -20,1718 +21,189 @@
...
-
-/** \brief   Set the vertex fog color.
-    \ingroup pvr_fog
+    \todo 
+        - Auxiliary accumulation buffer
+        - RTT texture sizes
+        - Multipass rendering
+        - pvr_poly_ctx_t::specular2
+            - specular enabled within modifier?
+        - Missing fog mode
```
andressbarajas, line 30: Oh I see you bundled all the todo on the main header

kernel/arch/dreamcast/include/dc/pvr/pvr_fog.h
```diff
@@ -0,0 +1,122 @@
...
+    \param  g               Green value of the fog
+    \param  b               Blue value of the fog
+*/
+void pvr_fog_table_color(float a, float r, float g, float b);
+
+/** \brief   Set the vertex fog color.
+    \ingroup pvr_fog
+
+    This function sets the fog color for vertex fog. 0-1 range for all colors.
+    This function is currently not implemented, as vertex fog is not supported
```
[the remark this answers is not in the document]
gyrovorbis: Should already have one at the top of `pvr.h`! But yep!

gyrovorbis:
*pant* *cough* *wheeze*

SHE'S DONE!!!! Everything has been rebuilt, tested, redocumented, re-reviewed, etc... I know this is going to be a PITA to look at, but 90% of it was just moving stuff around, with the other 10% being trivial styling like using idiomatic C types.

kernel/arch/dreamcast/hardware/pvr/pvr_fog.c
```diff
@@ -229,10 +229,7 @@ void pvr_fog_table_color(float a, float r, float g, float b) {
 
 /* Set the fog vertex color */
 void pvr_fog_vertex_color(float a, float r, float g, float b) {
-    (void)a;
-    (void)r;
-    (void)g;
-    (void)b;
+    PVR_SET(PVR_FOG_VERTEX_COLOR, PVR_PACK_COLOR(a, r, g, b));
```
andressbarajas, line 232: I see that you still have the assert and commented out version of this line below.  Does it work now or is it still not supported?

gyrovorbis: Wow, good eye...

while I was refactoring, this one really bothered me, because I couldn't understand how it could be unsupported, tbh. Maybe this code was from before support for it got added to the polygon/sprite contexts which set the corresponding field in the polygon/sprite header? ...just not seeing why this wouldn't work.

...but since I haven't actually tested this, I'm going to go ahead and remove this line again... I have captured a `\todo` item for this, but if anyone wants to look into it for another PR, it might just magically work with modern KOS.

gyrovorbis, line 232:
```suggestion
```

doc/Doxyfile
```diff
@@ -217,7 +217,7 @@ JAVADOC_BANNER         = NO
 # requiring an explicit \brief command for a brief description.)
 # The default value is: NO.
```
tsymalla, line 218: Needs update since you changed the default value

darcagn: That's the default value for the upstream example Doxyfle, so we'll keep the text as is here.

andressbarajas:
No, it has the 2.2.0 tag on it so it will not be in the 2.1.0 release.

gyrovorbis:
Rebased and ready-to-go for `2.2.0`!

pcercuei:
I believe you could rename `<pvr/pvr_foo.h>` to `<pvr/foo.h>`.

I also think that "pvr_regs.h" and "pvr_prim.h" contain implementation details that should not be directly visible in the API - at least not by including <kos.h>. Stuff like register addresses etc. should not be part of the visible API.

Then, I think some fields should be renamed (and we can use unions to avoid breaking the API). For instance the `cxt.blend.dst_enable`, documented as "dest blending enable", and usable with `PVR_BLEND_ENABLE`, does actually something very differently (set the second accumulation buffer as the target).

Finally, and this can be part of a different PR, but it would be great to have the geometry header structures directly usable, without having to use a "context" structure that is "compiled" later.

gyrovorbis:
> I believe you could rename `<pvr/pvr_foo.h>` to `<pvr/foo.h>`.
> 
> I also think that "pvr_regs.h" and "pvr_prim.h" contain implementation details that should not be directly visible in the API - at least not by including <kos.h>. Stuff like register addresses etc. should not be part of the visible API.
> 
> Then, I think some fields should be renamed (and we can use unions to avoid breaking the API). For instance the `cxt.blend.dst_enable`, documented as "dest blending enable", and usable with `PVR_BLEND_ENABLE`, does actually something very differently (set the second accumulation buffer as the target).
> 
> Finally, and this can be part of a different PR, but it would be great to have the geometry header structures directly usable, without having to use a "context" structure that is "compiled" later.

I 100% agree with you on every front. I simply cut this PR off at not doing anything at all to break the API or add any new features, because it was already becoming a beast.

I don't know if those things should be a part of this PR or not, but agreed, we should totally do it for 2.2.0!

andressbarajas:
I have verified that these changes do not break the gldc or pvr examples.

andressbarajas:
^ These two PRs should fix all compile errors for the examples after this PR has been merged.

andressbarajas:
@QuzarDC Anything we can do to help push this along to being merged?

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -17,1773 +18,206 @@
...
-    \ingroup pvr_fog
+    \remark
+    This file is used for pretty much everything related to the PVR, from memory
+    management to actual primitive rendering.
 
-    This function sets the fog color for vertex fog. 0-1 range for all colors.
-    This function is currently not implemented, as vertex fog is not supported
-    by KOS. Calling this function will cause an assertion failure.
+    \todo 
+        - Auxiliary accumulation buffer
```
pcercuei, line 26: That is already supported.

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -17,1773 +18,206 @@
...
+    primitive lists. There is a separate list for each transparency type for
+    both primitives (polygons or sprites) and modifier volumes. 
+*/
+typedef enum pvr_list {
+    PVR_LIST_OP_POLY = 0,  /**< Opaque polygon list */
+    PVR_LIST_OP_MOD  = 1,  /**< Opaque modifier list */
+    PVR_LIST_TR_POLY = 2,  /**< Translucent polygon list */
+    PVR_LIST_TR_MOD  = 3,  /**< Translucent modifier list*/
+    PVR_LIST_PT_POLY = 4,  /**< Punch-thru polygon list */
+    PVR_LIST_COUNT   = 5   /**< Number of list types */
```
pcercuei, line 88: No need for the explicit numbering. The same goes for every other enum.

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -1819,29 +253,32 @@ void pvr_mem_stats(void);
...
+    @{
 */
 
-/** \brief   Is vertex DMA enabled?
-    \ingroup pvr_vertex_dma
+/** Is PVR vertex DMA enabled?
     
-    \return                 Non-zero if vertex DMA was enabled at init time
+    \retval  true       Vertex DMA was enabled at init time
+    \retval  false      Vertex DMA was not enabled at init time
```
pcercuei, line 277: Maybe `\return True if vertex DMA was enabled at init time, false otherwise`

kernel/arch/dreamcast/include/dc/pvr/pvr_context.h
```diff
@@ -0,0 +1,612 @@
...
+/** \brief    PowerVR texture sampling modes
+    \ingroup  pvr_ctx_texture
+*/
+typedef enum __packed pvr_filter {
+    PVR_FILTER_NONE       = 0,   /**< \brief No filtering (point sample) */
+    PVR_FILTER_NEAREST    = 0,   /**< \brief No filtering (point sample) */
+    PVR_FILTER_BILINEAR   = 2,   /**< \brief Bilinear interpolation */
+    PVR_FILTER_TRILINEAR1 = 4,   /**< \brief Trilinear interpolation pass 1 */
+    PVR_FILTER_TRILINEAR2 = 6    /**< \brief Trilinear interpolation pass 2 */
+} pvr_filter_t;
```
pcercuei, line 202: I would actually use 0/1/2/3 here, because it is clear that bit 0 does not encode for the filter; and fixing it in pvr_poly_compile() and friends. That shouldn't break the API.

kernel/arch/dreamcast/include/dc/pvr/pvr_fog.h
```diff
@@ -0,0 +1,134 @@
...
+    responsibility to set up the `PVR_FOG_DENSITY` register by calling
+    pvr_fog_far_depth() with an appropriate value. The table passed in should
+    have `129` entries, where the `0`th entry is farthest from the eye and the last
+    entry is nearest. Higher values = heavier fog.
+
+    \param  tbl1            The table of fog values to set
+
+    \sa pvr_fog_table_exp2(), pvr_fog_table_exp(), pvr_fog_table_linear()
+*/
+void pvr_fog_table_custom(const float tbl1[]);
```
pcercuei, line 128: Either:
`const float *table`
Or:
`const float table[129]`

kernel/arch/dreamcast/include/dc/pvr/pvr_prim.h
```diff
@@ -0,0 +1,359 @@
...
+#define PVR_TA_PM3_TXRFMT          GENMASK(30, 21)
+/** @} */
+
+/** PVR polygon header.
+
+    This is the hardware equivalent of a rendering context; you'll create one of
+    these from your pvr_poly_cxt_t and use it for submission to the hardware.
+*/
+typedef struct pvr_poly_hdr {
+alignas(32)
```
pcercuei, line 117: `alignas(32)` is not... aligned?

pcercuei (changes requested):
A few things I noted too, is that some doxygen stuff (\defgroup, \see etc.) was dropped between the main pvr.h and the new headers.

kernel/arch/dreamcast/include/dc/pvr.h
gyrovorbis: You're right. At the time I had no idea how it was actually used, and everyone told me it wasn't something they knew how to use either... just need probably better docs and/or example.

kernel/arch/dreamcast/include/dc/pvr.h
gyrovorbis: You're right, but... now that I looked back, I remembered why I did it... Because their values actually do matter when the PVR itself is expecting them to be equal to certain things, and the values are going directly to the HW. I was worried not making them explicit would make it look like it's safe to change the values while explicitly enumerating would make it more clear that they are fixed?

...but I'm not very strong on this opinion. I actually dislike explicit enumeration like this in general. If you don't think this is an exception that warrants it, then I'm going with your vote, because I don't feel strongly either way.

pcercuei: The way I see it... those enums represent hardware values and as such shouldn't be changed, but:
- if for some reason we need new values we'd add them to the end, so it wouldn't be a problem;
- a PR that casually re-orders the enum entries would most likely cause breakages and wouldn't pass peer review (not my review at least).

kernel/arch/dreamcast/include/dc/pvr/pvr_prim.h
gyrovorbis: I don't know why, in my head it looked better like that, but now that you mention it, it actually looks like shit to me, so... Reverting. lol.

kernel/arch/dreamcast/include/dc/pvr.h
gyrovorbis: Done.

QuzarDC (commented):
Sorry for the lateness of these. I even wrote them a month ago when I was working on #874 but just hadn't hit commit. Almost all of them relate to the pieces I left out.

kernel/arch/dreamcast/hardware/pvr/pvr_fog.c
```diff
@@ -144,10 +144,9 @@ union ieee32_t {
 };
 
 /* useful macros */
-#define ABS(n) ( (n)<(0.0f) ? (-n):(n) )
-#define NEG(n) ( (n)>(0.0f) ? (-n):(n) )
-#define CLAMP01(x) \
-    ( (x)<(0.0f) ? ((x) = (0.0f)) : ((x)>(1.0f) ? ((x)=(1.0f)) : (x)) )
+#define ABS(n)     ((n) < (0.0f) ? (-n):(n))
+#define NEG(n)     ((n) > (0.0f) ? (-n):(n))
+#define CLAMP01(x) ((x) < (0.0f) ? ((0.0f)) : ((x) > (1.0f) ? ((1.0f)) : (x)))
```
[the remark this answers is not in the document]
pcercuei: One thing I would add: unless you need to write something that needs macro constructs, please always use static inline functions instead.

pcercuei (changes requested):

QuzarDC:
Worked through a cleaning up and splitting off a majority of the changes here in #874 #934 #1051 and #1288 . There's still quite a bit with our current `pvr.h` being double the size of the one here, but it's still half the size as it was when it started.

I've added the new label 'Revisit' to keep this searchable with other closed PRs that have useful changes, but it doesn't seem like there's any value to holding it open as a draft any further work would certainly be a new PR and just reference this.

PR #541, opened 2024-05-03
Allow the patched toolchain to support -m4-nofpu

QuzarDC:
As on tin. The startup.S becomes crt1.S in the toolchain. While it's likely pointless to have KOS support that mode, with a few minor changes, it can be used to build dcload-serial.

Needs extra things for obj-c to work, so keeping in draft briefly to see if I can get that sorted out.

Edit: Have rebased this and updated the notes with it to indicate the limitations (can't build KOS with it, no support for objc/cpp).

### DIFF
--- a/kernel/arch/dreamcast/kernel/startup.S
+++ b/kernel/arch/dreamcast/kernel/startup.S
@@ -109,6 +109,7 @@ memchk_done:
 	stc	vbr,r1
 	mov.l	r1,@r0
 
+#ifndef __SH4_NOFPU__
 	! Save FPSCR
 	mov.l	old_fpscr_addr,r0
 	sts	fpscr,r1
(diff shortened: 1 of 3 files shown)


QuzarDC:
@darcagn , would you have any interest in trying to figure out objc with this? With this you can add -m4-nofpu and it'll build for C but seemingly not for objc (and I think in gcc 15 c++ doesn't build either). The error I got seems to be an extremely generic 'something failed before this point' `error link tests are not allowed after gcc_no_executables` and maybe you'd have some ideas of where to look to adjust. 

I had also held back on adding any changes to the chain till you'd updated it, so now added a note there about possibly using it (similar to the plain 'm4' which we also don't really support).

ljsebald (approved):

PR #544, opened 2024-05-04
Fanalyzer ffixes

QuzarDC:
Some small fixes based on -fanalyzer feedback.

The vmufs one is a bit tricky, but vmu_dir_find checks the member filename for 0 to test if it's in use. Without 0-initializing the dir mem, it'll randomly go into the string comparison portion. I'm not 100% sure that it can make a difference since it must first be overwritten by vmu_block_read, but I don't think that's guaranteed to match the full size of the dir buffer.


gyrovorbis (approved):
These all look great... Went through them one-by-one, there's nothing wrong that I see, so I guess I'll go ahead and merge, because it's pretty trivial fixes.

PR #549, opened 2024-05-05
fb_console use memcpy4 and memset4

Tchan0:
There's no good reason to use memcpy2/memset2 here:
- All video modes are a multiple of 4, so fb_w is too,
- there is also * 2,
- the borders are a multiple of 4,
- the font height is a multiple of 4.

### DIFF
--- a/kernel/arch/dreamcast/util/fb_console.c
+++ b/kernel/arch/dreamcast/util/fb_console.c
@@ -76,10 +76,10 @@ static int fb_write(int c) {
         /* If going down a line put us over the edge of the screen, move
            everything up a line, fixing the problem. */
         if(cur_y + FONT_CHAR_HEIGHT > max_y) {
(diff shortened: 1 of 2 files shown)


andressbarajas:
@Tchan0 Nice find.  How would I test this?  Is there an existing example? Im guessing just using bfont_draw_str() to fill the screen wont hit this code?

darcagn:
> @Tchan0 Nice find. How would I test this? Is there an existing example? Im guessing just using bfont_draw_str() to fill the screen wont hit this code?

Just set the framebuffer as the output device:

`dbgio_dev_select("fb");`

darcagn (approved):

Tchan0:
@andressbarajas: just what @darcagn writes indeed.
I wrote a simple loop that writes 400 times a simple printf with a counter after dbgio_dev_select("fb");
With memset2/memcpy2, it took 76 seconds.
With memset4/memcpy4, it took 39 seconds

PR #550, opened 2024-05-06
Implemented C23 stdlib features for <time.h>

gyrovorbis:
- New macro for C23 compliance of the file
- Macros were redone to check for existence of C and C++ macros before "dereferencing"
- New C23 time base types
- `asctime()` and `ctime()` are now deprecated in C23
- C23 added `timespec_getres()` for getting resolution of a time base
- C23 added `gmtime_r()` and `localtime_r()` reentrant time converters
- Implemented timespec API by mapping it directly to POSIX clock API
- Removed `clock_gettime()`'s reliance on `timespec_get()` for `CLOCK_REALTIME`, so that the dependency could be reversed.
- Removed whitespace from RTC
- Enabling C23 compatibility in C++23, regardless of whether the C++ standards committee ever does interop, because I queried a bunch of devs, and they think it's "good guy" stdlib thing to do....
- Implemented `timegm()` which was added from POSIX in C23. It takes a broken-down UTC time and converts it into a Unix `time_t`, which is exactly what `mktime()` does, except its source broken-down time is in the local timezone... So given the fact that we have no timezone on DC, and our localtime is assumed to already be UTC, they are literally the exact same operation... This will have to be updated when we start actually attempting to retrieve a timezone from flash storage if it's present within web browser save data, but for now, given what we currently support, this is the correct implementation.


include/kos/time.h
```diff
@@ -35,50 +34,95 @@
...
+   #ifndef _POSIX_MONOTONIC_CLOCK
+      #define _POSIX_MONOTONIC_CLOCK 1
+   #endif
+
+   #ifndef _POSIX_CPUTIME
+      #define _POSIX_CPUTIME 1
+   #endif
+
+   /* We do NOT support thread-specific CPU time!
+   #ifndef _POSIX_THREAD_CPU_TIME
```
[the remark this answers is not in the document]
gyrovorbis: Good catch, my friend!!!

gyrovorbis:
Now that the Per-thread CPU time PR has been merged, I've enabled this in C23 via the `timespec_get()` API as well.

kernel/libc/c11/timespec_get.c
```diff
@@ -2,23 +2,75 @@
...
+}
+
+/*  C23/POSIX timegm() converts a UTC-based broken-down time to a unix
+    timestamp... and since we have no timezone on the Dreamcast (yet),
+    our local time is already UTC, which means this is going to be
+    equivalent to using mktime()! 
+*/
+time_t timegm(struct tm *timeptr) {
+    return mktime(timeptr);
+}
```
ljsebald, line 75: Why is this in this file? Please move it to its own file.

include/kos/time.h
```diff
@@ -35,48 +34,94 @@
...
-/* Microsecond resolution for clock(), per POSIX standard */
-#define CLOCKS_PER_SEC 1000000
+/* =============== Enable the following for >=c11, >=c++17 =================== */
+#if (defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 201112L)) || \
+    (defined(__cplusplus)      && (__cplusplus >= 201703L))
 
-/* C11 nanosecond-resolution timing. */
-extern int timespec_get(struct timespec *ts, int base);
+   /* Only supported base time in C11. */
+   #define TIME_UTC 1
```
ljsebald, line 50: Please don't indent things in like this inside #if blocks. We don't do that anywhere else in the code, so please don't do it here. And **especially** don't indent it by 3 spaces like this.

Note: this applies to the rest of the file too, of course.

ljsebald (changes requested):

kernel/libc/c11/timespec_get.c
```diff
@@ -2,23 +2,65 @@
...
+
+/* C11 timespec_get() -> POSIX clock_gettime(). */
+int timespec_get(struct timespec *ts, int base) {
+    return timespec_posix_adapter(ts, base, clock_gettime);
 }
+
+/* C23 timespec_getres() -> POSIX clock_getres(). */
+int timespec_getres(struct timespec *ts, int base) {
+    return timespec_posix_adapter(ts, base, clock_getres);
+}
```
ljsebald, line 66:
```suggestion
}

```

ljsebald (approved):

PR #551, opened 2024-05-08
Some cleanups in Maple

QuzarDC:
This is a set of small changes aimed at folding repeated code, removing unnecessary code, and standardizing bits that seem to only be different by accident. None of these should cause much functional change as  even the changes from while to if would only ever matter in exceptional cases.

- The vast majority of maple command functions attempt to lock the maple frame and if they fail they return EAGAIN to flag that it should be retried. A small set of ones were using a while loop passing the thread until the frame became unlocked. This seemed to be just a result of old copy-pasting. The one exception is `vmu_block_write_internal` that has a wrapper around it and behaves differently (I may also do the same there but just with the outer loop checking if it actually retried or just couldn't lock).

- It did not make sense that we were forcing 'status_valid' to all devices that do not have a periodic function. Currently pulling a dev_status on them would return garbage. Instead, remove those and the stubs that set status_valid =1 on attach. By checking if the driver has a periodic, we know if it will have a status that it can return, so we filter against that when retrieving that status. In this way, we reduce some overhead and maintain status_valid as an actual descriptor of status as valid or not.

- Every driver init function was checking if the driver had already been initiated before calling the driver_reg. With #476 the change it seemed to make sense to check for multiple drivers on the same function as a way to prevent double-init. The right way though I think is to move that double-init check into reg and remove the one from DCPlaya. There's no reason the driver architecture should prohibit two drivers from being attached for the same function type, it just can't have two of the same driver in the list.

- Every driver poll reply function would have the large block accessing the frame gated by a check that frm->dev was valid. I've inverted the logic on that to reduce nesting. In the same check I made more standard the assert of the reply size vs the cond variable (in one place it was checking sizeof the data unit differently, in one it was missing). To facilitate this, I migrated mouse.c/.h to use stdints as we have been doing throughout the codebase.

- In order to do the one above for the vmu button polling, corrected the vmu_cond_t to have a representation of the dummy bytes that are returned.

- Fixed a possible (maybe really impacting rumble?) corruption issue in `maple_dma_irq_hnd` where it would remove completed frames within the `TAILQ_FOREACH` to dispatch frames.


pcercuei:
It breaks device detection?

This is hello.c, on a dreamcast with a lot of stuff plugged:
```
KallistiOS Git revision v2.0.0-1381-g1aa38997:
  mer. 08 mai 2024 14:36:24 CEST
  paul@frank:/home/paul/dev/dreamcast/kos
  sh-elf-gcc (GCC) 15.0.0 20240427 (experimental)
thd: pre-emption enabled, HZ=100
maple: active drivers:
  DMA Buffer at ac0788c0
dc-load console support enabled
maple: attached devices:

Hello world!

arch: exit return code 0
arch: shutting down kernel
maple: final stats -- device count = 0, vbl_cntr = 8, dma_cntr = 6
```

QuzarDC:
> It breaks device detection?
> 
> This is hello.c, on a dreamcast with a lot of stuff plugged:
> 
> ```
> KallistiOS Git revision v2.0.0-1381-g1aa38997:
>   mer. 08 mai 2024 14:36:24 CEST
>   paul@frank:/home/paul/dev/dreamcast/kos
>   sh-elf-gcc (GCC) 15.0.0 20240427 (experimental)
> thd: pre-emption enabled, HZ=100
> maple: active drivers:
>   DMA Buffer at ac0788c0
> vid_set_mode: 640x480 VGA with 1 framebuffers.
> dc-load console support enabled
> maple: attached devices:
> 
> Hello world!
> 
> arch: exit return code 0
> arch: shutting down kernel
> maple: final stats -- device count = 0, vbl_cntr = 8, dma_cntr = 6
> vid_set_mode: 640x480 VGA with 1 framebuffers.
> ```

Yes. I had copy+pasted the logic for the driver detection wrong (the ! in the check in driver_reg). I had already fixed it locally, but haven't finished the work I was doing here, which is why it's still draft and not ready for review.

kernel/arch/dreamcast/hardware/maple/vmu.c
```diff
@@ -85,33 +85,33 @@ static void vmu_poll_reply(maple_state_t *st, maple_frame_t *frm) {
...
-                              (cooked->dpad_up    << 1) | /* down */
-                              (cooked->dpad_down  << 0) | /* up */
-                              (cooked->dpad_left  << 3) | /* right */
-                              (cooked->dpad_right << 2);  /* left */
+    if(!frm->dev)
+        return;
 
-        }
+    /* Verify the size of the frame and grab a pointer to it */
+    //assert(resp->data_len == 7);
```
andressbarajas, line 92: This line still meant to be commented out?

QuzarDC: Yea, I'm not familiar enough with the vmu button input stuff to have uncommented it and given it the same treatment. I'm can't be the same as the rest ( `assert(sizeof(vmu_cond_t) == ((resp->data_len - 1) * sizeof(uint32_t)));` ) because vmu_cond_t is only one byte.

This might be something @gyrovorbis could comment on, but I figured the proper thing would be to de-magic that number. Perhaps vmu_cond_t is actually meant to be 7 bytes and the rest is just dummy data, or not useful? (Like in mouse_cond_t).

gyrovorbis: Yeah, actually that's what it looks like. The BIOS is just writing 0s for the rest of the bytes.

QuzarDC: So, turns out that the 7 is just outright wrong. `resp->data_len` comes back as '2' which is the minimum size. It has one uint32 for the functions mask, then one for the payload. Only the first byte of the payload is used so the rest is dummy data. This makes a lot of sense since it just doesn't have anything else to add but the units are in uint32s for the response.

I've updated the vmu_cond_t to properly have those dummy bytes, which needed a bit of adjustment. Then the same assert pattern was possible as in every other polling function.

@gyrovorbis It would be great if you could test if this still works. I just don't have any controller that has the buttons exposed, and hopefully you still have the test program you wrote for this :) . Some day I'll take a hacksaw to one of my spare controllers to be able to fiddle with this more.

QuzarDC: > This line still meant to be commented out? I see that it existed in the old source. Do you know if we still need it or not?

Thanks for pointing it out. It was less work than I'd expected, and it really helped standardize that driver to match the rest. AND document the hardware slightly better :)

andressbarajas:
Did some testing. 

1. Stock controller => Works
2. Keyboard => Works
3. Mouse => Doesnt work (could be faulty controller; used mouse example in libdream)
4. vmu => Works
5. Light gun => Works
6. Dreameye => Strange output from basic DreamEye example:

```maple: attached devices:
  A0: Dreamcast Camera Flash  Devic (01000000: Controller)
  A1: Dreamcast Camera Flash LDevic (00080000: Camera)
  A2: Dreamcast Camera Flash LDevic (00080000: Camera)
  A3: Dreamcast Camera Flash LDevic (00080000: Camera)
  A4: Dreamcast Camera Flash LDevic (00080000: Camera)
  A5: Dreamcast Camera Flash LDevic (00080000: Camera)

KallistiOS Dreameye Test program
Attempting to find a connected Dreameye device...
Attempting to grab the number of saved images...
Image Count is valid -- (0)
Attempting to grab the first image.
Error was: -1```
7. Rumble => Detects but Rumble example didnt make it rumble I tried different values and two differe rumble packs. Here is some output:
```Rumble: 0x30000000!
Rumble: 0x30000000!
Rumble: 0x30200000!
Rumble: 0x30200000!
Rumble: 0x30907000!
Rumble: 0x30907000!
Rumble: 0x30907000!
Rumble: 0x30907500!
Rumble: 0x30907500!
Rumble: 0x30907500!
Rumble: 0x30907500!
Rumble: 0x30907500!
Rumble: 0x30907606!
Rumble: 0x30907606!
Rumble: 0x30907606!
Rumble: 0x30907606!
Rumble: 0x30907606!
Rumble: 0x30907606!
Rumble: 0x30907602!
Rumble: 0x30907602!
Rumble: 0x30907605!
Rumble: 0x30907605!
Rumble: 0x70907605!
Rumble: 0x70907605!
Rumble: 0x90907605!
Rumble: 0x90907605!
Rumble: 0x95907605!
Rumble: 0x95907a05!
Rumble: 0x95917a05!
Rumble: 0x95957a05!
Rumble: 0x95957a05!
Rumble: 0x95d97a05!
Rumble: 0x95d97a05!
Rumble: 0x95d97a05!
Rumble: 0x95d97f05!
Rumble: 0x95d97f05!
Rumble: 0x95d97f05!```

**Comment by gyrovorbis** [comment]
Interesting... I actually am seeing @QuzarDC's rumble example freeze on me after playing with it a bit and sending a rumble command.

**Comment by QuzarDC** [comment]
> Interesting... I actually am seeing @QuzarDC's rumble example freeze on me after playing with it a bit and sending a rumble command.

That's just so strange. But it is rumbling right? Any chance to control test against master to ensure it's not just general regression. The difference for puru should only be that in the two places that call attach/periodic, instead of getting a function that does nothing they test NULL and don't call the function that does nothing. And that now if you call dev_status on them you'll get NULL instead of a pointer to garbage status data (which nobody was doing anyways).

I didn't even think to test puru since the changes there were trivial :/

**Comment by andressbarajas** [comment]
Im experiencing the same rumble issues on master.  Maybe my rumbles paks are dead?

**Comment by gyrovorbis** [comment]
Shit... let me try real quick on master...

**Comment by gyrovorbis** [comment]
WHAT THE.... RED ALERT!!!!! I'm getting a freeze when playing around with the rumble example on `master` still!!!

**Comment by QuzarDC** [comment]
I took the example down to a minimal little 'send the rumble command, wait, send another' and it works fine. 

```#include <stdio.h>
#include <stdint.h>

#include <kos.h>

int main(int argc, char *argv[]) {
    
    maple_device_t *purudev = maple_enum_type(0, MAPLE_FUNC_PURUPURU);
 
    purupuru_rumble_raw(purudev, 0x3339f010);
    printf("Rumble: 0x%lx!\n", 0x3339f010);
    
    timer_spin_sleep(50000);
    
    purupuru_rumble_raw(purudev, 0);
    printf("Rumble: 0x%lx!\n", 0);
    
    timer_spin_sleep(100);
    
    purupuru_rumble_raw(purudev, 0x42153810);
    printf("Rumble: 0x%lx!\n", 0x42153810);
    
    timer_spin_sleep(50000);
    
    purupuru_rumble_raw(purudev, 0);
    printf("Rumble: 0x%lx!\n", 0);
    
    timer_spin_sleep(100);

    return 0;
}
```
I also tried running the example with only known valid effects and it works just fine.
```
maple: attached devices:
  A0: Dreamcast Controller           (01000000: Controller)
  A1: Visual Memory                  (0e000000: Clock, LCD, MemoryCard)
  A2: Puru Puru Pack                 (00010000: JumpPack)

pvr: disabling vertical scaling for VGA
Rumble: 0x3339f010!
Rumble: 0x3339f010!
Rumble Stopped!
Rumble: 0x3339f010!
Rumble Stopped!
Rumble: 0x3339f010!
Rumble Stopped!
Rumble: 0x3339f010!
Rumble Stopped!
Rumble Stopped!
Rumble: 0x3339f010!
Rumble Stopped!
Rumble Stopped!
Rumble: 0x3339f010!
Rumble Stopped!

arch: exit return code 0
arch: shutting down kernel
maple: final stats -- device count = 3, vbl_cntr = 8799, dma_cntr = 8798
fs_romdisk: unmounting image at 0x8c03789c from /rd
quzar@Centurion-IV:~/dcdev/KallistiOS/examples/dreamcast/rumble$
```

My output when I got it to freeze. It did not capture the last effect sent, which indicated that it never returned from `purupuru_rumble_raw`:
```
maple: attached devices:
  A0: Dreamcast Controller          (01000000: Controller)
  A1: Puru Puru Pack                (00010000: JumpPack)

pvr: disabling vertical scaling for VGA
Rumble: 0x0!
Rumble: 0x4000000!
Rumble: 0x4000100!
Rumble: 0x4000104!
Rumble: 0x4000114!
Rumble: 0x4002114!
Rumble: 0x4012114!
Rumble: 0x4022114!
^Cmake: *** [Makefile:25: run] Interrupt
```

As an additional pointer, the vmu_beep example also seems to work just fine, which is identical in the structure and its use of font/libparallax. My best guess at the moment is that there may be something deeper in either the inner workings of maple or perhaps specifically in the rumbling, that we've not addressed before to do with the behavior when sending invalid effect values.

I'll see about opening a new issue just for this. Likely the thing to do will be to rewrite the rumble example to take advantage of the newer macros we have for setting the specific effects and such and enforcing what we know about the limits in those. It *might* be the case that doing so resolves the issue, which would lend credence to the notion that it's to do with invalid effect values (and even if it's still the case, it would lead to a more useful example).

QuzarDC:
Accidentally moved it back to draft. I intend to test out the mouse and dreameye as well when I can to validate @andressbarajas 's reports for those as well.

If it turns out that all the behavior is also present in current master, then I'll move this out of draft.

ljsebald (approved):

PR #552, opened 2024-05-08
maple: Rework device detection, allocation and status buffers

pcercuei:
This PR updates the maple code to use smaller amounts of dynamically allocated memory, instead of huge amounts of statically allocated memory.

- Status buffers are now allocated to the right size;
- Maple device structures are only created for connected devices.

This required to change the detection mechanism, as it was now impossible to use a different maple frame for each possible combination of [port]x[unit]. The periodic detection will now use its own statically allocated maple frame and will only detect unit 0s; sub-devices will be detected using the unit 0's frame.

Additionally, maple devices are now properly detached on shutdown.

This saves about 48 KiB of static RAM.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
@@ -83,8 +83,7 @@ static void maple_hw_init(void) {
     /* Initialize other misc stuff */
     maple_state.vbl_cntr = maple_state.dma_cntr = 0;
     maple_state.detect_port_next = 0;
-    maple_state.detect_unit_next = 0;
-    maple_state.detect_wrapped = 0;
+    maple_state.scan_ready_mask = 0;
     maple_state.gun_port = -1;
     maple_state.gun_x = maple_state.gun_y = -1;
 
@@ -150,7 +149,7 @@ void maple_wait_scan(void) {
     maple_device_t  *dev;
 
     /* Wait for it to finish */
-    while(maple_state.detect_wrapped < 1)
+    while(maple_state.scan_ready_mask != 0xf)
         thd_pass();
 
     /* Enumerate everything */
--- a/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
@@ -37,23 +37,8 @@
 /* Initialize Hardware (call after driver inits) */
 static void maple_hw_init(void) {
     maple_driver_t *drv;
-    int p, u;
 
     dbglog(DBG_INFO, "maple: active drivers:\n");
-
-    /* Reset structures */
-    for(p = 0; p < MAPLE_PORT_COUNT; p++) {
-        maple_state.ports[p].port = p;
-
-        for(u = 0; u < MAPLE_UNIT_COUNT; u++) {
-            maple_state.ports[p].units[u].port = p;
-            maple_state.ports[p].units[u].unit = u;
-            maple_state.ports[p].units[u].valid = 0;
-            maple_state.ports[p].units[u].dev_mask = 0;
-            maple_state.ports[p].units[u].frame.queued = 0;
-            maple_state.ports[p].units[u].frame.state = MAPLE_FRAME_VACANT;
-        }
-    }
 
     TAILQ_INIT(&maple_state.frame_queue);
 
@@ -154,9 +139,9 @@ void maple_wait_scan(void) {
 
     for(p = 0; p < MAPLE_PORT_COUNT; p++) {
         for(u = 0; u < MAPLE_UNIT_COUNT; u++) {
-            dev = &maple_state.ports[p].units[u];
+            dev = maple_enum_dev(p, u);
 
-            if(dev->valid) {
+            if(dev) {
                 dbglog(DBG_INFO, "  %c%c: %.30s (%08lx: %s)\n",
                        'A' + p, '0' + u,
                        dev->info.product_name,
--- a/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
@@ -132,10 +132,7 @@ void maple_hw_shutdown(void) {
     /* Free any attached devices */
     for(cnt = 0, p = 0; p < MAPLE_PORT_COUNT; p++) {
         for(u = 0; u < MAPLE_UNIT_COUNT; u++) {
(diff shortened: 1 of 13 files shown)


kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
```diff
@@ -96,7 +80,8 @@ static void maple_hw_init(void) {
 
     /* Hook the necessary interrupts */
     maple_state.vbl_handle = vblank_handler_add(maple_vbl_irq_hnd, &maple_state);
-    asic_evt_set_handler(ASIC_EVT_MAPLE_DMA, maple_dma_irq_hnd, &maple_state);
+    asic_evt_request_threaded_handler(ASIC_EVT_MAPLE_DMA, maple_dma_irq_hnd,
```
QuzarDC, line 83: maple_dma_irq_hnd isn't safe to be run outside of an irq. It's not a "one line change" to set this into threading, even if it does 'just run'.

maple_dma_irq_hnd changes the frame queue, presuming it's inside an irq. Considering you weren't aware that most of this was happening in an irq context, please review the irq handler for concurrency implications beyond the most obvious one I pointed out.

QuzarDC (changes requested):

kernel/arch/dreamcast/include/dc/maple.h
```diff
@@ -277,8 +276,10 @@ typedef struct maple_device {
     maple_frame_t           frame;          /**< \brief One rx/tx frame */
     struct maple_driver     *drv;           /**< \brief Driver which handles this device */
 
-    volatile int            status_valid;   /**< \brief Have we got our first status update? */
-    uint8                   status[1024];   /**< \brief Status buffer (for pollable devices) */
+    uint8                   probe_mask;     /**< \brief Mask of sub-devices left to probe */
+
+    volatile uint8          status_valid;   /**< \brief Have we got our first status update? */
+    uint8                   *status;        /**< \brief Status buffer (for pollable devices) */
```
gyrovorbis, line 282:
```suggestion
    uint8                   status[];        /**< \brief Status buffer (for pollable devices) */
```
Can we use this as just a FAM so we can over-allocate the entire struct as one contiguous segment in one-go instead of needing two `malloc()` calls?

EDIT: Shit, can't really gracefully change the allocation management to work with a FAM in the code review, so if you go with this method, you'll have to handle it yourself. lol.

pcercuei:
It turns out using the unit 0's frame to detect sub-devices only worked the very first time; after that the unit 0's periodic callback would always queue its frame, which means it was always unavailable for the detection loop.

Therefore I also modified the code so that the detection mechanism will always use the dedicated static maple frame.

kernel/arch/dreamcast/include/dc/maple.h
pcercuei: Resolved and force-pushed.

gyrovorbis:
@pcercuei Yo, any way you can rebase this on master and fix those conflicts? The gainz here are so great that I want to start testing this branch out, so we can use it for GTA3 and Sonic Mania...

kernel/arch/dreamcast/hardware/maple/maple_irq.c
```diff
@@ -18,72 +19,84 @@
...
     }
 }
 
+static void vbl_chk_next_subdev(maple_state_t *state, maple_frame_t *frm, int p) {
+    maple_device_t *dev = maple_enum_dev(p, 0);
+    int u;
+
+    if (dev && dev->probe_mask) {
+        u = __builtin_ffs(dev->probe_mask);
+        dev->probe_mask &= ~(1 << (u - 1));
```
QuzarDC, line 68: It may be worthwhile to add a macro for this pattern, as your refactor makes it appear more than just the once.

pcercuei:
Rebased on top of master (63b9b9637).

andressbarajas:
@pcercuei With the 2.1.0 release behind us, older PRs like this one can be prioritized for merging.  I see one unresolved conversation.

kernel/arch/dreamcast/hardware/maple/maple_irq.c
pcercuei: I don't really see what macros you would want to see here.

gyrovorbis:
While I don't think this has been tested exhaustively with every Maple device type, it has definitely gotten some test time in `dev` with GTA3 at least... just noting...

kernel/arch/dreamcast/hardware/maple/maple_irq.c
```diff
@@ -113,31 +128,41 @@ static void vbl_autodet_callback(maple_state_t *state, maple_frame_t *frm) {
...
             dbglog(DBG_KDEBUG, "maple: attach on device %c%c\n",
                    'A' + p, '0' + u);
 #endif
 
+            if (irq_inside_int() && !malloc_irq_safe()) {
+                /* We can't create a device now. Fail silently as the device
+                 * will be re-probed in the next loop of the periodic IRQ. */
+                maple_frame_unlock(frm);
+                return;
+            }
```
QuzarDC, line 156: This check, or something similar, should also be needed to protect the `vbl_chk_disconnect` in the if above it (and where it's called from chk_subdevs. That is what automatically calls `maple_driver_detach` which does the `free()` to the attach's `calloc`.

kernel/arch/dreamcast/hardware/maple/maple_irq.c
QuzarDC: Will resolve this. As described in the discord, I was mistakenly interpreting this as encoding some of the specific logic for maple. In the future it would be great to set up some repo-wide basic access macros to cover stuff (like `BIT()`) but there's no reason to require it here.

kernel/arch/dreamcast/hardware/maple/maple_irq.c
pcercuei: I can move this block at the top of the `vbl_autodet_callback()` function.

QuzarDC: Yea, I guess that makes sense. I was thinking it to be narrowly applied, but the only thing that can be done here if it fires in the middle of a malloc is to update it's function data which is kind of pointless. And either way that check is just to protect against an edge case so hopefully it wouldn't cause significant issue. 

Fortunately that would just mean waiting 4 frames until the autodetect fires again, and is only for device attach/remove so it shouldn't be much of an issue at all.

pcercuei: Done.

QuzarDC (approved):

PR #554, opened 2024-05-09
Added dc-chain Configuration Option for C++'s Timezone Database

gyrovorbis:
Imagine my shock when I found out we have an entire freaking IANA timezone database working on Dreamcast, by default with GCC-14:

Screenshot from 2024-05-08 23-50-51

As much as this is amazing, yeah, obviously it should be a configuration option in the toolchain that gets disabled by default.

- Added new config option, `libstdcxx_tzdb`, for specifying how you want C++'s `std::chrono::tzdb` to be populated: 
    - `0` - None (only UTC and GMT supported)
    - `1` - Automatic (system DB + downloading IANA DB) 
    - `Path` - Custom path to directory containing `tzdata.zi` database file + downloading IANA DB.
- Default option is to disable the DB for C++.


utils/dc-chain/Makefile.default.cfg
```diff
@@ -204,6 +204,19 @@ newlib_c99_formats=1
 # performance.
 #newlib_opt_space=1
 
+#######################
+##### C++ OPTIONS #####
+#######################
```
darcagn, line 209:
```suggestion
###################
### C++ OPTIONS ###
###################
```

darcagn (changes requested):

darcagn (approved):

PR #556, opened 2024-05-09
Update KOS_PORTS path so its prettier in error messages

andressbarajas:
This is done so we see prettier error messages in kos-ports if a port fails to build.


andressbarajas:
This is to fix the lack of realpath on MacOS mentioned here: https://github.com/KallistiOS/kos-ports/pull/66

andressbarajas:
Another route we can take is to have a ```export DC_BASE="/opt/toolchains/dc"``` and use it through the rest of the environ.sh.sample file

doc/environ.sh.sample
```diff
@@ -43,7 +43,7 @@ export KOS_BASE="/opt/toolchains/dc/kos"
 #
 # Specifies the path to the KOS-ports directory
 #
-export KOS_PORTS="${KOS_BASE}/../kos-ports"
+export KOS_PORTS="$/opt/toolchains/dc/kos-ports"
```
darcagn, line 46:
```suggestion
export KOS_PORTS="/opt/toolchains/dc/kos-ports"
```

darcagn (approved):

PR #557, opened 2024-05-09
Update gitignore

andressbarajas:
This PR adds two items to .gitignore:
1.  utils/bincnv/bincnv.dSYM/ : This always shows up when compiling kos for me.
2.  examples/dreamcast/kgl/nehe/nehe26/data/txt2bin : When the makefile in data folder ran.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ kernel/arch/dreamcast/sound/arm/stream.drv
 lib/dreamcast/*.a
 addons/lib/dreamcast/*.a
 utils/bincnv/bincnv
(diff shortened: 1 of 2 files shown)


gyrovorbis:
I kind of feel like, at this point in time, we should just remove the `-g` flag for producing debug symbols from the `Makefile` for this tool... Not sure why this single one is randomly being built with debug symbols being produced, but none of the others are...

andressbarajas:
@gyrovorbis Great idea.  Done

gyrovorbis (approved):

PR #561, opened 2024-05-09
Bumped KOS's Default C and C++ Versions to 11

gyrovorbis:
Yes, this is quite an opinionated change, but enough people have brought it up and have talked about wanting to do it, that I thought now might be a good time to get it in the open...

We're hoping to move KOS to `C11`, because we keep running into things that we're missing from `C99` and would like to leverage, such as:
- `_Alignas()`
- `_Alignof()`
- `_Noreturn`
- `_Generic()` (okay, maybe this one's just me ;) )
- anonymous structures/unions
- `_Static_assert()` (we keep reinventing this)

There are a bunch of arguments that have been made that even the Linux kernel is `C11`, but I'm wondering if, assuming people are okay with this, if we should just go to `C17`, since it's been supported on all compilers, even the ones we deprecated, and it's seen as just a "bugfix" revision to fix up what `C11` added? 

Also... Honestly what is this C++ version even for? It's  not like we have C++ code in KOS or anything, and even GCC9 was defaulting to C++17... so if we're okay with bumping it up even higher, because it's mostly serving as a default for *external code*, I think we should go to 17. 

Yes, I verified that everything builds and works great with C11, at least.

### DIFF
--- a/environ_base.sh
+++ b/environ_base.sh
@@ -41,8 +41,8 @@ export KOS_CPPFLAGS="${KOS_CPPFLAGS} ${KOS_INC_PATHS_CPP} -fno-operator-names"
 # Which standards modes we want to compile for
 # Note that this only covers KOS itself, not necessarily anything else compiled
 # with kos-cc or kos-c++.
-export KOS_CSTD="-std=c99"
-export KOS_CPPSTD="-std=gnu++98"
+export KOS_CSTD="-std=c11"
+export KOS_CPPSTD="-std=gnu++11"
 
 export KOS_GCCVER="`kos-cc -dumpversion`"
 
--- a/environ_base.sh
+++ b/environ_base.sh
@@ -41,7 +41,7 @@ export KOS_CPPFLAGS="${KOS_CPPFLAGS} ${KOS_INC_PATHS_CPP} -fno-operator-names"
 # Which standards modes we want to compile for
 # Note that this only covers KOS itself, not necessarily anything else compiled
 # with kos-cc or kos-c++.
-export KOS_CSTD="-std=c17"
+export KOS_CSTD="-std=gnu17"
 export KOS_CPPSTD="-std=gnu++17"
 
 export KOS_GCCVER="`kos-cc -dumpversion`"
--- a/environ_base.sh
+++ b/environ_base.sh
@@ -41,8 +41,8 @@ export KOS_CPPFLAGS="${KOS_CPPFLAGS} ${KOS_INC_PATHS_CPP} -fno-operator-names"
 # Which standards modes we want to compile for
 # Note that this only covers KOS itself, not necessarily anything else compiled
 # with kos-cc or kos-c++.
-export KOS_CSTD="-std=c11"
-export KOS_CPPSTD="-std=gnu++11"
+export KOS_CSTD="-std=c17"
+export KOS_CPPSTD="-std=gnu++17"
 
 export KOS_GCCVER="`kos-cc -dumpversion`"
 


gyrovorbis:
Actually... why aren't we building the kernel with GNU extensions enabled? ...we're already using them, it's just that GCC is sloppily letting us get away with it, cuz we aren't using the `-pedantic` flag or whatever...

Just look at how many statement expressions are around the codebase for like the assembly intrinsics.

Also we have to jump through hoops and shenanigans to manually `#define POSIX_SOUCE_VERSION` or whatever in certain KOS files where we want to use POSIX functions. Also, I'm seeing, upon enabling them, that we might be duplicating a few symbols exported or declared by Newlib in KOS without knowing it...

If we're this nice pseudo OS that at this point goes out of it's way to manually implement certain POSIX calls, like `clock_gettime()`, `sysconf()`, etc, shouldn't we default build to having GNU extensions and POSIX crap enabled?

environ_base.sh
```diff
@@ -41,8 +41,8 @@ export KOS_CPPFLAGS="${KOS_CPPFLAGS} ${KOS_INC_PATHS_CPP} -fno-operator-names"
 # Which standards modes we want to compile for
 # Note that this only covers KOS itself, not necessarily anything else compiled
 # with kos-cc or kos-c++.
-export KOS_CSTD="-std=c99"
-export KOS_CPPSTD="-std=gnu++98"
+export KOS_CSTD="-std=c11"
+export KOS_CPPSTD="-std=gnu++11"
```
QuzarDC, line 45:
```suggestion
export KOS_CSTD="-std=c17"
export KOS_CPPSTD="-std=gnu++17"
```

After reading through the Defect Reports for 17 vs 11 I'd agree with your assessment and if we're going to move to 11 should move to 17 instead.

Tested this against the experimental toolchain as well as our legacy one. Found some other issues with newer code in the legacy toolchain (#579 and https://github.com/KallistiOS/kos-ports/pull/67 ) but none caused by this.

QuzarDC (dismissed):
I'd sign off on this, especially since it would mean that we could move away from a number of compiler-specific attributes in favor of the language features that replace them, but would certainly want @ljsebald to be able to give feedback on this as it is an opinionated change (and I'm sure there are some repercussions that may need considering).

darcagn:
I'm also very much in favor of defaulting to C11/C17, but just for consideration, since we agreed not to commit to moving away from multi-arch just yet, are there any other platforms we'd have in mind for which C11 support would be limited?

gyrovorbis:
> I'm also very much in favor of defaulting to C11/C17, but just for consideration, since we agreed not to commit to moving away from multi-arch just yet, are there any other platforms we'd have in mind for which C11 support would be limited?

I know for sure the PSP/PS2 guys just got up GCC14.1.0, along with the libDragon N64 guys. Then the Saturn scene is largely on 13.2.0, building their own stuff. I don't know for sure about the GBA, but I would assume they can easily build the latest GCC version, given how common that ARM target is.

Oh, then the PSVita scene was on GCC 10.x last I checked... Don't know of anyone who wouldn't have C17 support. Looks like it was added in GCC8.1.

gyrovorbis:
What do you guys think about `gnu17` vs `c17`? I believe we should go for the former, as it's truly what we're using internally, and we definitely want people to have access to all of the POSIX stuff externally.

QuzarDC:
> > I'm also very much in favor of defaulting to C11/C17, but just for consideration, since we agreed not to commit to moving away from multi-arch just yet, are there any other platforms we'd have in mind for which C11 support would be limited?
> 
> I know for sure the PSP/PS2 guys just got up GCC14.1.0, along with the libDragon N64 guys. Then the Saturn scene is largely on 13.2.0, building their own stuff. I don't know for sure about the GBA, but I would assume they can easily build the latest GCC version, given how common that ARM target is.
> 
> Oh, then the PSVita scene was on GCC 10.x last I checked... Don't know of anyone who wouldn't have C17 support. Looks like it was added in GCC8.1.

Just checked devkitpro for all the arm stuff for gba. They've got gcc14 so they're good :) 

Thanks @darcagn for keeping in mind the multi-arch. I had just figured it'd be a problem for implementing them, but it's great to know that shouldn't be a problem at all.

QuzarDC:
> What do you guys think about `gnu17` vs `c17`? I believe we should go for the former, as it's truly what we're using internally, and we definitely want people to have access to all of the POSIX stuff externally.

It seems we were simply on c98 vs gnu98 due to @ljsebald  narrowing it down here: https://github.com/KallistiOS/KallistiOS/commit/da33c868b249bedd99fffb425fade3a25579e42d . I think the notion was that it had been expanded to the gnu version in order to support some piece of code and once that was adjusted to more standard c it was clawed back.

I don't think there's an explicit reason to do so other than merely to default to only use the narrowest spec unless otherwise needed. And that was likely due to the older languishing of the toolchains and the desire to support even back to gcc2/3 in a limited way.

As a further note onto c11 vs c17, according to https://gcc.gnu.org/onlinedocs/gcc/Standards.html all the fixes made for the c17 bugfixes are actually applied in c11 (as they were all merely implementation clarifications). The only difference is that if we specify c17 we get a higher `__STDC_VERSION__` 

Additionally, if no standard is specified gnu17/gnu++17 are the defaults since gcc11 (so we should still specify them to cover gcc9.3.0 ).

darcagn:
Well, given that all of the arches we'd be likely to target in the future have no problem with gnu17, given all of the compiler versions we'd target would have no problem with gnu17, given the Linux kernel is on gnu11 and has infinitely more legacy baggage to consider than we do, I can't think of any reason we shouldn't move forward with this. I'm personally a big fan of the improvements in C11/C23 so I would really like to see us at least bumped up to C11/C17.

QuzarDC (approved):

environ_base.sh
```diff
@@ -41,8 +41,8 @@ export KOS_CPPFLAGS="${KOS_CPPFLAGS} ${KOS_INC_PATHS_CPP} -fno-operator-names"
 # Which standards modes we want to compile for
 # Note that this only covers KOS itself, not necessarily anything else compiled
 # with kos-cc or kos-c++.
-export KOS_CSTD="-std=c99"
-export KOS_CPPSTD="-std=gnu++98"
+export KOS_CSTD="-std=c17"
```
darcagn, line 44:
```suggestion
export KOS_CSTD="-std=gnu17"
```

darcagn (changes requested):

darcagn (approved):

QuzarDC:
> I'd sign off on this, especially since it would mean that we could move away from a number of compiler-specific attributes in favor of the language features that replace them, but would certainly want @ljsebald to be able to give feedback on this as it is an opinionated change (and I'm sure there are some repercussions that may need considering).

I'm sticking with this. I approve as well but would still prefer to have Blue's imprimatur. I think it wouldn't be too much a burden to have this sit for a day or two now that we got good convo on it.

QuzarDC (approved):

PR #562, opened 2024-05-09
Fix mke2fs example with Raylib Installed

gyrovorbis:
In a cruel twist of fate, if you have Raylib installed into KOS-ports, the mke2fs example will no longer build... Why? Because the include for "utils.h" will pick up the one in kos-ports from Raylib before picking up the one in the ext2fs addon directory.

- Swapped the order for the Makefile's extra include directory flag to ensure that it comes before kos-ports

### DIFF
--- a/examples/dreamcast/sd/mke2fs/Makefile
+++ b/examples/dreamcast/sd/mke2fs/Makefile
@@ -8,7 +8,7 @@ OBJS = mke2fs.o
 
 # We need the private headers from libkosext2fs, since we're not using the
 # facilities of fs_ext2 here.
-KOS_CFLAGS += -I$(KOS_BASE)/addons/libkosext2fs -Werror -W -std=gnu99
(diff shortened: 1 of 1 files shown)


QuzarDC (approved):
I don't see any issue with this, *but* this is perhaps something we should address as well in libkosext2fs to not have such a generic header as well as with Raylib to do the same.

As kos-ports grows I can see this becoming a more common kind of conflict that will not always be so easily avoidable.

gyrovorbis:
> I don't see any issue with this, _but_ this is perhaps something we should address as well in libkosext2fs to not have such a generic header as well as with Raylib to do the same.
> 
> As kos-ports grows I can see this becoming a more common kind of conflict that will not always be so easily avoidable.

I completely agree with this.

ljsebald:
I'm confused as to why this was an issue. Is raylib installing things directly into `include` without putting them in its own directory like the other ports do?

PR #503, opened 2024-04-09
Keyboard Cleanup + Event-Driven API

gyrovorbis:
Massive refactor of the keyboard Driver and API with a lot of new features. Everything that used to work still works and has been tested. The existing methods for processing input (ie: key press queue) are all there and work fine (and are now configurable). The ONLY thing that I did change is the names of the struct members in `kbd_state_t`, because I really feel like it had to be done to make the polling API pristine...

- Cleaned up C typing with typedefs, enums, unionized bitfields, and booleans
- Added an event-driven callback mechanism for registering a callback to be invoked whenever a key changes states (similar to how Qt, GTk, and high-level widgeting frameworks process key events)
- Made `kbd_state_t` a whole lot more user-friendly, as it's a public structure and is supposed to be reached into when polling for key states directly.
- Moved a bunch of crap that was cluttering the header and `kbd_state_t` from the public API to the private source file, where it's hidden from the user. This includes an "extended" `kbd_private_state_t` structure which contains all of the public members PLUS hidden private members only accessible from within the driver.
- `kbd_state_t`'s "key matrix" has been modified to store both the current frame and previous frame's key states, which allows for the driver to check for state changes and emit events accordingly. This added zero extra space overhead, since we were storing key states within a byte anyway, and I'm only using two bits of it.
- Added utility function for converting raw key to ASCII values without needing to go through a device
- Made repeated key press interval configurable 
- Adjusted libDream's keyboard example to show off event-driven handler mechanism
- Redid the entirety of the Doxygen with revamped topics, better organization, and example code snippets

PR #565, opened 2024-05-10
Added Versioning to KOS

gyrovorbis:
OKAY, this is something I can't freaking believe we haven't done already that absolutely needs to be done. 

I'm certain we want the following:

`#define KOS_VERSION` which is a bit-packed `uint32_t` of `MAJOR`, `MINOR`, `PATCH` components which can be used directly at compile-time with regular comparison operators to gate things within a preprocessor directive:

```
#if KOS_VERSION >= KOS_VERSION_MAKE(2, 0, 1)
    /* Do some shit only supported by 2.0.1 and above here. */
#endif
```

I'm also positive that we want to have the *runtime* equivalent of this, because you could be compiling with and including a header that has one version, yet be linking against another:
```
if(kos_version() >= KOS_VERSION_MAKE(2, 0, 1)) {
    /* RUNTIME CHECK */
 }
 ```
 
 Finally, I included `KOS_VERSION_STRING` for an easy way to print the compile-time version plus `kos_version_string()` for the runtime equivalent:
 ```c
printf("KOS Compile-Time Version: %s\n", KOS_VERSION_STRING);
printf("KOS Run-Time Version: %s\n", kos_version_string());
```

Now what I'm uncertain of:
- Where should this header/source file should live?
- Where should the top-level version info be stored? Should it be in this header itself, as it is now, or should it propagate to fill in these fields from an outer script that processes some top-level `version.txt`?
- How do we propagate this into the KOS startup banner?


kernel/version.c
```diff
@@ -0,0 +1,9 @@
+#include <kos/version.h>
+
+kos_version_t kos_version(void) {
+    return KOS_VERSION;
+}
+
+const char* kos_version_string(void) {
+    return KOS_VERSION_STRING;
+}
```
andressbarajas, line 9:
```suggestion
}

```

include/kos/version.h
```diff
@@ -0,0 +1,42 @@
...
+#include <stdint.h>
+
+typedef uint32_t kos_version_t;
+
+kos_version_t kos_version(void);
+const char* kos_version_string(void);
+
+__END_DECLS
+
+#endif /* __KOS_VERSION_H */
```
andressbarajas, line 42:
```suggestion
#endif /* __KOS_VERSION_H */

```

kernel/version.c
```diff
@@ -0,0 +1,9 @@
+#include <kos/version.h>
+
+kos_version_t kos_version(void) {
+    return KOS_VERSION;
+}
+
+const char* kos_version_string(void) {
```
andressbarajas, line 7:
```suggestion
const char *kos_version_string(void) {
```

include/kos/version.h
```diff
@@ -0,0 +1,42 @@
...
+
+#include <kos/cdefs.h>
+__BEGIN_DECLS
+
+#include <stdint.h>
+
+typedef uint32_t kos_version_t;
+
+kos_version_t kos_version(void);
+const char* kos_version_string(void);
```
andressbarajas, line 38:
```suggestion
const char *kos_version_string(void);
```

kernel/Makefile
```diff
@@ -4,14 +4,14 @@
 # (c)2001 Megan Potter
```
andressbarajas, line 4:
```suggestion
# Copyright (C) 2001 Megan Potter
```

include/kos/version.h
```diff
@@ -0,0 +1,42 @@
+#ifndef __KOS_VERSION_H
+#define __KOS_VERSION_H
+
+#define KOS_VERSION_MAJOR   2
+#define KOS_VERSION_MINOR   0
```
andressbarajas, line 5:
```suggestion
#define KOS_VERSION_MINOR   1
```

maishuji:
Hello, I think it would be nice to add tags for versions in github
It would be useful for example for building docker images of kos, targeting a stable/fixed version of kos.
https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository

gyrovorbis:
> Hello, I think it would be nice to add tags for versions in github It would be useful for example for building docker images of kos, targeting a stable/fixed version of kos. https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository

Actually, this is the plan moving forward, once we're done with the v2.1.0 release that we're all working towards. It'll get a tag then a binary release.

Although I guess nothing is stopping us from going back and doing the same for the previous v2.0.0 version release, which I believe did get that treatment on SourceForge and elsewhere from BlueCrab. Not positive, though, that was before many of our time contributing to KOS.

pcercuei:
> I'm also positive that we want to have the _runtime_ equivalent of this, because you could be compiling with and including a header that has one version, yet be linking against another:

In which universe would you do that?

gyrovorbis:
> > I'm also positive that we want to have the _runtime_ equivalent of this, because you could be compiling with and including a header that has one version, yet be linking against another:
> 
> In which universe would you do that?

Every mainstream large library I've ever worked with has had that in C or C++, Qt, SDL, OpenGL, etc. How else would you be able to verify that you compiled against and linked against the same versions? Especially if you're doing dynamic linking against such a library? 

The user could very easily have pulled down KOS but not rebuilt it, hence the compile-time and link-time versions would not match when they queried from their app.

QuzarDC:
Under what circumstances would you want to have your codebase support older versions of KOS explicitly? For other libraries the intent is that the end user may have precompiled libraries on their local system and dynamically linked in, or even that different underlying platforms the library is supported on are at different versions/feature levels. But since we always statically link KOS it doesn't seem to apply. This seems like the kind of thing that may seem like an issue due to working on the KOS codebase that would never find use for any actual *user* of the codebase rather than developer of the codebase. Maybe I'm just not seeing it though.

The one use that I do see though is as an internal runtime check to help validate that the headers and binaries are coherent to prevent issues with users wondering why things are broken when they forgot to rebuild KOS.

gyrovorbis:
> Under what circumstances would you want to have your codebase support older versions of KOS explicitly?
Literally right here is an example of it:
https://github.com/Kazade/GLdc/blob/ab0ea41876f04c3d661568a1712f5ce2fc594c25/GL/platforms/sh4.c#L35

Only look at the crazy bullshit Kazade has to go through to determine the KOS version, because we have failed to gracefully version it like a normal C library.

QuzarDC:
> > Under what circumstances would you want to have your codebase support older versions of KOS explicitly?
> 
> Literally right here is an example of it:
> 
> https://github.com/Kazade/GLdc/blob/ab0ea41876f04c3d661568a1712f5ce2fc594c25/GL/platforms/sh4.c#L35
> 
> Only look at the crazy bullshit @Kazade has to go through to determine the KOS version, because we have failed to gracefully version it like a normal C library.
> 
> This kind of thing can be extremely common in `kos-ports` or any library that utilizes KOS internally.

So, that doesn't answer the question you quoted. That is an example of a codebase supporting older versions of KOS explicitly. What I don't get is *why*? What circumstances exist that would cause you to want to continuously support both behaviors? Is there a contingent out there that want to be constantly updating their GLdc but never updating KOS?

I think the core issue is more just not having versioned releases at all. Since *everything* is KOS 2.0.0 it's not possible to easily say 'use the latest KOS' because that may not be the user's chosen known-stable version. Once we do have regular stable cut versions (that this presumes and requires) there shouldn't be the unknown barrier to be keeping updated.

Of course, we should still have such a mechanism because with regular releases there should be mechanisms to see that release version. But I don't know that we should encourage, especially with regular stable releases, people staying on older versions. Rather if there's some reason to do so we should be working to address it.

pcercuei:
> The user could very easily have pulled down KOS but not rebuilt it, hence the compile-time and run-time versions would not match when they queried from their app.

If that's the case I think it would be better to have a mechanism which enforces that the KOS version (down to the git hash) set in the header always match the one of the library.

Kazade:
> So, that doesn't answer the question you quoted. That is an example of a codebase supporting older versions of KOS explicitly. What I don't get is _why_? What circumstances exist that would cause you to want to continuously support both behaviors? Is there a contingent out there that want to be constantly updating their GLdc but never updating KOS?

So it is certainly the case that GLdc needs to support different versions of KOS. Updating KOS is a wider-reaching change than updating GLdc. For example, I'm not in a position (yet) to risk updating KOS in Driving Strikers because I'm concerned it might destabilise the networking code or whatever, so I'm holding back updating whereas others will be using up-to-date GLdc on up-to-date KOS. Having a version check like this will help GLdc handle both without having to do weird things as linked above.

However! I think there is a bigger issue here (and essentially what you're saying Quazer) in that **we introduced an incompatible change without having a KOS release**. For us to make breaking changes in KOS we need both regular tagged releases *AND* a version mechanism like this one. How that versioning works (and whether we include a commit hash or count) is one for BlueCrab. 

My personal opinion is that KOS should move away from mega releases every 5 years, and have a more regular release cycle - but that's like, just my opinion man. :)

As for the actual change here - it looks good to me, I like it, and I'm not too fussed about the whole commit-hash-versioning (because it inevitably ends up causing a dependency on git to generate it).

QuzarDC:
> > So, that doesn't answer the question you quoted. That is an example of a codebase supporting older versions of KOS explicitly. What I don't get is _why_? What circumstances exist that would cause you to want to continuously support both behaviors? Is there a contingent out there that want to be constantly updating their GLdc but never updating KOS?
> 
> So it is certainly the case that GLdc needs to support different versions of KOS. Updating KOS is a wider-reaching change than updating GLdc. For example, I'm not in a position (yet) to risk updating KOS in Driving Strikers because I'm concerned it might destabilise the networking code or whatever, so I'm holding back updating whereas others will be using up-to-date GLdc on up-to-date KOS. Having a version check like this will help GLdc handle both without having to do weird things as linked above.
> 
Ok, that makes sense. Wanting to not have to force all users to stay on the version you're using by addressing a breaking incompatibility. Thanks for explaining as I was just at a loss.

> However! I think there is a bigger issue here (and essentially what you're saying Quzar) in that **we introduced an incompatible change without having a KOS release**. For us to make breaking changes in KOS we need both regular tagged releases _AND_ a version mechanism like this one. How that versioning works (and whether we include a commit hash or count) is one for BlueCrab.
> 
> My personal opinion is that KOS should move away from mega releases every 5 years, and have a more regular release cycle - but that's like, just my opinion man. :)
> 
Yea, I've been pushing for 2.1.0, and it does look like we'll finally get it in the next month or so. After that ... constipation is out of the way, the plan is to move into something like a 6 month cycle as we now have the community support to be having frequent enhancements and fixes as well as testing. With that it should be much less of an issue to be able to have a user just pull a specific version maybe even with precompiled bins and only need to move into dailys if they want to be involved or find a specific issue.

QuzarDC (commented):
This doesn't really need much documentation at all, but it would be useful to show the use pattern you expect. From it as it stands it seems like it would be:

`if(kos_version() <= KOS_VERSION_MAKE(2,1,0))` / `#if (KOS_VERSION <= KOS_VERSION_MAKE(2,1,0))`

Both seem like mouthfulls though. Maybe something like `if(KOS_MIN(2,1,0))` / `#if KOS_MIN(2,1,0)` or something more compact would make a lot of sense?

Otherwise looks great and will be integral to being able to mitigate issues with versions in the future. Additionally it'll smooth over the maintenance of kos-ports. Now, for instance, https://github.com/KallistiOS/kos-ports/pull/70 could use defines so that there's no point between the current version and the release of the version with #503 where it is broken :)

Are there any other places in the current codebase where you think this might be applicable? I'm not entirely sure what the mechanism was for setting those top-of-file ##version## tags, but it would be good to make sure there isn't already some partial implementation of this around (that should likely just get taken out).

include/kos/version.h
```diff
@@ -0,0 +1,42 @@
...
+
+#define KOS_VERSION_STRING \
+    KOS_VERSION_MAKE_STRING(KOS_VERSION_MAJOR, \
+                            KOS_VERSION_MINOR, \
+                            KOS_VERSION_PATCH)
+
+#define KOS_VERSION_MAKE(major, minor, patch) \
+    (((kos_version_t)(major) << 24) | \
+     ((kos_version_t)(minor) << 8)  | \
+     ((kos_version_t)(patch)))
```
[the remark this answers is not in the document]
Kazade: I'm surprised @gyrovorbis didn't make a struct of bitfields for this ;)

gyrovorbis:
OKAY OKAY. I went a little hard on this API and spent like well over an hour documenting the living it out of it into its own Doxygen subsystem/module, but I think based on the feedback, it's now pristine (unless you disagree with the naming convention). 

The Doxygen screenshots should say it all for me:
image
image
image
image
image

NOTES:
- I have examples on exactly how to do both compile-time preprocessor and run-time conditional branch checks based on the current KOS version.
-  I have generalized the underlying macros for the version encodings and version checks into something that can be reused by KOS-ports or user applications to version their own stuff with the same scheme... because why not? Other than it took me a damn year to document it all. lol.
- Right now the actual KOS version is hardcoded in `version.h`:
```
#define KOS_VERSION_MAJOR   2   /**< KOS's current major revision number. */
#define KOS_VERSION_MINOR   0   /**< KOS's current minor revision number. */
#define KOS_VERSION_PATCH   1   /**< KOS's current patch revision number. */
```
I'm not so sure that I see any way around having the version number in this file either... as its not like a header file can include it from elsewhere, so any external script that does something with the version should probably be written to pull it from this header file as the one source of truth.
- I propose we update the KOS banner that gets printed when you start KOS to the following:
```
KallistiOS v2.0.1 
   Git Revision 1433-g12501132-dirty
      Sat Jun  8 04:27:40 AM CDT 2024
  falco@slutserv:/opt/toolchains/dc/kos
  sh-elf-gcc (GCC) 14.1.0
  ```
  This way we have the actual public, user-facing version as the first thing anyone sees, *along with* the commit ID information we have now... but dear god, do not make me have to be the one to make the `banner.sh` script somehow pull that information from `version.h`... Oh, god...

File: include/kos/version.h [hunk 8]
```diff
@@ -0,0 +1,424 @@
...
+
+    \param  major    Major version component.
+    \param  minor    Minor version component.
+    \param  patch    Patch version component.
+
+    \retval true     KOS's version is the same or higher.
+    \retval false    KOS's version is lower.
+
+*/
+#define KOS_VERSION_MIN(major, minor, patch) \
```
Kazade, line 142: Just a suggestion, but perhaps: KOS_VERSION_GEQUAL(...) / LEQUAL / LESS / EQUAL ?

include/kos/version.h
```diff
@@ -0,0 +1,424 @@
...
+
+    Used to create a version identifier at compile-time from its components.
+
+    \param major    Major version component.
+    \param minor    Minor version component.
+    \param patch    Patch version component.
+
+    \returns        Packed version identifier.
+*/
+#define KOS_VERSION_MAKE(major, minor, patch) \
```
Kazade, line 212: Is _MAKE necessary? Maybe KOS_VERSION(x, y, z) is enough ?

Kazade: Actually, ignore this, it conflicts with kos_version()

Kazade:
Just left one minor comment, it occurs to me that normally you want to do one of the following:

 - Check that the KOS version is greater than or equal to a version that introduced a feature
 - Check that the KOS version is less than a version that introduced a breaking change
 - Check that a KOS version is exact

MIN/MAX doesn't feel right here, because for the second one you'll need to write the previous version than the version you're writing against. What I mean is... say KOS 3.0 introduces a breaking change, it's nicer to write KOS_VERSION_LESS_THAN(3, 0, 0) *not* KOS_VERSION_MAX(2, 9, 9) 

Obviously opinions may vary :D

gyrovorbis:
> Just left one minor comment, it occurs to me that normally you want to do one of the following:
> 
> * Check that the KOS version is greater than or equal to a version that introduced a feature
> * Check that the KOS version is less than a version that introduced a breaking change
> * Check that a KOS version is exact
> 
> MIN/MAX doesn't feel right here, because for the second one you'll need to write the previous version than the version you're writing against. What I mean is... say KOS 3.0 introduces a breaking change, it's nicer to write KOS_VERSION_LESS_THAN(3, 0, 0) _not_ KOS_VERSION_MAX(2, 9, 9)
> 
> Obviously opinions may vary :D

You're exactly right... I was actually thinking that myself as I wrote the `KOS_MAX()` macro, that it's not really ideal that it's the version *before* the break, and then your suggestions are actually more consistent in terms of convention with one called `KOS_EQUAL()`, as it's the comparison name... On it!

QuzarDC:
> Just left one minor comment, it occurs to me that normally you want to do one of the following:
> 
>     * Check that the KOS version is greater than or equal to a version that introduced a feature
> 
>     * Check that the KOS version is less than a version that introduced a breaking change
> 
>     * Check that a KOS version is exact
> 
> 
> MIN/MAX doesn't feel right here, because for the second one you'll need to write the previous version than the version you're writing against. What I mean is... say KOS 3.0 introduces a breaking change, it's nicer to write KOS_VERSION_LESS_THAN(3, 0, 0) _not_ KOS_VERSION_MAX(2, 9, 9)
> 
> Obviously opinions may vary :D

`KOS_MIN(3,0,0)` is just an inverted `KOS_VERSION_LESS_THAN(3,0,0)`. It's 'this is the minimum version required' or GTE. My notion was that for 1, you would use that and for 2 you'd just do `!(KOS_MIN(3,0,0))`. The MAX, I don't know what the use case would be for it.

gyrovorbis:
OKAY. So I took everyone's feedback into account... We like what @Kazade noted about typically wanting a check for being LESS THAN a current version rather than having a maximum version... but then @QuzarDC also contrived a situation where a MAX could still be useful... So I went with both... lol. 

What I did, though, was for the sake of brevity and clarity in the naming scheme of these version checkers was to move away from naming them after the comparison operator names and instead gave them a naming scheme that I think is more intuitive and is also much less verbose:
```c
/* COMPILE-TIME VERSION CHECKERS */
#define KOS_VERSION_ABOVE(major, minor, patch) // >
#define KOS_VERSION_MIN(major, minor, patch) // >=
#define KOS_VERSION_IS(major, minor, patch) // ==
#define KOS_VERSION_MAX(major, minor, patch) // <=
#define KOS_VERSION_BELOW(major, minor, patch) // <

/* RUN-TIME VERSION CHECKERS */
bool kos_version_above(uint8_t major, uint16_t minor, uint8_t patch); // >
bool kos_version_min(uint8_t major, uint16_t minor, uint8_t patch); // >=
bool kos_version_is(uint8_t major, uint16_t minor, uint8_t patch); // ==
bool kos_version_max(uint8_t major, uint16_t minor, uint8_t patch); // <=
bool kos_version_below(uint8_t major, uint16_t minor, uint8_t patch); // <

**Review by Kazade — APPROVED** [review]

**Comment by gyrovorbis** [comment]
OKAY OKAY OKAY. I got off my lazy ass and rigged up the shell scripting stuff! It's all done now! EVERYTHING in terms of KOS versioning comes from `kos/include/kos/version.h` as the single source of truth for the overall version. These lines in particular:
```
#define KOS_VERSION_MAJOR   2   /**< KOS's current major revision number. */
#define KOS_VERSION_MINOR   1   /**< KOS's current minor revision number. */
#define KOS_VERSION_PATCH   0   /**< KOS's current patch revision number. */
```
Those are the numbers that get reflected across the entire environment now as the current version of KOS. You want to bump up the current version? You change those `#define`s. So then:

- KOS_VERSION_MAJOR, KOS_VERSION_MINOR, KOS_VERSION_PATCH, and KOS_VERSION are now all added to everyone's environment as environment variables that get pulled directly from this header in `environ.base.sh`.
- Updated `make_banner.sh` to add a release version header to the banner:
```
KallistiOS v2.1.0
  Git revision: v2.0.0-1441-gd5b5c4c2
  Sat Jun  8 04:04:07 PM CDT 2024
  falco@slutserv:/opt/toolchains/dc/kos
  sh-elf-gcc (GCC) 14.1.0
```
- Updated `version.sh` utility script that replaces the `##version##` string in a single file with the given version to now take the version from the environment, which itself is taken from `version.h`, so they always stay in-sync.

QuzarDC:
Can/will approve, but hope to hear from @Kazade . It looks like this now can satisfy any comparison that's desired against a version number. I'd want anyone else to be able to take a quick look at the scripts as I'm just garbage with shell scripting so can't be confident in my review of them.

One extra little thing that might be an exciting bit of change if you'd be ok with it @gyrovorbis . Perhaps leave all 3 as uint8_ts and have a reserved extra char. We shouldn't hold anything back for it, but that might be a place to fold in extra sorts of comparisons/flags in the future (I'm thinking the possibility of a bitfield holding git dirty/clean, fork/official, stable/daily, etc which could make things easier to validate via their banner or perhaps issue notification).

gyrovorbis:
Sorry for taking so long to respond to this... I'm back at it... 

> One extra little thing that might be an exciting bit of change if you'd be ok with it @gyrovorbis . Perhaps leave all 3 as uint8_ts and have a reserved extra char. We shouldn't hold anything back for it, but that might be a place to fold in extra sorts of comparisons/flags in the future (I'm thinking the possibility of a bitfield holding git dirty/clean, fork/official, stable/daily, etc which could make things easier to validate via their banner or perhaps issue notification).

OoOOo okay, yeah, that totally works for me. Will make the change for it.

gyrovorbis:
@QuzarDC Alrighty, it is done. I didn't really mention it or add any code to encode into that reserved 1 byte region, but I ensured it is reserved and that all ranges for the `patch`, `minor`, and `'major` components mention they're specifically only 1 byte each. 

Also went ahead and made damn sure to add a few warnings to the docs that you are not supposed to directly compare encoded version values together, just in case. lol.

QuzarDC (approved):

PR #566, opened 2024-05-10
Avoid conditionally-compiled code & fix NAOMI build

pcercuei:
Conditionally-compiled code (wrapped by #ifdef/#ifndef guards) is
problematic; because it is very difficult to compile-test all possible
combinations of the various flags verified by the guard macros.

Instead, evaluate compile-time constants, which will cause the compiler
to compile the exact same code independently of the configuration, which
means if it compiles for one configuration, it will compile for all
configurations. The compiler will then be smart enough to
garbage-collect dead code, dropping everything not included in the
current configuration.

### DIFF
--- a/include/kos/platform.h
+++ b/include/kos/platform.h
@@ -1,0 +1,28 @@
+/* KallistiOS ##version##
+
+   include/kos/platform.h
+   Copyright (C) 2024 Paul Cercueil
+
+*/
+
+/** \file   kos/platform.h
+    \brief  Platform detection macros.
+    \author Paul Cercueil
+*/
+
+#ifndef __KOS_PLATFORM_H
+#define __KOS_PLATFORM_H
+
+#ifdef __NAOMI__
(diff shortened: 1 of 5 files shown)


include/kos/platform.h
```diff
@@ -0,0 +1,28 @@
+/* KallistiOS ##version##
+
+   include/kos/once.h
```
gyrovorbis, line 3:
```suggestion
   include/kos/platform.h
```

gyrovorbis (commented):
I knew you were right, but I had to convince myself in Compiler Explorer anyway, and yeah, you're right. I can go ahead and merge this, because it's a trivial change with how we're handing the mechanism I'm familiar with that's breaking a build. 

As soon as that one tiny change is fixed in the header comment, I"ll merge.

pcercuei:
> As soon as that one tiny change is fixed in the header comment, I"ll merge.

Done.

gyrovorbis (approved):

PR #569, opened 2024-05-11
Linking to libm by Default in environ_base.sh.

gyrovorbis:
Because I don't think it's a good idea not to. It's an annoying little nuance of KOS-specific libc/Newlib/GCC toolchain configuration that it's a separate library, it leaks its way out to having to be added manually and intrusively to CMake, and I've just never had to care how or where the C standard library's math got implemented in a build system...

PLUS you have to add the lib in a particular order due to circular dependendencies, which is a PITA, plus it'll get garbage collected if you don't use it, etc, etc.


andressbarajas:
I remember a lot of examples relying on linking to it too.  By adding it here, will that cause interference if it is linked again?

gyrovorbis:
> I remember a lot of examples relying on linking to libm too. By adding it here, will that cause interference if it is linked again?

Nope. It's silently ignored... but I guess as part of this PR, I can go back and remove the duplicates?

gyrovorbis:
k, I'll stop being lazy... I removed the explicit `-lm` linker line from every example and confirmed that everything compiles fine...

...which, looking at the sheer amount of examples which need that library to do anything with, I definitely still firmly believe should be automatic... It's a transient dependency of our `libc` that just happens to be implemented as a separate static library, although it's not on many systems and environments... and expecting the user to resolve transient dependencies of internal dependencies sounds like a bad idea to me.

I'm also seeing the meson build guys wanting to do the same thing, so that meson doesn't require `-lm` only on certain build environments as we have to in CMake currently.

QuzarDC (approved):

PR #570, opened 2024-05-11
Fix Cache functions

andressbarajas:
This fixes the following:

1. All the flush, inval, purge functions did not check for value 0 entered for count.  This would make the function execute one iteration of loop when we are not supposed to.  
2. For count values < 32 the dcache flush/purge functions would loop infinitely.


gyrovorbis:
Pinging @DC-SWAT to weigh in, cuz this is also his knowledge domain.

kernel/arch/dreamcast/kernel/cache.s
```diff
@@ -27,6 +27,11 @@
 ! r5 is count
     .align 2
 _icache_flush_range:
+    ! Exit early if no blocks to flush
+    mov      #0, r0
+    cmp/eq   r0, r5
+    bt       .iflush_exit
+    
     mov.l    ifr_addr, r0
     mov.l    p2_mask, r1
```
pcercuei, line 36:
```
tst  r5,r5
mov.l ifr_addr, r0
bt/s .iflush_exit
mov.l p2_mask, r1
```
Then it adds zero cycle penalty.

kernel/arch/dreamcast/kernel/cache.s
```diff
@@ -90,6 +97,11 @@ _icache_flush_range:
...
     .align 2
 _dcache_inval_range:
+    ! Exit early if no blocks to inval
+    mov      #0, r0
+    cmp/eq   r0, r5
+    bt       .dinval_exit
+
     ! Get ending address from count and align start address
     add      r4, r5
     mov.l    align_mask, r0
```
pcercuei, line 107:
```
tst r5,r5
add r4,r5
bt/s .dinval_exit
mov.l align_mask, r0
```
Just one cycle penalty.

kernel/arch/dreamcast/kernel/cache.s
```diff
@@ -115,26 +128,29 @@ _dcache_inval_range:
...
+    mov      #0, r0
+    cmp/eq   r0, r5
+    bt       .dflush_exit    
 
     ! Compare with flush_check
-    mov.w    flush_check, r2
+    mov.l    flush_check, r2
     cmp/hi   r2, r5
-    bt       _dcache_flush_all  ! If lines > flush_check, jump to _dcache_flush_all
+    bt       _dcache_flush_all  ! If count > flush_check, jump to _dcache_flush_all
```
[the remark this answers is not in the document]
pcercuei: Actually, I think you can rewrite the whole thing between `_dcache_flush_range` and `.dflush_loop` like this:

```
    ! Check that 0 < size  < flush_check
    tst      r5, r5
    mov.l    flush_check, r2

    bt/s     .dflush_exit
    mov.l    align_mask, r0

    cmp/hi   r2, r5
    add      r4, r5 ! Get ending address from count

    bt/s     _dcache_flush_all  ! If lines > flush_check, jump to _dcache_flush_all
    and      r0, r4 ! Align start address to 32 bytes
```

This should need only 4 cycles as each couple execute in parallel.

kernel/arch/dreamcast/kernel/cache.s
```diff
@@ -170,26 +186,29 @@ _dcache_flush_all:
...
+    mov.l    purge_check, r2
     cmp/hi   r2, r5
-    bt       _dcache_purge_all  ! If lines > purge_check, jump to _dcache_purge_all
+    bt       _dcache_purge_all  ! If count > purge_check, jump to _dcache_purge_all
 
-    ! Align start address
+    ! Get ending address from count and align start address
+    add      r4, r5
     mov.l    align_mask, r0
     and      r0, r4
```
pcercuei, line 202: You can do the exact same as for `_dcache_flush_range`:
```
    ! Check that 0 < size < purge_check
    tst      r5, r5
    mov.l    purge_check, r2

    bt/s     .dpurge_exit
    mov.l    align_mask, r0

    cmp/hi   r2, r5
    add      r4, r5 ! Get ending address from count

    bt/s     _dcache_purge_all  ! If lines > purge_check, jump to _dcache_purge_all
    and      r0, r4 ! Align start address to 32 bytes
```

kernel/arch/dreamcast/kernel/cache.s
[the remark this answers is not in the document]
pcercuei: Disregard my first comment - the `bt` will execute in parallel just fine. So in my 4 code suggestions you can use `bt` instead of `bt/s`.

kernel/arch/dreamcast/kernel/cache.s
pcercuei: Actually even better:
```
tst r5,r5
mov.l align_mask, r0
bt .dinval_exit
add r4,r5
```
Then you avoid the stall cycle caused by using r0 right after it's loaded from memory.

pcercuei:
So on one hand I don't believe the 0-size check is really necessary - it should be the responsibility of the caller to not pass dumb values. On the other hand, when ordered properly those checks don't have any runtime penalty.

Overall that's a good change and it looks like it works just fine.

pcercuei:
LGTM

kernel/arch/dreamcast/kernel/cache.s
```diff
@@ -24,11 +24,15 @@
 ! taken from sh-stub.c.
```
DC-SWAT, line 24: Not sure if this is the right comment now.
This has been completely rewritten in the last year.

DC-SWAT (approved):

kernel/arch/dreamcast/kernel/cache.s
andressbarajas: Updated

QuzarDC (approved):

PR #571, opened 2024-05-11
Added Makefile Rules for Doxygen Documentation

gyrovorbis:
- docs: generates documentation
- docs_clean: removes autogenerated documentation files
- docs_open: opens index.html from docs in the default browser


Makefile
```diff
@@ -33,6 +34,16 @@ distclean: clean
 	-rm -f lib/$(KOS_ARCH)/*
 	-rm -f addons/lib/$(KOS_ARCH)/*
 
+docs:
+	cd $(KOS_BASE)/doc; \
```
andressbarajas, line 38:
```suggestion
	doxygen $(KOS_BASE)/doc/Doxyfile
```

Haven't test this but can we do this way instead of changing directory?

darcagn: Yes, this way works and is better.

Makefile
```diff
@@ -33,6 +34,16 @@ distclean: clean
 	-rm -f lib/$(KOS_ARCH)/*
 	-rm -f addons/lib/$(KOS_ARCH)/*
 
+docs:
+	doxygen $(KOS_BASE)/doc/Doxyfile
+	doxygen
```
darcagn, line 39:
```suggestion
```

darcagn (changes requested):

darcagn (approved):

PR #572, opened 2024-05-11
Updated the FPU Exception Example

gyrovorbis:
Located at: `kos/examples/dreamcast/basic/fpu/exc`.

- There's no need for inline assembly anymore, thanks to the compiler intrinsic for fpscr implemented by Oleg endo.
- Show off how to reproduce even more different exception types.
- Tested for both -m4-single-only and -m4-single


andressbarajas (approved):
Looks great.  Glad you utilized the builtin fpscr functions.

QuzarDC (approved):
Double checked to make sure that the builtin is supported on 9.3.0, and it is! Looks good.

PR #574, opened 2024-05-14
More dc-chain improvements

darcagn:
- libgccjit can be built in GCC pass1, so I've moved it there. This has two benefits:
  - if building libgccjit separately from the rest of the toolchain, it no longer will take two passes
  - if building libgccjit as part of one's main toolchain, the `--enable-host-shared` compiler performance penalty is no longer incurred as this won't apply to pass2
- Fix up D and Ada language selection variables to better fit variables introduced in the last PR
- As requested on Discord, `make clean` behavior does _not_ remove archives (previously this behavior was found in `make clean-keep-archives`), and `make distclean` _does_ remove archives (previously this behavior was found in `make clean`)
- Because `Makefile.cfg` edits are tracked by git, the default options are now stored in `Makefile.default.cfg`. If a user wishes to alter the settings, they may copy `Makefile.default.cfg` to `Makefile.cfg` which is untracked through `.gitignore`. If a user does not wish to make edits, or erroneously edits `Makefile.default.cfg` without copying it, the settings stored in `Makefile.default.cfg` will apply in the absence of `Makefile.cfg`.


utils/dc-chain/Makefile.default.cfg
```diff
@@ -22,7 +25,7 @@
 # - gccrs-dev:    GCC fork for development of the GCCRS Rust compiler.
 # - rustc-dev:    GCC fork for development of the libgccjit rustc GCC codegen.
 # If unsure, select stable. See README.md for more detailed descriptions.
-toolchain_profile=stable
+toolchain_profile=rustc-dev
```
QuzarDC, line 28: Should these and the other settings have changed as well? I'm thinking you accidentally put up your own Makefile.cfg that you were testing with?

darcagn: Yeah, I force-pushed the defaults back. 😅

QuzarDC (approved):

PR #576, opened 2024-05-14
Add vid_clear() support for PM_RGB888P format

andressbarajas:
Since the PM_RGB888P format is 24-bits I had to find the lowest amount of bytes that is a full repeatable pattern and would fit on a 32-bit boundary.  

[B G R B] [G R B G] [R B G R] , [B G R B] [G R B G] [R B G R], ETC ...

[B G R B] [G R B G] [R B G R] = 12 bytes

The lowest amount of bytes is 12 so I split that up into three different 32-bit values([B G R B] [G R B G] [R B G R]) that will hold the rgb pattern. The  I cycled through the three 32-bit values until all the data is copied.

There does seem to be an issue with the PM_RGB888P format in KOS because on my dreamcast displays color but there are vertical black lines.  This occurs if you just set it to all white as well using:

```sq_set32(vram_l, 255, (vid_mode->width * vid_mode->height) * vid_pmode_bpp[PM_RGB888P]);```


andressbarajas:
Whats holding up this PR is that KOS doesn't support PM_RGB888P correctly right now. Im gonna look into that issue before I make this ready for review.

QuzarDC:
> Whats holding up this PR is that KOS doesn't support PM_RGB888P correctly right now. Im gonna look into that issue before I make this ready for review.

To be sure, I only just added in the video type when rebuilding video.c for completeness, but aside from adding the bit in the vid_clear to note, no effort was undertaken to look for code paths that would use it outside of video.c. The bfont drawing would likely choke on it and any number of other places that start with the presumption of 16 bit. I had actually thought to rather include a warning when choosing it about the lack of support.

It's my understanding that some of the PVR stuff just doesn't work anyways outside of 16bit mode, so if there are parts that seem mysteriously not working, you may want to validate in 08888 mode.

andressbarajas:
I think once this PR (https://github.com/KallistiOS/KallistiOS/pull/697) is merged, I make this one ready for review.

PR #580, opened 2024-05-18
Stack Protector environ.sh Option + Example Tweak

gyrovorbis:
So I'm only now stumbling upon the stack protector... and I think it's pretty badass and is extremely useful... I figured we should expose a config option to the laymen who won't even know this bad-boy exists...

I also found that the example didn't actually work correctly with LTO enabled, so I fixed that and tweaked it to be more of an automated test so we can use it as such later.

### DIFF
--- a/doc/environ.sh.sample
+++ b/doc/environ.sh.sample
@@ -143,6 +143,17 @@ export KOS_CFLAGS="${KOS_CFLAGS} -O2"
 export KOS_CFLAGS="${KOS_CFLAGS} -fomit-frame-pointer"
 #export KOS_CFLAGS="${KOS_CFLAGS} -fno-omit-frame-pointer -DFRAME_POINTERS"
 
+# Stack Protector
+#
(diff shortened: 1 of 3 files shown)


darcagn (approved):

QuzarDC:
You should take a look at this in context of your signals PR. Remember that the newlib implementation of the function is intended to raise(SIGABRT). So once signal/raise is implemented it may make sense to just ditch the patch that makes this weak/removes raise and implement the override behavior as a signal handler (which seems to be how other platforms would use this).

https://github.com/KallistiOS/KallistiOS/blob/e53988facb6269cbb09d3c5c1bdd4c9ef35beef9/utils/dc-chain/patches/newlib-4.4.0.20231231-kos.diff#L236-L245

darcagn:
Apparently I reviewed this the other day and forgot to merge? I'm merging now, but @QuzarDC raises an interesting point as we might want to look at removing some of the Newlib patches if we can adjust things on our end.

PR #581, opened 2024-05-18
BIOS Font Cleanup, Refactor, and More Ergonomic API

gyrovorbis:
The BIOS font is extremely useful; however, it has always bothered me for one reason: it's a pain in the ass to draw text with, due to not supporting `printf()`-style formatting... It always requires a temporary buffer, `snprintf()` or something, then a final draw of the aggregate string... for absolutely  no reason... The same API can be extended to support printf()-style formatters via variadic arguments...

- Added `printf()`-style formatting to draw string functions
- Added va_list variant for drawing strings
- Completely overhauled doxygen
- Made everything idiomatic and pretty C with stdint, stdbool, enums, etc.
- Updated all examples that were doing two-step string construction to just call directly into the new printf-style formatter API


kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -258,16 +263,16 @@ unsigned char bfont_draw_ex(uint8 *buffer, uint32 bufwidth, uint32 fg, uint32 bg
     /* Increment over the height of the font. 3bytes at a time (2 thin or 1 wide row) */
     for(y = 0; y < BFONT_HEIGHT; y+= (2-wide),ch+=((BFONT_THIN_WIDTH*2)/8)) {
         /* Do the first row, or half row */
-        word = (((uint16)ch[0]) << 4) | ((ch[1] >> 4) & 0x0f);
-        buffer = (uint8*)bfont_draw_one_row((uint16*)buffer, word, opaque, fg, bg, bpp);
+        word = (((uint16_t)ch[0]) << 4) | ((ch[1] >> 4) & 0x0f);
+        buffer = (uint8_t*)bfont_draw_one_row((uint16_t*)buffer, word, opaque, fg, bg, bpp);
```
andressbarajas, line 267:
```suggestion
        buffer = (uint8_t *)bfont_draw_one_row((uint16_t *)buffer, word, opaque, fg, bg, bpp);
```

kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -258,16 +263,16 @@ unsigned char bfont_draw_ex(uint8 *buffer, uint32 bufwidth, uint32 fg, uint32 bg
...
 
         /* If we're thin, increment to next row, otherwise continue the row */
         if(!wide) buffer += ((bufwidth - BFONT_THIN_WIDTH)*bpp)/8;
 
         /* Do the second row, or second half */
-        word = ((((uint16)ch[1]) << 8) & 0xf00) | ch[2];
+        word = ((((uint16_t)ch[1]) << 8) & 0xf00) | ch[2];
 
-        buffer = (uint8*)bfont_draw_one_row((uint16*)buffer, word, opaque, fg, bg, bpp);
+        buffer = (uint8_t*)bfont_draw_one_row((uint16_t*)buffer, word, opaque, fg, bg, bpp);
```
andressbarajas, line 275:
```suggestion
        buffer = (uint8_t *)bfont_draw_one_row((uint16_t *)buffer, word, opaque, fg, bg, bpp);
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -26,267 +28,296 @@
...
-#define BFONT_DBUTTON               BFONT_DREAMCAST_SPECIFIC+(14*BFONT_WIDE_WIDTH*BFONT_HEIGHT/8)
-#define BFONT_XBUTTON               BFONT_DREAMCAST_SPECIFIC+(15*BFONT_WIDE_WIDTH*BFONT_HEIGHT/8)
-#define BFONT_YBUTTON               BFONT_DREAMCAST_SPECIFIC+(16*BFONT_WIDE_WIDTH*BFONT_HEIGHT/8)
-#define BFONT_ZBUTTON               BFONT_DREAMCAST_SPECIFIC+(17*BFONT_WIDE_WIDTH*BFONT_HEIGHT/8)
-#define BFONT_LTRIGGER              BFONT_DREAMCAST_SPECIFIC+(18*BFONT_WIDE_WIDTH*BFONT_HEIGHT/8)
-#define BFONT_RTRIGGER              BFONT_DREAMCAST_SPECIFIC+(19*BFONT_WIDE_WIDTH*BFONT_HEIGHT/8)
-#define BFONT_STARTBUTTON           BFONT_DREAMCAST_SPECIFIC+(20*BFONT_WIDE_WIDTH*BFONT_HEIGHT/8)
-#define BFONT_VMUICON               BFONT_DREAMCAST_SPECIFIC+(21*BFONT_WIDE_WIDTH*BFONT_HEIGHT/8)
+/** \brief Size of a row for JISX-0208 characters */
+#define JISX_0208_ROW_SIZE  94
```
andressbarajas, line 67:
```suggestion
#define JISX_0208_ROW_SIZE          94
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+*/
+void bfont_draw_str_ex(void *b, uint32_t width, uint32_t fg, uint32_t bg, uint8_t bpp,
+                       bool opaque, const char *fmt, ...) __printflike(7, 8);
+
+/** \brief   Draw a full formatted string to any sort of buffer (with va_args).
+
+    This function is equivalnet to bfont_draw_str_ex(), except that the variadic
+    argument list is passed via a pointer to a va_list.
+
+    \param b                The buffer to draw to.
```
andressbarajas, line 481:
```suggestion
    \param  b                The buffer to draw to.
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+    \param width            The width of the buffer in pixels.
```
andressbarajas, line 482:
```suggestion
    \param  width            The width of the buffer in pixels.
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+    \param fg               The foreground color to use.
```
andressbarajas, line 483:
```suggestion
    \param  fg               The foreground color to use.
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+    \param bg               The background color to use.
```
andressbarajas, line 484:
```suggestion
    \param  bg               The background color to use.
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+    \param bpp              The number of bits per pixel in the buffer.
```
andressbarajas, line 485:
```suggestion
    \param  bpp              The number of bits per pixel in the buffer.
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+    \param opaque           If non-zero, overwrite background areas with black,
```
andressbarajas, line 486:
```suggestion
    \param  opaque           If non-zero, overwrite background areas with black,
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+                            otherwise do not change them from what they are.
```
andressbarajas, line 487:
```suggestion
                             otherwise do not change them from what they are.
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+    \param fmt              The printf-style format string to draw.
```
andressbarajas, line 488:
```suggestion
    \param  fmt              The printf-style format string to draw.
```

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -413,14 +465,35 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+    \param ...              Additional printf-style variadic arguments
```
andressbarajas, line 489:
```suggestion
    \param  ...              Additional printf-style variadic arguments
```

andressbarajas:
Sorry I went a little overb. ard with the spacing. I sa.  a difference in the file and went ham. Then I checked other headers and they have the same issue too.  Feel free to ignore the spacing in the documentation.

kernel/arch/dreamcast/include/dc/biosfont.h
andressbarajas, line 489: Should this be var_args ? or does ... still work?

gyrovorbis: uhhhhhhhhhhhhh..... good question.

gyrovorbis: Yeah, it works as intended. Just checked.

kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -214,17 +217,19 @@ uint16 *bfont_draw_one_row(uint16 *b, uint16 word, uint8 opaque, uint32 fg, uint
...
-unsigned char bfont_draw_ex(uint8 *buffer, uint32 bufwidth, uint32 fg, uint32 bg, uint8 bpp, uint8 opaque, uint32 c, uint8 wide, uint8 iskana) {
-    uint8 *ch;
-    uint16 word;
-    uint8 y;
+size_t bfont_draw_ex(void *buf, uint32_t bufwidth, uint32_t fg, uint32_t bg, 
+                     uint8_t bpp, bool opaque, uint32_t c, bool wide, bool iskana) {
+    uint8_t *ch;
+    uint16_t word;
+    uint8_t y;
+    uint8_t *buffer = (uint8_t*)buf;
```
andressbarajas, line 232:
```suggestion
    uint8_t *buffer = (uint8_t *)buf;
```

kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -285,29 +290,31 @@ unsigned char bfont_draw_ex(uint8 *buffer, uint32 bufwidth, uint32 fg, uint32 bg
 }
 
 /* Draw half-width kana */
-unsigned char bfont_draw_thin(void *b, uint32 bufwidth, uint8 opaque, uint32 c, uint8 iskana) {
-    return bfont_draw_ex((uint8 *)b, bufwidth, bfont_fgcolor, bfont_bgcolor, (bfont_32bit ? (sizeof (uint32)) : (sizeof (uint16))) << 3, opaque, c, 0, iskana);
+size_t bfont_draw_thin(void *b, uint32_t bufwidth, bool opaque, uint32_t c, bool iskana) {
+    return bfont_draw_ex((uint8_t *)b, bufwidth, bfont_fgcolor, bfont_bgcolor,
```
QuzarDC, line 294:
```suggestion
    return bfont_draw_ex(b, bufwidth, bfont_fgcolor, bfont_bgcolor, 
```
Now that the underlying function is also void*, there's no need for this.

kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -285,29 +290,31 @@ unsigned char bfont_draw_ex(uint8 *buffer, uint32 bufwidth, uint32 fg, uint32 bg
...
+size_t bfont_draw_thin(void *b, uint32_t bufwidth, bool opaque, uint32_t c, bool iskana) {
+    return bfont_draw_ex((uint8_t *)b, bufwidth, bfont_fgcolor, bfont_bgcolor, 
+                         (bfont_32bit ? (sizeof (uint32_t)) : (sizeof (uint16_t))) << 3, opaque, c, 0, iskana);
 }
 
 /* Compat function */
-unsigned char bfont_draw(void *buffer, uint32 bufwidth, uint8 opaque, uint32 c) {
-    return bfont_draw_ex((uint8 *)buffer, bufwidth, bfont_fgcolor, bfont_bgcolor, (bfont_32bit ? (sizeof (uint32)) : (sizeof (uint16))) << 3, opaque, c, 0, 0);
+size_t bfont_draw(void *buffer, uint32_t bufwidth, bool opaque, uint32_t c) {
+    return bfont_draw_ex((uint8_t *)buffer, bufwidth, bfont_fgcolor, bfont_bgcolor,
```
QuzarDC, line 300:
```suggestion
    return bfont_draw_ex(buffer, bufwidth, bfont_fgcolor, bfont_bgcolor, 
```

kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -285,29 +290,31 @@ unsigned char bfont_draw_ex(uint8 *buffer, uint32 bufwidth, uint32 fg, uint32 bg
...
+size_t bfont_draw(void *buffer, uint32_t bufwidth, bool opaque, uint32_t c) {
+    return bfont_draw_ex((uint8_t *)buffer, bufwidth, bfont_fgcolor, bfont_bgcolor, 
+                        (bfont_32bit ? (sizeof (uint32_t)) : (sizeof (uint16_t))) << 3, opaque, c, 0, 0);
 }
 
 /* Draw wide character */
-unsigned char bfont_draw_wide(void *b, uint32 bufwidth, uint8 opaque, uint32 c) {
-    return bfont_draw_ex((uint8 *)b, bufwidth, bfont_fgcolor, bfont_bgcolor, (bfont_32bit ? (sizeof (uint32)) : (sizeof (uint16))) << 3, opaque, c, 1, 0);
+size_t bfont_draw_wide(void *b, uint32_t bufwidth, bool opaque, uint32_t c) {
+    return bfont_draw_ex((uint8_t *)b, bufwidth, bfont_fgcolor, bfont_bgcolor,
```
QuzarDC, line 306:
```suggestion
    return bfont_draw_ex(b, bufwidth, bfont_fgcolor, bfont_bgcolor, 
```

kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -285,29 +290,31 @@ unsigned char bfont_draw_ex(uint8 *buffer, uint32 bufwidth, uint32 fg, uint32 bg
...
-   support is disabled (BFONT_CODE_ISO8859_1).
-   XXX: Seems like this can be shrunk to use uint8 for nChr/Mask/Flag and
-    getting rid of nMask.
-   */
-void bfont_draw_str_ex(void *b, uint32 width, uint32 fg, uint32 bg, uint8 bpp, uint8 opaque, const char *str) {
-    uint16 nChr, nMask, nFlag;
-    uint8 *buffer = (uint8 *)b;
+void bfont_draw_str_ex_va(void *b, uint32_t width, uint32_t fg, uint32_t bg, uint8_t bpp, bool opaque, 
+                          const char *fmt, va_list *var_args) {
+    char string[1024 * 4];
```
QuzarDC, line 312: Just wondering if there's any specific rationale for the size of `string` here.

QuzarDC (changes requested):

kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -119,15 +122,15 @@ uint32 sjis2jis(uint32 sjis) {
...
-uint32 euc2jis(uint32 euc) {
+static uint32_t euc2jis(uint32_t euc) {
     return euc & ~0x8080;
 }
 
 /* Given an ASCII character, find it in the BIOS font if possible */
-uint8 *bfont_find_char(uint32 ch) {
-    uint8   *fa = syscall_font_address();
+uint8_t *bfont_find_char(uint32_t ch) {
+    uint8_t   *fa = syscall_font_address();
```
DC-SWAT, line 131: The font address is static, we can call `syscall_font_address()` once, and not every character.

kernel/arch/dreamcast/hardware/biosfont.c
andressbarajas: I think the screen can fit a total of 1060 of the thin width bios font characters.

gyrovorbis: Ah, good call. I wasn't sure how many could fit onscreen, so that seemed okay, but what @andressbarajas is suggesting is far more precise.

kernel/arch/dreamcast/hardware/biosfont.c
gyrovorbis: Good point! I'm caching it now.

QuzarDC: > Good point! I'm caching it now.

It doesn't interfere directly here, but you should keep in mind that the font locking is for access to the data region rather than for the calling of the font_address syscall. Caching the result is a good idea, but abstracts the value from the syscall a bit, so it seemed worthwhile to call that out.

kernel/arch/dreamcast/hardware/biosfont.c
QuzarDC: > Ah, good call. I wasn't sure how many could fit onscreen, so that seemed okay, but what @andressbarajas is suggesting is far more precise.

So I wasn't sure and dug in more. C99 standard says the max size for a printf buffer should be at least 4095, so maybe that was in your mind when you picked it. :) I wonder if there's a define to pull that kind of thing out of newlib.

andressbarajas:
@QuzarDC What are the changes you are requesting in this one?

andressbarajas (approved):
This all looks good to me. Will retest tonight and then work on the followup to add newlines, tabs, and maybe handle spaces differently.

andressbarajas:
Retested all examples besides the mruby one.  It all looks good.

QuzarDC:
After the whole chunk of research with #599 I think this needs some minor readjustment. Instead of adding the `...` param to the current `bfont_draw_str`, I think it should be refactored to something like `bfont_draw_printf`. Take a look at the paradigm here: https://github.com/KallistiOS/KallistiOS/blob/7f261a396099cfd4cbbd771108bf670914568ed6/kernel/debug/dbgio.c#L151-L184 . 

With the rewrite of the debug in #599 , you could use the `dbio_write_str` transparently with dbgio "fb" as both are safe of malloc. If that behavior is changed then it becomes impossible to do in malloc and unsafe to do in interrupts.

gyrovorbis:
> After the whole chunk of research with #599 I think this needs some minor readjustment. Instead of adding the `...` param to the current `bfont_draw_str`, I think it should be refactored to something like `bfont_draw_printf`. Take a look at the paradigm here:
> 
> https://github.com/KallistiOS/KallistiOS/blob/7f261a396099cfd4cbbd771108bf670914568ed6/kernel/debug/dbgio.c#L151-L184
> 
> .
> With the rewrite of the debug in #599 , you could use the `dbio_write_str` transparently with dbgio "fb" as both are safe of malloc. If that behavior is changed then it becomes impossible to do in malloc and unsafe to do in interrupts.

Is this because you found that the va_list processing is inherently using `malloc()` internally, and doing it like this gives you a form that is malloc()-free?

QuzarDC:
> Is this because you found that the va_list processing is inherently using `malloc()` internally, and doing it like this gives you a form that is malloc()-free?

Right, either the va_args/list or the underlying vsnprintf (which matches dblog) were not safe for malloc and neither were ever guaranteed to be. So having the printf-like version is fine and really nice to use, but it needs to be refactored some so that we still have access to draw without it. The way that you've factored it out here even the underlying call that the top-level makes has the vsnprintf so there's no way to *just* write a string.

gyrovorbis:
> > Is this because you found that the va_list processing is inherently using `malloc()` internally, and doing it like this gives you a form that is malloc()-free?
> 
> Right, either the va_args/list or the underlying vsnprintf (which matches dblog) were not safe for malloc and neither were ever guaranteed to be. So having the printf-like version is fine and really nice to use, but it needs to be refactored some so that we still have access to draw without it. The way that you've factored it out here even the underlying call that the top-level makes has the vsnprintf so there's no way to _just_ write a string.

Well I'll be damned... the only time in my entire life where I agree that duplicating the API like this is warranted... Yeah, good point...

kernel/arch/dreamcast/hardware/biosfont.c
```diff
@@ -354,16 +370,39 @@ void bfont_draw_str_ex(void *b, uint32 width, uint32 fg, uint32 bg, uint8 bpp, u
...
+   */
+void bfont_draw_str_ex(void *b, uint32_t width, uint32_t fg, uint32_t bg, uint8_t bpp, 
+                       bool opaque, const char *fmt, ...) {
+    va_list var_args;
+    va_start(var_args, fmt);
+
+    bfont_draw_str_ex_va(b, width, fg, bg, bpp, opaque, fmt, &var_args);
+
+    va_end(var_args);
+}
```
QuzarDC, line 387:
```suggestion
    DONT COMMIT ME OR QUZAR WONT BE ABLE TO REVIEW
int bfont_draw_printf(void *b, uint32_t width, uint32_t fg, uint32_t bg, uint8_t bpp, 
                       bool opaque, const char *fmt, ...) {
    char string[1088];
    int rv = 0;
    va_list var_args;
    va_start(var_args, fmt);
    rv = vsnprintf(string, sizeof(string), fmt, *var_args);
    va_end(var_args);    
    
    bfont_draw_str_ex(b, width, fg, bg, bpp, opaque);
    return rv;
}
```

Something like this is what I'm suggesting. It's almost identical to how dbgio does it, and this way you can even provide the return from vsnprintf so that the user can know if they're going to be truncated or something.

gyrovorbis: K, done! I'm not returning the return value from vsnprintf(), though, because it's how big of a buffer you NEED to have to not be truncated, and it's not like they have any control over that buffer either, because it's internal. Seemed pointless.

gyrovorbis:
Rebased onto `master` and ready for `v2.2.0`!

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -26,267 +31,296 @@
...
+#define BFONT_JISX_0208_ROW1        BFONT_WIDE_START
+/** \brief Start of JISX-0208 Row 16-47 (Start of Level 1) in Font Block */   
+#define BFONT_JISX_0208_ROW16       BFONT_WIDE_START + (658 * BFONT_WIDE_WIDTH * BFONT_HEIGHT / 8)  
+/** \brief JISX-0208 Row 48-84 (Start of Level 2) in Font Block */
+#define BFONT_JISX_0208_ROW48       BFONT_JISX_0208_ROW16 + ((32 * JISX_0208_ROW_SIZE) * BFONT_WIDE_WIDTH * BFONT_HEIGHT / 8) 
+
+/** \brief Start of DC Specific Characters in Font Block */
+#define BFONT_DREAMCAST_SPECIFIC    BFONT_WIDE_START + (7056 * BFONT_WIDE_WIDTH * BFONT_HEIGHT / 8) 
+/** \brief Takes a DC-specific icon index and returns a character offset. */
+#define BFONT_DC_ICON(offset)       BFONT_DREAMCAST_SPECIFIC + ( (offset) * BFONT_WIDE_WIDTH * BFONT_HEIGHT / 8)
```
QuzarDC, line 83: While you're macro-ing ...`BFONT_WIDE_WIDTH * BFONT_HEIGHT / 8` would  be a good candidate. But if there's a reason to leave it out, that's fine too. bytes-per-char.

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -26,267 +31,296 @@
...
-/** \defgroup vmu_icons Builtin VMU Icons
+/** \brief Builtin VMU Icons
     \ingroup  bfont_indicies
 
     Builtin VMU volume user icons. The Dreamcast's
     BIOS allows the user to set these when formatting the VMU.
+*/
+typedef enum bfont_vmu_icon {
+    BFONT_ICON_INVALID_VMU     = 0x00, /**< \brief Invalid */
+    BFONT_ICON_HOURGLASS_ONE   = 0x01, /**< \brief Hourglass 1 */
```
QuzarDC, line 125: Do these briefs make it much more readable in the generated docs or something? Cause they seem fairly redundant to me. If any of the define names were ambiguous at all, then maybe, but it's almost exactly just the text at `BFONT_ICON_*`

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -395,13 +443,20 @@ unsigned char bfont_draw_thin(void *buffer, uint32 bufwidth, uint8 opaque,
...
+/** @} */
+
+/** \name  String Drawing 
+    \brief Methods for rendering formatted text.
 
-    This function draws a NUL-terminated string in the set encoding to the given
+    @{
+*/
+
+/** \brief   Draw a full decorated string to any sort of buffer.
```
QuzarDC, line 457: What's 'decorated' mean in this context?

kernel/arch/dreamcast/include/dc/biosfont.h
```diff
@@ -415,14 +470,56 @@ unsigned char bfont_draw_wide(void *buffer, uint32 bufwidth, uint8 opaque,
...
+    \sa bfont_draw_str_ex_fmt()
+*/
+void bfont_draw_str_ex_vfmt(void *b, uint32_t width, uint32_t fg, uint32_t bg,
+                            uint8_t bpp, bool opaque, const char *fmt,
+                            va_list *var_args);
 
 /** \brief   Draw a full string to a buffer.
-    \ingroup bfont
 
-    This function draws a NUL-terminated string in the set encoding to the given
```
QuzarDC, line 425: It may be that it's more confusing than helpful, but `NUL` was purposeful in this context for the name of the character that is at encoding 0 (null) in string encoding schemes. To differentiate from NULL as an invalid pointer (even though of course they all have the same value and whatnot).

QuzarDC (approved):
We'd already hashed out the big stuff months ago. Sorry for taking so long on this, had confused it and the keyboard one that is soo much bigger.

Just a few tiny comments, pretty much none of it mandatory so take or leave but check them out.

QuzarDC (approved):

PR #583, opened 2024-05-19
Add lock-less implementation of compare-and-swap atomic

pcercuei:
This commit introduces an implementation of an atomic compare-and-swap, that does not use any locking mechanism.

Instead, a tiny ASM function is added, with labels delimiting the beginning and end of the atomic section.

In the unlikely event that a thread is preempted in that tiny time frame, the next time it is scheduled its PC address will be reset to the beginning of the atomic section.

This compare-and-swap atomic can in turn be used to implement lock-less C11 atomics, which can therefore be used both inside and outside IRQ contexts.

### DIFF
--- a/include/kos/atomic.h
+++ b/include/kos/atomic.h
@@ -1,0 +1,38 @@
+/* KallistiOS ##version##
+
+   include/kos/atomic.h
+   Copyright (C) 2024 Paul Cercueil
+*/
+
+/** \file    kos/atomic.h
+    \brief   Atomic compare-and-swap support.
+    \ingroup kthreads
+
+    This file contains KOS' atomic API.
+ */
+
+#ifndef __KOS_ATOMIC_H
+#define __KOS_ATOMIC_H
+
+#include <sys/cdefs.h>
+__BEGIN_DECLS
+
+#include <stdint.h>
+
+/** \brief  Atomically compare and swap values
+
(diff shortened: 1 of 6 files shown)


gyrovorbis:
By the way, dude, I had a look at what you're doing... I must say it's pretty fucking clever... 

I've spoken with the libUltra N64 guys a bunch about their MIPS CPU and what it can do, and I swear it just has a few assembly instructions that are doing essentially this.

pcercuei:
This compare-and-swap can be used to implement all 32-bit (and lower) atomics. It may or may not be better than the imask thing. Sadly, I wrote this thinking that it could support > 32bit atomics, but it cannot. So whether or not this change is useful is up for debate.

QuzarDC:
So I realize that this may have been violated elsewhere recently (and I'll see if I can track some others down), but platform specific solutions should be under their arch folder (\kernel\arch\dreamcast\kernel) rather than in the general kernel level (\kernel\thread). Does it seem likely that others would be able to implement in the same way? If not then we may want a slightly abstracted check to bypass the threading  `arch_check_atomic` or such that the architecture should provide if  it has such a capability.

pcercuei:
@QuzarDC yes, the ASM part is ASM just to be able to delimitate properly the critical section.

PR #587, opened 2024-05-23
DBGIO Cleanup + Auxiliary Output Devices

gyrovorbis:
Sick of your `stdout` options being:
     1. dcload-only: Too bad your output doesn't show up by default in emulators, and it looks like something is wrong to users.
     2.  fb-only: Too bad then you can't read your output in the console
     
Presenting:`dbgio_aux_select("fb")` Which allows you to select an optional dbgio device for auxiliary output which is just a secondary output device... As oppposed to `dbgio_dev_select("fb");` which makes it the primary device with no other output.

- Cleaned up dbgio API, source file, and drivers
- Got rid of annoying declarations to private internal structures only
  used in init.c
- Added API for auxiliary output device
- Modified API to allow for runtime dbgio device registry

### DIFF
--- a/include/kos/dbgio.h
+++ b/include/kos/dbgio.h
@@ -1,20 +1,21 @@
 /* KallistiOS ##version##
 
    kos/include/dbgio.h
-   Copyright (C)2000,2004 Megan Potter
-
+   Copyright (C) 2000, 2004 Megan Potter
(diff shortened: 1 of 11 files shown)


QuzarDC:
Closing out since #1284 and #1291 replace its functionality.

PR #588, opened 2024-05-23
Fixed Warnings from realpath() Signature Mismatches

gyrovorbis:
- The `realpath()` signature declared within Newlib and the stdlib headers with POSIX extensions has both parameters as pointers with the `__restrict` keyword, versus `realpath.c` which was lacking `__restrict` and made the second argument an array.
- Also removed an unnecessary (and now incorrect) forward declaration in `fs.c` which is no longer necessary now that Newlib is declaring it for us.


andressbarajas:
@QuzarDC since these changes are mirrored on your new branch.  Do you mind merging this one?

QuzarDC (approved):

PR #589, opened 2024-05-23
Better error/warning handling in some examples

QuzarDC:
Fixing warnings in examples, and cleaning up data before shutdown. Some from just -fanalyzer others from the malloc debugging in #599 

- Ensuring that we do get memory allocation back before writing to that memory. In the serpent_dma example I also replaced the overallocation for alignment purposes with a memalign (perhaps this should be one of the more standard/idiomatic functions?)
- Setting pragmas to ignore the explicit intentional bad code in the mmu and stackprotection examples. I put these at the narrowest scope because it seemed very much like they should be intentionally visible for the user rather than covering the whole file and being unobtrusive at the top and bottom (as we have before just to have cleaner compiles in the main KOS code).
- Free data and close files.
- Correct crash in serpent_dma


examples/dreamcast/2ndmix/2ndmix.c
```diff
@@ -398,6 +398,11 @@ int load_pcx(char *pcxdata) {
     int num_bytes;
     struct  pcx_hdr pcxh;
 
+    if(image == NULL) {
+        printf("Image not allocated\r\n");
```
gyrovorbis, line 402:
```suggestion
        fprintf(stderr, "Image not allocated\r\n");
```
So I am always super anal about making sure that *error* messages go to `stderr` rather than `stdout`. It's not buffered, so it displays immediately, plus when we get testing automation set up on GitHub, build runners listen for messages on `stderr` when trying to diagnose issues with a build that can flag it as a failure.

examples/dreamcast/network/httpd/httpd.c
```diff
@@ -299,7 +311,8 @@ void *client_thread(void *p) {
     fs_close(f);
 
 out:
-    free(buf);
+    if(buf != NULL)
```
gyrovorbis, line 314: Was this flagged or something? Because it's actually valid to `free()` a `NULL` in C/C++. It's just a nop, essentially.

gyrovorbis (commented):
Couple tiny little comments... Happy with how much prettier and more pristine everything looks.

examples/dreamcast/network/httpd/httpd.c
QuzarDC: Whoops! Nope, what was flagged with something else in the same path and I just had the possibility of double-free rather than freeing NULL in mind -_-; Good catch.

examples/dreamcast/2ndmix/2ndmix.c
QuzarDC: I'd just say that it's out of scope here. There's already plenty of error messages in 2ndmix and they all just go to printf. If it is to be changed it should be an overhaul for the whole file.

Where does dbglog fit into that scheme? We have quite a few examples using that too.

andressbarajas:
It all looks good.  I tested all examples edited except for httpd.  These are probably outside the scope of this PR but the serpent_dma crashes for me with this:
```Serpent sample: press START to exit
allocated 14080 bytes for 20 stacks, 20 + 2 slices, and 32-byte pvr_vertex_t
allocated 14080 bytes for 20 stacks, 20 + 2 slices, and 32-byte pvr_vertex_t
Unhandled exception: PC 20736e35, code 1, evt 00e0
 R0-R7: 00000000 ffffffff 000717e0 8c4b0520 00000000 8cfffeb0 8cfffeb0 8cfffead
 R8-R15: 4622f900 736e3531 36313535 39303733 37303434 37363434 8c83ef2c 8cffff8c
 SR 40000000 PR 20736e35
Stack Trace: frame pointers not enabled!
kernel panic: unhandled IRQ/Exception
arch: aborting the system
andressbarajas@GameDev serpent_dma % 
```

and the nullptr example just hangs with no output to terminal.

QuzarDC:
> It all looks good. I tested all examples edited except for httpd. These are probably outside the scope of this PR but the serpent_dma crashes for me with this:
> 
> ```
> allocated 14080 bytes for 20 stacks, 20 + 2 slices, and 32-byte pvr_vertex_t
> allocated 14080 bytes for 20 stacks, 20 + 2 slices, and 32-byte pvr_vertex_t
> Unhandled exception: PC 20736e35, code 1, evt 00e0
>  R0-R7: 00000000 ffffffff 000717e0 8c4b0520 00000000 8cfffeb0 8cfffeb0 8cfffead
>  R8-R15: 4622f900 736e3531 36313535 39303733 37303434 37363434 8c83ef2c 8cffff8c
>  SR 40000000 PR 20736e35
> Stack Trace: frame pointers not enabled!
> kernel panic: unhandled IRQ/Exception
> arch: aborting the system
> andressbarajas@GameDev serpent_dma % 
> ```
> 
> and the nullptr example just hangs with no output to terminal. Im using GCC 13.2 on an old Mac. This happens on master branch as well :^/

Funnily enough, like @gyrovorbis , I hadn't hit this when I sanity tested after my change, but recompiling now it popped up. Used stackprotector to quickly track it down. Expanded a static buffer that was too small and added some protection against this happening in the future.

examples/dreamcast/vmu/vmu_beep/beep.c
```diff
@@ -39,18 +40,25 @@
 
 KOS_INIT_FLAGS(INIT_DEFAULT);
 
+plx_font_t *fnt;
```
gyrovorbis, line 43:
```suggestion
static plx_font_t *fnt;
```

examples/dreamcast/vmu/vmu_beep/beep.c
```diff
@@ -39,18 +40,25 @@
...
+plx_fcxt_t *cxt;
```
gyrovorbis, line 44:
```suggestion
static plx_fcxt_t *cxt;
```

examples/dreamcast/vmu/vmu_beep/beep.c
```diff
@@ -39,18 +40,25 @@
 
 KOS_INIT_FLAGS(INIT_DEFAULT);
```
gyrovorbis, line 41:
```suggestion
```
Might as well get rid of it, since it's no longer needed for the defaults.

examples/dreamcast/vmu/vmu_beep/beep.c
```diff
@@ -39,18 +40,25 @@
...
 
+plx_font_t *fnt;
+plx_fcxt_t *cxt;
+
+void cleanup(void) {
```
gyrovorbis, line 46:
```suggestion
static void cleanup(void) {
```

gyrovorbis (commented):
Teensy little `static` places.

examples/dreamcast/vmu/vmu_beep/beep.c
QuzarDC: It was never needed for the defaults. Don't remember why it was tossed in.

gyrovorbis (approved):
k, looks good to me now!

PR #591, opened 2024-05-24
Add Makefile rule to check minimum GCC version

darcagn:
If `$KOS_GCCVER_MIN` is defined in your project's `Makefile`, this addition will check the current `$KOS_GCCVER` against it to determine if your GCC version meets the minimum necessary for the project.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -1,7 +1,8 @@
 # KallistiOS ##version##
 #
 # Makefile.rules
-# (c)2000-2001 Megan Potter
+# Copyright (c) 2000, 2001 Megan Potter
+# Copyright (c) 2024 Eric Fradella
 #
 
 # Global KallistiOS Makefile include
@@ -73,4 +74,22 @@ romdisk.o: romdisk.img
 	$(KOS_BASE)/utils/bin2o/bin2o romdisk.img romdisk romdisk_tmp.o
 	$(KOS_CC) -o romdisk.o -r romdisk_tmp.o $(KOS_LIB_PATHS) -Wl,--whole-archive -lromdiskbase
 	rm romdisk_tmp.o
+endif
+
+# Define KOS_GCCVER_MIN in your Makefile if you want to enforce a minimum GCC version.
+ifdef KOS_GCCVER_MIN
+  ifeq ($(shell \
+    awk 'BEGIN { \
+        split("$(KOS_GCCVER_MIN)", min, "."); \
+        split("$(KOS_GCCVER)", cur, "."); \
+        if (cur[1] > min[1] || \
+           (cur[1] == min[1] && cur[2] > min[2]) || \
+           (cur[1] == min[1] && cur[2] == min[2] && cur[3] >= min[3])) { \
+            print 1; \
+        } else { \
+            print 0; \
+        } \
+    }'), 0)
+    $(error A minimum GCC version of $(KOS_GCCVER_MIN) is required, but $(KOS_GCCVER) is currently in use.)
+  endif
 endif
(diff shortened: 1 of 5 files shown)


darcagn:
After submitting the PR, I realized I wasn't entirely sure where this should go (`Makefile.rules`, `Makefile.prefab`, `Makefile`, etc...)
So let me know if there's a more desirable location for it.

QuzarDC:
So ... this doesn't seem like the right way to go about this. 

By moving to c17 as standard, we've already cut off support for gcc 4 and 5. The listing for supported languages in 4.7.4 only goes up to c11: https://gcc.gnu.org/onlinedocs/gcc-4.7.4/gcc/Standards.html (and it's listed as 'limited' support, in 4.9.4 it's listed as substantial, but it still didn't support c17 and was buggy for DC work). I'm not sure what versions of 5 people might be using with the DC but we never had patches for them, and the last version of it also didn't support c17 ( https://gcc.gnu.org/onlinedocs/gcc-5.5.0/gcc/Standards.html ).

Then for the newer versions, this is entirely proxying for language support. Shouldn't the right thing be to query the compiler to see if it knows the minimum language support needed? By doing that we could also then have an easier generic way to know if we support c++ or objc or any of the exciting new languages being supported like fortran.

gyrovorbis:
> So ... this doesn't seem like the right way to go about this.
> 
> By moving to c17 as standard, we've already cut off support for gcc 4 and 5. The listing for supported languages in 4.7.4 only goes up to c11: https://gcc.gnu.org/onlinedocs/gcc-4.7.4/gcc/Standards.html (and it's listed as 'limited' support, in 4.9.4 it's listed as substantial, but it still didn't support c17 and was buggy for DC work). I'm not sure what versions of 5 people might be using with the DC but we never had patches for them, and the last version of it also didn't support c17 ( https://gcc.gnu.org/onlinedocs/gcc-5.5.0/gcc/Standards.html ).
> 

I mean, we have code *all over the place* for GCC 4 and even under which all needs to be cleaned up, and I had just assumed that would happen sometime in a follow-up PR.

FIG A: environment at least attempting to work with GCC2 and 3: https://github.com/KallistiOS/KallistiOS/blob/b0662100c30161bb25c9af862b5b774e6743c981/environ_base.sh#L50

FIG B: `init.c` doing all sorts of extra shit for GCC4: https://github.com/KallistiOS/KallistiOS/blob/b0662100c30161bb25c9af862b5b774e6743c981/kernel/arch/dreamcast/kernel/init.c#L33

> Then for the newer versions, this is entirely proxying for language support. 

That's not really true necessarily, though, across the board. For the FPU exception test, I am checking for GCC5 or greater, because that's when `builtin_sh_fpscr()` was added to GCC as a builtin function. 

> Shouldn't the right thing be to query the compiler to see if it knows the minimum language support needed? By doing that we could also then have an easier generic way to know if we support c++ or objc or any of the exciting new languages being supported like fortran.
I errrrr uhhhh..... is that a thing you can do? ....cuz yeah, that would be pretty damn sweet.... lol. And yeah, I've been thinking that as well.... what do we do if they disabled C++ or ObjC in their toolchains....

You guys have ideas on how to query that crap programmatically?

darcagn:
> You guys have ideas on how to query that crap programmatically?

The best solution I have right now is to check if the following succeeds or fails:

`echo "int main(void) { return 0; }" | kos-cc -std=$(1) -fsyntax-only -x $(2) -`

where `$(1)` is the desired standard, and `$(2)` is the desired language. Though the piped text would have to be different for a different language.

ljsebald (approved):

ljsebald:
I don't have a problem with this as it stands now, but it is a bit awkward to use it only to gate off language features...

This is basically one of those places where a more fancy build system would be what we'd want if we were really concerned with that kind of stuff. That said, I don't think we really are overly concerned with that kind of stuff, so I'm gonna go ahead and merge this as-is. It's useful functionality and it doesn't hurt anything.

PR #592, opened 2024-05-24
Separate Memory Manager API + sbrk() Update

gyrovorbis:


kernel/arch/dreamcast/include/arch/mm.h
```diff
@@ -0,0 +1,350 @@
...
+    \retval 0   Success
+*/
+int mm_init(void);
+/** \endcond */
+
+/** @} */
+
+__END_DECLS
+
+#endif
```
andressbarajas, line 350:
```suggestion
#endif


```

kernel/arch/dreamcast/include/arch/mm.h
```diff
@@ -0,0 +1,351 @@
...
+int mm_brk_set_capacity(size_t bytes);
+
+/** \brief Sets the current program break segment position.
+
+    Attempts to set the mm_brk_current() pointer directly.
+
+    \param new_pos  New current program break position pointer.
+
+    \retval 0       success: 
+    \retval -1      Failure: (`errno` is set to `ENOMEM`)
```
pcercuei, line 282: Why set errno? ...
Just return the (negative) errno code directly.

QuzarDC:
The changes here were basically implemented in #1286 . @gyrovorbis if there was more from here you'd want to get in, I figure it'd have to start from scratch anyways since there's so many changes from when this was done.

PR #593, opened 2024-05-25
Headers

andressbarajas:
1. Removed #include <kos.> from mintr.  It was needed.  Recompiled and retested PPP example and no issues.
2. Added #include <kos/oneshot_timer.h> to kos.h


gyrovorbis (approved):
Sweet!

PR #594, opened 2024-05-26
Add GPROF support to KOS

andressbarajas:
### Adding GPROF support to KOS

I’ve been working on getting GPROF support integrated into KOS and wanted to share what I’ve done so far. Here’s a rundown of the process and some thoughts on what we might need to do next.

**What I’ve Done:**

1. Implemented GPROF Methods:

- I added the necessary GPROF methods: mcount, monstartup, and mcleanup.
- Also added a 10ms timer to track the PC of the main thread.
- These changes are in koslib (gmon.c). Let me know if you think they should go somewhere else like newlib.

2. Calling the Methods:

- monstartup() runs before main().
- mcleanup() runs after main() finishes.
- mcount gets called every time a function starts.

**The Official Way:**

The official way to handle this is detailed in the crt1.s file that comes with our toolchain ([crt1.S in GCC](https://github.com/gcc-mirror/gcc/blob/master/libgcc/config/sh/crt1.S)). However, our startup.s file ([startup.s in KOS](https://github.com/KallistiOS/KallistiOS/blob/241724aaf43c5962ff16f214985872895bbdf757/kernel/arch/dreamcast/kernel/startup.s)) replaces this.

In the original crt1.s, there are sections with #ifdef PROFILE that handle:

- Calling monstartup and mcleanup.
- Using trapa # 33 for mcount calls when a function is entered.

To get this working, we’d need to:

- Update startup.s to include those #ifdef PROFILE sections.
- Modify the toolchain to use gcrt1.s (compiled with -DPROFILE) when the -pg flag is used.

Example Makefile entries:

```
gcrt1.o: $(srcdir)/config/sh/crt1.S
  $(gcc_compile) -c -DPROFILE $<
```

```
#define STARTFILE_SPEC \
  "%{!shared: %{pg|p|profile:%{static:gcrt0.o%s;:gcrt1.o%s};static:crt0.o%s;:crt1.o%s}} \
   crti.o%s %{static:crtbeginT.o%s;shared:crtbeginS.o%s;:crtbegin.o%s}"
#endif
```
Note: We can’t use -pg with -fomit-frame-pointer (our default setting).

**My Alternative Approach:**
To avoid all the hassle with the toolchain and startup.s, I used:

- -finstrument-functions to call mcount.
- __attribute__ ((constructor)) to call monstartup.
- __attribute__ ((destructor)) to call mcleanup.

I added a new addon library libgprof that users can link against. Just use:

```
CFLAGS = -g -finstrument-functions
LDFLAGS = -lgprof
```
This seems to be working, but I’m still ironing out some issues. This PR is a draft for now. If we decide to go with the official method, I’ll need some help with that.

Thanks to @darcagn and @cepawiel for pointing me in the right direction.


addons/libgprof/gprof.c
```diff
@@ -0,0 +1,49 @@
...
+    Note: This implementation requires the following compilation and linking flags:
+
+    CFLAGS = -g -finstrument-functions
+    LDFLAGS = -lgprof
+*/
+
+#include <stdint.h>
+
+/* Start and End address for .text portion of program */
+#define BASE_ADDRESS 0x8c010000
```
cepawiel, line 20: This should be tweaked slightly for NAOMI (0x8c020000) support. Both addresses are in `environ_dreamcast.sh` currently as part of `KOS_LDFLAGS`. Not sure if it makes more sense to modify the environ script and pass a define as a CFLAG or to just put both addresses here behind an ifdef. Alternatively I think you should be able to use `__executable_start` (similar to `_etext` below) which is provided by the linker script, might very slightly change the asm.

kernel/libc/koslib/gmon.c
```diff
@@ -0,0 +1,272 @@
...
+    /* Initialize histogram thread related members */
+    cxt->running_thread = true;
+    cxt->histogram_thread = thd_create(false, histogram_callback, NULL);
+    thd_set_prio(cxt->histogram_thread, PRIO_DEFAULT / 2);
+
+    /* Turn profiling on */
+    moncontrol(true);
+}
+
+void _mcleanup(void) {
```
cepawiel, line 204: I think this should probably just be a single underscore. I did see GCC uses 2 for that symbol (https://github.com/gcc-mirror/gcc/blob/master/libgcc/config/sh/crt1.S#L184) but not sure there's a specific need to match since we use our own startup asm, but perhaps I'm missing something?

addons/libgprof/gprof.c
cepawiel: Ah I missed that the only place `BASE_ADDRESS` gets used is in the setup function, probably just simplest to use `__executable_start` if that works. I initially assumed it was something being used for every sample and figured GCC could generate better code with a constant over a linker symbol.

andressbarajas: I'll use the __executable_start.  Thanks for pointing this out.  Forgot about Noami

kernel/libc/koslib/gmon.c
andressbarajas: No specific need.  Just wanted to match.  I'll make it use one.

include/kos/gmon.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include <sys/cdefs.h>
+__BEGIN_DECLS
+
+#include <stdbool.h>
+
+/** \defgroup debugging_gprof GPROF
+    \brief    Utilities for GPROF
+    \ingroup  debugging
+
```
gyrovorbis, line 28: Could you write a brief little summary of how to enable this and do the profiling here if it's not too long?

kernel/libc/koslib/gmon.c
```diff
@@ -0,0 +1,386 @@
...
+        return;
+
+overflow:
+        cxt->state = GMON_PROF_ERROR;
+        //ERR("mcount: call graph buffer size limit exceeded, gmon.out will not be generated\n");
+        return;
+    }
+}
+
+void _monstartup(uintptr_t lowpc, uintptr_t highpc) {
```
gyrovorbis, line 251: It looks like on POSIX at least, this is a *public function* which you can call to only profile a certain range of functions? https://man.freebsd.org/cgi/man.cgi?apropos=0&sektion=3&query=monstartup&manpath=FreeBSD+7.0-current&format=html

Is it as simple as just declaring this as public within the header to support this use-case? If so, I'd do it. If it's way more work than that, whatever, outside the scope of this PR, can come later. :)

kernel/libc/koslib/gmon.c
```diff
@@ -0,0 +1,386 @@
...
+    cxt->ncounters = (cxt->textsize + HISTFRACTION - 1) / HISTFRACTION;
+    counter_size = cxt->ncounters * sizeof(HIST_COUNTER_TYPE);
+
+    cxt->nfroms = (cxt->textsize + HASHFRACTION - 1) / HASHFRACTION;
+    froms_size = cxt->nfroms * sizeof(uint16_t);
+
+    cxt->nnodes = (cxt->textsize * ARCDENSITY) / 100;
+    nodes_size = cxt->nnodes * sizeof(gmon_node_t);
+
+    printf("Low: %p, High: %p\n", (void *)cxt->lowpc, (void *)cxt->highpc);
```
gyrovorbis, line 276:
```suggestion
    printf("Low: %p, High: %p\n", (void *)cxt->lowpc, (void *)cxt->highpc);
```
I would probably change these to like `dbglog(DBG_VERBOSE, ...);` so that people like Shirobon don't bitch that KOS is too chatty... I would actually make the first log line something like `dbglog(DBG_NOTICE, "Enabling gprof from <%p to %p>\n", ...);` or something...

I'm of the opinion that we should be logging more *major events* such as starting to profile, a controller being added/removed, etc as `DBG_NOTICE` with subsequent more detailed output/information as `DBG_VERBOSE` or `DBG_DEBUG` so that we can capture useful logging info at different levels of granularity.

kernel/arch/dreamcast/kernel/startup.s
```diff
@@ -180,8 +185,98 @@ normal_exit:
...
+	mov.l	r6, @-r15
+	mov.l	r7, @-r15
+	sts.l	pr, @-r15
+
+	sts pr, r4      ! Save previous program counter to r4 (frompc)
+	stc spc,r5	    ! Save current program counter to r5 (selfpc)
+
+	! The code sequence emitted by gcc for the profiling trap is
+	! .long 0   --- This is a 4-byte no-op placeholder to align trapa
+	! trapa #33 --- This is a 2-byte instruction
```
gyrovorbis, line 230: If this is going to be reserving `TRAPA #33` specifically, maybe we should put a `#define` for the TRAPA index in the public header somewhere, so people will know it's reserved and don't step on it, and/or it can easily be configured later?

...unless this is some "no shit" standard TRAPA number everyone else knows about except me... In which case I'd still probably note it in the Doxygen.

kernel/libc/koslib/gmon.c
```diff
@@ -0,0 +1,386 @@
+
```
gyrovorbis, line 1: Copyright header here.

kernel/libc/koslib/gmon.c
andressbarajas: Didnt realize that monstartup should be public. Honestly, I dont see many people doing it that way since you still need portions of your code compiled with -pg to even produce a call graph.  I'll think about how I would support this use case.

kernel/libc/koslib/gmon.c
```diff
@@ -0,0 +1,386 @@
...
+    arc.frompc = from_addr;
+    arc.selfpc = node->selfpc;
+    arc.count = node->count;
+    fwrite(&arc, sizeof(gmon_arc_t), 1, out);
+
+    /* Traverse the right subtree */
+    traverse_and_write(out, cxt, node->right, from_addr);
+}
+
+static void *histogram_callback(void *arg) {
```
gyrovorbis, line 128: Mind naming this something like `histogram_thread(..)` or `histogram_exec(..)` or something to distinguish between just a one-off callback and something that is a persistent thread callback which isn't just immediately returning?

kernel/libc/koslib/gmon.c
andressbarajas: My initial plan was to remove these after I was done debugging so that is why formatting is booboo. I will look into formatting them nicer and wrapping them in dbglog.

kernel/arch/dreamcast/kernel/startup.s
andressbarajas: Yeah I can do that.  The `trapa #33` is reserved by the GCC compiler for sh.

kernel/libc/koslib/gmon.c
andressbarajas: Didn't think of it that way. Makes sense. histaogram_thread it is.

andressbarajas:
Alright, its ready for review.  Currently we have two different ways of doing gprof profiling. Using 
```
CFLAGS = -g -finstrument-functions -fno-inline 
LDFLAGS = -lgprof
```
is the preferred method right now. I worked on supporting the official flags but can sometimes experience a crash if you mix it with like a -O0 flag:
```
CFLAGS = -pg -fno-inline 
```

I also need help in editing the toolchain to support the -pg flag.  All in all I need people to help me test and play around with it.

andressbarajas:
Spoke too soon.  Im getting a runtime issue.

PR #595, opened 2024-05-27
Minor cleanups in FS

QuzarDC:
EDIT: Refreshed these changes from spring for the current situation. Collection of mostly formal changes.

- `fs_map_hnd` already sets the errno on its bad paths, so avoid duplicating that.
- Fix typo in dirent documentation.
- Import defines from genromfs and romdisk documentation to fs_romdisk. Demagic volumename/filename maxes.
- Add romdisk write and unlink stubs. Not sure that they're helpful at all, but they enforce more accurate behavior based on the POSIX definitions for what the errnos should be set to.
- Cleanups around shutdowns for some FSes and conversion of some fh values to bools.

### DIFF
--- a/kernel/fs/fs.c
+++ b/kernel/fs/fs.c
@@ -333,19 +333,16 @@ static fs_hnd_t * fs_map_hnd(file_t fd) {
 /* Close a file and clean up the handle */
 int fs_close(file_t fd) {
     int retval;
-    fs_hnd_t *hnd = fs_map_hnd(fd);
+    fs_hnd_t *h = fs_map_hnd(fd);
 
-    if(!hnd) {
-        errno = EBADF;
-        return -1;
-    }
+    if(!h) return -1;
 
     /* Deref it and remove it from our table */
-    retval = fs_hnd_unref(hnd);
+    retval = fs_hnd_unref(h);
 
     /* Reset our position */
-    if(hnd->refcnt == 0)
-        hnd->idx = 0;
+    if(h->refcnt == 0)
+        h->idx = 0;
 
     fd_table[fd] = NULL;
     return retval ? -1 : 0;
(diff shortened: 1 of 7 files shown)


kernel/fs/fs_socket.c
```diff
@@ -298,8 +305,8 @@ int socket(int domain, int type, int protocol) {
         }
     }
 
-    /* If i is NULL, we got through the whole list without finding anything. */
-    if(!i) {
+     /* If i is NULL, we got through the whole list without finding anything. */
+    if(i != NULL) {
```
Firnis, line 309: Sholdn't it be `i == NULL`?

QuzarDC: Absolutely! Thank you.

andressbarajas:
Why the code style change from if(!h) => if(h == NULL) ?

kernel/fs/fs.c
```diff
@@ -745,10 +739,7 @@ static int fs_vfcntl(file_t fd, int cmd, va_list ap) {
     fs_hnd_t *h = fs_map_hnd(fd);
     int rv;
 
-    if(!h) {
-        errno = EBADF;
```
gyrovorbis, line 749: In case you didn't see it in Discord:

Why was this removed? Was it not how POSIX handles it? Screwing up the upstream?

QuzarDC:
> Why the code style change from if(!h) => if(h == NULL)? Just curious what is the coding style we should use for future PRs

Whoops. I'll find and fix that. Unintentional. For style I'd say only if it's really specifically important to demonstrate `NULL` vs other values, but that'd be rare.

gyrovorbis (approved):
Awesome! Looks good!

PR #567, opened 2024-05-10
Add watchdog timers + use watchdog timer in modem code

pcercuei:
A watchdog timer is a timer that will trigger an action (through a
pre-registered callback) after a timeout expires. The code using a
watchdog timer will typically reset it periodically to prevent the
action from happening.

Note that this code compiles but it UNTESTED! I don't have the infrastructure to test modem code.

PR #596, opened 2024-05-28
Add missing __END_DECLS

QuzarDC:
This fixes the compilation of libtsunami examples, and other c++ code that broke after #567 . Additionally cleaned up other instances of the same issue.


gyrovorbis (approved):
Good find!

PR #598, opened 2024-05-29
Work around MMU issue with SQs

pcercuei:
This change makes it possible to use SQs in a program that uses the MMU,
by disabling the MMU in sq_lock(), and re-enable it again in
mmu_restore().

This means that memory-mapped data must not be accessed while the SQ is
locked. This should be fine, as it wasn't working properly before, so
it's not a regression.

### DIFF
--- a/kernel/arch/dreamcast/hardware/sq.c
+++ b/kernel/arch/dreamcast/hardware/sq.c
@@ -8,6 +8,7 @@
 */
 
 #include <arch/cache.h>
+#include <arch/mmu.h>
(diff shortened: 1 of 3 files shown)


gyrovorbis (approved):
Looks good to me... as was mentioned, this was straight up broken with the MMU before, so regardless of if we want to handle this differently some day, this is fixing what we have going on atm.

DC-SWAT:
Also SQ can be disabled in MMUCR register, bit 9.

PR #599, opened 2024-05-29
Malloc leak and debug fixes

### LINKED ISSUE #290: KOS_DEBUG level 2 or enabling KM_DBG_VERBOSE hang the DC
This is a very low priority issue which impacts the power-user build configuration variables within `/kos/include/kos/opts.h`, but I still want to document this so it can eventually get fixed.

When KOS_DEBUG level 2 (or greater) is enabled, `KM_DBG_VERBOSE` becomes enabled (or it can be enabled directly), this enables a bunch of additional `printf()` calls all over `/kos/kernel/libc/koslib/malloc.c`. Currently with these enabled, the DC immediately hangs without so much as printing the regular KOS banner as soon as the application launches.

Creating a dummy macro to replace `printf()` with a NOP fixes the issue. So does changing the log to go to the framebuffer; however, going to the regular debug log still crashes... 

My theory is that these forms of logging are requiring internal buffer allocations which are calling into `malloc()` which are recursively calling back into the log functions, which call back into `malloc()`, etc... Not sure how this worked previously, but if the logging mechanism itself requires an allocation, we have to find a different mechanism for logging allocation debug messages.

QuzarDC:
These resolve a series of issues I have found while attempting to debug malloc behavior. With these changes we now have the capability to use the `KM_DBG_VERBOSE` flag again, and the hello example (as just a baseline init/deinit of KOS) is no longer leaking 2208 bytes on exit.

- Rearrange the allocation of dcload-ser's buffer in `fs_dcload_init_console` to prevent an unnecessary malloc from happening.
- Shutdown threading using the destroy function to free threads rather than manually removing individual pieces of their data. This stops TLS allocations from leaking.
- Use `LIST_FOREACH_SAFE` in places that remove/free the elements of the list. This is not safe to do in the normal `LIST_FOREACH` and can cause corruption or broken lists.
- Replace `vsprintf` for `vsnprintf` to back dgbio and dbglog. Since we are using a static buffer this helps guard against buffer overflows. Additionally check to ensure it worked before trying to print the buffer. 
- The one possibly problematic change was remove the use of `fs_write(1` to back dbgio. I couldn't tell by walking back through the repo why it was done quite like this. It seems that it may have been just to try to more visibly reflect "this is going to stdout" but fs_write has a shortcut in it that directly sends back to the same dbgio backend function.

If anyone wants, these could get split off into their own PR, but this helps to validate/verify the above fixes. These should be safer than they seem because none should change any behavior in malloc outside of when `KM_DBG` is set:
- Check if the current thread is null before derefing to get its tid when KM_DBG is set. Now in `malloc_stats()` and malloc debug output thread tid 0 represents memory allocated when no threading is running.
- Use static strings for the function names stored in the debug structs.
- Refactor block checking to deduplicate the checks done in `free`, `realloc`, and `mem_check_block`.
- Replace printfs with custom string compositing to dbgio. This partially resolves #290 for `KM_DBG_VERBOSE`. This was required because `printf` series functions (including our `dbglog`) may use malloc in them.


QuzarDC:
Of note, if I set the verbose debug on malloc it seems to either crash or lock up immediately. Additionally I am still seeing though with the 9.3.0 toolchain that data is leaking (though not with 15.0.0) in just the plain hello example with `INIT_MALLOCSTATS` enabled.
image

I believe the size 12 calloc is the allocation of the rootfs fh here: https://github.com/KallistiOS/KallistiOS/blob/edcd7615c1d7ef0c88f8c0e601b9e11668b010e8/kernel/fs/fs.c#L55-L57 which, it makes sense that it would be leaked, but it doesn't make sense that it would even be accessed at all in hello world.

I don't think I'll actually be solving all these before setting this to ready, but wanted to take some extra notes to spin off different issues.

andressbarajas:
Here is the issue about the KM_DBG_VERBOSE crash/lockup: https://github.com/KallistiOS/KallistiOS/issues/290

QuzarDC:
> Here is the issue about the KM_DBG_VERBOSE crash/lockup: #290

Thanks for that, I wasn't aware. Added a touchup to the dbglog core that I thought might help avoid possible mallocs inside the debug output, but it didn't seem to make a difference.

andressbarajas:
@QuzarDC What are the best examples to test this with?

QuzarDC:
> @QuzarDC What are the best examples to test this with?

Unfortunately most of it is internal/debug paths or the shutdown paths. I actually walked through about a dozen of our examples enabling `INIT_MALLOCSTATS` to see the impact before/after (before, it would show more memory still 'in use'). I've actually got a separate PR that I held back from this one with all the examples that I also made sure to add proper cleanups to (closing files and freeing memory).

I think anything with lots of file I/O and text out would likely be the best since those are the kinds of public paths most changed here. Testing out building anything with KM_DBG set would also be good to work out the cleanups in there.

QuzarDC:
Figured out how to fix the KM_DBG_VERBOSE issue, so going to polish that up and fold it into the existing changes here since they're all so related.

kernel/libc/koslib/malloc.c
```diff
@@ -1673,6 +1673,105 @@ typedef struct memctl {
...
+        strcat(dbg_print_buffer, " bytes @ 0x");
+        itoa((uint32)m, (dbg_print_buffer + strlen(dbg_print_buffer)), 16);
+    }
+
+    strcat(dbg_print_buffer, "\n");
+
+    dbgio_write_str(dbg_print_buffer);
+}
+
+int mem_check_block_int(memctl_t * ctl, int source) {
```
andressbarajas, line 1716:
```suggestion
int mem_check_block_int(memctl_t *ctl, int source) {
```

kernel/libc/koslib/malloc.c
```diff
@@ -1673,6 +1673,105 @@ typedef struct memctl {
...
+#include <stdlib.h>
+char dbg_print_buffer[256];
+
+enum func_type_names { name_MALLOC = 0, name_REALLOC = 1, name_MEMALIGN = 2, name_CALLOC = 3,
+    name_FREE = 4, name_CHECKALL = 5, name_CHECK = 6, name_STATS = 7};
+
+static const char * func_type[8] =
+    {"malloc", "realloc", "memalign", "calloc", "free", "check_all", "check", "stats"};
+
+void dbg_print_thd_addr_action(tid_t thread, uint32 addr, void* m, size_t s, uint8 which) {
```
andressbarajas, line 1690:
```suggestion
void dbg_print_thd_addr_action(tid_t thread, uint32 addr, void *m, size_t s, uint8_t which) {
```

kernel/libc/koslib/malloc.c
```diff
@@ -1673,6 +1673,105 @@ typedef struct memctl {
...
+#define get_cur_tid_safe ((thd_current == NULL) ? (tid_t)0 : thd_current->tid)
+
+#include <kos/dbgio.h>
+#include <stdlib.h>
+char dbg_print_buffer[256];
+
+enum func_type_names { name_MALLOC = 0, name_REALLOC = 1, name_MEMALIGN = 2, name_CALLOC = 3,
+    name_FREE = 4, name_CHECKALL = 5, name_CHECK = 6, name_STATS = 7};
+
+static const char * func_type[8] =
```
andressbarajas, line 1687:
```suggestion
static const char *func_type[8] =
```

kernel/arch/dreamcast/fs/fs_dcload.c
```diff
@@ -513,17 +513,20 @@ void fs_dcload_init_console(void) {
...
-    if(dcload_wrkmem) {
-        if(dclsc(DCLOAD_ASSIGNWRKMEM, dcload_wrkmem) == -1) {
-            free(dcload_wrkmem);
-            dcload_type = DCLOAD_TYPE_IP;
-            dcload_wrkmem = NULL;
-        }
-        else {
-            dcload_type = DCLOAD_TYPE_SER;
+    /* dcload IP will always return -1 here. Serial will return 0 and make
+    no change since it already holds 0 as 'no mem assigned */
```
ljsebald, line 518:
```suggestion
      no change since it already holds 0 as 'no mem assigned */
```

kernel/libc/koslib/dbglog.c
```diff
@@ -39,13 +40,11 @@ void dbglog(int level, const char *fmt, ...) {
         spinlock_lock(&mutex);
 
     va_start(args, fmt);
-    (void)vsprintf(printf_buf, fmt, args);
+    i= vsnprintf(printf_buf, sizeof(printf_buf), fmt, args);
```
ljsebald, line 43:
```suggestion
    i = vsnprintf(printf_buf, sizeof(printf_buf), fmt, args);
```

ljsebald (changes requested):
Just a few minor things that I noticed....

ljsebald (approved):

PR #600, opened 2024-06-02
Diagnose help on reg dumps

andressbarajas:
1. Added KOS_ADDR2LINE
2. When dumping registers on an unhandled exception, generate a terminal command the user can easily enter to diagnose the issue.

This hasn't been tested.  I have been having issues generating an exception.


QuzarDC:
For me, running the fpu/exc example outputs 
    
    Original Value:
        FPSCR [40000]:
                FR     = 0
                SZ     = 0
                PR     = 0
                DN     = 1
                Cause  = 0
                Enable = 0
                Flag   = 0
                RM     = 0
 which would indicate that all exceptions are currently disabled. I tested enabling them by setting this:    
    
    /* Set all exceptions to enabled */
     __builtin_sh_set_fpscr(fpscr_start | 0xf80);
    fpscr_start = __builtin_sh_get_fpscr();
    printf("Exceptions Enabled:\n%s", fpscr_stringify(fpscr_start, buffer, sizeof(buffer)));
    
Before running the tests and it seems to work, but on the first exception it just hangs. Start still pulls it out, but they are disabled by defaut via startup.s: https://github.com/KallistiOS/KallistiOS/blame/34f7e943d311cd0105eb012571e7fcda41d3370d/kernel/arch/dreamcast/kernel/startup.s#L118-L122
    
It's possible the existing handler for them just doesn't work correctly since they've always been disabled.

andressbarajas:
Thanks for testing! Besides the issue that you created a PR for, are there any improvements you see I could make to this PR?

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -83,22 +83,71 @@ int trapa_set_handler(irq_t code, irq_handler hnd, void *data) {
...
+
+#ifdef FRAME_POINTERS
+        while(fp != 0xffffffff) {
+            if((fp & 3) || (fp < 0x8c000000) || (fp > _arch_mem_top))
+                break;
+
+            dbglog(DBG_DEAD, " %08lx", arch_fptr_ret_addr(fp));
+            fp = arch_fptr_next(fp);
+        }
+#endif
```
QuzarDC, line 147: This is just a second `arch_stk_trace_at(fp, 0)` without the little little separator prints. Why print it twice?

QuzarDC:
> Thanks for testing! Besides the issue that you created a PR for, are there any improvements you see I could make to this PR?

The inclusion of the addr2line is great, had been using it throughout my malloc testing, being able to just copy+paste from the dbgio is really user-friendly. I made a comment in the code on what seems to be just a duplication. 

After further thought I realized that this is only for *unhandled* exceptions. So in addition to enabling exceptions for FPU in the fpu_exc example, it was necessary to uninstall the existing FPU exception handler: `irq_set_handler(EXC_FPU, NULL, NULL);` to trigger this.

kernel/arch/dreamcast/kernel/irq.c
andressbarajas: I'm constructing a single terminal command that they can copy paste. That is why the info is repeated

QuzarDC: Ohhhhhhhh! That makes perfect sense. I had missed that they were going on the same line as the rest. That makes it so much better. Though then it almost seems like this should just live inside `arch_stk_trace`. I guess it wouldn't have access to the pc/pr there, and the only other place it's used is in asserts and it wouldn't be as helpful with them since you already know the line.

Got it!

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -83,22 +83,71 @@ int trapa_set_handler(irq_t code, irq_handler hnd, void *data) {
...
+            return "Data TLB miss (write)";
+        case EXC_DTLB_PV_READ:
+            return "Data TLB protection violation (read)";
+        case EXC_DTLB_PV_WRITE:
+            return "Data TLB protection violation (write)";
+        case EXC_FPU:
+            return "FPU exception";
+        case EXC_INITIAL_PAGE_WRITE:
+            return "Initial page write exception";
+        default:
```
QuzarDC, line 113:
```suggestion
        case EXC_INITIAL_PAGE_WRITE:
            return "Initial page write exception";
        case EXC_TRAPA:
            return "Unconditional trap (trapa)";
        case EXC_USER_BREAK_PRE:
        case EXC_USER_BREAK_POST:
            return "User break";
        default:
```
Added in the irq_completion_code ones, just to round it out. Not sure if they'd be visible in the same unhandled exception flow, but I don't see why not.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -83,22 +83,71 @@ int trapa_set_handler(irq_t code, irq_handler hnd, void *data) {
...
+        case EXC_DATA_ADDRESS_READ:
+            return "Data address error (read)";
+        case EXC_DATA_ADDRESS_WRITE:
+            return "Data address error (write)";
+        case EXC_DTLB_MISS_READ:
+            return "Data TLB miss (read)";
+        case EXC_DTLB_MISS_WRITE:
+            return "Data TLB miss (write)";
+        case EXC_DTLB_PV_READ:
+            return "Data TLB protection violation (read)";
```
QuzarDC, line 106:
```suggestion
        case EXC_ITLB_MISS:
        case EXC_DTLB_MISS_READ:
            return "Instruction or Data(read) TLB miss";
        case EXC_DTLB_MISS_WRITE:
            return "Data(write) TLB miss";
        case EXC_ITLB_PV:            
        case EXC_DTLB_PV_READ:
            return "Instruction or Data(read) TLB protection violation";
```
There are a few of the codes that have two exceptions mapped to them. Or contextually do the iTLB codes not make sense to include?

QuzarDC (commented):
Was able to test this generating an unhandled fpu exception, then walked through the book on the exceptions. I marked out the possible gaps in the string table. Good to approve after you check out the comments (even if you don't incorporate them, since it might just not make sense to).

kernel/arch/dreamcast/kernel/stack.c
```diff
@@ -31,18 +31,27 @@ void arch_stk_trace(int n) {
...
+        
+        if(n == 0) {
+            /* Get the return address from the function pointer */
+            fp = arch_fptr_ret_addr(fp);
 
-        if(n <= 0) {
-            dbgio_printf("   %08lx\n", arch_fptr_ret_addr(fp));
+            /* Validate the return address */
+            if(!arch_valid_address(fp)) {
+                dbgio_printf("   %08lx   (invalid return address)\n");
```
QuzarDC, line 51:
```suggestion
                dbgio_printf("   %08lx   (invalid return address)\n", fp);
```

QuzarDC (approved):

PR #601, opened 2024-06-02
Correct bitmasks in fpu_exc stringify

QuzarDC:
Very tiny change. When testing for #600 noticed that the wrong values were being printed for the fpscr data. All were set to the same 6-bit size, but according to the table in the SH4 manual:
image
and the bit shifting being done, the flags and enabled are each 5 bits.


gyrovorbis (approved):
You are correct, and I am an idiot. Thank you, kindly!

PR #604, opened 2024-06-04
Avoid infinite loop in mouse example (also, is mouse broken?)

QuzarDC:
Currently, if you start up the mouse example with no mouse, you will go into an inescapable loop. With this adjustment you can now still exit from the example by pressing Start on a controller (if no controller was plugged in, it would always just exit).

Perhaps more importantly, I can't seem to get it to work at all with my Treamcast mouse and @andressbarajas reported similar here: https://github.com/KallistiOS/KallistiOS/pull/551#issuecomment-2118666206 .

If anyone has a functioning mouse it would be great to get it tested but not a prerequisite for this since the behavior only strictly needs a controller to validate (I may open a matching issue but I hope someone is able to give it a quick try).

### DIFF
--- a/examples/dreamcast/libdream/mouse/mouse.c
+++ b/examples/dreamcast/libdream/mouse/mouse.c
@@ -13,10 +13,6 @@ void mouse_test(void) {
 
         if(!cont) continue;
 
-        mouse = maple_enum_type(0, MAPLE_FUNC_MOUSE);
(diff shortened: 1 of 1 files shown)


andressbarajas (approved):

gyrovorbis (approved):
Looks good.

PR #605, opened 2024-06-05
Fixed false failure in once_test on 16MB RAM

gyrovorbis:
- Accidentally allocated too many goddamn threads only testing on 32MB units
- Somehow allocation was failing gracefully on 16MB, the program never crashed, and it just returned a real failure code...
- Running out of memory allocating a thread then trying to join on a NULL later is apparently pristinely protected against. Impressive.
- Modified test to allocate 475 threads on 16MB and 950 on 32MB.
- Fixed the error code assignment when failing to join the thread, as it was printing "1" rather than "-1."
- Also enabled `INIT_MALLOCSTATS` to see what the final RAM stats look like with all that thread allocating.

### DIFF
--- a/examples/dreamcast/basic/threading/once/once_test.c
+++ b/examples/dreamcast/basic/threading/once/once_test.c
@@ -23,7 +23,7 @@
 #include <dc/maple/controller.h>
 
 #define UNUSED __attribute__((unused))
-#define THD_COUNT 600
+#define THD_COUNT (475 * (DBL_MEM? 2 : 1))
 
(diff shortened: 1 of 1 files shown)


andressbarajas:
I was the one who found this issue and tested that 475 threads work on 16mb dreamcast

gyrovorbis:
> I was the one who found this issue and tested that 475 threads work on 16mb dreamcast

YEP. I still need to set up my second dev Dreamcast which will be stock 16MB.

darcagn (approved):

PR #607, opened 2024-06-05
Fix texturing and closing opened files

andressbarajas:
This PR fixes texturing in the CPP version of the gltest.  Also we need to close opened files.


gyrovorbis:
I'm adding this to the current milestone for the current release, because actually this is pretty much bugfixing/cleaning up existing code and examples.

QuzarDC (approved):

PR #608, opened 2024-06-05
Threaded Controller Callbacks v2

QuzarDC:
This is a replacement of #209 and is intended to avoid the button callback system running entirely in an irq context. This severely limited what could be used in it and caused a massively problematic shutdown (hangs, crashes) when the callback was used for it's most common purpose, to exit.

The basic operation is unchanged: call the function to install a callback, call it with `NULL` as the callback to clear the existing one. Now though, you can install any number of them that you want. Ones with `0` as the address, which are catch-alls will run after any that specify specific controllers. Passing `NULL` as the callback will uninstall all currently installed on a port.

Right now it strictly maintains the existing API, which never actually stated only one could be installed. I've intentionally not specified too much of the details of the backing implementation in the API documentation in order to keep it open still for expansion in later feature PRs (allow testing against non-button inputs, add/remove/own the installed callback object, and perhaps have the option for shared vs individual threads per callback).

Small changes were needed to existing threading, worker, and init code to facilitate this, which are all explained in review comments. They were primarily to be able to cleanly shut down the callback threads or to allow one of those threads to cleanly shutdown KOS.


pcercuei:
> I needed to add an exception in worker threads to prevent a deadlock situation where you're cleaning up the worker from its own thread, which seems like a valid condition, especially if they'll be useful for this. @pcercuei , perhaps you might have some input into their usage here.

Yeah. Don't do that. Instead, what I suggest is to allocate a single worker thread in init code that handles all input callbacks, instead of having one worker per callback. That would simplify the code quite a bit.

pcercuei:
Well you have a function named `cont_shutdown`,  you can destroy the worker in there, no?

QuzarDC:
> Well you have a function named `cont_shutdown`, you can destroy the worker in there, no?

If the worker's callback is `exit()` (the most common use case) then calling `thd_worker_destroy` would deadlock because you `thd_join` to the current thread. This is why I added the check to avoid that.

pcercuei:
@QuzarDC well, if you don't join the thread, you leak memory. And since you free up "worker" without joining the thread first, you'll have a use-after-free in the thread.

pcercuei:
@QuzarDC the way I see to solve this issue:
- The deadlock with `thd_join()` with the current thread should actually not happen. IMHO `thd_join(thd_current);` should always return an error.
- You should always call `thd_join()` anyway for non-detachable threads, to free up their memory. So just use a detachable thread for your worker, using `thd_worker_create_ex`. Note that in that case `thd_join()` will silently fail and it won't deadlock.
- Handle the worker being destroyed from within its own callback by moving the `free(worker);` to the end of `thd_worker_thread()`, before the return.

QuzarDC:
> @QuzarDC the way I see to solve this issue:
> 
>     * The deadlock with `thd_join()` with the current thread should actually not happen. IMHO `thd_join(thd_current);` should always return an error.
> 
>     * You should always call `thd_join()` anyway for non-detachable threads, to free up their memory. So just use a detachable thread for your worker, using `thd_worker_create_ex`. Note that in that case `thd_join()` will silently fail and it won't deadlock.
> 
>     * Handle the worker being destroyed from within its own callback by moving the `free(worker);` to the end of `thd_worker_thread()`, before the return.

Thank you for the suggestions. Moving the free would certainly help even just for the generic instance of requesting a detached worker. I had considered resolving the issue via detaching but had not seen yet how to work around the worker effectively holding itself. This'll do it.

I had gotten bogged down in the fact that KOS already has an issue in general cases (not just with worker threads) with a thread that is not the main kernel thread being the one to shut down the threading system (in the more general case, calling `exit()`). I was mistakenly attributing that issue merely to this workflow, but just generally spawning a thread and calling exit will do it.

So it seems I'll be spinning a PR or two off this for general updates to the threading system. Which is good.

kernel/arch/dreamcast/include/dc/maple/controller.h
```diff
@@ -209,21 +205,16 @@ typedef void (*cont_btn_callback_t)(uint8_t addr, uint32_t btns);
...
+    your callback. Any callback with addr==0 will be installed to the end of
+    the list of callbacks and will run after callbacks with the same btns but
+    a specified address.
 
     \param  addr            The controller to listen on (or 0 for all ports). 
                             This value can be obtained by using maple_addr().
     \param  btns            The buttons bitmask to match.
     \param  cb              The callback to call when the buttons are pressed.
+                            Passing NULL will uninstall all callbacks on the
+                            addr/btn combination.
```
pcercuei, line 217: `addr/btns`

QuzarDC:
It seems there's still a significant issue here. When I spin off multiple workers and try to destroy them on shutdown, it hangs on attempting to destroy the first worker (unless it's the current thread).

As far as I can tell, it's deadlocking at the thd_join now, though not every time, sometimes it makes it up to the worker thread returning. That indicates to me that it's some sort of race condition likely involving two things waiting on each other.

So far I think I identified some small gaps in that could cause this but patching over either seemed to have no impact:
1) In `thd_worker_destroy` I believe irqs need to be disabled before setting the flag and the genwait_wake. Otherwise it's possible that the thread frees worker prior to its use in the genwait_wake. Additionally if the thread makes it all the way to exiting, it could get destroyed before being woken. Which leads me to the second gap.
2) if `thd_destroy` gets called on a thread that is waiting on an object, then there is no mechanism that removes it from the genwait queue on destruction. So the next time that object gets woken, the stored thread will be popped out of the queue and set to running despite already being destroyed.

I think these were both legitimate, but it's really gotten my mind in a knot so any feedback would be appreciated.

kernel/thread/worker.c
```diff
@@ -29,7 +29,7 @@ static void *thd_worker_thread(void *d) {
     for (;;) {
         flags = irq_disable();
 
-        if (!worker->pending)
+        if ((!worker->pending) && (!worker->quit))
```
QuzarDC, line 32: Documenting the case, since I won't remember later:
1A. The worker was woken and has started it's routine, setting worker->pending to false
2B. A thread/interrupt calls `thd_worker_destroy(worker), setting worker->quit to true (and genwait_wake_one, which fails as the worker isn't waiting)
3A. Worker A finishes its routine, looping, and reaches if !pending so it genwaits and deadlocks.

QuzarDC, line 32: Documenting the case, since I won't remember later:
1A. The worker was woken and has started it's routine, setting `worker->pending` to false
2B. A thread/interrupt calls `thd_worker_destroy(worker)`:
    setting `worker->quit` to true (and `genwait_wake_one`, which fails as the worker isn't waiting)
    then `thd_join(worker->thd)` which falls through to `genwait_wait(worker->thd)`
3A. Worker A finishes its routine, looping, and reaches if !pending so it `genwait_wait(worker)` and deadlocks.

So, being signaled to quit must also prevent waiting for a new pending wake.

kernel/thread/thread.c
```diff
@@ -565,6 +565,11 @@ int thd_destroy(kthread_t *thd) {
        and unblock them. */
     genwait_wake_all(thd);
 
+    /* If this thread was waiting on anything, remove it from
+       the genwait queue. */
+    if(thd->wait_obj)
+        genwait_wake_thd(thd->wait_obj, thd, 0);
```
QuzarDC, line 571: 1A. A thread is made to wait on object Z.
2B. Thread B destroys thread A.
3B. Thread B tries to wake anything waiting on object Z. Thread A is still in the genwait queue for Z so when it is unqueued, an already-destroyed thread is added to the running queue and a bunch of use-after-free happens writing to update the thread.

kernel/thread/worker.c
```diff
@@ -89,13 +90,18 @@ void thd_worker_wakeup(kthread_worker_t *worker) {
 }
 
 void thd_worker_destroy(kthread_worker_t *worker) {
+    uint32_t flags;
+
     assert(worker != NULL);
 
+    flags = irq_disable();
+
```
QuzarDC, line 98: Without IRQ disabled here the following sequence can happen:
1A. Thread A calls `thd_worker_destroy(B)`. Sets `worker->quit = true`
2B. Worker B was awake already, and checks if `worker->quit == true` then breaks and frees itself.
3A. Thread A now continues on to try signaling worker B with `genwait_wake_one(B)` which is a use-after-free.

An alternative course may be (and it might be required anyways but I haven't constructed a scenario yet for it) to expand the irq disable in the worker thread to encompass the check against worker->quit, ensuring that a woken worker always checks to quit before moving on. This approach would also favor acting on the wake rather than the destroy effectively interrupting the attempt to wake.

kernel/thread/worker.c
```diff
@@ -89,13 +90,20 @@ void thd_worker_wakeup(kthread_worker_t *worker) {
...
     worker->quit = true;
     genwait_wake_one(worker);
+    thd = worker->thd;
 
-    thd_join(worker->thd, NULL);
-    free(worker);
```
QuzarDC, line 98: When a worker is destroyed, `thd_worker_destroy` attempts to wake the worker then `thd_join` it. In the case of a detached thread or trying to join oneself (worker's routine calls `thd_shutdown`), the `thd_join` will fail and we may then call the `free(worker)` before the worker's routine has started. This would lead to using worker after free.

kernel/thread/worker.c
```diff
@@ -89,13 +90,20 @@ void thd_worker_wakeup(kthread_worker_t *worker) {
...
+    uint32_t flags;
+    kthread_t *thd;
+
     assert(worker != NULL);
 
+    flags = irq_disable();
+
     worker->quit = true;
     genwait_wake_one(worker);
+    thd = worker->thd;
```
QuzarDC, line 102: We are caching this to avoid the chance that once irqs are reenabled the worker thread finishes all on its own and worker is now freed before calling `thd_join`

kernel/thread/thread.c
```diff
@@ -838,6 +843,10 @@ int thd_join(kthread_t *thd, void **value_ptr) {
     if(thd == NULL)
         return -1;
 
+    /* If you wait for yourself, you'll never leave */
+    if(thd == thd_current)
+        return -4;
```
QuzarDC, line 848: This allows the graceful cleanup of `thd_join`ing threads that the system is using without having to worry if that thread itself was what requested the shutdown. I think generally as well, it's a good idea to avoid possible deadlocks.

At some point we likely need to refactor the exit paths in order for them to be assured to run in one of the system threads.

kernel/thread/thread.c
QuzarDC, line 571: If a user were to call `thd_destroy` on a thread that is waiting, it would otherwise not have been removed from the waiting queue and be returned as freed memory onto the runnable list. This also ensures that the genwait list is left clean after `thd_shutdown`

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -231,10 +231,15 @@ void  __weak arch_auto_shutdown(void) {
...
     snd_shutdown();
-    timer_shutdown();
     hardware_shutdown();
+    /* XXX: We should investigate shrinking this irq_disabled
+       time. Until then, all these shut downs happen with
+       irqs disabled which prevents things like safely joining
+       threads or sending cleanup commands to hardware.
+    */
+    irq_disable();
+    timer_shutdown();
```
QuzarDC, line 242: As in the note, shrinking this irq_disabled to ensure our shutdowns are not limited by that is valuable in itself. Here though, we could not properly shut down threads for the maple callbacks using `thd_join` with irqs disabled. Doing so would just send the callback threads into a genwait_wait that it would never return from (as there was no more context switching). `timer_shutdown` had to be moved for the same reason, as turning off the timers would also prevent context switching and cause a `genwait_wait` to be an infinite loop.

None of the other shutdowns in `hardware_shutdown` should be impacted, as most do little or nothing. Also `net_shutdown` was already happening prior to the irq disable in order to be able to clean up the net thread.

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -8,11 +8,15 @@
 #include <arch/arch.h>
 #include <dc/maple.h>
 #include <dc/maple/controller.h>
-#include <string.h>
+#include <kos/mutex.h>
+#include <kos/worker_thread.h>
```
pold500, line 12: I wonder if we can use standard C17 threads instead of KOS threads?

pcercuei: This is inside KOS though, so it makes sense to use KOS threads.

C11/C17 threads are implemented on top of KOS' threads so it would be adding a layer of indirection for no real reason.

gyrovorbis: Yeah, we are the ones providing C17 threads... using KOS threads.

kernel/arch/dreamcast/hardware/maple/controller.c
```diff
@@ -25,9 +29,21 @@ typedef struct cont_cond {
     uint8_t joy2y;     /* second joystick Y */
 } cont_cond_t;
 
-static cont_btn_callback_t btn_callback = NULL;
-static uint8_t btn_callback_addr = 0;
-static uint32_t btn_callback_btns = 0;
+typedef struct cont_callback_params {
```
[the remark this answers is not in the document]
gyrovorbis: The problem with that is that the entire user-facing API was written prefixing everything with `cont_` and `CONT_`, so this would be a whole breaking API change... Something more to note for 3.0 with the other API breaks we want to make, if that's what we want to do.

gyrovorbis (approved):
Alrighty. I've thoroughly reviewed it and have personally played with it now, doing a bunch of questionable shit like complex maple stuff, printf()-ing, sleep()-ing, and other gross things that should never have worked with the previous mechanism, and I'm satisfied that it works.

The only other thing is what we discussed in the Discord server yesterday: exiting from that callback thread is not safe. Not because of anything going on in this PR, but because it has literally never been safe... and we can address this in a follow-up PR.

PR #609, opened 2024-06-05
Improve environ.sh handling of built-in flag for fast math

darcagn:
This simply explicitly enables builtins if the fast math section is enabled in the environ.sh.
Before the user would need to manually uncomment/comment the `-fno-builtin` line.


gyrovorbis (approved):
Good call, actually. It's good to have these toggles be independent of one-another.

PR #611, opened 2024-06-05
pthread: Add dummy pthread_atfork() for CMake detection

pcercuei:
Add a stub for pthread_atfork(), whose purpose is only for CMake's FindThreads to detect threading support in KallistiOS properly.


gyrovorbis (approved):
Going to go ahead and approve of this, since it looks like Crabby's new pthreads will not be a part of the v2.1.0 release, and this little change single-handedly fixes the "pthreads don't work with freaking CMake" pain-point, making the current implementation much more useful while we wait for Crabby's new one to be ready for the next release.

PR #613, opened 2024-06-06
Add DC_TOOLS_BASE env variable for external tools

darcagn:
Somehow, despite adding this path to the `PATH` env variable and installing things to it (`mrbc` in kos-ports, `dc-tool-ip`, etc.), we don't actually have a defined environment variable for the path where tools external to KOS, but installed and used by KOS, are supposed to be installed. 

This has also caused some issues for me, for example in `mruby`, the directory where `mrbc` is to be installed is defined as so:
```make
MRUBY_INSTALL_DIR = $(KOS_CC_BASE)/../bin
```
which means when I do custom compiler work and install toolchains to weird paths and build kos-ports, I end up later finding `bin/mrbc` littered around my Dreamcast tools. 😆 

This PR adds `DC_TOOLS_BASE` to the `environ.sh` sample and better places where the paths are checked and added. I chose `DC_` instead of `KOS_` as a prefix since these tools are not actually part of KOS and are not KOS-specific (similar to how our ARM toolchain is not patched specifically for KOS and is `DC_ARM_BASE`).

A follow-up PR will come for the `mrubyc` kos-port.

### DIFF
--- a/environ_dreamcast.sh
+++ b/environ_dreamcast.sh
@@ -1,6 +1,16 @@
 # KallistiOS environment variable settings. These are the shared pieces
 # for the Dreamcast(tm) platform.
 
+# Add the default external DC tools path if it isn't already set.
+if [ -z "${DC_TOOLS_BASE}" ] ; then
+    export DC_TOOLS_BASE="/opt/toolchains/dc/bin"
+fi
+
+# Add the external DC tools dir to the path if it is not already.
+if [[ ":$PATH:" != *":${DC_TOOLS_BASE}:"* ]]; then
+  export PATH="${PATH}:${DC_TOOLS_BASE}"
+fi
+
 # Default the SH4 floating point precision if it isn't already set.
 if [ -z "${KOS_SH4_PRECISION}" ] ; then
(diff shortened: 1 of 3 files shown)


environ_dreamcast.sh
```diff
@@ -1,6 +1,16 @@
 # KallistiOS environment variable settings. These are the shared pieces
 # for the Dreamcast(tm) platform.
 
+# Add the default external DC tools path if it isn't already set.
+if [ -z "${DC_TOOLS_BASE}" ] ; then
+    export DC_TOOLS_BASE="/opt/toolchains/dc/bin"
```
ljsebald, line 6: I'd rather we didn't put any absolute paths in here... Since this is a dc-specific thing as of now and this is the assumed default at this point, it should be fine to do it this way as a default.

```suggestion
    export DC_TOOLS_BASE="${KOS_CC_BASE}/../bin"
```

ljsebald (dismissed):

ljsebald (approved):

sizious:
Thank you for that, I completely forgot this stuff, I was indeed thinking the same as you (having a specific variable for that kind of tools)!

PR #406, opened 2023-12-05
Faster IRQ context saving / restore

pcercuei:
Works with 2ndmix, which means it totally works for everything and everybody, of course :)

PR #614, opened 2024-06-06
Improve diagnose message

andressbarajas:
1. We didnt handle invalid PC and PR addresses. We do now by excluding them from the list.
2. Use improved address check to make sure addresses are in the text section of program.
3. Impoved spacing by adding a newline before "Encountered ..."
4. your_program.elf => prog.elf because its easier to edit after copying.
5. Add options to demangle C++ symbols, show function names, print inlined function calls

### DIFF
--- a/kernel/arch/dreamcast/include/arch/arch.h
+++ b/kernel/arch/dreamcast/include/arch/arch.h
@@ -40,6 +40,10 @@ extern uint32 _arch_mem_top;
 #define _arch_mem_top   ((uint32) 0x8d000000)
 #endif
 
+/** \brief  Start and End address for .text portion of program. */
+extern char _executable_start;
+extern char _etext; 
+
 #define PAGESIZE        4096            /**< \brief Page size (for MMU) */
 #define PAGESIZE_BITS   12              /**< \brief Bits for page size */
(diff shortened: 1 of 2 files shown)


gyrovorbis:
I'll see if I can play with this in the next couple days to move this review along!

gyrovorbis (approved):
I'm pretty sure the logic for `arch_stk_trace_at()` is incorrect for traversing the stack, as I can't seem to get an actual correct stack trace from anything... Just 1 layer deep with an invalid return address. Neither can the people I'm talking to...

That being said, has nothing to do with these changes, which do work quite well. We'll have to address that in a follow-up ticket. Approved.

QuzarDC:
> I'm pretty sure the logic for `arch_stk_trace_at()` is incorrect for traversing the stack, as I can't seem to get an actual correct stack trace from anything... Just 1 layer deep with an invalid return address. Neither can the people I'm talking to...
> 
> That being said, has nothing to do with these changes, which do work quite well. We'll have to address that in a follow-up ticket. Approved.

That's kind of where I ended up at the last time I tried to dig into it. I had suspected #406 as that shuffled around the context saving, mostly because it seems like that should have needed to cause some changes to the `arch_stk_trace_at` define and friends.

PR #615, opened 2024-06-06
Fix building libstunami examples on GCC9&10

cepawiel:
libtsunami examples fail to build on GCC9&10 due to same issue fixed in KallistiOS/kos-ports/pull/71

### DIFF
--- a/examples/dreamcast/tsunami/banner/Makefile
+++ b/examples/dreamcast/tsunami/banner/Makefile
@@ -7,6 +7,10 @@ TARGET = banner.elf
 OBJS = banner.o romdisk.o
 KOS_ROMDISK_DIR = romdisk
 
+# GCC versions prior to 11.0 have incomplete C++17 support and need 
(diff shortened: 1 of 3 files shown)


gyrovorbis (approved):
Good call.

QuzarDC:
But isn't that already the default in the repo? Do the default flags not get applied for these examples?

https://github.com/KallistiOS/KallistiOS/blob/f92103258b6d0943bc657de1199eb9a5d8b4087d/environ_base.sh#L46-L47

gyrovorbis:
> But isn't that already the default in the repo? Do the default flags not get applied for these examples?
> 
> https://github.com/KallistiOS/KallistiOS/blob/f92103258b6d0943bc657de1199eb9a5d8b4087d/environ_base.sh#L46-L47

NOPE. That's just internal for KOS. I had the same thought as well until @cepawiel showed me where they actually get used.

PR #618, opened 2024-06-08
IRQ Handler Propagation Chain, Partial Nesting, Example/Test, Documentation, and Cleanup

gyrovorbis:
This is a pretty massive PR that I worked my ass off on after taking a deep dive into `irq.h` and `irq.c` and noticing several things that seemed janky. I realized that with a few slight changes to the core IRQ handler and a few additions to the API, we could have an extremely robust IRQ handling system with multiple extension points. I then realized with a tiny bit of work, we could get the foundation laid for supporting interrupt nesting later on. While I was at it, I threw in some micro-optimizations such as getting rid of unnecessary conditionals plus throwing in some `__likely` and `__unlikely` compiler directives. 

I worked extremely hard to not break anything in the API or change any behavior of any existing IRQs. I tested the hell out of it with every example that I thought would thoroughly flex it plus other apps I had locally. I also profiled the crap out of it and saw no change or a teensy improvement. Furthermore I documented the living hell out of the entire IRQ system and made what I deem to be an extremely strong example and verification scenario for the IRQ handler chain that also offers a very useful general IRQ handling mechanism to C++ developers which they can use in their projects.

- [x] Cleaned up API and made everything proper, idiomatic C: `irq_t` is now an `enum`, proper `<stdint.h>` types are used, `<stdbool.h>` is used, etc.
- [x] Cleaned up register accesses in `irq.c` to use the same paradigm as the other drivers.
- [x] Introduced new API additions for retrieving the active interrupt, checking whether it has handled the IRQ or not, or signaling an IRQ to be handled.
- [x] Made `irq_handle_exception()` check whether the exception has been handled or not before propagating it to next one in the chain. 
- [x] Made `EXC_UNHANDLED_EXC` actually recoverable now, so that it can be used as a catch-all or even a way to show a fancy exception screen rather than just causing an immediate kernel panic and aborting the program.
- [x] Introduced new internal structure, `irq_state`, which forms a stack-based linked list of its instances which get pushed onto the stack every time an IRQ is entered and popped from the stack every time one is exited. This construct gives us the C-side basis for supporting generalized nested interrupts. In the meantime, before the remainder of the work is done on the ASM side, at least an `EXC_DOUBLE_FAULT` exception can now query for two layers worth of exception information in its handler (although its still not recoverable since there's only one IRQ context instance). 
- [x] Micro-optimized central IRQ entry point/dispatcher: `irq_handle_exception()`
- [x] Broke out `TRAPA` sub-API into its own dedicated `trap.h` header file which I intend to go to town on fleshing out similarly in a follow-up PR.
- [x] Created C++23-based example and verification scenario to test out the whole IRQ dispatcher and handler propagation scheme,
- [x] Profiled with 3rdmix, Twada's bloom demo, pvrmark_strips_direct.
- [X] Redid Doxygen for entire API.

Here's a look at the new Doxygen:
image

Here's a screenshot of the test scenario:
Screenshot from 2024-06-07 16-43-34

Here's a screenshot of `pvrmark_strips_direct` pushing over 3.04 million polygons/sec even with the new IRQ handler features, which is the same, or slightly better performance than we were seeing with the less flexible system in place:
Screenshot from 2024-06-08 00-33-11


kernel/arch/dreamcast/include/arch/trap.h
```diff
@@ -0,0 +1,103 @@
...
+*/
+trapa_handler trapa_get_handler(trapa_t trap, void **data);
+
+/** @} */
+
+/** @} */
+
+__END_DECLS
+
+#endif
```
andressbarajas, line 103:
```suggestion
#endif

```

gyrovorbis:
Alrighty, updated and rebased onto `master`, ready for `v2.2.0`. I know this is a big-ass scary change nobody asked for (yet), but it's been doing its thing in `dev`, has been tested with KOS for awhile now, and is doing fine with GTA3... She's feeling pretty solid. Also seeing zero effect in `pvrmark_strips_direct` benchmarks, so I do believe that despite having added more functionality, flexibility, and complexity to this whole mechanism, the thing that I optimized and cleaned up have canceled out any performance overhead.

QuzarDC:
Now @gyrovorbis whenever you could rebase this one I'll take a look :) It should be less than half the number of lines of changes and they should pretty much all be the new functionality.

As I noted on the discord, if the return of `irq_inside_int` is changed, make sure to propagate the new intended behavior out to places like this that were using the return:
https://github.com/KallistiOS/KallistiOS/blob/45525d79ba9ec9a3dd17f45a8ed3c4cf8bf77053/kernel/thread/mutex.c#L98-L105
It's fairly helpful for diagnostics.

gyrovorbis:
> Now @gyrovorbis whenever you could rebase this one I'll take a look :) It should be less than half the number of lines of changes and they should pretty much all be the new functionality.
> 
> As I noted on the discord, if the return of `irq_inside_int` is changed, make sure to propagate the new intended behavior out to places like this that were using the return:
> 
> https://github.com/KallistiOS/KallistiOS/blob/45525d79ba9ec9a3dd17f45a8ed3c4cf8bf77053/kernel/thread/mutex.c#L98-L105
> 
> 
> It's fairly helpful for diagnostics.

OH SHIT. Good call! I didn't even notice that anything was reaching into the return value ,which is now gone since I'm returning a `bool`! Updated to use the new API to just grab the active IRQ code.

kernel/arch/dreamcast/include/arch/irq.h
```diff
@@ -178,7 +201,7 @@ void irq_create_context(irq_context_t *context, uint32_t stack_pointer,
 
    |Type    | Description
    |--------|------------
-   |`RESET` | Caused by system reset. Uncatchable and fatal. Automatically branch to address `0xA0000000`.
+   |`RESET` | Caused by system reset. Uncatchable and fatal. Automatically branch to address `0xA0000000`.
```
QuzarDC, line 204: Extra trailing space.

kernel/arch/dreamcast/include/arch/irq.h
```diff
@@ -401,7 +559,7 @@ typedef void (*irq_handler)(irq_t code, irq_context_t *context, void *data);
     \param  data            A pointer that will be passed along to the callback.
     
     \retval 0               On success.
-    \retval -1              If the code is invalid.
+    \retval -1              If the source is invalid.
```
QuzarDC, line 562: This should be consistent across the handler set/get/type. Here you changed the comment to note `source` but didn't change the function's param to be `source` and in set handler it's `code` in the doc but `source` in the param.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -69,13 +149,16 @@ int irq_set_handler(irq_t code, irq_handler hnd, void *data) {
 }
 
 /* Get the address of the current handler */
-irq_handler irq_get_handler(irq_t code) {
+irq_handler irq_get_handler(irq_t code, void **data) {
     /* Make sure they don't do something crackheaded */
     if(code >= 0x1000 || (code & 0x000f))
         return NULL;
-
+
```
QuzarDC, line 156: ws

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -326,7 +403,7 @@ irq_context_t *irq_get_context(void) {
 void irq_create_context(irq_context_t *context, uint32_t stkpntr,
                         uint32_t routine, const uint32_t *args, bool usermode) {
     /* Clear out all registers. */
-    memset(context, 0, sizeof(irq_context_t));
+    bzero(context, sizeof(irq_context_t));
```
QuzarDC, line 406: Hadn't you ended up at the conclusion that bzero wasn't an improvement over memset 0 after all? This isn't used in a critical path at all right? Unless it was and there was a huge gain from bzero, I'd suggest against it.

Basically the same for the bzeros in irq_init, but init for sure isn't critical.

QuzarDC (changes requested):
A quick preliminary, mostly just whitespace and some naming. Then bzeros.

My apologies for the trouble of splitting this up, but now it looks *much* more digestible and I'll be able to get to a thorough review soon.

kernel/arch/dreamcast/kernel/irq.c
QuzarDC: For posterity found the comment: https://discord.com/channels/488332893324836864/614912396460556298/1249539168770592849 Apparently despite appearing more optimized the outputted asm is just worse.

With the bzero's gone I'm pretty sure the strings include can also be removed.

kernel/arch/dreamcast/include/arch/irq.h
```diff
@@ -15,426 +17,628 @@
...
-/** \defgroup irq_completion_codes  Completion Type
-    \brief                          IRQ completion type codes
+/** @} */
 
-    These exceptions are actually handled in-between instructions, allowing the
-    instruction that causes them to finish completely. The saved PC thus is the
-    value of the next instruction.
+/** Interrupt exception codes
+   
+   Dreamcast-specific exception codes. Used to identify the source or type of
```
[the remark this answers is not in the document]
gyrovorbis: LOL. U right! Changed that one.

I think I just got rid of the whitespace...thx.

kernel/arch/dreamcast/include/arch/irq.h
gyrovorbis: Shit. Looks like it should've stayed `code`. Thanks.

gyrovorbis:
> A quick preliminary, mostly just whitespace and some naming. Then bzeros.
> 
> My apologies for the trouble of splitting this up, but now it looks _much_ more digestible and I'll be able to get to a thorough review soon.

No need to apologize for breaking it up. Totally get it, and the majority of the time between breaking it up and now has been with the ball in my court. lol.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -38,22 +38,102 @@ struct trapa_cb {
...
+
+    const volatile struct irq_state *state = irq_state_n(level);
+    const irq_t irq = state? state->evt : 0;
+
+    irq_restore(imask);
+    return irq;
+}
+
+/* Have we handled the active interrupt? */
+bool irq_handled_int(size_t level) {
```
QuzarDC, line 122: When or how would this be used? It seems as though the dispatcher is manually checking the handled state at each layer, so it wouldn't be to know if you should chain then. Would it be something like a nested IRQ's handler determining how to behave based on if the handler it is interrupting has finished yet or not?

At first glance I thought 'sure, API, expose that' but I'm just not clear on how you'd use it. It doesn't seem (which was another first thought) that it is meant to allow multiple handlers to run on the same IRQ and behave differently if one before handled it, since the dispatcher gates propagation based on handling (or maybe I'm misunderstanding that?). Then I thought 'maybe for a handler to have multiple paths based on if the IRQ it's handling has been handled?' but it would be able to track that locally.

kernel/arch/dreamcast/include/arch/irq.h
```diff
@@ -255,60 +278,102 @@ typedef enum irq_exception {
...
+    \retval >0              Exception code for Nth IRQ context.
+    \retval 0               No IRQ context is active.
+
+    \sa irq_inside_int(), irq_int_depth()
+*/
+irq_t irq_active_int(size_t level);
+
+/** Returns whether the current IRQ has been handled.
+
+    Used to determine whether the active interrupt has been accepted by a
```
QuzarDC, line 330: Isn't it only the active interrupt when level=0?

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -352,6 +429,7 @@ static void irq_def_timer(irq_t src, irq_context_t *context, void *data) {
     (void)src;
     (void)context;
     (void)data;
```
QuzarDC, line 431: no longer needed with the use of it in timer_clear.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -381,15 +459,17 @@ int irq_init(void) {
...
 
     /* Default to not in an interrupt */
-    inside_int = 0;
+    irq_state_current = NULL;
 
-    /* Set a default timer handler */
+    /* Set default timer handlers */
     irq_set_handler(EXC_TMU0_TUNI0, irq_def_timer, (void *)0);
+    irq_set_handler(EXC_TMU1_TUNI1, irq_def_timer, (void *)1);
+    irq_set_handler(EXC_TMU2_TUNI2, irq_def_timer, (void *)2);
```
QuzarDC, line 472: Magic numbers?! `TMU0` `TMU1` `TMU2`

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -38,22 +38,102 @@ struct trapa_cb {
...
 /* Are we inside an interrupt? */
-static int inside_int;
-int irq_inside_int(void) {
-    return inside_int;
+bool irq_inside_int(void) {
+    return !!irq_state_current;
+}
+
+/* What's the active IRQ? */
+irq_t irq_active_int(size_t level) {
```
QuzarDC, line 111: Like what I mentioned elsewhere the name/comment seems incongruous with what it does. It would only be the active IRQ if level=0.

Maybe this is all because it's sort of been expanded to make way for the possibility of nesting and what's being described is what it can do now?

QuzarDC (commented):
Have some big questions about how some of the API works. Then some really minor whitespace/magic numbers/docs bits.

I don't see any issue at all with the current behavior as it would work now, but it seems some of the API that would be leveraged for handlers *might* just need minor adjustment to names/comments or such.

kernel/arch/dreamcast/kernel/irq.c
gyrovorbis: OH. Good call.

kernel/thread/thread.c
```diff
@@ -851,10 +851,9 @@ int thd_join(kthread_t *thd, void **value_ptr) {
     if(thd == thd_current)
         return -4;
 
-    if((rv = irq_inside_int())) {
+    if(irq_inside_int()) {
         dbglog(DBG_WARNING, "thd_join(%p) called inside an interrupt with "
-               "code: %x evt: %.4x\n", (void *)thd, ((rv >> 16) & 0xf),
-               (rv & 0xffff));
+               "code: [%x]\n", (void *)thd, irq_active_int(0));
         return -1;
```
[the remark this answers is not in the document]
gyrovorbis: > I was wrong on this (I think). Confused level and depth, which are the reverse of each other.
Ah yes, you are correct here. I've used the terms `depth` vs `level` for `irq_int_depth()` and `irq_active_int()` to draw that distinction.

I was also unsure about potentially losing the `evt` ID for the code, but... The truth is all you need is the code to fully identify an interrupt, because, assuming you need it, you can always look up the `evt` ID for it... Just wasn't sure whether it warranted an entirely new API with a new `evt_t` (to be similarly proper) to add such information, but... if you want, I can do it.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -201,90 +287,81 @@ static void irq_dump_regs(int code, irq_t evt) {
...
-        }
-        else {
-            timer_clear(TMU2);
+        if(hnd->hdl)
+            hnd->hdl(EXC_DOUBLE_FAULT, irq_srt_addr, hnd->data);
+        
+        /* Panic if it went unhandled. */
+        if(!irq_state.handled) {
+            irq_dump_regs(code, irq_state.evt);
+            arch_panic("double fault");
```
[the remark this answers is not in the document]
gyrovorbis: YES! Absolutely right!

kernel/thread/thread.c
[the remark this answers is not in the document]
gyrovorbis: OH. I did think that actually making that detail public would further add to the confusion of distinguishing between "general exceptions," and "interrupts," because then we're mixing the API which has opted for IRQ to be the universal term with the implementation detail of the SH4 docs, where that is not the case.

kernel/arch/dreamcast/include/arch/irq.h
gyrovorbis: Perhaps I misused the term `active`. In my head, in the case of multiple nested IRQs, I was thinking you'd consider each IRQ in this "stack" as *active*... 

And yeah, right now, it is only useful to ever call with a level of 0; however, the API was written to allow for nested interrupts in the future, so it's future-proof and won't have to break then.

kernel/arch/dreamcast/kernel/irq.c
gyrovorbis: Yeah, so the API is supposed to provide for the future nesting of interrupts so that there is potentially a *stack of active interrupts*. 

I think I agree, though, that maybe the comment there and on `irq_handled_int(size_t level)` and `irq_active_int(int level)` seem off, as if there is only ever a single one active, when the plan is to support multiple... updated.

kernel/arch/dreamcast/kernel/irq.c
gyrovorbis: So this was really one of those things where I do believe 99% of the time this function will not be useful; however, I do think it's worth having for the sake of API completeness, and because I do think there could eventually be a few very niche scenarios where the user may need to know whether the exception was accepted or rejected.

For instance, perhaps the "handler" thing they have is also propagating something through multiple functions or multiple regions of code which could accept or deny, and the top-level dispatcher on the user-side might need to know which state the IRQ is in: handled or not.

or perhaps whatever is installed to handle the exception is wanting to check whether the exception type is one which is handled by default or not, and do something different accordingly.

gyrovorbis:
> Have some big questions about how some of the API works. Then some really minor whitespace/magic numbers/docs bits.
> 
> I don't see any issue at all with the current behavior as it would work now, but it seems some of the API that would be leveraged for handlers _might_ just need minor adjustment to names/comments or such.

Okay, so I have gone through and finally addressed every single point of yours. I'm leaving a few of them unresolved to just ensure that you are in agreement with how I addressed them, because I don't want to be presumptuous. :)

The biggest and most important thing is the confusion regarding the API and documentation's vocabulary used to refer to exceptions/interrupts/events. I believe I have addressed it as well as we possible can without literally having to rewrite the API to name it `exc_xxx()` rather than `irq_xxx()` to be consistent with the manual.

What I did is write a big note in the top-level API docs stating explicitly the differences in nomenclature between this API and the SH4 documentation. I also went back through and made sure to very explicitly call out the couple of places where our usage of the terms "interrupt" or "exception" are not simply interchangeable and are referring to specific interrupt or general exception subtypes as they are classified in the SH4 manual.

Also retested and rebenchmarked to make sure I didn't do anything dumb. :)

kernel/arch/dreamcast/include/arch/irq.h
QuzarDC: Ah, ok, that makes sense. I was thinking 'active' as 'the one the handler is currently processing'. It makes perfect sense though to consider the whole stack as active. I have a feeling the term 'raised' would have a role here, but it makes enough sense as-is.

QuzarDC:
I was ready to merge and went back to retest the example and it seems that it just hangs for me now after:

`Testing accepting the exception in GLOBAL handler...
Caught exception: EXC_FPU from Global Handler!`

image


This might have to do with other PRs that went in the last update you made from `master` as I haven't modified my toolchain or build settings since the last time I tested. I tried to do a bit of debugging but the modernity of the C++ used makes my antique brain hurt. Does it work fine for you? Maybe your last run of sanity tests were on general stuff and not re-running your example?

I'll see if I can test this out of a different environment, to rule out things like gcc 15.0.0 being the cause.

QuzarDC:
Turns out my other system was on stable, so couldn't build the example for this. After rebuilding my whole environment to 14.2.0 I now get this:
image
And the example infinitely loops catching unknown exceptions.

Is it still working fine for you @gyrovorbis ?

pcercuei:
TBH this PR makes me uneasy. The idea that until now all IRQs handlers were assumed as running in atomic context, and this PR changes that, sounds like opening a big can of worms.

From a technical point of view it's very interesting, but I also don't understand why we'd want nested interrupts - I just don't see the benefits, but I can definitely see the drawbacks.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -351,7 +428,7 @@ void irq_create_context(irq_context_t *context, uint32_t stkpntr,
 static void irq_def_timer(irq_t src, irq_context_t *context, void *data) {
     (void)src;
     (void)context;
-    (void)data;
+    timer_clear((int)data);
```
QuzarDC, line 431: I was checking through again to see for any causes of the weird behavior and to see for other things that could be split out. The purpose of this was to replace the forced checking of the timer IRQs in the main handler right? Because it seems like this method is just strictly different from the original behavior. Before the timer was being cleared prior to its handler being called regardless of whether there was a handler or not. Now the handler itself is being made responsible for that clearing.

For our `TMU2` timer, that works out, saving us a clear every time as its handler is performing the clearing.

For `TMU1` we just never enable the interrupts for it, and if they were enabled then the old behavior would have messed with our ms spin timer anyways.

For `TMU0` it means that the last loop when the timer has completed doesn't have the timer be cleared, though it does get cleared on the next call through.

So it seems that it makes no difference for any of these, and it wasn't noted in any documentation that I can find that we by default clearing timers. This could make a difference for any software that was repurposing the timers and relying on the general handler to clear them, so despite reaching the end of my analysis on it and finding that it's pretty much fine, figured I'd make sure to note it in case it comes up (and in case I missed something ...)

It also occurs to me that we should probably have a minimal inner `void timer_clear` that just twiddles the register (or an outer one that returns the value) as none of our uses of it actually read the return and all simply use it for the purpose of clearing the underflow.

QuzarDC:
> TBH this PR makes me uneasy. The idea that until now all IRQs handlers were assumed as running in atomic context, and this PR changes that, sounds like opening a big can of worms.
> 
> From a technical point of view it's very interesting, but I also don't understand why we'd want nested interrupts - I just don't see the benefits, but I can definitely see the drawbacks.

I had been teetering on the edge of that same notion, but seemed to think of some very specific use being described in discord chat that I can no longer recall. With all the updates and cleanups already moved into the split off PRs, it may be that the rest of the API expansion is something that's simply not ripe without some specific use. Still pending would be some of the 'micro-optimizations' and I went and added some further comment on that.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -201,90 +287,81 @@ static void irq_dump_regs(int code, irq_t evt) {
...
     }
 
-    /* If there's a handler, call it */
-    {
-        hnd = &irq_handlers[evt >> 4];
-        if(hnd->hdl != NULL) {
-            hnd->hdl(evt, irq_srt_addr, hnd->data);
-            handled = 1;
+    /* If there's a global handler, it goes first */
+    if(__unlikely(global_irq_handler.hdl))
```
QuzarDC, line 333: It occurs to me that this is much more likely to be a compile-time thing. Either your software is going to be installing a handler (which is unlikely) or it just won't. Rather than 'most of the time irqs won't have a global handler', it should be 'most of the time no global handler will be installed'. 

Could this be done instead by having the global handler be a weak function (or similar method) that is simply optimized away if it is not ever set?

gyrovorbis:
We kind of half-assedly already did support multiple layers of IRQs--we just simply immediately shit our pants when the second IRQ was raised and were never even able to determine what the cause of the double-fault was within the double fault handler (which also was not recoverable and was bordering on useless).

Also by far the main point of this entire PR is the propagation chain and handler which has not been split off or replicated from this PR.

As it stands, our ability to handle and chain IRQs is extremely primitive and inflexible. What is the point of a "global handler," if a single callback must literally handle every single IRQ, including those for the TMUs and everything ever? Why can't I simply "observe" the IRQs and only handle those which I am interested in? As it stands a global handler is something that is so cumbersome and inflexible to use that it's basically a worthless construct.

Then there's zero way to actually create a generalized unhandled exception handler, since the entire system aborts regardless of whethere there's an undhandled exception that was just handled or not. Again, this construct and mechanism makes no sense and is worthless.

The 3-tiered IRQ propagation mechanism devised here provides the EXACT same default behavior as was there originally while simultaneously allowing you to passively "observe" (but not necessarily handle) interrupts (what if I just want to increment a counter or log something and not handle it)? Then it also allows you to implement a generic unhandled exception handler which could be something like drawing the exception to screen or writing it to a log or something.

kernel/arch/dreamcast/kernel/irq.c
gyrovorbis: YEP!!!! That's the exact approach I'm taking in the dca3-kos fork atm and am planning to PR back as soon as I've seen the codegen is optimal. This will also get rid of the branch check here, and I think the same should be done for the "unhandled exception handler" scenario as well. Will make this not only more flexible, but faster too!

QuzarDC:
> We kind of half-assedly already did support multiple layers of IRQs--we just simply immediately shit our pants when the second IRQ was raised and were never even able to determine what the cause of the double-fault was within the double fault handler (which also was not recoverable and was bordering on useless).
> 
Isn't that part entirely solved merely by having `inside_int` replaced by `irq_current`? Then during the double fault you have access to the whole previous irq state?

> Also by far the main point of this entire PR is the propagation chain and handler which has not been split off or replicated from this PR.
> 
Right, but the usefulness of that is exactly what is being called into question. What I was describing was to leave the storage of the one-deep in place for exactly what's described above without filling in an API that we don't have a use for yet.

> As it stands, our ability to handle and chain IRQs is extremely primitive and inflexible. What is the point of a "global handler," if a single callback must literally handle every single IRQ, including those for the TMUs and everything ever? Why can't I simply "observe" the IRQs and only handle those which I am interested in? As it stands a global handler is something that is so cumbersome and inflexible to use that it's basically a worthless construct.
> 
The contention is with the nesting aspect (handling an IRQ while already handling another) not the chaining (calling multiple handler functions to respond to a single irq). With just the updates to the `get_handlers` wouldn't this already be possible? 
```
irqX_handler, irqY_handler;
irqX_data, irqY_data;

my_XY_handler() {
  do_stuff;
  if(irqX && irqX_handler) irqX_handler(irqX_data);
  do_other_stuff;
  if(irqY && irqY_handler) irqY_handler(irqY_data);
}

install_my_handler() {
  irqX_handler = irq_get_handler(irqX, &irqX_data);
  irqY_handler = irq_get_handler(irqY, &irqY_data);
  irq_set_handler(irqX, my_XY_handler, NULL);
  irq_set_handler(irqY, my_XY_handler, NULL);
}
```

> Then there's zero way to actually create a generalized unhandled exception handler, since the entire system aborts regardless of whethere there's an undhandled exception that was just handled or not. Again, this construct and mechanism makes no sense and is worthless.
> 
If you wanted a generalized unhandled exception handler couldn't you do something like the below?

```
my_generalized_unhandled_handler;

install_my_handler() {   
   for(code = 0x000; code<0x100; code++) {
       if(!irq_get_handler(code<<4, NULL)
           irq_set_handler(my_generalized_unhandled_handler);
    }
}
```
   
   
> The 3-tiered IRQ propagation mechanism devised here provides the EXACT same default behavior as was there originally while simultaneously allowing you to passively "observe" (but not necessarily handle) interrupts (what if I just want to increment a counter or log something and not handle it)? Then it also allows you to implement a generic unhandled exception handler which could be something like drawing the exception to screen or writing it to a log or something.

Maybe I'm just not seeing it. How does the nesting facilitate the chaining? Am I way off in my examples above? Couldn't the passive 'observation' mode functionality be enabled *just* by removing the `handled=1` after the running of the global handler and having that one API component added?

gyrovorbis:
PR rebased, example fixed. Everything should be good to go!

pcercuei (changes requested):

kernel/arch/dreamcast/include/arch/irq.h
```diff
@@ -93,7 +126,8 @@ __BEGIN_DECLS
     The size of this structure should be less than or equal to the
     \ref REG_BYTE_CNT value.
 */
-typedef __attribute__((aligned(32))) struct irq_context {
+typedef struct irq_context {
+alignas(32)
```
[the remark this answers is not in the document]
QuzarDC: For ref #919 . When the toolchain imports `thread.h` to get the kos thread model stub, it brings in `irq.h` and for whatever reason that's part of the toolchain building is not C11 compatible.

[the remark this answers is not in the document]
gyrovorbis: Shiiiiiiit. Yep, this was an accident with the merge... Good catch.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -36,22 +36,102 @@ struct trapa_cb {
...
 };
 
+/* Linked list of IRQ states, one is pushed onto the stack
+   every time the top-level ISR is entered. */
+struct irq_state {            // SIZE
+    bool           handled;   // 1 byte  /* mov.b only has 15-byte displacement */
+    uint8_t        code;      // 1 byte
+    uint16_t       evt;       // 2 bytes
+    volatile struct
+        irq_state *previous;  // 4 bytes
```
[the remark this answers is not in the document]
gyrovorbis: Whoops. `irq_state_t`, didn't get a typedef until later on, and I forgot to update.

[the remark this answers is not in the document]
gyrovorbis: I didn't want to bother having to forward declare it just so I could pre-typedef in order to be able to use it within its own definition...  Will get rid of newline.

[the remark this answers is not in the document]
gyrovorbis: Well, actually, I didn't typedef it there because it was internal? I thought that's what you preferred? You kind of started convincing me... lol.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -199,73 +279,81 @@ static void irq_dump_regs(int code, irq_t evt) {
 /* The C-level routine that processes context switching and other
    types of interrupts. NOTE: We are running on the stack of the process
    that was interrupted! */
-volatile uint32_t jiffies = 0;
-void irq_handle_exception(int code) {
+__cold void irq_handle_exception(int code) {
```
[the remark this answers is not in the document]
gyrovorbis: Complete fuck-up. Was testing to see the impact on that keyword and forgot to swap it back to `__hot`. Thanks.

kernel/arch/dreamcast/include/arch/irq.h
```diff
@@ -265,45 +296,72 @@ typedef enum irq_exception {
...
 
-    Provides an API for accessing the state of the current IRQ context such
-    as the active interrupt or whether it has been handled.
+    \retval true               If interrupt handling is in progress.
+    \retval false              If normal processing is in progress.
 
-    @{
+    \sa irq_int_depth(), irq_active_int()
 */
+bool irq_inside_int(void);
```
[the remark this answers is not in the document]
gyrovorbis: Ah, yeah, fair.

QuzarDC:
@gyrovorbis  - I did the backend work to clean up and rebase this in the branch https://github.com/KallistiOS/KallistiOS/tree/irq_cleanup_final . If it looks good to you, you could go ahead and just rename it locally and force push to update this to that. Or a new PR could be opened. It was basically just down to the nesting.

QuzarDC:
Did one more round to get the whitespace cleanups and the one extra error check into master out of that `irq_cleanup_final`. The branch is now just 3 commits with the changes for IRQ nesting and the example.

Gonna go ahead and also move this to draft.

QuzarDC:
Closing this one as the vast majority of it was incorporated. There's still the [`irq_cleanup_final`](https://github.com/KallistiOS/KallistiOS/tree/irq_cleanup_final) branch that I made that has just the final update left of nesting, but that'd even need to be completely rewritten at this point to fit with the current codebase.

To make sure the ideas here don't just disappear (kind of the reason it's been left as draft) I've created a new label called `Revisit`.  I've added it on here and to others where we would want to give another try to in the future or that have enhancements that were never taken.

PR #619, opened 2024-06-08
Clean up CD-ROM code

pcercuei:
Change the macro lists to enums, and use enums as return codes / parameters for API functions.

Note the slight behavior change, `cdrom_change_datatype()` will no longer accept -1 as the "use default values" wildcard.

### DIFF
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -55,7 +55,7 @@ typedef int gdc_cmd_hnd_t;
 /* The G1 ATA access mutex */
 mutex_t _g1_ata_mutex = MUTEX_INITIALIZER;
 
-static int cdrom_get_track_type(void)
+static enum cd_track_type cdrom_get_track_type(void)
 {
     uint32_t params[4];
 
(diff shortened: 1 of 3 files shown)


kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -371,21 +374,21 @@ int cdrom_cdda_play(uint32 start, uint32 end, uint32 repeat, int mode) {
 }
 
 /* Pause CDDA audio playback */
-int cdrom_cdda_pause(void) {
+enum cd_cmd_response cdrom_cdda_pause(void) {
     int rv;
```
DC-SWAT, line 378: In this case, the variable "rv" can be made of this type.
I would also add typedef for enums to avoid using the word "enum" everywhere.

QuzarDC:
Assigned to myself to see about harmonizing some of the overlapping changes here and in #619 and #739. All great additions but would need some changes and don't want to shunt that effort to the submitters.

DC-SWAT:
Please add typedefs for enums.

QuzarDC:
> Please add typedefs for enums.

Out of curiosity, is that for style reasons, or is it because you have use cases where you'd want to be making typed vars to hold the same data (or some other reasons)?

I was likely to update these changes to do so, but was trying to determine how we use them in the rest of the codebase and if we'd have reason to not do so in specific cases, so am open to opinions on the matter.

DC-SWAT:
> > Please add typedefs for enums.
> 
> Out of curiosity, is that for style reasons, or is it because you have use cases where you'd want to be making typed vars to hold the same data (or some other reasons)?
> 
> I was likely to update these changes to do so, but was trying to determine how we use them in the rest of the codebase and if we'd have reason to not do so in specific cases, so am open to opinions on the matter.

I don't like to expand the type so much. Previously it was just "int", and now it's "enum some_long_name". It would be more convenient not to write at least enum like "some_long_name_t".
Besides, typedefs are used in many places. They are just mostly for structures, since instead of enums there were usually defines.

QuzarDC:
Closing this in favor of the methodology used in #778  of using typedef, propagating the types to the syscalls that own them, and preserving compatibility.

PR #621, opened 2024-06-09
Update startup

andressbarajas:
Each time startup.o is built we can copy it over so whenever there is an issue with startup.s, the toolchain doesnt have to be rebuilt to fix it.

### DIFF
--- a/kernel/arch/dreamcast/kernel/Makefile
+++ b/kernel/arch/dreamcast/kernel/Makefile
@@ -18,12 +18,12 @@ COPYOBJS += uname.o
 OBJS = $(COPYOBJS) startup.o
 SUBDIRS =
 
-CRT1_DIR = $(shell $(KOS_CC) $(KOS_SH4_PRECISION) -print-file-name=crt1.o | sed 's/crt1.o//')
+CRT1_PATH = $(shell $(KOS_CC) $(KOS_CFLAGS) -print-file-name=crt1.o)
 
 myall: $(OBJS)
 	-cp $(COPYOBJS) $(KOS_BASE)/kernel/build/
 	-rm banner.h authors.h banner.o uname.o
-	-cp startup.o $(CRT1_DIR)crt1.o
+	-cp startup.o $(CRT1_PATH)
 
 include $(KOS_BASE)/Makefile.prefab
 
--- a/kernel/arch/dreamcast/kernel/Makefile
+++ b/kernel/arch/dreamcast/kernel/Makefile
@@ -23,7 +23,7 @@ CRT1_DIR = $(shell $(KOS_CC) $(KOS_SH4_PRECISION) -print-file-name=crt1.o | sed 
 myall: $(OBJS)
 	-cp $(COPYOBJS) $(KOS_BASE)/kernel/build/
 	-rm banner.h authors.h banner.o uname.o
-	cp startup.o $(CRT1_DIR)crt1.o
+	-cp startup.o $(CRT1_DIR)crt1.o
 
 include $(KOS_BASE)/Makefile.prefab
 
--- a/kernel/arch/dreamcast/kernel/Makefile
+++ b/kernel/arch/dreamcast/kernel/Makefile
@@ -18,9 +18,12 @@ COPYOBJS += uname.o
 OBJS = $(COPYOBJS) startup.o
 SUBDIRS =
 
+CRT1_DIR = $(shell $(KOS_CC) $(KOS_SH4_PRECISION) -print-file-name=crt1.o | sed 's/crt1.o//')
+
 myall: $(OBJS)
 	-cp $(COPYOBJS) $(KOS_BASE)/kernel/build/
 	-rm banner.h authors.h banner.o uname.o
+	cp startup.o $(CRT1_DIR)crt1.o
 
 include $(KOS_BASE)/Makefile.prefab
 


kernel/arch/dreamcast/kernel/Makefile
```diff
@@ -18,9 +18,12 @@ COPYOBJS += uname.o
 OBJS = $(COPYOBJS) startup.o
 SUBDIRS =
 
+CRT1_DIR = $(shell $(KOS_CC) $(KOS_SH4_PRECISION) -print-file-name=crt1.o | sed 's/crt1.o//')
+
 myall: $(OBJS)
 	-cp $(COPYOBJS) $(KOS_BASE)/kernel/build/
 	-rm banner.h authors.h banner.o uname.o
+	-cp startup.o $(CRT1_DIR)crt1.o
```
darcagn, line 26:
```suggestion
CRT1_PATH = $(shell $(KOS_CC) $(KOS_CFLAGS) -print-file-name=crt1.o)

myall: $(OBJS)
	-cp $(COPYOBJS) $(KOS_BASE)/kernel/build/
	-rm banner.h authors.h banner.o uname.o
	-cp startup.o $(CRT1_PATH)
```

Why remove `crt1.o` with `sed` and then add it back on again?

`$(KOS_CFLAGS)` also would be better than using `$(KOS_SH4_PRECISION)`, just in case the user has other things in their toolchain setup that change the effective multilib path, like endianness.

darcagn (approved):

darcagn:
I'm approving this because I believe this should work correctly, and it would be nice to merge soon because we had a bug in `startup.s` that just had a fix merged and users now have to recompile the toolchain if we don't merge this. @ljsebald or @QuzarDC do either of you know if there'd be any negative consequence to this?

QuzarDC:
> I'm approving this because I believe this should work correctly, and it would be nice to merge soon because we had a bug in `startup.s` that just had a fix merged and users now have to recompile the toolchain if we don't merge this. @ljsebald or @QuzarDC do either of you know if there'd be any negative consequence to this?

No concerns on that part. If anything, this might also help startup behave better if someone is messing around with linker-related stuff. Just few nitty gritty bits:
 - Would any permissions be required for that copy to go through correctly (especially if it's overwriting an existing copy)? 
 - Does it always rebuild even if not changed?
 - Would/should it be removed by a clean or distclean?

gyrovorbis:
I also have some PRs I want to make for KOS 2.2.0 which will be easier with this... I added command-line argument propagation and passing to `arch_exec()` locally, but then abandoned it in a local branch when I realized the damn `startup.s` changes are toolchain-level and just didn't want to have to deal with it...

ljsebald:
> > I'm approving this because I believe this should work correctly, and it would be nice to merge soon because we had a bug in `startup.s` that just had a fix merged and users now have to recompile the toolchain if we don't merge this. @ljsebald or @QuzarDC do either of you know if there'd be any negative consequence to this?
> 
> No concerns on that part. If anything, this might also help startup behave better if someone is messing around with linker-related stuff. Just few nitty gritty bits:
> 
> * Would any permissions be required for that copy to go through correctly (especially if it's overwriting an existing copy)?
> * Does it always rebuild even if not changed?
> * Would/should it be removed by a clean or distclean?

The permissions thing is definitely a **big** potential issue here. There needs to be something done to deal with that before this can be merged, IMO. For reference, this would almost certainly fail to work on my setup, and probably the way that many Linux/BSD users have their toolchains set up as well (i.e, if you had to do `sudo` to install your toolchain, then this will almost certainly fail unless you were to run `sudo make` to build KOS -- which is something I always explain to my students that you should never have to do when building even something like the Linux kernel).

QuzarDC:
> > > I'm approving this because I believe this should work correctly, and it would be nice to merge soon because we had a bug in `startup.s` that just had a fix merged and users now have to recompile the toolchain if we don't merge this. @ljsebald or @QuzarDC do either of you know if there'd be any negative consequence to this?
> > 
> > 
> > No concerns on that part. If anything, this might also help startup behave better if someone is messing around with linker-related stuff. Just few nitty gritty bits:
> > 
> > * Would any permissions be required for that copy to go through correctly (especially if it's overwriting an existing copy)?
> > * Does it always rebuild even if not changed?
> > * Would/should it be removed by a clean or distclean?
> 
> The permissions thing is definitely a **big** potential issue here. There needs to be something done to deal with that before this can be merged, IMO. For reference, this would almost certainly fail to work on my setup, and probably the way that many Linux/BSD users have their toolchains set up as well (i.e, if you had to do `sudo` to install your toolchain, then this will almost certainly fail unless you were to run `sudo make` to build KOS -- which is something I always explain to my students that you should never have to do when building even something like the Linux kernel).

Perhaps it could be made fail-safe? That way it's merely strictly better than not having it, it will attempt to update if possible, but if not silently fail. Then expose something like `make startup` so that if it doesn't work automatically you can do the narrow `sudo make startup` just to patch that? And maybe even have the error just echo out a text saying that if you want that updated to call `sudo make startup`? I have almost no skill with makefiles so not sure if that's feasible.

This is mostly a concern for those of us working *on* KOS rather than *with* KOS. So the biggest problem would just be ensuring it doesn't cause a simple `make` to throw an error.

andressbarajas:
I believe it should silently fail because there is a hyphen in front of the cp.

darcagn:
> Perhaps it could be made fail-safe? That way it's merely strictly better than not having it, it will attempt to update if possible, but if not silently fail. Then expose something like `make startup` so that if it doesn't work automatically you can do the narrow `sudo make startup` just to patch that? And maybe even have the error just echo out a text saying that if you want that updated to call `sudo make startup`? I have almost no skill with makefiles so not sure if that's feasible.

I agree with this. Like @andressbarajas mentioned, the PR as-is will silently fail. The current behavior of 'never tell the user their crt1 is out of date' is far worse, and we can make an explicit 'update the crt1' target that users can invoke `sudo` upon if needed. One concern here is how we'll know the crt1 is outdated to tell the user they need to invoke the special target. Can we check the hash (i.e. are `startup.o`/`crt1.o` reproducible builds)?

The other thing is how this would affect Docker container setups where KOS and the toolchain are separated.

PR #622, opened 2024-06-09
Thread bugfixes: Stack ownership, Genwait object release upon death

gyrovorbis:
These are two really important bugfixes that I think need to get into `master` ASAP for the upcoming release. One was found by me, and one was found by @QuzarDC, who is currently busy with other concurrency stuff atm, so I figured I'd take care of it while I was handling the other bug...

In regards to the first bugfix: the kernel stack duplication... Upon pondering feedback from @QuzarDC and @ljsebald, who didn't like the user-facing flag for signalling to the kernel whether it is supposed to free the stack or not, and after studying how POSIX threads handle the issue, I came to agree with them... that shit's janky and unnecessary.

What we are doing instead is just utilizing one more bit in our thread's existing flags (no new storage anywhere) that gets set *internally* when the kernel allocates a new stack for a new thread.  Then upon destruction, if the flag is set, the kernel calls `free()` on the thread's stack pointer. With this new model, the user does nothing differently, and it's just assumed that when you provide your own stack to the kernel, since you've allocated it yourself, you're managing its lifetime yourself and are freeing it when the thread dies yourself. This seems to be how POSIX works, and it's simpler and more intuitive. 

1) Stack ownership
    - We were previously creating a thread for the main kernel thread
      using thd_create() which gave it a duplicated, default-initialized
32KB stack rather than having it utilize its static 64KB stack range.
    - To utilize its own stack, we have to internally maintain whether
      or not we are the owner of the stack (ie we allocated it), so that
we know not to free it.
    - This model also fixes issues created by supporting POSIX threads
      via the pthread API which assign their own stacks.
    - The new model/rule: you allocate and provide your own stack, you
      clean up your own stack too!
2) Genwait cleanup upon thread death
    - We were properly notifying other threads waiting upon the dying
      thread that they should wake up.
    - We were NOT properly cancelling any pending genwait objects that the
      dying thread was awaiting! So it could potentially later try to
wake up + notify a dead thread when the object becomes available.
    - We now remove the thread's genwait object (if it has one) upon its
      death.
      
 EDIT: Oh, shaved a few (like... barely anything) clock cycles off of some stuff by swapping from `memset(ptr, value, size)` to `bzero(ptr, size)`, which is like... molecularly faster (as verified via Compiler Explorer), because, at very least, a `value` argument is not present, not inspected, and is not wasting a register.
 
 Also cleaned up some comments a bit and added an actual `#define` for how large the kernel stack should be.

### DIFF
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -409,7 +409,7 @@ static void *thd_create_tls_data(void) {
     assert(!((uintptr_t)tcbhead % 8)); 
 
     /* Since we aren't using either member within it, zero out tcbhead. */
-    bzero(tcbhead, sizeof(tcbhead_t));
+    memset(tcbhead, 0, sizeof(tcbhead_t));
 
     /* Initialize .TDATA */
     if(tdata_size) { 
@@ -430,7 +430,7 @@ static void *thd_create_tls_data(void) {
         assert(!((uintptr_t)tbss_segment % tbss_align));
            
         /* Zero-initialize tbss_segment. */
-        bzero(tbss_segment, tbss_size);
+        memset(tbss_segment, 0, tbss_size);
     }
 
     /* Return segment head: this is what GBR points to. */
@@ -475,7 +475,7 @@ kthread_t *thd_create_ex(const kthread_attr_t *restrict attr,
 
         if(nt != NULL) {
             /* Clear out potentially unused stuff */
-            bzero(nt, sizeof(kthread_t));
+            memset(nt, 0, sizeof(kthread_t));
 
             /* Initialize the flags to defaults immediately. */
             nt->flags = THD_DEFAULTS;
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -40,8 +40,6 @@ also using their queue library verbatim (sys/queue.h).
 
 */
 
-#define THD_KERNEL_STACK_SIZE (64 * 1024)
-
 /* TLS Section ELF data - exported from linker script. */
 extern int _tdata_start, _tdata_size;
 extern int _tbss_size;
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -40,6 +40,8 @@ also using their queue library verbatim (sys/queue.h).
 
 */
 
+#define THD_KERNEL_STACK_SIZE (64 * 1024)
+
 /* TLS Section ELF data - exported from linker script. */
 extern int _tdata_start, _tdata_size;
 extern int _tbss_size;
@@ -407,7 +409,7 @@ static void *thd_create_tls_data(void) {
     assert(!((uintptr_t)tcbhead % 8)); 
 
     /* Since we aren't using either member within it, zero out tcbhead. */
-    memset(tcbhead, 0, sizeof(tcbhead_t));  
+    bzero(tcbhead, sizeof(tcbhead_t));
 
     /* Initialize .TDATA */
     if(tdata_size) { 
@@ -428,7 +430,7 @@ static void *thd_create_tls_data(void) {
         assert(!((uintptr_t)tbss_segment % tbss_align));
            
         /* Zero-initialize tbss_segment. */
-        memset(tbss_segment, 0, tbss_size);
+        bzero(tbss_segment, tbss_size);
     }
 
     /* Return segment head: this is what GBR points to. */
@@ -444,7 +446,7 @@ kthread_t *thd_create_ex(const kthread_attr_t *restrict attr,
     tid_t tid;
     uint32_t params[4];
     int oldirq = 0;
-    kthread_attr_t real_attr = { 0, THD_STACK_SIZE, NULL, PRIO_DEFAULT, NULL };
+    kthread_attr_t real_attr = { false, THD_STACK_SIZE, NULL, PRIO_DEFAULT, NULL };
 
     if(attr)
         real_attr = *attr;
@@ -473,7 +475,10 @@ kthread_t *thd_create_ex(const kthread_attr_t *restrict attr,
 
         if(nt != NULL) {
             /* Clear out potentially unused stuff */
-            memset(nt, 0, sizeof(kthread_t));
+            bzero(nt, sizeof(kthread_t));
+
+            /* Initialize the flags to defaults immediately. */
+            nt->flags = THD_DEFAULTS;
 
             /* Create a new thread stack */
             if(!real_attr.stack_ptr) {
@@ -484,6 +489,9 @@ kthread_t *thd_create_ex(const kthread_attr_t *restrict attr,
                     irq_restore(oldirq);
                     return NULL;
                 }
+
+                /* Since we allocated the stack, we own the stack! */
+                nt->flags |= THD_OWNS_STACK;
             }
             else {
                 nt->stack = (uint32_t*)real_attr.stack_ptr;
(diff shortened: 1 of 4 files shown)


QuzarDC:
Need to do a proper review but here are some partial notes:

It looks like `bzero` is generally deprecated: https://pubs.opengroup.org/onlinepubs/009695399/functions/bzero.html and it's suggested to replace calls to it with just the same memset 0. Not sure if that means that it's shaky to rely on. It's also not used anywhere else in the codebase. It seems perfectly fine, but not sure what the implications of that 'legacy' status are.

Did we have any reproducible issue being caused by the kernel stack swapping out like we were doing? I still have a poor time understanding why our creating a new stack and abandoning the startup one wasn't just wrecking things all over (other than perhaps just coincidence and general stability in the code that comes prior to threading being initialized).

To note for the genwait issue. I originally misunderstood the ability of that to be a problem, though it still seems like sensible cleanup to do. There are only 4 normal paths that allow `thd_destroy` to be called: reaper, detach, join, and shutdown.
1) In reaper, STATE_ZOMBIE must be set to get destroyed, which only gets set when a detached thread exits. So it couldn't have been gen_waiting.
2) In detach STATE_FINISHED must be set to get destroyed, which only gets set when a joinable thread exits. So it couldn't have been gen_waiting.
3) In join we only destroy after the thread wakes up from gen_wait, so it can't be waiting.
4) In shutdown, this is applicable, but we are also shutting down threading to never come back, so we can't have any kind of gen_waiting possible. (PS - since genwait and thd_shutdown are independent you could get into some shenanigans if you were to shutdown and init one or the other. It could be a good exercise to shut down then reinitialize threading to be able to shake out bugs and dependencies in its shutdown.)

What does make it sensible though is that `thd_destroy` is public. We should consider removing that from public access in the future. It's only current use in the codebase is to destroy the networking thread rather than joining to it if it is shut down in an interrupt.

gyrovorbis:
> Need to do a proper review but here are some partial notes:
> 
> It looks like `bzero` is generally deprecated: https://pubs.opengroup.org/onlinepubs/009695399/functions/bzero.html and it's suggested to replace calls to it with just the same memset 0. Not sure if that means that it's shaky to rely on. It's also not used anywhere else in the codebase. It seems perfectly fine, but not sure what the implications of that 'legacy' status are.
> 

I was about to argue that `bzero()` saves a cycle per word over `memset(,0,)`, but when I want to go make *damn sure* before making such a bold claim, I made a stark realization... 

Guess I'll post my stupidity here for others to see... At least at `-O3` there's ONE instruction difference.

`bzero()`:
Screenshot from 2024-06-09 20-32-25
`memset(,0,)`:
Screenshot from 2024-06-09 20-32-16
The key is... as the `nop` and `ldl.l` in `bzero()` is dual-issued, so too is the `mov` and `ldl.l` in `memset(,0,)`... meaning they're the same goddamn thing. I apologize for my fake news!

> Did we have any reproducible issue being caused by the kernel stack swapping out like we were doing? I still have a poor time understanding why our creating a new stack and abandoning the startup one wasn't just wrecking things all over (other than perhaps just coincidence and general stability in the code that comes prior to threading being initialized).
> 

Actually, the ONLY issue I can find that would've manifested from this mistake is that you'd be gated to a 32KB main thread stack, while you should've gotten 64KB, and the thread system would've enforced it upon context switch, so it would've been even detected properly as knowing the smaller boundary of the new stack.

There shouldn't actually be any other issue created by this obscure bug, which is why it went undetected for so long. The logic up until the first context switch would've happily been executing on the original 64KB stack, but then upon returning to the main thread, we would've swapped execution to the new 32KB stack and kept going along happily... Then you'd *think* we'd hit some issue returning to DC-Load or something with its initial stack frames, but nope. As far as I can tell, it would've just sat there preserved in the original 64KB stack, ready to go... Then the return address back to that stack is stored in ASM in `startup.s` and is jumped to directly when we exit `main()`, rather than a real function return, so I don't see how we would've encountered any real issue with this...

> To note for the genwait issue. I originally misunderstood the ability of that to be a problem, though it still seems like sensible cleanup to do. There are only 4 normal paths that allow `thd_destroy` to be called: reaper, detach, join, and shutdown.
> 
> 1. In reaper, STATE_ZOMBIE must be set to get destroyed, which only gets set when a detached thread exits. So it couldn't have been gen_waiting.
> 2. In detach STATE_FINISHED must be set to get destroyed, which only gets set when a joinable thread exits. So it couldn't have been gen_waiting.
> 3. In join we only destroy after the thread wakes up from gen_wait, so it can't be waiting.
> 4. In shutdown, this is applicable, but we are also shutting down threading to never come back, so we can't have any kind of gen_waiting possible. (PS - since genwait and thd_shutdown are independent you could get into some shenanigans if you were to shutdown and init one or the other. It could be a good exercise to shut down then reinitialize threading to be able to shake out bugs and dependencies in its shutdown.)
> 
> What does make it sensible though is that `thd_destroy` is public. We should consider removing that from public access in the future. It's only current use in the codebase (outside of threading internals) is to destroy the networking thread rather than joining to it if it is shut down in an interrupt.

1) I was actually unaware of the additional nuance behind the 4 points you've just enumerated, which are all really good. I just knew that, yeah, as you said, `thd_destroy()` could screw it up if the user called it... which brings me to my next point...

2) I think you're 100% exactly right. I didn't even think about it, but wtf thread system even has the user destroy it like that? That's what `thd_join` is for, or even when running detached, the reaper is the one to clean it up.

I think you're correct in that it should really be a private function 100% for sure. Does that need to happen right now for this release? No idea, might break the API. Dunno if we wanna do that yet or just take note... but I'd go so far as to say that *all* of the actual `kthread` struct itself should be private and moved to the source file, with only an opaque handle being declared in the public header.

Why? Because this is one of the best APIs in KOS in terms of having great accessors for managing everything you need. There's just not a good argument for having the user able to reach into the internals of a thread like that when we provide such a good API, and it could cause issues down the line when we want to change those internals... anyway, something to think about some day. :)

kernel/thread/thread.c
```diff
@@ -475,6 +477,9 @@ kthread_t *thd_create_ex(const kthread_attr_t *restrict attr,
             /* Clear out potentially unused stuff */
             memset(nt, 0, sizeof(kthread_t));
 
+            /* Initialize the flags to defaults immediately. */
+            nt->flags = THD_DEFAULTS;
+
```
QuzarDC, line 482: This is good to have, even though it's already implicitly set by the memset above it and `THD_DEFAULTS` being 0. Really helps keep the reasoning from getting obscured.

kernel/thread/thread.c
```diff
@@ -40,6 +40,8 @@ also using their queue library verbatim (sys/queue.h).
 
 */
 
+#define THD_KERNEL_STACK_SIZE (64 * 1024)
+
```
QuzarDC, line 44: This seems like it should be in some place where it can be shared with the use of the same value here: https://github.com/KallistiOS/KallistiOS/blob/395d99a9421aed367d7d277065080301c1a1d54d/kernel/arch/dreamcast/kernel/mm.c#L50 . To me it's arch specific and should live in the arch.h along with mem-top and such.

An alternative would be to say that KOS guarantees a 64k stack and we leave this in thread.h to share back with the arch-specific `mm`. I don't think that makes as much sense though as for something like GBA you'd likely want much less space (would be 1/5th the total ram).

QuzarDC (commented):
Looks good. Just had one bit that I think should get changed, but if you have good reason not to that could still work.

QuzarDC:
Tested it out in this form and saw it work fine. There's a small chance it corrected an error I had been testing in httpd but I'm not sure it was entirely reproducible beforehand.

Leaving this open, just to see if @ljsebald has commentary since he had reviewed the initial implementation via discord. I believe though that this addresses the feedback from that quite well.

QuzarDC (approved):

DC-SWAT:
Instead of `bzero` and `memset` can be used `memset4` here.

PR #623, opened 2024-06-10
RFC: Preliminary support for alternative memory managers

pcercuei:
This draft adds experimental support for using the TLSF memory manager (https://github.com/mattconte/tlsf) instead of the one included in KOS.

To test it:
  `make KOS_USE_TLSF=1`

This PR is not really meant to be merged as-is, its purpose is more to open a discussion about the design of KallistiOS, and to be able to compare the built-in memory manager with other implementations; especially since the built-in memory manager has the infamous reputation of being buggy.


gyrovorbis:
@DC-SWAT You are going to want to see this one for getting rid of your `malloc()`/`free()` issue.

DC-SWAT:
Good to see.
Also will be good if added backward compatibility for memalign.

gyrovorbis:
> Good to see. Also will be good if added backward compatibility for memalign.

Yeah, that's a good point... even though it's deprecated by `posix_memalign()`, it's still very widely used, and we should probably still support it by just having it call into `aligned_alloc()` or something...

pcercuei:
Well, `memalign()` didn't go anywhere, did it?

gyrovorbis:
> Well, `memalign()` didn't go anywhere, did it?

OH. I had just seen that you swapped it over to `aligned_alloc()` in several places and assumed that it was being implemented by the old allocator or something for some reason... but yeah, I guess it would be a Newlib thing. lmao.

Issue #543, opened 2024-05-04
Suggestion to change default debug log level


ppxxcc:
As seen in 

https://github.com/KallistiOS/KallistiOS/blob/master/kernel/libc/koslib/dbglog.c

```c
/* Default kernel debug log level: if a message has a level higher than this,
   it won't be shown. Set to DBG_DEAD to see basically nothing, and set to
   DBG_KDEBUG to see everything. DBG_INFO is generally a decent level. */
   
int dbglog_level = DBG_KDEBUG;
```

By default KOS will output every superfluous detail to stdout. Change video mode? It goes right to stdout.
```
vid_set_mode: 320x240 NTSC with 1 framebuffers.
pvr: enabling vertical scaling for non-VGA
```

I think this is a quite absurd default, how many other APIs have you used which prints internal details (not even errors, just regular information) to stdout by default? At least making `DBG_INFO` the default would be a much more sane default, if not `DBG_DEAD`.

gyrovorbis:
Just kidding. I actually agree with you. I also think a few other details, like "DMA buffer at: 0xblah" should also be hidden as verbose/debug-level information that has no relevance to the user during both startup and shutdown...

gyrovorbis:
While we're on the topic, is there anything useful you WANT to see? I'm kind of thinking there's a bunch of useful shit we're not logging for `DBG_NOTICE`, like when Maple devices are attached and removed, for example, which would actually be useful to log and capture...

QuzarDC:
I'd generally agree, we basically already have it:
https://github.com/KallistiOS/KallistiOS/blob/dbdc05cf9f64a32b679285c2bb5bb196546407d7/doc/environ.sh.sample#L112-L118

But I think there may be some happier medium between the default high debug we have now and no debug for retail purpose that I understand that define to be providing.

gyrovorbis:
> I'd generally agree, we basically already have it:
> 
> https://github.com/KallistiOS/KallistiOS/blob/dbdc05cf9f64a32b679285c2bb5bb196546407d7/doc/environ.sh.sample#L112-L118
> 
> But I think there may be some happier medium for a default between the default high debug we have now and no debug for retail purpose that I understand that define to be providing.

Oh, there we go... Should we be actually changing what the default log level gets initialized to in KOS for debug versus release builds? Because I don't think we're doing anything like that with that definition--it's just that `assert()` is honoring it... 

[...]

PR #626, opened 2024-06-11
Ignore `dbglog` with `NDEBUG` set and adjust default logging level

QuzarDC:
EDIT: Scope reduced to just address default logging level and NDEBUG condition (as that would be the most common usage of static levels anyways).
Intended to partially address #543 

Drop dbglog if `NDEBUG` is set and make `DBG_INFO` the default level, as it's noted to be 'generally good' and the higher levels give a lot of noise for typical users.

I would think with this new setup that the strings that are passed to `dbglog` would be optimized out with `NDEBUG` set, but that doesn't seem to be the case. Any suggestions on how to tweak it to better ensure that would be welcome. This should be able to significantly reduce the size of `NDEBUG` programs since we have so many `dbglog` messages. I might be related to it using vargs, or perhaps I'm just not testing correctly. The only difference I was able to see is the 1kb removal of the static buffer with `NDEBUG`

### DIFF
--- a/kernel/libc/koslib/dbglog.c
+++ b/kernel/libc/koslib/dbglog.c
@@ -15,6 +15,19 @@
 #include <kos/fs.h>
 #include <arch/spinlock.h>
 
+#ifdef NDEBUG
+
+/* Don't set the debug level, since we have no debug. */
+void dbglog_set_level(int) {
+    return;
+}
+
+void dbglog(int, const char *, ...) {
+    return;
+}
+
+#else   /* NDEBUG */
+
 /* Not re-entrant */
 static char printf_buf[1024];
 static spinlock_t mutex = SPINLOCK_INITIALIZER;
@@ -22,7 +35,7 @@ static spinlock_t mutex = SPINLOCK_INITIALIZER;
 /* Default kernel debug log level: if a message has a level higher than this,
    it won't be shown. Set to DBG_DEAD to see basically nothing, and set to
    DBG_KDEBUG to see everything. DBG_INFO is generally a decent level. */
-int dbglog_level = DBG_KDEBUG;
+int dbglog_level = DBG_INFO;
 
 /* Set debug level */
 void dbglog_set_level(int level) {
@@ -55,3 +68,5 @@ void dbglog(int level, const char *fmt, ...) {
     if(level >= DBG_ERROR && !irq_inside_int())
         spinlock_unlock(&mutex);
 }
+
+#endif  /* NDEBUG */


kernel/libc/koslib/dbglog.c
```diff
@@ -21,9 +21,14 @@ static spinlock_t mutex = SPINLOCK_INITIALIZER;
    DBG_KDEBUG to see everything. DBG_INFO is generally a decent level. */
 int dbglog_level = DBG_KDEBUG;
 
-/* Set debug level */
+/* Set debug level if DBG_DYNLOG is enabled */
 void dbglog_set_level(int level) {
+#ifdef DBG_DYNLOG
     dbglog_level = level;
+#else
+    dbgio_write_str("dbglog: Define DBG_DYNLOG to change log level.\n \
```
gyrovorbis, line 29:
```suggestion
#include <assert.h>

    static_assert(false, "dbglog: Define DBG_DYNLOG to change log level.\n \
```
Now that we're on a revision of C from this century, we can make checks like this compile-time. I would move it outside of the function too to global scope.

gyrovorbis:
GOOD ONE!

kernel/libc/koslib/dbglog.c
QuzarDC: Oh, sure. I was basically just having it as the softest warning possible but still building. The check could always be compile time and just using `#warning` too. Don't know if anyone's actually ever used the function anyways (maybe in lieu of figuring out the defines and setting it at the start of main).

ljsebald:
I would imagine that the majority of the time if people are looking for debug messages to be garbage collected at compile/link time, they're probably looking for all of them to be collected -- not just those above a certain level (i.e, making a release build where you really probably don't need debug logging at all).

In my opinion, essentially instead of doing this (which is an API change that I don't particularly agree with -- making the logging level static by default), the inverse logic should be true -- a macro set at compile time as a way to make the log level static (or to disable all logging all together), rather than to require a macro be set to follow the behavior that we already have.

pcercuei:
Not really fond of the hijack of the `NDEBUG` macro.
It already has a function, to disable asserts.
Here you are adding a new behaviour to what is a C11 / POSIX 2008 macro. Why not simply use a new macro?

QuzarDC:
> Not really fond of the hijack of the `NDEBUG` macro. It already has a function, to disable asserts. Here you are adding a new behaviour to what is a C11 / POSIX 2008 macro. Why not simply use a new macro?

I had no idea it was a standard define. It's used elsewhere in the codebase (though not much, mostly in comments) for the same purpose and described in our `environ.sh` sample as disabling 'additional debugging checks' as well as asserts:
https://github.com/KallistiOS/KallistiOS/blob/09b44b0d5e40e9302a71c9269e7df36c08377f1a/doc/environ.sh.sample#L121-L127

so I understood it to be for this purpose. Since that's not the case, this would be *way* better folded into the `INIT_QUIET` init flag, which already reports to disable dbgio (and thus also already disables high-level dbglog messages which use dbgio rather than stdout).

Issue #629, opened 2024-06-14
ping example leaks memory on shutdown


QuzarDC:
I tried following the calls back into the networking code but could not see where this was happening. This doesn't appear in the 'basic' test so it's not *just* from enabling networking, it must be something happening behind the icmp. Unfortunately the newlib calls that cause it are merely to `balloc` inside of stdlib/mprec so it could still be any number of stdlib functions that are behind it.

I don't think this should be blocking to v2.1.0, but found it during testing and wanted to make sure it got documented.

```KallistiOS Git revision v2.0.0-1433-g9b1cc962-dirty:
  Fri Jun 14 06:46:10 AM EDT 2024
  quzar@Centurion-IV.:/home/quzar/dcdev/KallistiOS
  sh-elf-gcc (GCC) 15.0.0 20240522 (experimental)
thd: pre-emption enabled, HZ=100
maple: active drivers:
    Dreameye (Camera): Camera
    Sound Input Peripheral: Microphone
    PuruPuru (Vibration) Pack: JumpPack
    VMU Driver: Clock, LCD, MemoryCard
    Mouse Driver: Mouse
    Keyboard Driver: Keyboard
    Controller Driver: Controller
    Lightgun: LightGun
  DMA Buffer at ac0a2be0
vid_set_mode: 640x480 VGA with 1 framebuffers.
lan: no device detected
dc-load console support enabled
maple: attached devices:
  A0: Dreamcast Controller           (01000000: Controller)
  A1: Visual Memory                  (0e000000: Clock, LCD, MemoryCard)
  A2: Puru Puru Pack                 (00010000: JumpPack)
  B0: Dreamcast Camera Flash  Device (01000000: Controller)
  B1: Dreamcast Camera Flash LDevice (00080000: Camera)
  B2: Dreamcast Camera Flash LDevice (00080000: Camera)
  B3: Dreamcast Camera Flash LDevice (00080000: Camera)
  B4: Dreamcast Camera Flash LDevice (00080000: Camera)
  B5: Dreamcast Camera Flash LDevice (00080000: Camera)
bba: MAC Address is 00:d0:f1:02:a5:a6
bba: link lost
bba: link stable
net_dev_init: detected 1 usable network device(s)

72 bytes from 192.168.1.1: icmp_seq=0 ttl=64 time=5.150 ms
72 bytes from 192.168.1.1: icmp_seq=1 ttl=64 time=2.854 ms
[...]

PR #630, opened 2024-06-14
Clean up after all the allocations done in serpent_dma

QuzarDC:
As on tin. We shouldn't offer examples that leak memory. The example still leaks memory, but in a similar way as described in #629 which may share a root cause and seems like it may be related to the use of printfs. Am adding the details to that issue rather than here.


gyrovorbis (approved):
Nice catch.

PR #631, opened 2024-06-14
Avoid returning error code when examples are too old for toolchain version

cepawiel:
Currently the following command will fail on older toolchain versions `cd $KOS_BASE/examples/dreamcast && make` as the error in `Makefile.rules` would be detected as a build failure and return a failure error code. This change makes the targets just print a message about skipping the build so the user is notified while allowing a success return code.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -91,3 +91,6 @@ $(shell \
 	}')
 endef
 
+define KOS_GCCVER_MIN_WARNING
+@echo "Skipping $(TARGET) build as current GCC version ($(KOS_GCCVER)) is less than $(KOS_GCCVER_MIN)."
+endef
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -76,20 +76,18 @@ romdisk.o: romdisk.img
 	rm romdisk_tmp.o
 endif
 
-# Define KOS_GCCVER_MIN in your Makefile if you want to enforce a minimum GCC version.
-ifdef KOS_GCCVER_MIN
-  ifeq ($(shell \
-    awk 'BEGIN { \
-        split("$(KOS_GCCVER_MIN)", min, "."); \
-        split("$(KOS_GCCVER)", cur, "."); \
-        if (cur[1] > min[1] || \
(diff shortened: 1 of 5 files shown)


QuzarDC:
Good idea. It might make sense to make the pattern for warning or the whole 'if block' be baked into the makefile rules as well though, if that's possible, rather than repeated in each makefile that needs it.

cepawiel:
I moved the warning message into `Makefile.rules` to avoid some amount of copy/pasting. I attempted to move more of the if block into the `Makefile.rules` but wasn't able to come up with any great solutions, maybe I'm just missing an obvious solution though.

gyrovorbis:
Anybody else know how to deduplicate some of the stuff in the individual Makefiles? Maybe a way to override/BS the `all` and other targets if the GCC_VERSION is less than expected? I'd like to see the logic centralized and less work having to be done in each individual `Makefile`.

That being said, I'm absolute shit at bash. I have no idea if it's even possible to deduplicate further, and if that's the case, that's fine with me.

Any ideas? @darcagn @andressbarajas @pcercuei ?

QuzarDC:
> Anybody else know how to deduplicate some of the stuff in the individual Makefiles? Maybe a way to override/BS the `all` and other targets if the GCC_VERSION is less than expected? I'd like to see the logic centralized and less work having to be done in each individual `Makefile`.
> 
> That being said, I'm absolute shit at bash. I have no idea if it's even possible to deduplicate further, and if that's the case, that's fine with me.
> 
> Any ideas? @darcagn @andressbarajas @pcercuei ?

Yea, and just to be clear, I am shit at it too and was suggesting it merely as a general observation not to say that there *is* a way to do so.

gyrovorbis (approved):
Alrighty, we polled the audience, and nobody knew how to do this any cleaner... Been holding up @cepawiel's CI stuff here long enough, so I'm going to merge... Obviously if anyone has a sudden stroke of Makefile/shell brilliance for how to clean it up, a follow-up PR is welcome! :rofl:

PR #634, opened 2024-06-17
Add . & .. to filesystems and filesystem example

andressbarajas:
I added support for .. folder in ramdisk, vmu, cd, pty, and dev.  I also added a filesystem example to test these changes.  The example needs more work so I will leave in draft for now. Feel free to still review though.


examples/dreamcast/filesystem/browse/browse.c
```diff
@@ -0,0 +1,379 @@
...
+}
+
+static int button_pressed(unsigned int current_buttons, unsigned int changed_buttons, unsigned int button) {
+    if (changed_buttons & button) {
+        if (current_buttons & button)
+            return 1;
+    }
+
+    return 0;
+}
```
QuzarDC, line 379:
```suggestion
static bool button_pressed(unsigned int current_buttons, unsigned int changed_buttons, unsigned int button) {
    if (changed_buttons & button) {
        if (current_buttons & button)
            return true;
    }

    return false;
}
```
or maybe a slightly less readable
```suggestion
static bool button_pressed(unsigned int current_buttons, unsigned int changed_buttons, unsigned int button) {
    return !!(changed_buttons & current_buttons & button);
}
```

examples/dreamcast/filesystem/browse/browse.c
```diff
@@ -0,0 +1,379 @@
...
+    int x = 20 + 24, y = 24;
+    int color = 1;
+    int count = 0;
+    char str[80];
+
+    bfont_set_foreground_color(0xFFFFFFFF);
+    bfont_set_background_color(0x00000000);
+    for(int i=0;i<num;i++) {
+        if(directory_contents[i].is_dir) {
+             sprintf(str, "%-40s%s", directory_contents[i].filename, "< DIR >");
```
QuzarDC, line 345:
```suggestion
             snprintf(str, "%-40s%s", strlen(str), directory_contents[i].filename, "< DIR >");
```

Otherwise long filenames will overrun the buffer.

examples/dreamcast/filesystem/browse/browse.c
```diff
@@ -0,0 +1,379 @@
...
+
+    if(highlight_yes) {
+        bfont_set_foreground_color(0x00000000);
+        bfont_set_background_color(0xFFFFFFFF);
+    }
+    else {
+        bfont_set_foreground_color(0xFFFFFFFF);
+        bfont_set_background_color(0x00000000);
+    }
+    y += 24;
```
QuzarDC, line 311:
```suggestion
    y += BFONT_HEIGHT;
```
Maybe you'd rather set as a local define (`LINE_STRIDE` or something?) if the name doesn't seem right, but it seems quite a bit in many uses of fixed numbers you could use the defines from bfont as they reflect the bfont height/width.

QuzarDC (commented):
I know this is still in draft, but figured I'd give a bit of feedback on the examples.

Also, is there a significant hindrance to having `.` and `..` implemented at the top level `fs_` stuff? Romdisk is the only one that requires it to be baked in because `genromfs` creates the files as symlinks inside of each dir. Perhaps a new bit in `vfs_handler_t::flags` to denote if the fs will handle dots on its own, if that was the issue.

andressbarajas:
Marking this as ready for review.  I think the SD issues are unrelated to my example or my changes.  Logged an issue here:  https://github.com/KallistiOS/KallistiOS/issues/635.  One more change I probably need to make is to mark fat and ext2fs also as providing the . and .. directories but I cant test that currently.

andressbarajas:
@QuzarDC I made updates to romdisk and dev:

1. Romdisk: Instead of checking for the type, we check if the name matches . or ... and did away with the hardlink check.
2. Dev: We only reset the position when it was the last reference.  This fixes the bug in GhettoPlay Vorbis where the dev directory contents(random, urandom, null) disappear for repeat visits to the dev directory.  They only show the first time.

gyrovorbis:
Guys, is this "v2.2.0 upcoming feature" material or is this "bugfix for v2.1.0" material? Asking so I know which milestone to tag it in and whether I should be testing/helping to review this one atm or not.

andressbarajas:
Moving this to draft. I'll pull out the dev and rd bug fixes to their own branches. Also I found some corner cases and I want to support those so more work needs to be done.

andressbarajas:
Okay, I moved all code changes related to other FS fixes to their own PRs.  Did away with the flags suggested earlier because there where edge cases that needed to be handled.  Now all logic is inside fs_readdir().

gyrovorbis:
Guys, is this slated for `v2.1.0`? I saw it as kind of a bug that the VFSes did not all have consistent `.` and `..` directories, but what do you guys think? If so, I can get this tested... @andressbarajas looks like you got a few pending comments, though.

andressbarajas:
@gyrovorbis @QuzarDC Any pending comments?

gyrovorbis:
> @gyrovorbis @QuzarDC Any pending comments?

Yeah, there's a couple of comments from me about missing copyright and header comments on I think 2 files... other than that, no, but I want to test the shit out of it.

QuzarDC:
Had gone over these changes before, and it looks fine. Just needs rebasing and the changelog adjusted to 2.2.0

andressbarajas:
@QuzarDC Done

QuzarDC (approved):

Issue #305, opened 2023-09-11
STDIO file redirection does not work


gyrovorbis:
A low priority bug was reported by Ian Michael which I wanted to capture, because it's a perfectly reasonable thing to want to do with KOS and is something that might be an easy fix that just never got looked into.

Anyway, standard C allows you to redirect the standard output file streams like so:
```
#include <stdio.h>

int main(int argc, char* argv[]) {
    // Open a file for writing 
    FILE *output_file = fopen("/pc/output.txt", "w");

    if (output_file == NULL) {
        perror("Error opening output file");
        return 1;
    }

    // Redirect stdout to the output file
    if (freopen("/pc/output.txt", "w", stdout) == NULL) {
        perror("Error redirecting stdout");
        return 1;
    }

    // Now, any text printed using printf will be written to the output file
    printf("This text will be written to the file.\n");
    printf("redirecting printf to a file.\n");

    // Close the output file to flush the data
    fclose(output_file);
    
    return 0;
}
```
However, it seems as though this fails to produce any output to "output.txt" with KOS, as somewhere down the code path stdout/stderr appears to be hardcoded.

QuzarDC:
This is explicitly noted as a TODO: https://github.com/KallistiOS/KallistiOS/blob/master/kernel/fs/fs.c#L367 . I had called it out in https://github.com/KallistiOS/KallistiOS/issues/224 as part of the reason to document fs_pty. To my estimation fs_pty would be better served as an addon, or to just be removed unless there's anyone out there who has ever used it (could never find evidence of that).

It has a separate issue in that /pty/ does not present a '.' in its dir structure, so if allowed in just an open file browser, you'd need to write in explicit support for returning to the previous directory.

pcercuei:
> It has a separate issue in that /pty/ does not present a '.' in its dir structure, so if allowed in just an open file browser, you'd need to write in explicit support for returning to the previous directory.

Honestly this is a problem with half of the VFS, that probably should be fixed.

gyrovorbis:
> > It has a separate issue in that /pty/ does not present a '.' in its dir structure, so if allowed in just an open file browser, you'd need to write in explicit support for returning to the previous directory.
> 
> Honestly this is a problem with half of the VFS, that probably should be fixed.

Agreed. I actually have it in my notes and on my todo list to add that to all of the roots.

tavianator:
[...]

Issue #224, opened 2023-05-28
Create example/test for fs_pty


QuzarDC:
As title. We technically support it, but I don't know if it's ever been used, or how it would be used. There is a reference to its use in [kernel/fs/fs.c 's fs_write](https://github.com/KallistiOS/KallistiOS/blob/master/kernel/fs/fs.c#L371) with a note to fix, and that may be a thing to do as well with an example of its use.

QuzarDC:
Thanks for taking this on @andressbarajas .

PR #636, opened 2024-06-19
Fix pty

andressbarajas:
Newlib likes fs_pty now.  Added dup, dup2 as wrappers around fs_dup & fs_dup2. Also added pipe() which is just kinda a wrapper around fs_pty_create.

Fixes these issues:  
1. https://github.com/KallistiOS/KallistiOS/issues/305
2. https://github.com/KallistiOS/KallistiOS/issues/224


QuzarDC (commented):
Looks great, and thank you for taking it up, it had been on my mind lately. It's amazing that the solution was basically 'get rid of all the hacks' and the rest behind it had already all been resolved.

I think that this might require https://github.com/KallistiOS/KallistiOS/blob/master/kernel/libc/newlib/newlib_isatty.c to be updated as well to reflect that fd 1/2 are no longer forced to be tty (I think it should be doable by just looking up the handler and test if it's the pty one).

And: 

https://github.com/KallistiOS/KallistiOS/blob/d035be9fd2f2d06e8bdf263e06bb6ff5966f424e/kernel/fs/fs.c#L213-L218
Might also need updating to no longer exclude them from cleanup, but I'm not certain. I added the exception because the function was never being called, but should have been part of shutdown. I think the reason I had to exclude them is because pty_shutdown doesn't attempt to just close the ptys, it just frees the fd table entries which leads to possible use after free.

Finally, it might also make sense to change [kernel/libc/koslib/dbglog.c ](https://github.com/KallistiOS/KallistiOS/blob/d035be9fd2f2d06e8bdf263e06bb6ff5966f424e/kernel/libc/koslib/dbglog.c#L46-48) back to using an `fs_write(1, ...)`. I had changed it because with the hack in place for fs_write, the two were functionally identical. Now though, the expectation should be that dbglog still outputs to stdout (or, perhaps in the future to an arbitrary fd).

kernel/libc/koslib/dup.c
```diff
@@ -0,0 +1,17 @@
...
+#include <unistd.h>
+#include <kos/fs.h>
+
+int dup(int oldfd) {
+    return fs_dup(oldfd);
+}
+
+int dup2(int oldfd, int newfd) {
+    return fs_dup2(oldfd, newfd);
+}
```
ljsebald, line 17: As silly as it is, please separate this out to its own file.

andressbarajas:
@QuzarDC I believe I addressed your suggestions.  For the fdtbl_destroy function I just ignored refcnt.  I figured we are shutting down, refcnt be damned.

andressbarajas:
@QuzarDC I made the changes you suggested to isatty, but it caused only part of the printf output to appear, with some remaining in the buffer. According to my research, isatty should return true for PTY. After reverting the changes, the printf messages displayed correctly as before.

QuzarDC:
> @QuzarDC I made the changes you suggested to isatty, but it caused only part of the printf output to appear, with some remaining in the buffer. According to my research, isatty should return true for PTY. After reverting the changes, the printf messages displayed correctly as before.

My point was for isatty to be able to return false if you have redirected stdin, stdout, or stderr to a file on a different vfs. Additionally to return true if you have opened an additional pty that isn't one of those three. I think in the test changes you made the logic was reversed.

So that this would work, for instance:
```
    isatty(STDOUT_FILENO); // Should return true, by default it's backed by a tty
    freopen("/pc/output.txt", "w", stdout); //Presume this works
    isatty(STDOUT_FILENO); // Should return false as stdout is no longer a tty, currently would return true
```
or this:
```
    file_t master_fd, slave_fd;
    fs_pty_create(NULL, 0, &master_fd, &slave_fd);
    isatty(master_fd); // Should return true, currently would return false as it's *only* checking the std fds
```
Maybe I've just misunderstood `isatty` if my examples are wrong?

andressbarajas:
@QuzarDC Oh I misunderstood.  I get it now.  Ill take care of it.

andressbarajas:
@QuzarDC Updated isatty()

kernel/libc/koslib/pipe.c
```diff
@@ -0,0 +1,37 @@
...
+    if(fs_pty_create(NULL, 0, &master_fd, &slave_fd) < 0)
+        return -1;
+
+    if(master_fd < 0)
+        return -1;
+
+    if(slave_fd < 0) {
+        fs_close(master_fd);
+        return -1;
+    }
```
QuzarDC, line 30: Sorry for the late stray review, I overlooked pipe the first time around.

I think, possibly, that this checking represents a gap in `fs_pty_create`s error handling. If either of those fds aren't valid, then doesn't that mean the pty creation failed? Or is that behavior allowed for some reason?

If that's not an issue, then this should try to close both, as master_fd being invalid doesn't mean that slave_fd is also invalid

andressbarajas: Youre right.  Looks to be a gap in fs_pty_creates error handling.  I'll take a look tomorrow.

andressbarajas: @QuzarDC I updated pipe() to count on fs_pty_create() to do the right error handling.  Updated fs_pty_create() to handle cleanup if the master or slave files fail to open.

QuzarDC:
I got to test this out and it looks like it works great except for one thing. Now all shutdown messages get dropped. Since we're now actually properly shutting down the pty connections, anything after that will be cut off from dbglog unless it's inside an interrupt (which would then bypass `fs_write`). Even with the changes made in #608 to the shutdown routine, you'd only get some of them and not all (especially not the final mallocstats).

I think there's two different things that could resolve this and aren't mutually exclusive, but I'm open to other ideas:
1) Update dbglog so that if `fs_write(stdout)` fails, it fails back to a direct dbgio write which would bypass possible redirection. Tested it and it worked fine. It seems to make sense that once the user code exits they should not expect stdio to still be redirected. Something like this, though perhaps shuffled around to look nicer:

```
    if(i > 0) {
        if(irq_inside_int() || 
            (fs_write(STDOUT_FILENO, printf_buf, strlen(printf_buf)) < 0))
                dbgio_write_str(printf_buf);        
    }
```
2) Rearrange the order of shutdowns so that stdio is shut down after most other things. This though could require a bit of rewriting to ensure that they don't use dynamic memory to not have mallocstats always show a leak, if it's put that far down. I have an issue I'm writing up for a general review of the shutdown process and order so I think I'll just add this as something to consider during that.

Oh, and tested with the mallocstats and it seems as though this stops some of the memory leaks that were fairly persistently showing up as pointing to pty as I showed here: https://github.com/KallistiOS/KallistiOS/issues/629#issuecomment-2168780306

andressbarajas:
Like idea number one but I think there is a third option. Undoing the redirection in the fs_pty_shutdown. I'll give that a try tonight

andressbarajas:
@QuzarDC I tested option number 1 and im still not seeing the shutdown debug :^/.  Do I have to compile KOS a certain way?

andressbarajas:
@QuzarDC Oops, nevermind it works.  I got dbgio_printf confused with dbglog.  Should I update dbgio_printf as well?:

```c
int dbgio_printf(const char *fmt, ...) {
    va_list args;
    int i;

    /* XXX This isn't correct. We could be inside an int with IRQs
      enabled, and we could be outside an int with IRQs disabled, which
      would cause a deadlock here. We need an irq_is_enabled()! */
    if(!irq_inside_int())
        spinlock_lock(&lock);

    va_start(args, fmt);
    i = vsnprintf(printf_buf, sizeof(printf_buf), fmt, args);
    va_end(args);

    if(i > 0) {
        if(irq_inside_int() || 
             (fs_write(STDOUT_FILENO, printf_buf, strlen(printf_buf)) < 0))
                 dbgio_write_str(printf_buf);        
    }

    if(!irq_inside_int())
        spinlock_unlock(&lock);

    return i;
}
```

QuzarDC:
> Like idea number one but I think there is a third option. Undoing the redirection in the fs_pty_shutdown. I'll give that a try tonight
That wouldn't fix the whole thing because the problem exists without redirection, but that would get things further down the road and it's not mutually exclusive to the others.

> @QuzarDC Oops, nevermind it works. I got dbgio_printf confused with dbglog. Should I update dbgio_printf as well? Before it was just using dbgio_write_str():
> 
> ```c
> int dbgio_printf(const char *fmt, ...) {
>     va_list args;
>     int i;
> 
>     /* XXX This isn't correct. We could be inside an int with IRQs
>       enabled, and we could be outside an int with IRQs disabled, which
>       would cause a deadlock here. We need an irq_is_enabled()! */
>     if(!irq_inside_int())
>         spinlock_lock(&lock);
> 
>     va_start(args, fmt);
>     i = vsnprintf(printf_buf, sizeof(printf_buf), fmt, args);
>     va_end(args);
> 
>     if(i > 0) {
>         if(irq_inside_int() || 
>              (fs_write(STDOUT_FILENO, printf_buf, strlen(printf_buf)) < 0))
>                  dbgio_write_str(printf_buf);        
>     }
> 
>     if(!irq_inside_int())
>         spinlock_unlock(&lock);
> 
>     return i;
> }
> ```

No, I don't think so, because that would end up just selectively hijacking things that go through dbgio_printf vs dbgio_write_str which should be at the same level. I suppose it could be added at the dbgio_write_str level, but I think that's all meant to be lower level than stdio. If you want to redirect dbgio then it can be via a replacement dbgio defined device rather than via stdio redirection.

andressbarajas:
@QuzarDC Yeah, the third option didnt pan out :^(. I implemented the change only to dbglog.

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+
+#include <kos/fs_pty.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+int main(int argc, char* argv[]) {
+    /* Create a PTY pair */
+    file_t master_fd, slave_fd;
```
gyrovorbis, line 15:
```suggestion
    file_t master_fd=NULL, slave_fd=NULL;
    int retval = EXIT_SUCCESS;
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+        }
+    } else if (bytes_read == 0) {
+        printf("No more data to read (EOF)\n");
+    } else {
+        /* Null-terminate and print the received message */
+        buffer[bytes_read] = '\0';
+        printf("Received 2nd message: %s\n", buffer);
+    }
+
+    /* Clean up resources */
```
gyrovorbis, line 76:
```suggestion
    /* Clean up resources */
cleanup:
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+    fs_close(master_fd);
```
gyrovorbis, line 77:
```suggestion
    if(master_fd) fs_close(master_fd);
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+    fs_close(slave_fd);
```
gyrovorbis, line 78:
```suggestion
    if(slave_fd) fs_close(slave_fd);
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+            return 1;
+        }
+    } else if (bytes_read == 0) {
+        printf("No more data to read (EOF)\n");
+    } else {
+        /* Null-terminate and print the received message */
+        buffer[bytes_read] = '\0';
+        printf("Received 2nd message: %s\n", buffer);
+    }
+
```
gyrovorbis, line 75:
```suggestion
    /* Jump over failure and only do clean-up */
    goto cleanup;
    
    /* Set exit code and clean up */
failure:
    retval = EXIT_FAILURE;

```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+    /* Clean up resources */
+    fs_close(master_fd);
+    fs_close(slave_fd);
+
+    printf("DONE!\n");
+
+    return 0;
```
gyrovorbis, line 82:
```suggestion
    return retval;
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+#include <stdio.h>
+#include <string.h>
+#include <errno.h>
+
+int main(int argc, char* argv[]) {
+    /* Create a PTY pair */
+    file_t master_fd, slave_fd;
+    if(fs_pty_create(NULL, 0, &master_fd, &slave_fd) < 0) {
+        fprintf(stderr, "Error creating PTY pair");
+        return 1;
```
gyrovorbis, line 18:
```suggestion
        goto failure;
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /* Set non-blocking mode on the slave_fd for testing */
+    int flags = fcntl(slave_fd, F_GETFL, 0);
+    if(fcntl(slave_fd, F_SETFL, flags | O_NONBLOCK) < 0) {
+        fprintf(stderr, "Error setting O_NONBLOCK");
+        fs_close(master_fd);
```
gyrovorbis, line 25:
```suggestion
        goto failure;
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+        fs_close(slave_fd);
```
gyrovorbis, line 26:
```suggestion
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+        return 1;
```
gyrovorbis, line 27:
```suggestion
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /* Write to the master end of the PTY */
+    const char *msg = "Hello from master!";
+    if(write(master_fd, msg, strlen(msg)) < 0) {
+        fprintf(stderr, "Error writing to master PTY");
+        fs_close(master_fd);
```
gyrovorbis, line 34:
```suggestion
        goto failure;
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+        fs_close(slave_fd);
```
gyrovorbis, line 35:
```suggestion
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+        return 1;
```
gyrovorbis, line 36:
```suggestion
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+
+    /* Read from the slave end of the PTY */
+    char buffer[128];
+    ssize_t bytes_read = read(slave_fd, buffer, sizeof(buffer) - 1);
+    if(bytes_read < 0) {
+        if(errno == EAGAIN) {
+            printf("No data available (non-blocking mode)\n");
+        } else {
+            fprintf(stderr, "Error reading from slave PTY");
+            fs_close(master_fd);
```
gyrovorbis, line 47:
```suggestion
            goto failure;
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+            fs_close(slave_fd);
```
gyrovorbis, line 48:
```suggestion
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+            return 1;
```
gyrovorbis, line 49:
```suggestion
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    /* Try and read again */
+    bytes_read = read(slave_fd, buffer, sizeof(buffer) - 1);
+    if(bytes_read < 0) {
+        if(errno == EAGAIN) {
+            printf("No more data available (non-blocking mode)\n");
+        } else {
+            fprintf(stderr, "Error reading from slave PTY");
+            fs_close(master_fd);
```
gyrovorbis, line 64:
```suggestion
            goto failure;
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+            fs_close(slave_fd);
```
gyrovorbis, line 65:
```suggestion
```

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
...
+            return 1;
```
gyrovorbis, line 66:
```suggestion
```

gyrovorbis (commented):
These are really trivial, and you can tell me to eff-off, but I really wanted to show you that if you use the alpha goto-as-a-structured-exception-undo-stack pattern, you can deduplicate all of the cleanup you have to do when exiting with an error...

This is a common idiom used in high-quality C code (including KOS, in at least Crabby's code), where a complex multi-step function which can fail and has to undo the previous steps will use goto to jump to a series of labels which fall through and perform clean-up in the reverse order with only a single clean exit point. It deduplicates code and help make sure you don't miss any cleanup or exit before undoing important stuff with a complex function like this.

examples/dreamcast/filesystem/pty/pty.c
```diff
@@ -0,0 +1,83 @@
+/* KallistiOS ##version##
+
+   pty.c
+   Copyright (C) 2024 Andress Barajas
+*/
+
```
gyrovorbis, line 6: There should be one more comment here that describes overall what the example/program does and how it works from a high-level.

kernel/libc/newlib/newlib_isatty.c
```diff
@@ -3,29 +3,43 @@
...
+    /* Make sure that stdin is shown as a tty, otherwise
+       it won't be set as line-buffered. */
+    if(fd == STDIN_FILENO)
+        return 1;
+
+    if(fstat(fd, &statbuf) == -1)
+        return 0;
+
+    /* Check if PTY */
+    if(statbuf.st_dev == PTY_DEV)
```
gyrovorbis, line 35:
```suggestion
    if(statbuf.st_dev == PTY_DEV)
```
Not gonna lie, like... This makes me kind of uncomfortable... In terms of design, the fact we're basically hard-coding this stuff as being TTY or not TTY by doing some intrusive inspection... and now `PTY_DEV` is something that isn't really part of the PTY interface and is a local private define rather than a public one that would at least keep the values synchronized.... 

I guess I'm fine doing this as part of this PR, but the *correct* solution really should be adding an `isatty()` virtual function/function pointer to the VFS interface itself so that each individual filesystem can uniformly report back whether it is to be treated as a TTY or not, and that logic remains encapsulated within each...

Also an alternative would be to add another flag to the statbuf or something...

gyrovorbis:
> > Like idea number one but I think there is a third option. Undoing the redirection in the fs_pty_shutdown. I'll give that a try tonight
> > That wouldn't fix the whole thing because the problem exists without redirection, but that would get things further down the road and it's not mutually exclusive to the others.
> 
> > @QuzarDC Oops, nevermind it works. I got dbgio_printf confused with dbglog. Should I update dbgio_printf as well? Before it was just using dbgio_write_str():
> > ```c
> > int dbgio_printf(const char *fmt, ...) {
> >     va_list args;
> >     int i;
> > 
> >     /* XXX This isn't correct. We could be inside an int with IRQs
> >       enabled, and we could be outside an int with IRQs disabled, which
> >       would cause a deadlock here. We need an irq_is_enabled()! */
> >     if(!irq_inside_int())
> >         spinlock_lock(&lock);
> > 
> >     va_start(args, fmt);
> >     i = vsnprintf(printf_buf, sizeof(printf_buf), fmt, args);
> >     va_end(args);
> > 
> >     if(i > 0) {
> >         if(irq_inside_int() || 
> >              (fs_write(STDOUT_FILENO, printf_buf, strlen(printf_buf)) < 0))
> >                  dbgio_write_str(printf_buf);        
> >     }
> > 
> >     if(!irq_inside_int())
> >         spinlock_unlock(&lock);
> > 
> >     return i;
> > }
> > ```
> 
> No, I don't think so, because that would end up just selectively hijacking things that go through dbgio_printf vs dbgio_write_str which should be at the same level. I suppose it could be added at the dbgio_write_str level, but I think that's all meant to be lower level than stdio. If you want to redirect dbgio then it can be via a replacement dbgio defined device rather than via stdio redirection.

Guys, I really, really think we shouldn't be doing this. One of the first things I did to save some RAM and to speed up `printf()` and logging in general was to completely get rid of this spinlock.... Why exactly do we need to spinlock here? Because we're sharing the same global 1KB buffer for every thread... Which... why are we doing that? Everything this routine needs to do for a given thread can be done on the stack without any of that extra complexity, both here and in `dbglog`:

```c
int dbgio_printf(const char *fmt, ...) {
   char printf_buf[1024];
    va_list args;
    int i;

    va_start(args, fmt);
    i = vsnprintf(printf_buf, sizeof(printf_buf), fmt, args);
    va_end(args);

    dbgio_write_str(printf_buf);        

    return i;
}
```

1) We're saving 1KB of persistent RAM which might never even need to get used or would only be used transiently and infrequently.
2) We're simplifying the logic and are inherently reentrant now just by making the buffer on the stack.

I do think we should make sure that there's not supposed to be a global `#define` for the `printf()` buffer size somewhere like in `limits.h` that we're supposed to advertise and abide by...

kernel/libc/newlib/newlib_isatty.c
andressbarajas: That makes sense.  I think it would be better to have that in another PR.  I think there are two open PRs currently that "cleanup" the fs.c file.  Maybe it can be done in one of those?  Im cool to create another PR to take the virtual function/function pointer approach after this one is merged.

andressbarajas:
@gyrovorbis Thanks for the review. Maybe after this PR is merged I can implement the dbgio_printf() and dbgio_log() changes you mentioned in another PR.

kernel/libc/newlib/newlib_isatty.c
QuzarDC: > Not gonna lie, like... This makes me kind of uncomfortable... In terms of design, the fact we're basically hard-coding this stuff as being TTY or not TTY by doing some intrusive inspection... and now `PTY_DEV` is something that isn't really part of the PTY interface and is a local private define rather than a public one that would at least keep the values synchronized....
> 
> I guess I'm fine doing this as part of this PR, but the _correct_ solution really should be adding a something like a `bool (*isatty)(file_t)` virtual function/function pointer to the VFS interface itself so that each individual filesystem can uniformly report back whether it is to be treated as a TTY or not, and that logic remains encapsulated within each...
> 
> Also an alternative would be to add another flag to the statbuf or something...
> 
> But like I said, probably can come later during a more clean-up/feature addition pass to the VFS, as I feel like there are other things that need to be added there as well.

If you look at how `isatty` is implemented in many platforms it often comes down to insane things like comparing to a list of known TTY driver device names. While I agree that `PTY_DEV` should stem directly from fs_pty, it seems silly to add a new virtual function that would only deviate/be used for a single driver. Unless it seems likely that we get other hardware-backed TTY implementations, but I really don't imagine that being the case.

Another flag in statbuf doesn't seem right either since that's so much part of the external stat system which normally has no such flags. Doing it at the vfs driver level as just a flag (rather than the virtual function that would basically return the value of such a flag) would be my preferred method, but even that seems like a lot for a singular exception that already has to have a lot of hardcoded stuff by standard.

gyrovorbis: I just looked at several platforms (because I have no idea how it's implemented), and two of them that I just looked at for libc and Linux are calling into `tcgetattr()` which makes a call into `ioctl()`., which... we actually do have in the VFS.

I guess I agree with you. I forgot `ioctl()` is there as a kind of one-off single-command thing that is optional to implement on most filesystems and can be safely ignored for things that aren't supported, but that doesn't require a whole function pointer entry in the driver. 

Here's `isatty()` for glibc/POSIX:
https://codebrowser.dev/glibc/glibc/sysdeps/posix/isatty.c.html#27
Here's the underlying `tcgetattr()` function:
https://codebrowser.dev/glibc/glibc/sysdeps/unix/bsd/tcgetattr.c.html#__tcgetattr

QuzarDC:
> Guys, I really, really think we shouldn't be doing this. One of the first things I did to save some RAM and to speed up `printf()` and logging in general was to completely get rid of this spinlock and static/global buffer... I tested the heck out of it too. Why exactly do we need to spinlock here? Because we're sharing the same global 1KB buffer for every thread... Which... why are we doing that? Everything this routine needs to do for a given thread can be done on the stack without any of that extra complexity, both here and in `dbglog`:
> 
> ```c
> int dbgio_printf(const char *fmt, ...) {
>     char printf_buf[1024];
>     va_list args;
>     int i;
> 
>     va_start(args, fmt);
>     i = vsnprintf(printf_buf, sizeof(printf_buf), fmt, args);
>     va_end(args);
> 
>     dbgio_write_str(printf_buf);        
> 
>     return i;
> }
> ```
> 
>     1. We're saving 1KB of persistent RAM which might never even need to get used or would only be used transiently and infrequently.
> 
>     2. We're simplifying the logic and are inherently reentrant now just by making the buffer on the stack.


That seems like a good point to be explored in a PR to update dbgio (no changes are being made to it in this one). I believe the locks are intended (though perhaps some other locking mechanism could be used or they could be moved further inward?) to prevent multiple concurrent prints from trampling over each other and keep the individual strings being printed contiguous. Without such a protection multiple threads outputting dbgio could end up with mangled output. With the standard dcload/text output it wouldn't make too much of a difference since at worst it would be strings inserted into others (`firsecondst ` rather than `first second`) but with `fb` it could really turn into garbled nonsense fast.

> 
> 
> I do think we should make sure that there's not supposed to be a global `#define` for the `printf()` buffer size somewhere like in `limits.h` that we're supposed to advertise and abide by...

C99 only gives a minimum required size to be able to handle which is 'at least 4095 characters per conversion'. That's 4095 per individual % param to my reading. In practice, that's why they rely on dynamic memory allocation rather than trying to make huge compliant buffers. There may be some extra posix bits about it, but I've never found any.

For our purposes I think it's sufficient to have an arbitrary size like 1024 and simply note that KOS' dbgio system only writes that amount before truncation. I'd imagine the only time this would come up would be with someone sort of abusing stdout redirection to use printf rather than fprintf.

gyrovorbis:
> > Guys, I really, really think we shouldn't be doing this. One of the first things I did to save some RAM and to speed up `printf()` and logging in general was to completely get rid of this spinlock and static/global buffer... I tested the heck out of it too. Why exactly do we need to spinlock here? Because we're sharing the same global 1KB buffer for every thread... Which... why are we doing that? Everything this routine needs to do for a given thread can be done on the stack without any of that extra complexity, both here and in `dbglog`:
> > ```c
> > int dbgio_printf(const char *fmt, ...) {
> >     char printf_buf[1024];
> >     va_list args;
> >     int i;
> > 
> >     va_start(args, fmt);
> >     i = vsnprintf(printf_buf, sizeof(printf_buf), fmt, args);
> >     va_end(args);
> > 
> >     dbgio_write_str(printf_buf);        
> > 
> >     return i;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > 1. We're saving 1KB of persistent RAM which might never even need to get used or would only be used transiently and infrequently.
> > 
> > 2. We're simplifying the logic and are inherently reentrant now just by making the buffer on the stack.
> > ```
> 
> That seems like a good point to be explored in a PR to update dbgio (no changes are being made to it in this one). I believe the locks are intended (though perhaps some other locking mechanism could be used or they could be moved further inward?) to prevent multiple concurrent prints from trampling over each other and keep the individual strings being printed contiguous. 

Every filesystem that I'm seeing that can do writes like this is already doing a lock before within their filesystem internally. I think what you said about "moving it inward" is probably what we want, so that the dbgio driver that just forwards to dcload doesn't need to lock (knowing the VFS driver will) while the FB debug driver would have the lock there... 

Check out how it works in the VFS in terms of locking at the lower layers:
https://github.com/KallistiOS/KallistiOS/blob/f34c9f30ac28fdae651da16cbc3618192f96626b/kernel/arch/dreamcast/fs/fs_dcload.c#L112
https://github.com/KallistiOS/KallistiOS/blob/f34c9f30ac28fdae651da16cbc3618192f96626b/kernel/arch/dreamcast/fs/fs_dclsocket.c#L364

> Without such a protection multiple threads outputting dbgio could end up with mangled output. With the standard dcload/text output it wouldn't make too much of a difference since at worst it would be strings inserted into others (`firsecondst ` rather than `first second`) but with `fb` it could really turn into garbled nonsense fast.
Right. Problem is this `spinlock` isn't actually protecting against that at all.. You can circumvent it and write concurrently to `dbgio` just using `dbgio_writestring()` directly instead of using `dbgio_printf()`. 

> 
> > I do think we should make sure that there's not supposed to be a global `#define` for the `printf()` buffer size somewhere like in `limits.h` that we're supposed to advertise and abide by...
> 
> C99 only gives a minimum required size to be able to handle which is 'at least 4095 characters per conversion'. That's 4095 per individual % param to my reading. In practice, that's why they rely on dynamic memory allocation rather than trying to make huge compliant buffers. There may be some extra posix bits about it, but I've never found any.
> 
> For our purposes I think it's sufficient to have an arbitrary size like 1024 and simply note that KOS' dbgio system only writes that amount before truncation. I'd imagine the only time this would come up would be with someone sort of abusing stdout redirection to use printf rather than fprintf.
Yeah, I agree... I had read about the 4k buffer default (which seems ridiculous) but wasn't sure about the specifics... I guess if it's *assumed* to be 4k, there's not some pretty `#define` we can provide to let the user know our internal limit?

kernel/libc/newlib/newlib_isatty.c
andressbarajas: Nice find!  I'll take a look later today to use ioctl() instead

andressbarajas: I probably jumped the gun but I did the above implementation posted by @gyrovorbis.  Had to create more files. @QuzarDC let me know :^p. I think this is the only thing left blocking.

QuzarDC (approved):
All the previous issues were addressed, and then some! Something that chooses not to rely on `isatty` could get to it the long way around through `ioctl` (and this sets us up to more easily provide other ioctl data, such as for /dev/fb)

PR #637, opened 2024-06-19
Add IRQ-safe mutex / rwsem functions, and scoped mutexes

pcercuei:
Introduce three new functions:
- `mutex_lock_irqsafe`
- `rwsem_write_lock_irqsafe`
- `rwsem_read_lock_irqsafe`

These functions will return -1 with errno == EWOULDBLOCK when running inside an IRQ context, while the mutex or semaphore has already been locked.

Also introduce two new macros:
- `mutex_lock_scoped`
- `mutex_lock_irqsafe_scoped`

Those two macros will lock a mutex, similarly to mutex_lock / mutex_lock_irqsafe respectively, with the difference that the mutex will automatically be unlocked once the execution exits the functional block in which these macros were called.

Using scoped mutexes is generally a good idea, because it ensures that the mutexes will always be unlocked on all exit points of a function, and as such it avoids the common mistake where a mutex isn't properly unlocked in one of the exit paths of a function.

This PR also updates some of the code in KallistiOS to use the new constructs. As a result of the cleanup, more than 500 lines of code have been dropped.


include/kos/mutex.h
```diff
@@ -167,6 +167,29 @@ int mutex_destroy(mutex_t *m);
 */
 int mutex_lock(mutex_t *m);
 
+/** \brief  Lock a mutex.
+
+    This function will lock a mutex, if it is not already locked by another
+    thread. If it is locked by another thread already, this function will block
+    until the mutex has been acquired for the calling thread.
+    This function can be from within an interrupt context. In that case, if the
```
andressbarajas, line 175:
```suggestion
    This function can be from within an interrupt context. In that case, if the
```
This function can be called?

ljsebald:
One minor comment... `EAGAIN` and `EWOULDBLOCK` are the same thing (and they do seem to be on just about every system out there that I've ever used -- considering even POSIX says that they're the same value most of the time). There's no reason to be changing back-and-forth between the two of them...

pcercuei:
I didn't know! I will update the code accordingly.

pcercuei:
@ljsebald Updated.

kernel/thread/rwsem.c
```diff
@@ -154,6 +161,13 @@ int rwsem_write_lock(rw_semaphore_t *s) {
     return rwsem_write_lock_timed(s, 0);
 }
 
+int rwsem_write_lock_irqsafe(rw_semaphore_t *s) {
+    if (irq_inside_int())
```
ljsebald, line 165:
```suggestion
    if(irq_inside_int())
```

kernel/thread/rwsem.c
```diff
@@ -107,6 +107,13 @@ int rwsem_read_lock(rw_semaphore_t *s) {
     return rwsem_read_lock_timed(s, 0);
 }
 
+int rwsem_read_lock_irqsafe(rw_semaphore_t *s) {
+    if (irq_inside_int())
```
ljsebald, line 111:
```suggestion
    if(irq_inside_int())
```

kernel/thread/mutex.c
```diff
@@ -82,6 +82,13 @@ int mutex_lock(mutex_t *m) {
     return mutex_lock_timed(m, 0);
 }
 
+int mutex_lock_irqsafe(mutex_t *m) {
+    if (irq_inside_int())
```
ljsebald, line 86:
```suggestion
    if(irq_inside_int())
```

pcercuei:
Updated again with the `\n` tokens added at the end of the doxygen lines which need it.

ljsebald (changes requested):
A few minor stylistic things real quick...

pcercuei:
Updated again.

Aside from the coding style fixes, I removed `mutex_lock_irqsafe_scoped()` which did not work properly, and the best way to implement it (if we do want it) has not been decided yet.

ljsebald (approved):

ljsebald:
Approving, but I think someone else should also take a look at this to make sure I didn't miss anything, since it does kinda change a bit of core functionality. @QuzarDC , perhaps?

kernel/arch/dreamcast/fs/fs_dclsocket.c
```diff
@@ -206,22 +206,14 @@ static void dcls_recv_loop(void) {
...
-
-    if(locked == -1 && irq_inside_int()) {
-        errno = EAGAIN;
-        return 0;
-    }
-    else if(locked == -1) {
-        mutex_lock(&mutex);
-    }
+    if(mutex_lock_irqsafe(&mutex))
+        return NULL;
```
Kazade, line 216: Should we be setting `errno = EAGAIN` here?

Kazade: There are other cases of this below, so assume I've written this comment on those too :p

kernel/net/net_udp.c
```diff
@@ -633,15 +592,8 @@ static void net_udp_close(net_socket_t *hnd) {
...
-    if(irq_inside_int()) {
-        if(mutex_trylock(&udp_mutex) == -1) {
-            errno = EWOULDBLOCK;
-            return;
-        }
-    }
-    else {
-        mutex_lock(&udp_mutex);
-    }
+    if(mutex_lock_irqsafe(&udp_mutex))
```
Kazade, line 595: Seems odd that we're not indicating any error here (no -1 return or errno set)

pcercuei: `errno` is set within `mutex_lock_irqsafe`.

QuzarDC (commented):
Only found one issue where return values seem to have gotten accidentally changed. Still pending testing against different toolchains and validating that the examples still work, but looks good!

pcercuei:
@QuzarDC Thank you for your review.

I fixed the return value issues in the TCP code as well as in the UDP code (which had the same problem) and force-pushed.

include/kos/mutex.h
```diff
@@ -255,6 +278,28 @@ int mutex_unlock(mutex_t *m);
...
 int mutex_unlock_as_thread(mutex_t *m, kthread_t *thd);
 
+/** \cond */
+static inline void __mutex_scoped_cleanup(mutex_t **m) {
+    if(*m)
+        mutex_unlock(*m);
+}
+
+#define ___mutex_lock_scoped(m, l) \
+    mutex_t *__scoped_mutex_##l __attribute__((cleanup(__mutex_scoped_cleanup))) = mutex_lock(m) ? NULL : (m)
```
gyrovorbis, line 288: Damn. I have used a pattern where I've had some macro have to declare an internal/hidden variable that is supposed to not be seen by the user, and I've had to try to think of a good name for it... but I never thought of using the source location to make it unique like that... Clever.

QuzarDC (approved):

PR #638, opened 2024-06-19
Add mutex_lock_scoped()

pcercuei:
The `mutex_lock_scoped()` macro will lock a mutex, similarly to mutex_lock, with the difference that the mutex will automatically be unlocked once the execution exits the functional block in which the macro was called.

The ramdisk FS code was updated to use it, as an example to how it can be helpful.

Using scoped mutexes is generally a good idea, because it ensures that the mutexes will always be unlocked on all exit points of a function, and as such it avoids the common mistake where a mutex isn't properly unlocked in one of the exit paths of a function.

### DIFF
--- a/include/kos/mutex.h
+++ b/include/kos/mutex.h
@@ -255,6 +255,37 @@ int mutex_unlock(mutex_t *m);
 */
 int mutex_unlock_as_thread(mutex_t *m, kthread_t *thd);
 
+/** \cond */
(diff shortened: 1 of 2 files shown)


pcercuei:
I decided to close this PR, and merge those two commits into my other PR: https://github.com/KallistiOS/KallistiOS/pull/637

For the reason that the network code in the other PR could use scoped locks as well.

PR #643, opened 2024-06-26
mutex: Fix mutexes not mutexing

pcercuei:
Imagine a scenario where thread #0 holds a lock. Thread #1 attempts to lock it, doesn't get it, and is put to sleep with genwait_wait(). Thread #0 wakes up, releases the lock, which triggers genwait_wake_one(), placing thread #1 back into the running queue. However, before leaving hand, thread #0 gets the lock once again. Thread #1 would then wake up, and incorrectly set itself as the lock owner, without checking first that the lock was free.

Address this issue by checking that the lock is indeed free after being awaken, and if not, go back to sleep.

### DIFF
--- a/kernel/thread/mutex.c
+++ b/kernel/thread/mutex.c
@@ -14,6 +14,7 @@
 #include <kos/dbglog.h>
 
 #include <arch/irq.h>
+#include <arch/timer.h>
 
 mutex_t *mutex_create(void) {
     mutex_t *rv;
@@ -83,6 +84,7 @@ int mutex_lock(mutex_t *m) {
 }
 
 int mutex_lock_timed(mutex_t *m, int timeout) {
+    uint64_t deadline = 0;
     int old, rv = 0;
 
     if((rv = irq_inside_int())) {
@@ -123,14 +125,31 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
         rv = -1;
     }
     else {
-        if(!(rv = genwait_wait(m, timeout ? "mutex_lock_timed" : "mutex_lock",
-                               timeout, NULL))) {
-            m->holder = thd_current;
-            m->count = 1;
-        }
-        else {
-            errno = ETIMEDOUT;
-            rv = -1;
+        if(timeout)
+            deadline = timer_ms_gettime64() + timeout;
+
+        for(;;) {
+            rv = genwait_wait(m, timeout ? "mutex_lock_timed" : "mutex_lock",
+                              timeout, NULL);
+            if(rv < 0) {
+                errno = ETIMEDOUT;
+                break;
+            }
+
+            if(!m->holder) {
+                m->holder = thd_current;
+                m->count = 1;
+                break;
+            }
+
+            if(timeout) {
+                timeout = deadline - timer_ms_gettime64();
+                if(timeout <= 0) {
+                    errno = ETIMEDOUT;
+                    rv = -1;
+                    break;
+                }
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


darcagn:
I was having a race problem with the Rust standard library's `ReentrantLock` which is implemented on top of `pthread_mutex_lock` using the `libpthread` beta KOS branch. Multiple threads would get the lock and then crash when asserting this was not the case.

I reimplemented the Rust `ReentrantLock` in C to eliminate the Rust and libpthread variables, and found the exact same problem occurred.

This PR fixes both scenarios. Thank you so much for finding this.

kernel/thread/mutex.c
```diff
@@ -123,14 +125,30 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
...
     else {
-        if(!(rv = genwait_wait(m, timeout ? "mutex_lock_timed" : "mutex_lock",
-                               timeout, NULL))) {
-            m->holder = thd_current;
-            m->count = 1;
-        }
-        else {
-            errno = ETIMEDOUT;
-            rv = -1;
+        if (timeout)
```
andressbarajas, line 128:
```suggestion
        if(timeout)
```

kernel/thread/mutex.c
```diff
@@ -123,14 +125,30 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
...
+            deadline = timer_ms_gettime64() + timeout;
+
+        for (;;) {
```
andressbarajas, line 131:
```suggestion
        for(;;) {
```

kernel/thread/mutex.c
```diff
@@ -123,14 +125,30 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
...
+            rv = genwait_wait(m, timeout ? "mutex_lock_timed" : "mutex_lock",
+                              timeout, NULL);
+            if (rv < 0) {
```
andressbarajas, line 134:
```suggestion
            if(rv < 0) {
```

kernel/thread/mutex.c
```diff
@@ -123,14 +125,30 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
...
+                errno = ETIMEDOUT;
+                break;
+            }
+
+            if (!m->holder) {
```
andressbarajas, line 139:
```suggestion
            if(!m->holder) {
```

kernel/thread/mutex.c
```diff
@@ -123,14 +125,30 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
...
+                m->holder = thd_current;
+                m->count = 1;
+                break;
+            }
+
+            if (timeout) {
```
andressbarajas, line 145:
```suggestion
            if(timeout) {
```

kernel/thread/mutex.c
```diff
@@ -123,14 +125,30 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
...
+                timeout = deadline - timer_ms_gettime64();
+                if (timeout <= 0) {
```
andressbarajas, line 147:
```suggestion
                if(timeout <= 0) {
```

pcercuei:
@andressbarajas thanks, I fixed the whitespace issues.

Also added a `rv = -1` below the second `errno = ETIMEDOUT` which was missing.

gyrovorbis:
@darcagn was most definitely able to create a repro case for this in minimal C, which we've PR'd as a KOS example/regressoin test here: https://github.com/KallistiOS/KallistiOS/pull/644

...So we can rest easy knowing that this shit won't happen again. lol.

ljsebald (approved):

PR #645, opened 2024-06-27
/dev rewinddir on close

andressbarajas:
Playing with the sound/ghettoplay-vorbis example.  I found that if I visited the /dev directory and then revisited it again, the directory contents would be missing.  This fixes that.


QuzarDC (approved):
Great catch! Thank you also for separating this out from the bigger PR.

Along with the discussion of this, I'd mentioned the problem that there's only one fd so only tracking for one readdir at a time. To track that for the future I opened #649 where I have a proposal for that which might fix edge cases in other places.

PR #646, opened 2024-06-27
/rd . & .. are directories

andressbarajas:
Playing with the browse example I have in another PR I noticed that . and .. folders are not recognized as directories.  This fixes that.


QuzarDC (approved):
Looks good, and thank you so much for the extra work to split this one out from your bigger one.

PR #647, opened 2024-06-27
/sd directories viewed as files

andressbarajas:
The directories and subdirectories in /sd would be considered files even with the O_DIR attr set.  This fixes that by setting the correct size and time.


gyrovorbis (approved):
Thanks. Yep, this appears to be in-line with how the rest of the VFS drivers handle sizing a directory.

PR #648, opened 2024-06-29
Replace realpath() for a better license

andressbarajas:
The purpose of this PR was to make our AUTHORS file a little lighter.  The current implementation of realpath() just normalized a path without checking if a path exists or handling symlinks like its supposed to do: https://pubs.opengroup.org/onlinepubs/009696799/functions/realpath.html.

With this new correct implementation, I found that I needed to create a replacement for how the old implementation behaved, fs_normalize_path().  

All-in-all, no behavior change. We just free up realpath() to work like it is supposed to.

### DIFF
--- a/kernel/libc/koslib/realpath.c
+++ b/kernel/libc/koslib/realpath.c
@@ -1,165 +1,143 @@
+/* KallistiOS ##version##
+
+   realpath.c
+
+   Copyright (C) 2024 Andress Barajas
+*/
+
 /*
- * Copyright (c) 2003 Constantin S. Svintsoff <kostik@iclub.nsu.ru>
- *
- * Redistribution and use in source and binary forms, with or without
- * modification, are permitted provided that the following conditions
- * are met:
- * 1. Redistributions of source code must retain the above copyright
- *    notice, this list of conditions and the following disclaimer.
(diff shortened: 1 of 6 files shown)


kernel/libc/koslib/realpath.c
```diff
@@ -1,165 +1,143 @@
...
- */
-
-#include <kos/limits.h>
-#include <sys/cdefs.h>
-#include <sys/param.h>
-#include <sys/stat.h>
-#include <kos/fs.h>
+    The support for handling removing symlink components in this function is 
+    currently disabled because readlink implementations don't exist for all 
+    but one file system (ext2fs).
```
QuzarDC, line 11: Would it make sense to leave it and just let the `readlink` (or `fs_readlink`) appropriately fail? Or is the notion to leave it excluded for now because there's insufficient ways to validate it without wider support? And maybe uncomment it once symlink support is added for /ram/ or /pc/ or such?

andressbarajas: Yes, leave it excluded for now because it is untested for now. Can uncommented later once symlink support is added to FS's that are more popular.

QuzarDC (approved):

PR #652, opened 2024-07-02
CDFS: Set errno on errors

UnknownShadow200:
The CDFS now appropriately sets errno in error cases.

I attempted to choose the most appropriate errno codes for each error case, although there may be more appropriate errno codes.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_iso9660.c
+++ b/kernel/arch/dreamcast/fs/fs_iso9660.c
@@ -603,19 +603,26 @@ static void * iso_open(vfs_handler_t * vfs, const char *fn, int mode) {
     (void)vfs;
 
     /* Make sure they don't want to open things as writeable */
-    if((mode & O_MODE_MASK) != O_RDONLY)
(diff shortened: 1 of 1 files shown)


gyrovorbis (approved):
So I'm going to approve of this, because I reviewed it, and yes it looks great, and these changes are absolutely going to be needed for this VFS to work with a bunch of higher-level APIs including C++ and Rust's filesystems... BUT I am not going to merge this yet, because @ljsebald has historically always had a better eye than I have for making sure my `errno` handling has been POSIX/Linux compliant... I'll let him merge it when he gets around to looking at it.

For the record, everyone, this is one of the main developers behind ClassiCube, the classic Minecraft reimagination in low-level C and running on a bazillion platforms. Very impressive stuff!

Thank you so much for this PR, and please do not hesitate to upstream anything else you may find that needs fixing or open tickets for us to fix. We are happy to support your fine work!

QuzarDC (approved):
Just the one note for posterity. Otherwise looks good! Thanks for the contribution, it's appreciated.

Issue #617, opened 2024-06-08
INIT_NET issues


andressbarajas:
I was playing around with @darcagn's [httpd-ack-next repo](https://github.com/darcagn/httpd-ack-next/tree/testing) and I noticed that when using the INIT_NET flag in KOS_INIT_FLAGS(INIT_DEFAULT | INIT_NET); it would show a black screen for 45-60 seconds before showing the main screen.  Looking at the [code](https://github.com/KallistiOS/KallistiOS/blob/f92103258b6d0943bc657de1199eb9a5d8b4087d/kernel/net/net_core.c#L185) I found that KOS was trying to establish a DHCP connection which Im not setup for so it takes 45-60 secs to fail before it goes to showing the main screen of the repo.  By updating this portion of net_init() from

```
if(net_default_dev) {
    /* Did we get a requested IP address? this normally happens over dcload-ip. */
    if(ip) {
        rv = net_dhcp_request(ip);
        if(rv < 0) {
            dbglog(DBG_DEBUG, "Failed to acquire the specified IP with DHCP\n");

            /* If that fails, set the address manually. Although gateway
               and dns etc. will also manually need setting */
            net_ipv4_parse_address(ip, net_default_dev->ip_addr);
            rv = 0;
        }
    }
    /* We didn't get a requested IP address, if we don't already have one
       set, then do so via DHCP. */
    else if(!net_default_dev->ip_addr[0])
        rv = net_dhcp_request(0);
}
```
to this:
```
if(net_default_dev) {
    /* Did we get a requested IP address? this normally happens over dcload-ip. */
    if(ip) {
            /* Set the address manually. Although gateway
               and dns etc. will also manually need setting */
            net_ipv4_parse_address(ip, net_default_dev->ip_addr);
            rv = 0;
        }
    }
    /* We didn't get a requested IP address, if we don't already have one
       set, then do so via DHCP. */
    else if(!net_default_dev->ip_addr[0])
        rv = net_dhcp_request(0);
}
``` the main screen shows up within 2-3 seconds of upload using dcload-serial.  I have a question (as a networking noob):

1. Why do we do a call to net_dhcp_request(ip); when given an IP when it can also be set manually? For people that are not setup for DHCP, it blocks the main thread and we have 45-60 seconds of black screen before it times out and sets it manually.

Tagging experts and people Ive talked with this about: @Kazade @darcagn @gyrovorbis

andressbarajas:
[...]

PR #653, opened 2024-07-02
Add the option to prefer static IPs over DHCP

andressbarajas:
This PR is developer focused and helps developers who don't have their setup configured for DHCP.  This PR was created to partially address this issue (https://github.com/KallistiOS/KallistiOS/issues/617). It might be a while before dcload-ip catches up to even implement the ideas in this issue.  Even with taking an asynchronous approach, static configurations will still have to wait for DHCP requests to timeout (45-60 seconds).  This PR gives preference to initialize the network device using static options before defaulting to DHCP.

1. Added new INIT_NET_STATIC kos init flag that will use static methods first to initialize the network device. 
2. Added new net_init_static() that will try static methods before default to DHCP.
3. No behavior change for people using INIT_NET flag or net_init().

It still needs more testing so I will leave it in draft and see what everyone thinks about this approach.

Oh I also fixed some references to DMAC registers by using the existing header file dmac.h instead of redefining in init.c

### DIFF
--- a/kernel/arch/dreamcast/kernel/init.c
+++ b/kernel/arch/dreamcast/kernel/init.c
@@ -24,6 +24,7 @@
 #include <dc/pvr.h>
 #include <dc/vmufs.h>
 #include <dc/syscalls.h>
+#include <dc/dmac.h>
 
 #include "initall_hdrs.h"
 
@@ -44,10 +45,6 @@ void (*__kos_init_early_fn)(void) __attribute__((weak,section(".data"))) = NULL;
 int main(int argc, char **argv);
 uint32 _fs_dclsocket_get_ip(void);
 
-#define SAR2    ((vuint32 *)0xFFA00020)
-#define CHCR2   ((vuint32 *)0xFFA0002C)
-#define DMAOR   ((vuint32 *)0xFFA00040)
-
 /* We have to put this here so we can include plat-specific devices */
 dbgio_handler_t * dbgio_handlers[] = {
     &dbgio_dcload,
@@ -58,10 +55,12 @@ dbgio_handler_t * dbgio_handlers[] = {
 };
 int dbgio_handler_cnt = sizeof(dbgio_handlers) / sizeof(dbgio_handler_t *);
 
+/* If headless(launching with -n option), dcload_type will always be 
+   DCLOAD_TYPE_NONE, so it will behave just like arch_init_net_no_dcload() */
 void arch_init_net_dcload_ip(void) {
     union {
-        uint32 ipl;
-        uint8 ipb[4];
+        uint32_t ipl;
+        uint8_t ipb[4];
     } ip = { 0 };
 
     if(dcload_type == DCLOAD_TYPE_IP) {
@@ -71,9 +70,14 @@ void arch_init_net_dcload_ip(void) {
                ip.ipb[2], ip.ipb[1], ip.ipb[0]);
         dbgio_disable();
     }
+    
+    /* Enable networking (and drivers) */
+    if((__kos_init_flags & INIT_NET_STATIC) == INIT_NET_STATIC)
+        net_init_static(ip.ipl);
+    else
+        net_init(ip.ipl);     
 
-    net_init(ip.ipl);     /* Enable networking (and drivers) */
-
+    /* If headless(-n) this will always be false */
     if(dcload_type == DCLOAD_TYPE_IP) {
         fs_dclsocket_init_console();
 
@@ -86,7 +90,10 @@ void arch_init_net_dcload_ip(void) {
 }
 
 void arch_init_net_no_dcload(void) {
-    net_init(0);
+    if((__kos_init_flags & INIT_NET_STATIC) == INIT_NET_STATIC)
+        net_init_static(0);
+    else
+        net_init(0);  
 }
 
 KOS_INIT_FLAG_WEAK(arch_init_net_dcload_ip, true);
@@ -206,7 +213,7 @@ int  __weak arch_auto_init(void) {
 
     KOS_INIT_FLAG_CALL(dcload_init);
 
-    if (!KOS_PLATFORM_IS_NAOMI)
+    if(!KOS_PLATFORM_IS_NAOMI)
         KOS_INIT_FLAG_CALL(fs_iso9660_init);
 
     KOS_INIT_FLAG_CALL(vmu_fs_init);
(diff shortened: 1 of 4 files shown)


andressbarajas:
Tagging @Kazade here because he has played with the network stack the most and my changes are based on the work he has done.

andressbarajas:
Tested and should be ready for review

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -71,9 +70,14 @@ void arch_init_net_dcload_ip(void) {
                ip.ipb[2], ip.ipb[1], ip.ipb[0]);
         dbgio_disable();
     }
-
-    net_init(ip.ipl);     /* Enable networking (and drivers) */
-
+    
+    /* Enable networking (and drivers) */
+    if((__kos_init_flags & INIT_NET_STATIC) == INIT_NET_STATIC)
```
gyrovorbis, line 75:
```suggestion
    if(__kos_init_flags & INIT_NET_STATIC)
```
When we only have one bit in a mask we're testing against (ie a bitflat), we just check it like that, since *any* nonzero value implies equality here. This is also consistent with the rest of the coedz.

gyrovorbis:
See... this is like... where if we had command-line args, you'd be like `--kos-net=STATIC` or something... I'm just not sure how I feel that we have to hardcode something like this into the binary itself which seems like it's going to vary from person-to-person and is instead a user-configuration/preference thing...

Dare I say... it almost seems like we should be trying to probe flash or internet settings or something on network initialization for this setting, rather than hardcoding it into the binary... but I dunno. I don't know what mechanism we would use to determine which method to use at runtime... We'll see what @Kazade thinks?

kernel/arch/dreamcast/kernel/init.c
andressbarajas: INIT_NET_STATIC is the ORed value of 0x100 and INIT_NET.  So would this give a false positive if INIT_NET is used instead of INIT_NET_STATIC?

andressbarajas:
I believe when it comes to game releases, the DHCP [net_init()] method will be used.  This PR is developer focused to give non-DHCP people the power to play with networking without waiting for DHCP timeout(45-60 secs).

darcagn:
I'm not really understanding the need for this change?

- If you're loading normally (i.e. not loading from dcload-ip), then
  - If you have static IP set in your netcfg, then it will use the static IP without waiting for DHCP timeout 
  - If you have DHCP set in your netcfg, then it will use DHCP
  - If you have no netcfg, then it will use DHCP (I think? Not sure here, but that's what it _should_ do IMO)

So there shouldn't be any problem there. But where it gets tricky is if you're loading from dcload-ip, then it will try to inherit the IP from dcload-ip, but because dcload-ip only manages an IP address for incoming connections and nothing else, KOS doesn't know if the IP is static or DHCP, lease length if DHCP, or default gateway, subnet mask, DNS server address, etc. 

What KOS does right now is just tries to grab this stuff via a DHCP request using the dcload-ip IP address, but I'm not sure if it is taking into account netcfg settings, but I believe you said you had no netcfg set up? Do you now?

IMO it _should_ check if DHCP is enabled in netcfg and DHCP request using the dcload-ip IP address if so, and if static IP is enabled use the settings from netcfg. I just don't know what the behavior should be if the dcload-ip IP address doesn't match the static IP in netcfg. 

But it seems like this PR is adding a new API that would *only* have any usefulness if *all* of the following are true:
 - the app is loaded via dcload-ip
 - the app makes no outgoing connections, only accepts incoming connections
 - the user (developer) has no DHCP server or doesn't want to use DHCP
 - the user (developer) does not want to manually set up networking and really wants an init flag for this _exact_ scenario
 - the user (developer) does not want to set up a netcfg (if we fix it to take into account netcfg settings)

If my assumptions above are true and I'm not missing something, that's waaay too niche of a scenario to justify adding this special function. Please let me know if I have overlooked something.

andressbarajas:
I dont not have a netcfg setup on my Dreamcast, yet.  I wanted to test the code first with no netcfg set (which requires other applications to do so. Im thinking there may be more Dreamcasts with no settings set so I think it is important to handle that case).



>   * If you have DHCP set in your netcfg, then it will use DHCP
Can you show me where?  All I found is this:  https://github.com/KallistiOS/KallistiOS/blob/eb3ed416c8ddf2038376e362a15939b057b37628/kernel/arch/dreamcast/hardware/network/broadband_adapter.c#L1081

>   * If you have no netcfg, then it will use DHCP (I think? Not sure here, but that's what it _should_ do IMO)

Yep, thats how it works.

> But it seems like this PR is adding a new API that would _only_ have any usefulness if _all_ of the following are true:
> 
> * the app is loaded via dcload-ip
> * the app makes no outgoing connections, only accepts incoming connections
> * the user (developer) has no DHCP server or doesn't want to use DHCP
> * the user (developer) does not want to manually set up networking and really wants an init flag for this _exact_ scenario
> * the user (developer) does not want to set up a netcfg (if we fix it to take into account netcfg settings)
> 
> If my assumptions above are true and I'm not missing something, that's waaay too niche of a scenario to justify adding this special function. Please let me know if I have overlooked something.

If the consensus is that its way too niche then yeah I agree that this change isnt needed.

andressbarajas:
> We should be accommodating to your scenario, but rearranging and expanding the API and creating init flags for this very niche scenario isn't the right way to do it. You should set a static IP configuration in your netcfg and then when KOS inits the network and inherits the IP from dcload-ip, I believe the correct behavior from KOS should be to check for a static IP in netcfg before reaching for DHCP (unless I'm missing a scenario in which this would cause a problem).

Yea, Im not too sure either.  If for some reason the program needs to use the same IP provided by DHCP from dcloadip.  I actually don't find a lot of use running INIT_NET without using the -n option for dctool which makes the ip always 0.

> Alternatively if you want to avoid all of that, just leave `INIT_NET` out of the application and then manually bring up the network functions with your static IP info. You were preventing the binary from being distributable to others by hardcoding the `INIT_NET_STATIC` into your binary anyway, so hardcoding your network settings shouldn't be any more of an inconvenience.

If you do that, providing an ip(non-zero) to net_init() will trigger a DHCP request first before manually setting IP.  If you do not provide an IP then it will try and use netcfg and if that wasnt setup then it will use dhcp(0). Which was the main motivation for this PR, try static first before trying DHCP.

> Since KOS's netcfg lets you specify a network configuration file from any dir, you could possibly even put your network configuration in `/pc/` and load it from there, I'm not sure if dcload-ip would have any issues retrieving it and delivering it to KOS before KOS itself inits network. Or since you're willing to bake it into your binaries just put the netcfg on a romdisk and load it. 

Can we still write to /pc/ when using the -n dctool option? Like I said earlier in this post, I haven't found a reason to use INIT_NET without it being headless because it causes issues. Have you had any luck using INIT_NET without -n?

PR #655, opened 2024-07-03
Replace redefined regs with dmac.h definitions

andressbarajas:
Found some redefined registers.


gyrovorbis (approved):
Good call. Those registers didn't even belong in the already-super-cluttered `init.c` anyway!

PR #656, opened 2024-07-03
Separate screenshot logic

andressbarajas:
Im working on a web utility lib that can possibly run in the background of the main application.  I think it would be cool if I can request screenshot data whenever I want outside of the main application.  So I separated the screenshot logic in the two parts.  One function that generates the data and the other that deals with writing that data to a file.  No change in current behavior besides providing vid_screen_shot_data(); for uses other than writing that data to a file.

### DIFF
--- a/include/kos/version.h
+++ b/include/kos/version.h
@@ -98,7 +98,7 @@
 */
 
 #define KOS_VERSION_MAJOR   2   /**< KOS's current major revision number. */
-#define KOS_VERSION_MINOR   0   /**< KOS's current minor revision number. */
+#define KOS_VERSION_MINOR   1   /**< KOS's current minor revision number. */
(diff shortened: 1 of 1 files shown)


gyrovorbis:
HEY. Stop making the APIs flexible and multi-purpose like that! 

jk jk. Totally reasonable API addition. lol.

QuzarDC (changes requested):
Great and useful idea. Aside from the minor comments, the function should also be added to the export files, as it'll be useful there.

One question though on the overall thing. Would it make more sense to have the function allocate/return *just* the data without the header and have `vid_screen_shot` handle writing the header? Since PPM is just raw pixels anyways I'd imagine that many wanting to use the new `data` function would be stripping off the ppm file header anyways to just get the data. It shouldn't be too bad to do two fwrites (first for header, then for rest of buffer).

PR #659, opened 2024-07-04
Iso9660: Read CD-ROM sectors using DMA

pcercuei:
Use the DMA to read sectors from the CD-ROM instead of reading them using PIO.

This frees the CPU to work on other tasks while the data transfer is ongoing.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_iso9660.c
+++ b/kernel/arch/dreamcast/fs/fs_iso9660.c
@@ -271,7 +271,7 @@ static int bread_cache(cache_block_t **cache, uint32 sector) {
     }
 
     /* Load the requested block */
-    j = cdrom_read_sectors(cache[i]->data, sector + 150, 1);
+    j = cdrom_read_sectors_ex(cache[i]->data, sector + 150, 1, CDROM_READ_DMA);
 
     if(j < 0) {
         //dbglog(DBG_ERROR, "fs_iso9660: can't read_sectors for %d: %d\n",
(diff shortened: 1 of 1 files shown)


DC-SWAT:
It would be great to do multi-sector reading, since if you read on 1 sector, the CPU usage is still high.

pcercuei:
> It would be great to do multi-sector reading, since if you read on 1 sector, the CPU usage is still high.

It's not really possible right now with the current ISO9660 filesystem code.

DC-SWAT:
> > It would be great to do multi-sector reading, since if you read on 1 sector, the CPU usage is still high.
> 
> It's not really possible right now with the current ISO9660 filesystem code.

I know, that's what I'm talking about :)
BTW, part of the code trying to implement this is still there.

gyrovorbis (approved):
Thanks!

kapodamy:
The DMA was disabled in commit https://github.com/pcercuei/KallistiOS/commit/44a62b490650c4da1fd386da594204aded37e9e1 (very old commit) due:
> XXX: DMA Mode may conflict with using a second G1ATA device. More testing is needed from someone with such a device.
¿its already fixed?

QuzarDC:
> The DMA was disabled in commit [44a62b4](https://github.com/KallistiOS/KallistiOS/commit/44a62b490650c4da1fd386da594204aded37e9e1) (very old commit) due:
> 
> > XXX: DMA Mode may conflict with using a second G1ATA device. More testing is needed from someone with such a device.
> 
> ¿its already fixed?

It wasn't disabled there, that commit also allowed it for the first time, it just wasn't made the default for reading due to potential conflict. This set of changes was primarily to align the cache data so that it could be written to by dma for iso9660, but also shouldn't have made DMA mandatory. That being said, there shouldn't be any negative behavior, as we use the g1ata mutex to prevent both cd and g1ata device from working at the same time. 

At the time I wrote the comment I was being overly cautious because g1ata support had been added during the time that I was working on and testing the change to add the dma reading mode generally.

Issue #654, opened 2024-07-03
KOS examples issues


andressbarajas:
I found issues with the following examples.  Most of them have on exit issues.  Forgive the formatting, I will fix later:

KOS_CFLAGS:
 -O2 -fomit-frame-pointer -fno-builtin -m4-single-only -ml -ffunction-sections -fdata-sections -matomic-model=soft-imask -ftls-model=local-exec -D__DREAMCAST__  -I/opt/toolchains/dc/kos/include -I/opt/toolchains/dc/kos/kernel/arch/dreamcast/include -I/opt/toolchains/dc/kos/addons/include -I/opt/toolchains/dc/kos-ports/include -D_arch_dreamcast -D_arch_sub_pristine -Wall -g

1. exec Seg fault: (DCLOAD IP ONLY)
maple: attached devices:
  A0: Dreamcast Controller           (01000000: Controller)
  C0: Dreamcast Mouse                (00020000: Mouse)
  D0: Keyboard                       (40000000: Keyboard)

Hello world from the exec.elf process
sub.bin mapped at 8c033c70, jumping to it!

arch: shutting down kernel
maple: final stats -- device count = 3, vbl_cntr = 4, dma_cntr = 4
vid_set_mode: 640x480IL NTSC with 1 framebuffers.
fs_romdisk: unmounting image at 0x8c033bf0 from /rd
Obviously bad packet, avoiding segfault
make: *** [run] Segmentation fault: 11

2. mmu/nullptr freeze (DCLOAD IP ONLY)

andressbarajas:
@QuzarDC @gyrovorbis I updated the list to remove the issues fixed by PR #660.

andressbarajas:
@QuzarDC I updated the list to remove issue fixed by PR #667

pcercuei:
>     2. mmu/nullptr freeze (Seems to be dcload-ip only?)

dcload-ip seems to have problems with MMU usage, yes.
Related ticket: https://github.com/sizious/dcload-ip/issues/22

andressbarajas:
Closing this issue now that the only remaining examples are broken when uploaded using dcload-ip.  So they are dcload-ip issues and not example issues.

PR #660, opened 2024-07-04
Fix example freezes/assertions

andressbarajas:
This seems to fix MOST of the freezes and the assertion errors mentioned in this issue: https://github.com/KallistiOS/KallistiOS/issues/654.

### DIFF
--- a/kernel/arch/dreamcast/kernel/init.c
+++ b/kernel/arch/dreamcast/kernel/init.c
@@ -249,12 +249,12 @@ void  __weak arch_auto_shutdown(void) {
 #if defined(__NEWLIB__) && !(__NEWLIB__ < 2 && __NEWLIB_MINOR__ < 4)
     fs_rnd_shutdown();
 #endif
+    fs_shutdown();
     fs_ramdisk_shutdown();
     KOS_INIT_FLAG_CALL(fs_romdisk_shutdown);
     fs_pty_shutdown();
     fs_null_shutdown();
     fs_dev_shutdown();
-    fs_shutdown();
     thd_shutdown();
     rtc_shutdown();
 }
--- a/kernel/fs/fs.c
+++ b/kernel/fs/fs.c
@@ -169,22 +169,18 @@ static void fs_hnd_ref(fs_hnd_t * ref) {
    to a raw handle is no longer applicable. This function may destroy the
    file handle, so under no circumstances should you presume that it will
    still exist later. */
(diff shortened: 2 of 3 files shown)


QuzarDC (approved):
So, just wanted to make sure to add some background here as to the issue and causes:

In #599 I had set `fs_shutdown` to call `fs_fdtbl_destroy` in order to prevent leaking unclosed handles. Prior to this `fs_fdtbl_destroy` was not used at all for some reason and `fs_shutdown` was just an empty stub. It was a loop over the full fd table using `fs_hnd_unref` to clear it out. Since it had never been used, I presumed it had just always been broken and forced it to skip fds 0-2 as I saw crashes on shutdown doing so.

In #636 I suggested that the the fixes to fs_pty might allow the exception for fds 0-2 to be removed. The function was then updated to force the close rather than to check refs, and it seems we all had misunderstood that the refs were not the number of open references to a *file* but to the file handle in the table. So the `fs_hnd_unref` *was* correct as the only way to get over 1 refcnt in a file handle is if `fs_dup` was used in order to have multiple entries filled with the same handle object. The only way that a loop over the table calling unref should leave any handles would be if there were some sort of corruption, and no combinations of public functions should allow it to happen.

With the shift to forcing the closing and freeing of each handle in the table, we were causing a double-free of any `dup`'d handles, namely fd 0 and 1 which are `dup`'d during initialization. Once #608 was merged, where multiple threads existing during shutdown was more common, that double-freeing expressed in all sorts of problems as noted in #654.

The actual root of the problem for all of it was much more simple. `fs_shutdown` was being called at the wrong time. It was being set to shut down after all the vfses had, which seemed cromulent, but was actually backwards. The top-level `fs_shutdown` should be called first, because it is intended to go through each table entry and call the registered handle's close function. This ensures that file handles left open by user code are cleanly closed out. In the way it was arranged previously, all the installed vfs handlers were being shut down and cleaned out *then* the top-level handles would try to close them.

So there are three functional changes in the PR:
1) Roll back the behavior of `fs_fdtbl_destroy` to how it was originally, before #599. This ensures that both all fds are cleaned up and that it's safe for use with dup'd files.
2) Have `fs_shutdown` run before shutting down any of the vfses.
3) The rearrangement in `fs_hnd_unref` ensures that even if the handle has no handler registered or the handler has no close function, the file handle still gets freed. Previously freeing the handle was locked behind the check for those.

PR #661, opened 2024-07-05
Added POSIX settimeofday()

gyrovorbis:
This function was already being declared by newlib, but for whatever reason was never implemented for SH... It's the more commonly used one compared to `clock_settime()`, is supported by the PSPSDK (but not us till now!), and can literally be implemented by forwarding the call straight through to `clock_settime()`. 

- Newlib declares this function for us but never bothers to defined it
  for our target.
- It can be defined trivially by just forwarding the call to
  `clock_settime()` and converting usecs to nsecs.
- Made `rtc_set_unix_sets()` return standard POSIX codes, so that
  `clock_settime()` and `settimeofday()` can simply forward the call and
have `errno` set automatically from the RTC driver (which now has `errno`
support)
    * updated documentation.


pcercuei:
Reading your description, isn't the real fix adding it to Newlib?

gyrovorbis:
> Reading your description, isn't the real fix adding it to Newlib?

Absolutely, I'm not going to lie... but like... This is the model we've used for LOTS of stuff that is technically a newlib issue or isn't implemented in Newlib yet is declared in Newlib...

We historically have not been playing ball with Newlib and upstreaming or anything. We've handled it on the KOS-side which we're in control of...

pcercuei:
You just said it was just not implemented for SH?

gyrovorbis:
Yeah, it's not implemented for SH. It's just prototyped for SH. I'm guessing the SH-Linux guys actually implement it on their end (using the SH4 RTC, unlike us having to use the AICA RTC on DC)? Then we just implement it on our end here like this for KOS, and there's not really any code reuse or shared code reason to have a common hook or implementation here?

darcagn (approved):

PR #663, opened 2024-07-06
Fixed GCC9+10 Build Failures: pthread_atfork()

gyrovorbis:
1) pthread_atfork()'s implementation was not declaring agument variables
   by name, which was not supported in ancient GCC9 and GCC10,
apparently.
2) pthread_atfork() was added as a symbol for CMake checks for pthreads
   to pass, yet we were never declaring the symbol publicly anywhere,
and implicit declarations of C functions are now deprecated in GCC14+.

### DIFF
--- a/include/pthread.h
+++ b/include/pthread.h
@@ -62,7 +62,8 @@ extern "C" {
         NOTE: RTEMS does not provide pthread_atfork().  */
 
 #if !defined(__rtems__)
-// #warning "Add pthread_atfork() prototype"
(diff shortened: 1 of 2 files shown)


pcercuei:
`s/agument/argument/` in the commit message (that will train you to git-rebase ;))

Aren't anonymous function arguments a C11 thing? Since we're compiling KallistiOS for C11 now...

QuzarDC (approved):

PR #664, opened 2024-07-06
cache: Fix ranges of i/d cache flush/inval

pcercuei:
The algorithms were flushing / invalidating the data using the start address in r4, and the end address (start + size) in r5.

They would then flush / invalidate using the address in r4, check "r5 >= r4", then add 0x20 to r4 and loop if true.

If we have r4 == 0x0 and r5 == 0x20, it means that the address 0x0 would be flushed / invalidated as expected, but the address 0x20 would also be flushed / invalidated. Even worse, the address 0x40 would be flushed too, as the range check is performed before the 0x20 increment.

This isn't a big deal when just flushing data to cache (it's just slightly more expensive), but is a huge problem when invalidating data. For instance, if we want to invalidate 32 bytes at a given address in preparation of a DMA read, we absolutely do not want to invalidate any data present after the end of the DMA area, or we risk data corruption.

Address this issue by updating the value of r4 before the range check, and using "r5 > r4" instead of "r5 >= r4" as the loop continue condition.

### DIFF
--- a/kernel/arch/dreamcast/kernel/cache.s
+++ b/kernel/arch/dreamcast/kernel/cache.s
@@ -66,7 +66,7 @@ _icache_flush_range:
     and      r3, r7
 
     add      #32, r4       ! Move on to next cache block
-    cmp/hs   r4, r5
+    cmp/hi   r4, r5
     bt/s     .iflush_loop
     mov.l    r7, @r6       ! *addr = data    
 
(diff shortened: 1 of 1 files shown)


andressbarajas (approved):
Thanks for the fixes.

gyrovorbis (approved):
Thank I you kindly for fixing the d-caches!!!

PR #668, opened 2024-07-08
Update BBA driver

andressbarajas:
Using dcache_inval_range instead should work now since zcrc fixed it


gyrovorbis (approved):
This looks like it was long overdue, and I stress tested the hell out of the BBA driver with Sonic Mania streaming everything from `/pc/` while spamming the hell out of `printf()`, seems solid!

Issue #666, opened 2024-07-08
Romdisk issue : undefined reference _romdisk_data (using cmake)


maishuji:
Hello, 
I' m not able to link romdisk anymore following kos update
Failed build can be found here : 
https://github.com/maishuji/dreamcast-in-devcontainer-example/actions/runs/9834310279/job/27145902985?pr=1
The issue is 

> undefined reference to `romdisk_data'

I tried to change _extern uint romdisk_ by _extern const uint romdisk_data_ in the main cpp, but still same issue, can you check this issue ? thanks !

I put bellow the logs : 

> -- The CXX compiler identification is GNU 14.1.1
> -- Detecting CXX compiler ABI info
> -- Detecting CXX compiler ABI info - done
> -- Check for working CXX compiler: /opt/toolchains/dc/kos/utils/build_wrappers/kos-c++ - skipped
> -- Detecting CXX compile features
> -- Detecting CXX compile features - done
> -- Configuring done (0.2s)
> -- Generating done (0.0s)
> -- Build files have been written to: /__w/dreamcast-in-devcontainer-example/dreamcast-in-devcontainer-example/build
> [ 20%] Generating romdisk.img
> 0    rom 1720418368       [0xffffffff, 0xffffffff] 37777777777, sz     0, at 0x0     
> 1    .                    [0x811     , 0x873f7   ] 0040755, sz     0, at 0x20    
> 1    ..                   [0x811     , 0x873f4   ] 0040755, sz     0, at 0x40     [link to 0x20    ]
> 1    default.txf          [0x811     , 0x873f8   ] 0100755, sz  9196, at 0x60    
> [ 40%] Generating romdisk_tmp.c
> [ 60%] Building CXX object src/clock_example/CMakeFiles/clock-example.dir/clock.cpp.obj
> [ 80%] Linking CXX executable clock-example.elf
> /opt/toolchains/dc/sh-elf/lib/gcc/sh-elf/14.1.1/../../../../sh-elf/bin/ld: /opt/toolchains/dc/kos/lib/dreamcast/libromdiskbase.a(romdiskbase.o):/opt/toolchains/dc/kos/kernel/romdisk/romdiskbase.c:11:(.data.__kos_romdisk+0x0): undefined reference to `romdisk_data'
> collect2: error: ld returned 1 exit status
> make[2]: *** [src/clock_example/CMakeFiles/clock-example.dir/build.make:105: src/clock_example/clock-example.elf] Error 1
[...]

PR #670, opened 2024-07-08
Fix cmake add_romdisk function

maishuji:
Hello, this is a fix proposal for the issue explained here :  https://github.com/KallistiOS/KallistiOS/issues/666
I compared it with the code from here, and made the change appropriately: https://github.com/KallistiOS/KallistiOS/blob/3f86969c5d4a0846a532823a208ea95a94879897/Makefile.rules#L73
I tested it on my project on linux

### DIFF
--- a/utils/cmake/dreamcast.cmake
+++ b/utils/cmake/dreamcast.cmake
@@ -31,6 +31,14 @@ if(NOT DEFINED KOS_CC_BASE)
     endif()
 endif()
 
+if(NOT DEFINED KOS_CC_PREFIX)
+    if(NOT DEFINED ENV{KOS_CC_PREFIX})
+        message(FATAL_ERROR "KOS_CC_PREFIX environment variable not found!")
+    else()
+        set(KOS_CC_PREFIX $ENV{KOS_CC_PREFIX})
+    endif()
+endif()
+
 ### Helper Function for Bin2Object ###
 function(kos_bin2o inFile symbol)
     # outFile is optional and defaults to the symbol name in the build directory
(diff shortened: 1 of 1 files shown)


pcercuei:
`cmake_add_romdisk` works perfectly fine in Bloom: https://github.com/pcercuei/bloom/blob/main/CMakeLists.txt#L318

I don't know what you are doing differently.

maishuji:
> @maishuji I think what happens is that CMake won't compile romdisk.o from romdisk.c because your `project()` line says that it is exclusively C++.
> 
> If you write instead:
> 
> ```cmake
> project(dc-cmake-examples LANGUAGES C CXX)
> ```
> 
> It will most likely work.

Hello, I checked, yes it works without the fix, when adding C language support in CMakeLists.txt as you mentionned.
Is there any benefit to keep the original behavior compared to the fix proposal ? I don't think it is intuitive having to add C support in a full CXX project, so if it is possible isn't it better to avoid having to do it?

pcercuei:
@maishuji see https://github.com/KallistiOS/KallistiOS/pull/677

gyrovorbis:
> > @maishuji I think what happens is that CMake won't compile romdisk.o from romdisk.c because your `project()` line says that it is exclusively C++.
> > If you write instead:
> > ```cmake
> > project(dc-cmake-examples LANGUAGES C CXX)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It will most likely work.
> 
> Hello, I checked, yes it works without the fix, when adding C language support in CMakeLists.txt as you mentionned. Is there any benefit to keep the original behavior compared to the fix proposal ? I don't think it is intuitive having to add C support in a full CXX project, so if it is possible isn't it better to avoid having to do it?

Ughghghg. Yeah, you're right. That behavior does kind of suck and is counter-intuitive to how I'd expect CMake to work... but on the other hand, building it the other way breaks with LTO enabled...

Actually... hold on... Can't we make an internal static lib target that has nothing but that romdisk `.c` file in it, which gets added to the linked libs for the main target, completely transparently in that CMake utility function?

PR #671, opened 2024-07-09
Network speedtest

andressbarajas:
Added a speed test example for BBA/LAN adapters that is based on httpd and Darc's work.  I will be MIA for a while due to work obligations so I created this PR in hopes that it will be useful and that if changes are needed it can be hijacked by someone to make further improvements/corrections.


QuzarDC (changes requested):
So the content looks good, the big thing that's missing is any info on expected use and such. I wouldn't expect significant detail, since this isn't quite meant to be an example on how to build your own stuff, but a few sentences in the header of handle_request.c or speedtest.c would go a long way.

andressbarajas:
@QuzarDC Addressed your request.

QuzarDC (approved):

PR #673, opened 2024-07-11
Don't allow dbgio until it is initted

QuzarDC:
Opened #798 to track the ongoing issue and moving this one along.

This change ensures that dbgio is marked as disabled until it is initted. Before this it was always on from the start which would break the rare use of it prior to `dbgio_init` being called (such as malloc debugging).


gyrovorbis (approved):
OMG GOOD FIND.

PR #674, opened 2024-07-13
Ensure thd_current does not get set to null

QuzarDC:
Before this `genwait_wait` was setting `thd_current` to NULL as a way to signal to the scheduler that it should not be re-enqueued. This would cause the thread time updating introduced in #506 to write null deref on any genwait_wait.

It seems very much like that check was pointless, and perhaps a relic of older threading modes. The setting to null would be used to set the variable `dontenq` to 1 which would exempt the thread from getting rescheduled. Each of the two checks that checked against that though would always then validate that the `thd_current->state == STATE_RUNNING` which is always false while `thd_current` is null because that was always followed by setting `thd_current->state == STATE_WAIT`.

So have entirely removed that set of checks that seemed to be redundant. This allows thd_current to not get set to null ever and `thd_update_cpu_time` to work correctly.

### DIFF
--- a/kernel/thread/genwait.c
+++ b/kernel/thread/genwait.c
@@ -20,11 +20,9 @@
 #include <assert.h>
 #include <errno.h>
 
-#include <arch/arch.h>
 #include <arch/timer.h>
 #include <kos/genwait.h>
 #include <kos/sem.h>
-#include <kos/dbgio.h>
 
 /* Our sleep queues table. This is also modeled after the BSD numbers. I
    figure if they've been using it as long as they have, they must be
--- a/kernel/thread/genwait.c
+++ b/kernel/thread/genwait.c
@@ -20,9 +20,11 @@
 #include <assert.h>
 #include <errno.h>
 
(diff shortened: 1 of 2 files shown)


andressbarajas:
@gyrovorbis You think it might fix the weirdness you were experiencing here:  https://github.com/KallistiOS/KallistiOS/pull/537 ?

QuzarDC:
> @gyrovorbis You think it might fix the weirdness you were experiencing here: #537 ?

None of the issues were noted in that draft, but what this would have caused is that any thread that called `gen_wait()` would have not had the time spent prior to it switching from the wait counted. This should have caused the per-thread cpu time tracking to be wrong in most real-world cases. The only time it would be correct would be if every context switch was due to the scheduler or `thd_pass()` (so never waiting on a mutex, never doing `thd_sleep()`, etc).

gyrovorbis (approved):
Holy SHIT. Good find! You can even see the timing accuracy improve in the general threading tests for per-thread CPU time... great!

PR #677, opened 2024-07-16
CMake: Support romdisks in C++ projects

pcercuei:
When the project is C++ only, any .c file added to the target sources will be ignored by CMake. This meant that the source file generated by kos_add_romdisk() macro would not be compiled.

Address this issue by changing the extension of the generated source file to .cpp when a C++ compiler is used.

This requires a small modification to bin2c so that it generates code that can also be compiled by a C++ compiler.

### DIFF
--- a/utils/bin2c/bin2c.c
+++ b/utils/bin2c/bin2c.c
@@ -26,7 +26,10 @@ void convert(char *ifn, char *ofn, char *prefix) {
     fseek(i, 0, SEEK_SET);
     setbuf(o, buf);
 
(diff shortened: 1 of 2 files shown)


gyrovorbis:
Oh shit, didn't realize you could query for whether a C++ compiler was present or not like that in CMake... very cool.

maishuji:
Tested on my CXX example, no issue :100:

gyrovorbis:
> Tested on my CXX example, no issue 💯

Thank you kindly for testing!

gyrovorbis (approved):
Sweet. This should solve it.

Issue #649, opened 2024-07-01
Have a vfs for 'virtual directory'


QuzarDC:
The current implementation of `/dev` does not support having multiple open readdirs. This stems from the use of a simple static single file handle for the /dev folder. The way to correct this seems to be to just have it generated so that each open will have its own, like a more regular vfs. 

Though, if it is to be expanded in that way, it may make sense instead to create a more generic `virtual directory` vfs that could cover `/dev` and `/` along with maybe `/vmu` since all of them effectively shouldn't be anything but a directory that has other vfs mounted under it. This would also give a better path and likely more sensical way to handle things like mounting an rd to subdir (if you were to mount a romdisk to `/rds/myrd/` what vfs owns `/rds/`?) .

pcercuei:
Can't we use the existing ramdisk code for that?

`/ram` could actually be moved to be the root (which would become R/W), allowing for directories to be created, to which other filesystems can then be mounted.

QuzarDC:
> Can't we use the existing ramdisk code for that?
> 
> `/ram` could actually be moved to be the root (which would become R/W), allowing for directories to be created, to which other filesystems can then be mounted.

That's not a bad idea. I hadn't gotten back to this after #818 which was a blocker and seemed like should be separate. What would still be needed is some distinction to allow unmounting a vfs to remove the virtual paths that were inserted just for it. Or perhaps that would be such a mess that they should just stick around (esp since it's kind of silly to cater to these 'cleanup'/'shutdown' edgecases).

PR #680, opened 2024-07-17
Another round of KOS_INIT_FLAG() modularization.

gyrovorbis:
The following new KOS_INIT_FLAG() values have been added; however, none of the defaults have changed, so they are all already enabled as part of INIT_DEFAULT.

1) INIT_FS_DEV
    - enable /dev/ VFS
2) INIT_FS_NULL
    - enable /dev/null VFS
3) INIT_FS_PTY
    - enable /pty
4) INIT_FS_RAMDISK
    - enable /ram
5) INIT_FS_RND
    - enable /dev/urandom
6) INIT_LIBRARY
    - enable dynamic library loading


QuzarDC:
So, one thing I can point out is that if you disable `/dev/` then navigating to the devices under `/dev/` will be broken. I think if we do make it modularizable, then we also need to have the dependencies forced where its only disabled if all dev vfs are disabled, or additionally where we have the name manager automatically set up virtual directories if we request names in subdirectories (related to #649). In that scheme then there'd also be no option to disable `/dev` as it would only appear if any device vfs is registered under it.

Also is it required that the functions in init calls only return void? Or is that unrelated?

pcercuei:
I actually wanted to do it, but @QuzarDC had a good argument against it: how many bytes is it saving? Is that really worth the effort? It's not like you're compiling out the network stack or something like that.

pcercuei:
@gyrovorbis You actually do not need a `INIT_FS_DEV` flag. Just use `INIT_FS_ALL` (without `INIT_FS_DEV`) in the `KOS_INIT_FLAG()` calls.

QuzarDC (changes requested):
As discussed in the comments (reiterating with review in order to ensure this is flagged as 'request changes'):
What kind of difference does this make to the memory usage? It seems very much like these would all save almost nothing. Unless somehow the library subsystem is forcing exports?

Additionally disabling `/dev/` *without* disabling the vfs under it could lead to issues with accessing them.

gyrovorbis:
I just ran two basic tests disabling pretty much everything on the `hello.c` example, and rebuilding it along with KOS at `-Os` for both `master` as the baseline minimal binary as well as this PR (after merging master into it, to be fair) to compare the sizes of the produced binaries with everything this PR offers as modularizable disabled. 

`hello` example on `master` with `KOS_INIT_FLAGS(INIT_IRQ)`:
```
Section .text, lma 0x8c010000, size 115772
Section .init, lma 0x8c02c43c, size 54
Section .fini, lma 0x8c02c474, size 42
Section .rodata, lma 0x8c02c4a0, size 36496
Section .eh_frame, lma 0x8c035330, size 292
Section .ctors, lma 0x8c0354d4, size 8
Section .dtors, lma 0x8c0354dc, size 8
Section .data, lma 0x8c0354e4, size 5172
Section .got, lma 0x8c036918, size 12
Transferred 157856 bytes at 60832051.839765 bytes / sec
```

`hello` example on `master` with `KOS_INIT_FLAGS(INIT_IRQ)`:
```
Section .text, lma 0x8c010000, size 77416
Section .init, lma 0x8c022e68, size 54
Section .fini, lma 0x8c022ea0, size 42
Section .rodata, lma 0x8c022ecc, size 7324
Section .eh_frame, lma 0x8c024b68, size 4
Section .ctors, lma 0x8c024bec, size 8
Section .dtors, lma 0x8c024bf4, size 8
Section .data, lma 0x8c024bfc, size 2928
Section .got, lma 0x8c02576c, size 12
Transferred 87796 bytes at 33691044.280329 bytes / sec
```

Soooo... essentially the savings yielded by disabling everything that this PR now makes optional:
- `.text`: 38356 bytes
- `.rodata`: 29172 bytes
- `.data`: 2244 bytes

This gives us back a total of `68.14` KB when built with `-Os`, which, while not tremendous`, is larger than the entire stack space for the main thread and could still be game changing for apps like GTA3, Sonic Mania, Xash3D, or some of the new ones fighting for every byte of RAM they can.

`-Os` was chosen as the pathological *worst-case* scenario to flex gainz in this PR, so it would give the lower boundary on what it can save you, as `-Os` produces way smaller code, so there's less code that can get optimized away...

Now what's compelling AF, is if I run the same test at `-O3`, we save a whopping `764.98` KB of RAM, which is pretty tremendous! 

I'm not under any illusion that all flags are contributing equally here, I just made everything configurable for consistency's sake. If you would like for me to take time measuring *individual* gainz and getting rid of the ones that are miniscule, that's fine, but I do believe that overall, the total gainz for this PR are worth pursuing...

pcercuei:
How did you count the bytes? The zero-initialized data will be put into the BSS section, which is not packed inside the executable, so maybe the space saved is bigger.

gyrovorbis:
> How did you count the bytes? The zero-initialized data will be put into the BSS section, which is not packed inside the executable, so maybe the space saved is bigger.

..........ugh. I'm an idiot. I was first adding up the bytes of the various segments transferred by DC-Tool, but then I realized they were equal to the total mentioned in the final "X bytes transferred" line, but... YEAH, NO SHIT, THERE IS NO SIZE RESERVED IN THE BINARY FOR THE .BSS SEGMENT.

include/kos/init.h
```diff
@@ -139,29 +139,33 @@ extern const void * __kos_romdisk;
...
-#define INIT_FS_PTY      0x00000200  /**< Enable support for PTY VFS */
-#define INIT_FS_RAMDISK  0x00000400  /**< Enable support for ramdisk VFS */
-#define INIT_FS_RND      0x00000800  /**< Enable support for /dev/urandom VFS */
-#define INIT_FS_DEV      0x00001000  /**< Enable support for /dev/ VFS */
-#define INIT_NO_SHUTDOWN 0x00002000  /**< Disable hardware shutdown */
+#define INIT_NET         0x00000002  /**< Enable built-in networking */
+#define INIT_MALLOCSTATS 0x00000004  /**< Enable malloc statistics */
+#define INIT_QUIET       0x00000008  /**< Disable dbgio */
+#define INIT_EXPORT      0x00000010  /**< Export kernel symbols */
+#define INIT_LIBRARY     0x00000020  /**< Enable support for dynamic libraries. */
```
DC-SWAT, line 160: `INIT_EXPORT` and `INIT_LIBRARY` is the same subsystem. Why does it need separate flags?

gyrovorbis: Ehhhhh... not quite. One of them is for exporting KOS's symbols for external libraries to make KOS calls, the other is for just loading external libraries themselves, which could be used without needing to link back to the main binary to make such KOS calls...

But now that you are mentioning it, I suppose that would be pretty silly, as... what the heck are you going to do in a dynamic library with access to essentially nothing? Okay, point taken. Will combine the two when I get home, but I would prefer `INIT_LIBRARY` then to handle both, as `INIT_EXPORT` doesn't seem as descriptive or to include the dynamic library loader itself, based on the name.

pcercuei: Better not break the API for no reason though.

gyrovorbis: Alrighty, that's legit with me. Good call.

DC-SWAT: > Ehhhhh... not quite. One of them is for exporting KOS's symbols for external libraries to make KOS calls, the other is for just loading external libraries themselves, which could be used without needing to link back to the main binary to make such KOS calls...
> 
> But now that you are mentioning it, I suppose that would be pretty silly, as... what the heck are you going to do in a dynamic library with access to essentially nothing? Okay, point taken. Will combine the two when I get home, but I would prefer `INIT_LIBRARY` then to handle both, as `INIT_EXPORT` doesn't seem as descriptive or to include the dynamic library loader itself, based on the name.

So you won't load the library without exports. I mean, it can't do much without the KOS, stdlib and other APIs that need to be exported.

gyrovorbis: Alrighty, values have been readjusted to be the same for `INIT_EXPORT` and `INIT_LIBRARY`.

ljsebald (approved):

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -183,20 +195,15 @@ int  __weak arch_auto_init(void) {
...
-
-/* The arc4random_buf() function used for random & urandom is only
-   available in newlib starting with version 2.4.0 */
-#if defined(__NEWLIB__) && !(__NEWLIB__ < 2 && __NEWLIB_MINOR__ < 4)
-    fs_rnd_init();          /* /dev/urandom etc. */
-#else
-#warning "/dev filesystem is not supported with Newlib < 2.4.0"
-#endif
+    if(__kos_init_flags & INIT_FS_ALL)
+        fs_init();                        /* VFS */
```
QuzarDC, line 199: So, this doesn't map quite right. Since `INIT_FS_ALL` doesn't capture loadable (SD/G1ATA) or arch-specific FSes (`/cd/` `/pc/` `/vmu/`) it would prevent initializing the vfs system if you don't want this set of FSes.

Now ... of course currently it's somewhat moot cause fs_init doesn't do anything, but that's totally a place that should be rewritten to have a dynamically allocated table which would be done via fs_init, and just formally it doesn't match. And ... since it doesn't do anything, gating it shouldn't really save any memory.

ljsebald: I've created an issue to track adjusting this (#990), rather than leaving this PR here to rot any longer.

PR #683, opened 2024-07-25
Update PVR_PACK_COLOR macro to handle expression params

andressbarajas:
Fix PVR_PACK_COLOR macro to ensure proper color packing by handling expressions passed to the macro


pcercuei:
Commit message could be `Fix PVR_PACK_COLOR macro` (with the same commit description).

andressbarajas:
Sorry, I did this through the web because I was away from my DC dev computer right before bed. I'll see if I can fix it now.

andressbarajas:
@pcercuei Done :) Thanks for pushing me

gyrovorbis (approved):
Everyone pay attention here. Textbook reason why we put macro args in parenthesis!

PR #685, opened 2024-07-28
Update minimum macOS version for toolchain build

andressbarajas:
Updated the CXX line to set -mmacosx-version-min to 10.14 from 10.7 to support building the toolchain on M1 Macs. This ensures compatibility with newer macOS versions.


gyrovorbis (approved):
Thanks for fixing the MacOS toolchains!

Issue #686, opened 2024-07-31
RAMdisk filenames with similar prefixes


azihassan:
Hello everyone, I hope you're doing well.

I ran into a strange issue with ramdisk files that start with the same values, for example "/ram/sv_0_hero" and "/ram/sv_0_heroitems" (save files from the Diablo game). It seems like this causes confusion to the fs_ramdisk code. Below is a working example. It mounts two files, /ram/foo and /ram/bar. It writes "Hello world from /ram/foo" and "Hello world from /ram/foo" to them respectively, reads them back to stdout, then unmounts them:

```c
#include <assert.h>
#include <stdio.h>
#include <malloc.h>
#include <string.h>
#include <kos/fs.h>
#include <kos/fs_ramdisk.h>
#include <libgen.h>


void listdir(const char *dir, int depth) {
	printf("Listing contents of DIR %s\n", dir);
	file_t d = fs_open(dir, O_RDONLY | O_DIR);
	dirent_t *entry;
	while(NULL != (entry = fs_readdir(d))) {
		bool isDir = entry->size == -1;
		printf("[%s] %s\n", isDir ? "DIR" : "FIL", entry->name);
		if(isDir) {
			char nextDir[1024];
			strcpy(nextDir, dir);
			strcat(nextDir, "/");
			strcat(nextDir, entry->name);
			printf("nextDir = %s, depth = %d\n", nextDir, depth);
			listdir(nextDir, depth + 1);
		}
	}
	fs_close(d);
}

bool writefile(const char* path, const char* contents) {
	char* baseName = basename(path);
	size_t size = strlen(contents) + 1;
	void* buffer = malloc(size);
	printf("malloc OK for path %s (baseName = %s) ? %s\n", path, baseName, buffer != NULL ? "true" : "false");
	memcpy(buffer, contents, size);
	int attach_result = fs_ramdisk_attach(baseName, buffer, size);
	return attach_result != -1;
}

bool freefile(const char* path) {
	char* baseName = basename(path);
	void *toFree;
	size_t ignore;
	int detach_result = fs_ramdisk_detach(baseName, &toFree, &ignore);
	free(toFree);
	printf("fs_ramdisk_detach result for path %s (baseName = %s) = %d\n", path, baseName, detach_result);
	return detach_result != -1;
}

void readfile(const char* path) {
	void* contents;
	ssize_t size = fs_load(path, &contents);
[...]

PR #687, opened 2024-07-31
Fix ramdisk_find

andressbarajas:
Fix ramdisk_find by making sure we also compare the filename lengths so false positives wont occur.

Fixes issue #686

### DIFF
--- a/kernel/fs/fs_ramdisk.c
+++ b/kernel/fs/fs_ramdisk.c
@@ -107,7 +107,7 @@ static rd_file_t *ramdisk_find(rd_dir_t *parent, const char *name, size_t namele
     rd_file_t   *f;
 
     LIST_FOREACH(f, parent, dirlist) {
-        if(!strncasecmp(name, f->name, namelen) && strlen(f->name) == namelen)
+        if((strlen(f->name) == namelen) && !strncasecmp(name, f->name, namelen))
             return f;
     }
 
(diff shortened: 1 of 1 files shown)


kernel/fs/fs_ramdisk.c
```diff
@@ -103,11 +103,11 @@ static mutex_t rd_mutex;
 
 /* Search a directory for the named file; return the struct if
    we find it. Assumes we hold rd_mutex. */
-static rd_file_t * ramdisk_find(rd_dir_t * parent, const char * name, int namelen) {
+static rd_file_t *ramdisk_find(rd_dir_t *parent, const char *name, size_t namelen) {
     rd_file_t   *f;
 
     LIST_FOREACH(f, parent, dirlist) {
-        if(!strncasecmp(name, f->name, namelen))
+        if(!strncasecmp(name, f->name, namelen) && strlen(f->name) == namelen)
```
QuzarDC, line 110: Good catch. Two minor suggestions. 1) toss parens around the strlen comparison so that it's clearer. 2) swap it to the front since it's a faster check to do.
Something like:
`if((strlen(f->name) == namelen) && !strncasecmp(name, f->name, namelen))`

andressbarajas: Done

QuzarDC (approved):

PR #688, opened 2024-07-31
Replace duplicated define for the size of the PVR RAM

QuzarDC:
As on tin. Though I recognized when adding the define that this wasn't the right place for it, I hadn't at the time found where it was already defined. Instead of redefining, pull from pvr.h


gyrovorbis (approved):
Good catch. lmao.

PR #689, opened 2024-08-02
Upgrade to GCC 14.2.0

darcagn:
* dc-chain: Update 14.1.0 profile to GCC 14.2.0
* dc-chain: Update 14.1.1-dev profile to GCC 14.2.1 (git)

KOS patches apply cleanly to both toolchains, they both build, and they build KOS, kos-ports, and examples without error. 2ndmix runs with no obvious issues.

Now that 14.2.0 is out, we should test all examples with it and if working with no regressions, mark it as `stable` toolchain for KOS master/2.2.0. Since this PR does not stabilize 14.2.0, this PR can go into 2.1.0 though, if desired.


gyrovorbis (approved):
Sweeeet. Built fine for me over here! Everything seems to be working great!

PR #694, opened 2024-08-06
Fixed Overflow with Writing PVR BG Plane

gyrovorbis:
- skmp (correctly) reported that our little loop to copy the internally populated `pvr_bkg_poly_t` polygon (which gets color info from `pvr_set_bg_color()`) was looping too far.
    * structure is only 60 bytes, we looped for 64 and then wrote
      another word even beyond that.
- swapped implementation to use `memcp4y()` based on `sizeof(pvr_bkg_poly_t)` which is safer.
- removed final `vrl[0x11] = 0` assignment, as it clearly wasn't necessary and doesn't appear to even be doing anything useful at-all. The BG polygon types are special and don't need end-of-list flags, and interpolate the 4th vertex, more like PVR sprites.
- Adjusted size of GB plane to not be hardcoded to 640x480 and to instead be the size of the framebuffer resolution
- Changed BG plane's Z coords from `0.2f`, which seems sketchy to `FLT_EPSILON` to push it further back.

### DIFF
--- a/include/kos/time.h
+++ b/include/kos/time.h
@@ -113,11 +113,13 @@ extern __time_t timegm(struct tm *timeptr);
 #endif
 
 /* Explicitly provided function declarations for POSIX clock API, since
-getting them from Newlib requires supporting the rest of the _POSIX_TIMERS
-API, which is not implemented yet. */
+   getting them from Newlib requires supporting the rest of the _POSIX_TIMERS
+   API, which is not implemented yet. */
 extern int clock_settime(__clockid_t clock_id, const struct timespec *ts);
 extern int clock_gettime(__clockid_t clock_id, struct timespec *ts);
 extern int clock_getres(__clockid_t clock_id, struct timespec *res);
+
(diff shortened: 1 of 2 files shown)


andressbarajas:
I think has been tested on GTA if that means anything

QuzarDC (approved):
Tested against the pvr examples and they still look as expected. Had discussed this one fairly extensively in the discord. Just two notes:
1) We really have to rewrite the way this works. For whatever reason we're dynamically determining where to place this data every frame and copying it there, despite it effectively being static. We should reserve some specific area of vram for this and just update it whenever the parameters might change (on calling vid_set_mode or changing the background color). And, of course, demagic those params for it.
2) Should it be `FLT_EPSILON`? For sure 0.2f is problematic because it's not well advertised and you can end up accidentally behind it, but would it be worthwhile for it to be configurable? That could wait till it gets a whole rewrite, but am not sure what, if any, use cases there could be for that.

PR #697, opened 2024-08-08
Fix 24-bit / 32-bit video modes

pcercuei:
These video modes were only usable when writing the framebuffer directly, and not when using the PVR.

These two commits fix 24 / 32-bit support in KallistiOS so that these video modes can be used.


gyrovorbis (approved):
Looks good, but I'll lset @QuzarDC merge, cuz this is more his area...

gyrovorbis (approved):
Cool. Finally!

PR #698, opened 2024-08-08
Fix assumptions about current video width & height in pvr_misc.c

gameblabla:
Current resolution may be higher or lower than 640x480 (Like 768x480 or 768x576 and other resolutions that may get added later or custom modelines set by user), fix hardcoded values in pvr_misc.c


gameblabla:
I realized https://github.com/KallistiOS/KallistiOS/pull/694 was fixing it too + other things, so oops...

gyrovorbis:
> I realized #694 was fixing it too + other things, so oops...

Tbh, if I had noticed you had taken care of that too, I wouldn't have added it to my PR... I didn't see this one, though. WHOOPS!

PR #699, opened 2024-08-09
vmu: Add function vmu_printf()

pcercuei:
This function can be used to print a string to every connect VMU, using the built-in VMU font.


gyrovorbis (approved):
Hell yeah

Issue #690, opened 2024-08-03
BIOS font deadlock with G1 ATA devices.


gyrovorbis:
@DC-SWAT just reported the following, which is indicative of a pretty serious problem with the new locking mechanism introduced for the BIOS font.. capturing it so we don't forget about it. 

![image](https://github.com/user-attachments/assets/57d5214e-cea4-4664-81d3-2beefd37a0db)
![image](https://github.com/user-attachments/assets/1fdcd92f-f69c-4b6d-aba1-c065480b234a)

gyrovorbis:
I have verified this should now be fixed on `master`! Closing!

PR #700, opened 2024-08-10
Remove g1ata mutex use in biosfont

andressbarajas:
Removed mutex from biosfont because it causes an issue in DS bootloader as well as give the ability to use biofonts inside of an IRQ context.

Removing the mutex fixes this issue: https://github.com/KallistiOS/KallistiOS/issues/690


gyrovorbis (approved):
Thanks, this will do!!!

PR #703, opened 2024-08-10
Make FD_SETSIZE configurable

andressbarajas:
1. Moves the three existing #define FD_SETSIZE 1024 to kos/opts.h
2. Allows the value to be configurable in environ.sh
3. Add some dbglog message if the value is set too low and the user gets a runtime error when trying to open a file.


gyrovorbis (approved):
Lookin' good.

PR #706, opened 2024-08-11
Downsized stack sizes for idle and reaper threads

gyrovorbis:
- Both the idle and reaper threads are constructed with the default, 32KB stack size.
- Neither one of these need 32KB, and both of them are doing deterministic things...
- Created static buffers for both of their stacks.
- Idle thread gets 64 bytes of stack space.
- Reaper gets 512 bytes of stack space.
- Total space saved: 63.4375KB.


QuzarDC:
Was there some specific calculation done in order to determine those stack sizes or was it more just 'shrink it till it breaks'?

gyrovorbis:
> Was there some specific calculation done in order to determine those stack sizes or was it more just 'shrink it till it breaks'?

I kind of summed up the variables I saw on the stack plus gave some wiggle room for stack frame which was just an overly conservative estimate I pulled out of my ass--then, yes--not gonna lie, it was wiggle around there until it breaks... lol.

QuzarDC (approved):

PR #707, opened 2024-08-11
Cleaned up + optimized pvrmark_strips_direct bench

gyrovorbis:
- The pvrmark_strips_direct benchmark is an EXTREMELY important example, both for our ability to benchmark the PVR API's absolute fastest render path, AND as the only example we have given to users for how to use it.
- The benchmark happens to do a bunch of dumb things that are artifically limiting performance here and are making runs inconstistent.
- Removed branch from critical path
- Stopped clearing out unused vertex attributes like u, v, and oargb
- Stopped performing printf logic witin the timed sections
- Cleaned up benchmark
- Increased the rate at which the INCREMENT phase increased poly count (because it takes forever to normalize, atm)
- Added brief description


examples/dreamcast/pvr/pvrmark_strips_direct/pvrmark_strips_direct.c
```diff
@@ -120,23 +125,30 @@ void do_frame(void) {
...
+    nextnum();
+    col = getnum(256);
+    nextnum();
+    vert = pvr_dr_target(dr_state);
+    vert->flags = PVR_CMD_VERTEX_EOL;
+    vert->x = x;
+    vert->y = y;
+    vert->z = z;
+    vert->argb = col | (col << 8) | (col << 16) | 0xff000000;
+    pvr_dr_commit(vert);
```
QuzarDC, line 144: Instead of duplicating this, would it make sense to just be a function that the loop calls and is called for cleanup? Could be inlined to not have the overhead. 

At the very least, it seems like that first vert could be part of the loop by just moving the seed adjusting part to the end of the loop rather than the start so that it uses the right initial values. That'd have the side effect of leaving the seeds incremented for that final vert too.

examples/dreamcast/pvr/pvrmark_strips_direct/pvrmark_strips_direct.c
```diff
@@ -69,17 +75,18 @@ void setup(void) {
     pvr_poly_compile(&hdr, &cxt);
 }
 
-int oldseed = 0xdeadbeef;
+
+#define nextnum() seed = seed * 1164525 + 1013904223;
+#define getnum(mn) (seed & ((mn) - 1))
+
```
QuzarDC, line 81: I know you didn't write this bit, but since they're always used in a pair, it'd make sense and declutter a lot to combine them. If you don't wanna though, that's fine too, no need to clean up everything just because you want to clean up part of it :)

QuzarDC (commented):
Looks good, but left some comments on possible further streamlining. They aren't mandatory, so feel free to resolve them if they're no good, and this can just go in.

examples/dreamcast/pvr/pvrmark_strips_direct/pvrmark_strips_direct.c
gyrovorbis: Yeah, you're right. Cleaned.

gyrovorbis:
Screenshot after the mods, can hit +1 million pps, has better colors, has better spread of sizes:

Screenshot 2024-09-27 17-46-58

QuzarDC (approved):

PR #711, opened 2024-08-14
Fix kos doxygen warnings

andressbarajas:
This PR fixes warnings that show up when building the docs using ```make docs``` in kos folder.


gyrovorbis (approved):
Thanks!

PR #714, opened 2024-08-15
Add an example of how to render lines with the PVR.

jnmartin84:
One of the things I needed for my port of Doom 64 to the Dreamcast was a way to draw lines with hardware acceleration.

I adapted a method of taking the normal to a line segment and drawing a quad along the line as a triangle strip.

I would have loved to have had this available as an example when I needed it.

Provided is a function for rendering lines with the PVR and an example of rendering opaque and transparent lines in an interleaved fashion using `pvr_list_prim` .

Hoping this proves useful to others and might be accepted to the KOS project.

### DIFF
--- a/examples/dreamcast/pvr/Makefile
+++ b/examples/dreamcast/pvr/Makefile
@@ -6,6 +6,7 @@
 
 all:
 	$(KOS_MAKE) -C plasma
+	$(KOS_MAKE) -C pvrline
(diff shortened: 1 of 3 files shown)


gyrovorbis (approved):
Tested. Works fantastically. Absolutely top tier example! Thanks so much and congrats on being a contributor!

PR #719, opened 2024-08-19
Remove FD_SETSIZE setting from environ.sh

darcagn:
Both BlueCrab and I pointed out that this setting belongs in opts.h and does not belong in environ.sh, yet PR#703 was merged anyway. This PR reverts the change to environ.sh.

### DIFF
--- a/doc/environ.sh.sample
+++ b/doc/environ.sh.sample
@@ -144,21 +144,6 @@ export KOS_CFLAGS="${KOS_CFLAGS} -O2"
 #
 #export KOS_CFLAGS="${KOS_CFLAGS} -freorder-blocks-algorithm=simple -flto=auto"
 
-# RAM-Saving Optimization
-#
-# Uncomment the line below to reduce the default size of FD_SETSIZE from the
-# default value of 1024 to 64. FD_SETSIZE defines the maximum number of files
-# that can be opened simultaneously, including files on the ramdisk, romdisk, 
-# CD, VMU, network sockets, etc. The default setting allows 1024 files to be
-# opened at once, which is typically more than most use cases require.
-#
-# By reducing this value, you can save approximately ~11.25 KB of RAM, which 
-# can be particularly useful when porting large games to the Dreamcast that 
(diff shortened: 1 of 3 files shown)


andressbarajas:
Can you also update the fs.c file, specifically:
```c
dbglog(DBG_ERROR, "fs_hnd_assign: Update -DFD_SETSIZE flag in \
                  environ.sh to support additional files being opened. Current \
                  limit is %d\n", FD_SETSIZE);
```

darcagn:
Updated. What was the purpose of only giving the warning if `FD_SETSIZE != 1024`?

andressbarajas:
Hmmm. Well my thought process was that 1024 is the default. So if the user goes into KOS and changes that value in opts.h and then get a runtime error of running out of FD entries that would be due to them modifying that value to number that is not the default.  It should probably be FD_SETSIZE < 1024 OR just remove the if statement altogether.  Im cool with either.  

My main focus was to report to the user if they changed the default (1024).  If they still get issue with the default(1024) they are doing something very wrong. At least I would think 1024 files/sockets open at once is crazy.

 Do you think it would be a good idea to refactor the comment of the entry you are removing from environ.sh and sticking it above the FD_SETSIZE definition?

andressbarajas:
Looks all good to me!

darcagn:
Updated to fix the description in `opts.h`.

> So if the user goes into KOS and changes that value in opts.h and then get a runtime error of running out of FD entries that would be due to them modifying that value to number that is not the default. It should probably be FD_SETSIZE < 1024 OR just remove the if statement altogether.

If we're going to check if we've run out of descriptors and inform the user, what difference does it make why they've run out of descriptors? It doesn't make sense to then hide that error if `FD_SETSIZE` is 1024 but show it if it's 1023 or 1025, so I removed it.

andressbarajas:
Your right. Currently having this issue in Doom64 where we are running out of fd entries.  Always printing out the error message is the best course of action.

gyrovorbis (approved):
Everything looks good based on our previous discussions.

PR #662, opened 2024-07-05
Support chained LBA28 DMA transfers in IDE and DMA transfers in FAT

pcercuei:
Add support for transferring more than 256 sectors worth of data using DMA on devices that don't support LBA48.

This works by cutting the DMA transfer into chunks of 256 sectors, that are transferred automatically: every time a 256-sectors transfer is done, the DMA complete IRQ will start the next one.

PR #722, opened 2024-08-21
Revert "Support chained LBA28 DMA transfers in IDE and DMA transfers in FAT"

QuzarDC:
Reverts KallistiOS/KallistiOS#662

Breaks the much more common LBA48 and possibly general paths. Failures reported both in dreamboot and dreamshell related to this. Needs to be well tested before being merged again. Simply running with it integrated and not having failures in daily usage is not sufficient.


darcagn (approved):

QuzarDC:
Not needed as a fix for the issue was provided in #723

PR #724, opened 2024-08-22
Fix KOS warning when compiling SDL2 using m4-single

andressbarajas:
While compiling SDL2 library I came across this warning when using m4-single instead of m4-single-only:

/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/arch/args.h:17:5: warning:"__SH4_SINGLE_ONLY__" is not defined, evaluates to 0 [-Wundef]
   17 | #if __SH4_SINGLE_ONLY__
      |     ^~~~~~~~~~~~~~~~~~~
    
This PR should fix the warning.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/args.h
+++ b/kernel/arch/dreamcast/include/arch/args.h
@@ -14,7 +14,7 @@
     \author Paul Cercueil
 */
 
-#if defined(__SH4_SINGLE_ONLY__) && __SH4_SINGLE_ONLY__
+#ifdef __SH4_SINGLE_ONLY__
 #define KOS_SH4_SINGLE_ONLY 1
 #else
 #define KOS_SH4_SINGLE_ONLY 0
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/include/arch/args.h
```diff
@@ -14,7 +14,7 @@
     \author Paul Cercueil
 */
 
-#if __SH4_SINGLE_ONLY__
+#if defined(__SH4_SINGLE_ONLY__) && __SH4_SINGLE_ONLY__
```
darcagn, line 17: Is there any reason we have to do both of these rather than a typical `#ifdef`?

```suggestion
#ifdef __SH4_SINGLE_ONLY__
```

andressbarajas: None

darcagn (approved):

PR #727, opened 2024-08-24
v2.1.0 Relnotes Update

QuzarDC:
My notion was that since v2.1.0 has gone so long and had so many changes, it was worthwhile to provide a summary of new functionality added since v2.0.0 with the same basic framework as the main readme's functionality overview. I tried to be thorough in going through old commits but I'm sure I've missed some things.

It would likely make sense to add sections for additional language support, toolchain updates, new examples, and new utils but I wasn't entirely sure about structuring those.

I welcome suggestions for rounding this out, and the only thing I want to avoid is it going so long that it would need its own summary.

### DIFF
--- a/doc/RELNOTES.md
+++ b/doc/RELNOTES.md
@@ -1,8 +1,55 @@
 KallistiOS ##version##
 Copyright (C) 2002, 2003 Megan Potter
 Copyright (C) 2012-2019 Lawrence Sebald
+Copyright (C) 2024 Donald Haase
 
 RELEASE NOTES for 2.1.0
+-----------------------
+
+# What's New in Version 2.1.0
+
+KOS v2.1.0 has been a long time in the making. As such, it seemed prudent to 
+provide an overview of the new functionality since v2.0.0 in 2013. We intend 
+to have more frequent versioned releases moving forward, so this kind of 
+information should be easily seen in the changelog.
+
+# Core Functionality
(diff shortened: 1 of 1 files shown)


andressbarajas:
There was romdisk changes where you don't need to specify the extern romdisk in the source and only in your makefile:

KOS_ROMDISK_DIR‎ = romdisk

QuzarDC:
> There was romdisk changes where you don't need to specify the extern romdisk in the source and only in your makefile:
> 
> KOS_ROMDISK_DIR‎ = romdisk

Sorry for not having responded sooner. That's a kind of thing I had in mind to do as a 'what has changed?' for API breaks and process changes, but trying to gather all of them proved to be a daunting task, so I limited this to just entirely new functionality. If anyone would want to take up digging out more of these (like the PVR init change, SQs being internally managed and used in various places, whole new steps for the toolchain building, the DR changes, etc) it'd be welcome as a companion to this.

Moving forward it's certainly something that we should track on PRs to make it easier to gather up for releases (new label maybe).

QuzarDC:
I'd left this in draft since I knew it'd conflict with darc's documentation updates #747 . Once those are in then this one can be adjusted to the new baseline and move forward.

QuzarDC:
Just going to fold this into #747 as there's no need to have independent updates both offered by me to the same file.

PR #730, opened 2024-08-27
Fix incorrect file handle checks

andressbarajas:
Fix incorrect file handle checks. This fixes the runtime freeze that is documented here (# 3): https://github.com/KallistiOS/KallistiOS/issues/654


darcagn (approved):

darcagn:
Looks like in libdream this was `iso_open()` which would return `0` on error, but in KOS this was switched to `fs_open()` which returns `-1` on error.

PR #734, opened 2024-09-04
Refactor sysinfo_icon/id

andressbarajas:
Made syscall_sysinfo_init() public and call it in arch_auto_init() to prepare for future calls to syscalls for info_icon and info_id. This should be done to reduce redundant calls to syscall_sysinfo_init() which only needs to be called once.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/syscalls.h
+++ b/kernel/arch/dreamcast/include/dc/syscalls.h
@@ -38,6 +38,14 @@ __BEGIN_DECLS
 #include <stdint.h>
 #include <sys/types.h>
 
+/** \brief   Inits data needed by sysinfo id/icon
+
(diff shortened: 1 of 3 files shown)


DC-SWAT (approved):

kernel/arch/dreamcast/include/dc/syscalls.h
```diff
@@ -38,6 +38,14 @@ __BEGIN_DECLS
 #include <stdint.h>
 #include <sys/types.h>
 
+/** \brief   Inits data needed by sysinfo id/icon
+
```
gyrovorbis, line 42:
```suggestion
    \note This is called automatically by KOS during initialization.
```

gyrovorbis (approved):
This is great to have. Thanks, guys!

PR #740, opened 2024-09-06
PVR Hybrid Submission Model

gyrovorbis:
Very first attempt at an initial implementation of a "hybrid" vertex submission strategy using the existing PVR APIs.

To use this method:
1) Enable DMA mode when initializing the PVR
2) Set the in-RAM vertex buffers for each list type to be deferred with
   the DMA (pvr_set_vertbuf()).
3) For the list type to be immediately submitted:
    - pvr_list_begin(LIST_TYPE);
    - pvr_dr_begin(dr_state);
    - /* COMMIT POLYGONS */
    - pvr_list_end(); //!!!! NOTICE YOU DO NOT CALL PVR_DR_END()!!!

These changes have been tested and work 100% with: 1) Regular pvr_prim() (not broken)
2) PVR DMA (not broken)
3) PVR DR (not broken)
4) Hybrid
    a. OP list submitted via PVR DR
    b. TR list submitted via PVR DMA

THERE IS CURRENTLY A KNOWN ISSUE WITH:
1) Hybrid
    a. OP list submitted via PVR DMA
    b. TR list submitted via PVR DR
*** Currently causes artifacts if a piece of geometry isn't getting submitted for each list type. ***

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_scene.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_scene.c
@@ -175,7 +175,13 @@ int pvr_list_finish(void) {
 
 #endif  /* !NDEBUG */
 
-    if(!pvr_state.dma_mode) {
+    /* Check for immediate submission:
+       A. If we are not in DMA mode, we must be submitting polygons
+          immediately.
+       B. If we are in DMA mode, yet we've used the PVR DR API with
+          the current list type, assume we're doing hybrid drawing and
+          are directly submitting this list type. */
+    if(!pvr_state.dma_mode || pvr_state.dr_used) {
         /* Release Store Queues if they are used */
         if(pvr_state.dr_used) {
             pvr_dr_finish();
@@ -271,8 +277,12 @@ int pvr_scene_finish(void) {
         b = pvr_state.dma_buffers + pvr_state.ram_target;
 
         for(i = 0; i < PVR_OPB_COUNT; i++) {
-            // Not enabled -> skip
-            if(!(pvr_state.lists_enabled & (1 << i)))
+            /* Check whether the current list type should be skipped:
+               A. We never enabled the list globally with pvr_init().
+               B. We never associated an in-RAM DMA vertex buffer with
+                  the given list type, because we're using hybrid
+                  rendering and submitted that list type directly. */
+            if(!(pvr_state.lists_enabled & (1 << i)) || !b->base[i])
                 continue;
 
             // Make sure there's at least one primitive in each.
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
@@ -40,6 +40,16 @@ static void dma_next_list(void *data) {
             // Get the buffers for this frame.
             b = pvr_state.dma_buffers + (pvr_state.ram_target ^ 1);
 
+            /* If we are in PVR DMA mode, yet we haven't associated a
+               RAM-residing vertex buffer with the current list
+               (because we submitted it directly, for example),
+               mark it as complete, so we skip trying to DMA it. */
+            if(!b->base[i]) {
+                pvr_state.lists_dmaed       |= 1 << i;
+                pvr_state.lists_transferred |= 1 << i;
+                continue;
+            }
+
             // Flush the last 32 bytes out of dcache, just in case.
             // dcache_flush_range((ptr_t)(b->base[i] + b->ptr[i] - 32), 32);
             dcache_flush_range((ptr_t)(b->base[i]), b->ptr[i] + 32);


kernel/arch/dreamcast/hardware/pvr/pvr_scene.c
```diff
@@ -190,6 +202,8 @@ int pvr_list_finish(void) {
         /* In case we haven't sent anything in this list, send a dummy */
         pvr_blank_polyhdr(pvr_state.list_reg_open);
 
+        sq_unlock();
```
DC-SWAT, line 205: Conflicts with `pvr_dr_finish` above.

gyrovorbis:
NOTE: This PR also fixes a DRASTIC performance regression with `pvr_prim()` which got introduced when we made `sq_cpy()` need to lock and unlock a mutex to claim ownership over the SQs...

The fix not only properly claims the SQs ,but also offers a 20% performance improvement over the original, non-regressed performance!

DC-SWAT (approved):

QuzarDC:
> NOTE: This PR also fixes a DRASTIC performance regression with `pvr_prim()` which got introduced when we made `sq_cpy()` need to lock and unlock a mutex to claim ownership over the SQs...
> 
> The fix not only properly claims the SQs ,but also offers a 20% performance improvement over the original, non-regressed performance!

How exactly does this offer a 20% performance improvement? Shouldn't there only be a performance hit for the mutex lock/unlock if you end up competing with something else that needs to use the SQs, which would now just be the one being blocked instead? The lock/unlock itself should be trivial unless it has to wait in line during any of the items.

gyrovorbis:
> > NOTE: This PR also fixes a DRASTIC performance regression with `pvr_prim()` which got introduced when we made `sq_cpy()` need to lock and unlock a mutex to claim ownership over the SQs...
> > The fix not only properly claims the SQs ,but also offers a 20% performance improvement over the original, non-regressed performance!
> 
> How exactly does this offer a 20% performance improvement? Shouldn't there only be a performance hit for the mutex lock/unlock if you end up competing with something else that needs to use the SQs, which would now just be the one being blocked instead? The lock/unlock itself should be trivial unless it has to wait in line during any of the items.

Nah, firstly, this is in such a critical path--the most critical path of any DC game--that this `pvr_prim()` routine is literally called tens to hundreds of thousands of times per frame... even a `NOP` here would make a difference, so actually having to do non-inlined function calls with branching to check whether you own a mutex is absolutely MASSIVE compared to anything you should be doing in your per-vertex T&L loop. It's analogous to adding the work for dynamic lighting calculations or something per-vertex.

The 20% performance improvement over the original is due to now calling `sq_fast_cpy()` directly rather than having it get called indirectly via `sq_cpy()` which also locks/unlocks mutexes and has a branch before eventually trickling into `sq_fast_cpy()`.

andressbarajas:
I think Darc originally discovered this regression when testing the GCC versions against each other.  This was confirmed to be the fix.

andressbarajas (approved):

QuzarDC:
> > > NOTE: This PR also fixes a DRASTIC performance regression with `pvr_prim()` which got introduced when we made `sq_cpy()` need to lock and unlock a mutex to claim ownership over the SQs...
> > > The fix not only properly claims the SQs ,but also offers a 20% performance improvement over the original, non-regressed performance!
> > 
> > 
> > How exactly does this offer a 20% performance improvement? Shouldn't there only be a performance hit for the mutex lock/unlock if you end up competing with something else that needs to use the SQs, which would now just be the one being blocked instead? The lock/unlock itself should be trivial unless it has to wait in line during any of the items.
> 
> Nah, firstly, this is in such a critical path--the most critical path of any DC game--that this `pvr_prim()` routine is literally called tens to hundreds of thousands of times per frame... even a `NOP` here would make a difference, so actually having to do non-inlined function calls with branching to check whether you own a mutex is absolutely MASSIVE compared to anything you should be doing in your per-vertex T&L loop. It's analogous to adding the work for dynamic lighting calculations or something per-vertex.
> 
> The 20% performance improvement over the original is due to now calling `sq_fast_cpy()` directly rather than having it get called indirectly via `sq_cpy()` which also has a branch for checking alignment requirements before eventually trickling into `sq_fast_cpy()`.
> 
> The peformance boost is extremely easy to test yourself. Just go run ANY pvrmark example. Against `2.1.0` you should see an astronomical performance improvement. Against earlier KOS, before we introduced the SQ lock, which I'm now doing less frequently, back when performance was good for `pvr_prim()`, you should also still see approximately a 20% improvement in peak pps.

Gotcha! Thanks for the context. That makes perfect sense.

QuzarDC:
Should the end of list transfers interrupt handler act differently in this path for hybrid mode?
https://github.com/KallistiOS/KallistiOS/blob/93519b5546411b0d6a723f74884753056212ff97/kernel/arch/dreamcast/hardware/pvr/pvr_irq.c#L212-L225

I'm specifically thinking of here:
https://github.com/KallistiOS/KallistiOS/blob/93519b5546411b0d6a723f74884753056212ff97/kernel/arch/dreamcast/hardware/pvr/pvr_misc.c#L146-L166

Since that was written presuming all lists are being dma'd, not sure if there's impact on the way the registers are being set. Tried to dig into some documentation but didn't get far. If it's been well considered then that's fine.

Another bit to clarify, for your step 2 of how to do this, you mean calling `pvr_set_vertbuf(my_dr_list, null, 0)` right (based on the check for the base pointer being null)? If so, that seems like a good thing to call out in the doc for the function as now a specific param value has extra implications. I'm also not sure how that ends up squaring away with this:
https://github.com/KallistiOS/KallistiOS/blob/93519b5546411b0d6a723f74884753056212ff97/kernel/arch/dreamcast/include/dc/pvr.h#L1821-L1823

gyrovorbis:
> Should the end of list transfers interrupt handler act differently in this path for hybrid mode?
> 
> https://github.com/KallistiOS/KallistiOS/blob/93519b5546411b0d6a723f74884753056212ff97/kernel/arch/dreamcast/hardware/pvr/pvr_irq.c#L212-L225
> 
> I'm specifically thinking of here:
> 
> https://github.com/KallistiOS/KallistiOS/blob/93519b5546411b0d6a723f74884753056212ff97/kernel/arch/dreamcast/hardware/pvr/pvr_misc.c#L146-L166
> 
> Since that was written presuming all lists are being dma'd, not sure if there's impact on the way the registers are being set. Tried to dig into some documentation but didn't get far. If it's been well considered then that's fine.

I'm fairly certain that these registers should still be configured the same way here when beginning the next frame regardless of whether you're even using the DMA or SQs to fill the destination vertex buffers or TA OBPs.

> 
> Another bit to clarify, for your step 2 of how to do this, you mean calling `pvr_set_vertbuf(my_dr_list, null, 0)` right (based on the check for the base pointer being null)? If so, that seems like a good thing to call out in the doc for the function as now a specific param value has extra implications. I'm also not sure how that ends up squaring away with this:
> 
> https://github.com/KallistiOS/KallistiOS/blob/93519b5546411b0d6a723f74884753056212ff97/kernel/arch/dreamcast/include/dc/pvr.h#L1821-L1823
Whoops. So by default all of the vertex buffers are already NULL. You only have to set them explicitly to a non-NULL value for the list types you wish to use with the DMA... This is actually consistent with how it worked before, it was just required for you to specify a non-NULL vertex buffer for every list type.

The real change is that now you can optionally *not do that* for any list type, so that the default, NULL, vertex buffer is used (and the driver knows you are DR submitting that list via the SQs)

QuzarDC:
> > Should the end of list transfers interrupt handler act differently in this path for hybrid mode?
> > https://github.com/KallistiOS/KallistiOS/blob/93519b5546411b0d6a723f74884753056212ff97/kernel/arch/dreamcast/hardware/pvr/pvr_irq.c#L212-L225
> > 
> > I'm specifically thinking of here:
> > https://github.com/KallistiOS/KallistiOS/blob/93519b5546411b0d6a723f74884753056212ff97/kernel/arch/dreamcast/hardware/pvr/pvr_misc.c#L146-L166
> > 
> > Since that was written presuming all lists are being dma'd, not sure if there's impact on the way the registers are being set. Tried to dig into some documentation but didn't get far. If it's been well considered then that's fine.
> 
> I'm fairly certain that these registers should still be configured the same way here when beginning the next frame regardless of whether you're even using the DMA or SQs to fill the destination vertex buffers or TA OBPs.

But that's not what's happening in the current code or your update. When DMA mode isn't set `pvr_sync_reg_buffer` is only called on init. I guess they'd just get the same 'null' written to them on every frame. It could make sense to make them conditional, but that'd probably just be messier and not save anything.

Ok, will give it another lookover then get it in :)

QuzarDC (approved):

PR #741, opened 2024-09-06
aica: Fix the volume LUT

skmp:
According to my understanding for the aica hardware, the volumes use a "fractional shift"/logaritmic way where the shift amount is specified in a fixed point 4.4 format.

This has the advantage that all volumes can be added together in the chip before the final volume calculation is performed, and it is able to express better volumes in the low end, where they are more important, with equivelant precision of 16 bits.

We should really let the sh4 side set the volume value directly in the future, in addition to linear volume, for better control.

Binary driver compiled by @andressbarajas as the arm toolchain won't build on my machine.


andressbarajas:
I tested these changes and they work.  Im able to hear audio with volume set to ~17 which I wasnt able to before.  Attaching
[stream.drv.prebuilt.zip](https://github.com/user-attachments/files/16910904/stream.drv.prebuilt.zip)
 the prebuilt file that should also be included in this PR and can be used by others to test this change if you dont have the ARM compiler built

kernel/arch/dreamcast/sound/arm/aica.c
```diff
@@ -30,31 +31,36 @@ void aica_init(void) {
...
+    13, 13, 12, 12, 12, 12, 12, 12, 11, 11, 11, 11, 11, 11, 11, 10,
+    10, 10, 10, 10, 10, 10, 9, 9, 9, 9, 9, 9, 9, 8, 8, 8,
+    8, 8, 8, 8, 8, 7, 7, 7, 7, 7, 7, 7, 7, 6, 6, 6,
+    6, 6, 6, 6, 6, 5, 5, 5, 5, 5, 5, 5, 5, 5, 4, 4,
+    4, 4, 4, 4, 4, 4, 4, 3, 3, 3, 3, 3, 3, 3, 3, 3,
+    2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 1, 1, 1, 1, 1, 1,
+    1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0
 };
 
 static inline int calc_aica_vol(int x) {
```
andressbarajas, line 62:
```suggestion
static inline uint8 calc_aica_vol(int x) {
```

andressbarajas, line 62:
```suggestion
static inline uint8 calc_aica_vol(uint8 x) {
```
@pcercuei Mentioned this change in discord.

kernel/arch/dreamcast/sound/arm/aica.c
```diff
@@ -30,31 +31,36 @@ void aica_init(void) {
...
+   the AICA chip
+
+    Calculated by
+        for (int i = 0; i < 256; i++)
+            if (i == 0)
+                logs[i] = 255;
+            else
+                logs[i] = 16.0 * log2(255.0 / i);
+   */
 static int logs[] = {
```
andressbarajas, line 43:
```suggestion
static uint8 logs[] = {
```

andressbarajas:
The above requested changes seem to work fine.  Here is the new prebuilt with them:
[stream.drv.prebuilt.zip](https://github.com/user-attachments/files/16913820/stream.drv.prebuilt.zip)

andressbarajas:
I played around a little bit more and updated the memcpy function in main.c to copy 4 bytes at a time instead of 1.  Seems to work fine.  Didn't have any alignment issues with my testing with the sfx and multi-stream examples.  
```cvoid * memcpy(void *dest, const void *src, size_t count) {
    uint32 *tmp = (uint32 *) dest;
    uint32 *s = (uint32 *) src;
    count = count / 4;

    while(count--)
        *tmp++ = *s++;

    return dest;
}```

gyrovorbis:
Jesus guys, fantastic work here!

skmp:
> The above requested changes seem to work fine. Here is the new prebuilt with them: [stream.drv.prebuilt.zip](https://github.com/user-attachments/files/16913820/stream.drv.prebuilt.zip)

Commited the changes and added this driver to the PR

skmp:
Updated desc, remove draft flag.

Kazade (approved):

PR #749, opened 2024-09-08
utils/pvrtex: fix and improve Makefile

pcercuei:
We must not add KOS specific include paths when compiling a *host* utility. This can cause the compilation to fail (as it did on my system).

A lot of the rules are unnecesary, the default rules can be used as long as the proper variable names are used.

The "all" rule is moved as the first rule, so that running "make" will be equivalent to running "make all".


gyrovorbis (approved):
Tested and didn't regress Ubuntu, everything is fine here.

For whatever reason internet connection to my Win11 VM is down, so I couldn't verify with DreamSDK, but... assuming it works, I'll approve. Just can't merge, cuz master is frozen. Might pull into `dev` so it can get some soaking and testing.

andressbarajas:
I had no issues compiling PVRTEX either before or after testing this PR, and I'm using a Mac.

PR #751, opened 2024-09-10
Modified PVR context compilation to be write-only.

gyrovorbis:
Modified the following PVR API routines to be WRITE-ONLY to their destination vertex header pointers:
    - pvr_poly_compile()
    - pvr_sprite_compile()
    - pvr_mod_compile()

This will allow us to compile headers "in-place" within the SQ-mapped pointers returned by pvr_dr_target(), for better gainz using the direct rendering API...

This is required since according to some docs, the SQs are write-only while others claim only 32-bit reads are valid... Either way, this was necessary to be safe.


pcercuei:
ACK. This fixes a bug I had in Bloom.

QuzarDC (approved):

PR #752, opened 2024-09-11
Fixed KOS's Makefile.rules: CFLAGS, CXXFLAGS, CPPFLAGS. 

gyrovorbis:
They were incorrect and were causing really annoying warnings on other projects, such as GTA3, where C++ was warning that "-std=gnu17" is only available in C++.

How does this happen? Because our Makefile.rules file was INCORRECTLY concatenating CFLAGS and CXXFLAGS and sending them to the C++ compiler.

This is incorrect. Those two are separate and should go to their respective compilers only; however, CPPFLAGS is supposed to be common, shared, preprocessor flags that go to both.

- Removed CFLAGS from C++ rules
- Added CPPFLAGS to C, C++, ObjC, and ObjC++ rules


Makefile.rules
```diff
@@ -11,36 +11,36 @@
...
-	kos-c++ $(CFLAGS) $(CXXFLAGS) -c $< -o $@
+	kos-c++ $(CXXFLAGS) $(CPPFLAGS) -c $< -o $@
 
 %.o: %.m
-	kos-cc $(CFLAGS) -c $< -o $@
+	kos-cc $(CFLAGS) $(CPPFLAGS) -c $< -o $@
 
 %.o: %.mm
-	kos-c++ $(CFLAGS) $(CXXFLAGS) -c $< -o $@
+	kos-c++ $(CFLAGS) $(CXXFLAGS) $(CPPFLAGS) -c $< -o $@
```
QuzarDC, line 26: Should this still have `$(CFLAGS)`. I'd think it would match the rest and that it was accidentally using CFLAGS rather than CPPFLAGS before, but am not sure if obj-c needs both for some reason.

gyrovorbis: ObjC is the one time apparently you do want CFLAGS to trickle over. It's because C++ has shitty C backwards compat, while the spec requires ObjC *strictly* comply with anything standard C can do, so any C flags or anything like that can also be consumed by ObjC.

QuzarDC (approved):

PR #754, opened 2024-09-12
Remove a lot of ASM from fmath_base.h

pcercuei:
Use the compiler's built-in functions to provide __fsin(), __fcos(), __ftan(), __fisin(), __ficos(), __fitan(), __fsincosr(), __fsincos(), __fsqrt() and __fsqrtr().

The same code (or better) will be generated as long as the -mfsrra and -mfsca flags are passed.

The exception is __fsincos() which generates slightly worse code because of a missed optimization in GCC.

That was verified in compiler explorer, but some bugs may have slipped through.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/fmath_base.h
+++ b/kernel/arch/dreamcast/include/dc/fmath_base.h
@@ -32,105 +32,24 @@ __BEGIN_DECLS
 #define F_PI 3.1415926f
 
 /** \cond */
-#define __fsin(x) \
-    ({ float __value, __arg = (x), __scale = 10430.37835f; \
-        __asm__("fmul   %2,%1\n\t" \
-                "ftrc   %1,fpul\n\t" \
-                "fsca   fpul,dr0\n\t" \
-                "fmov   fr0,%0" \
-                : "=f" (__value), "+&f" (__scale) \
-                : "f" (__arg) \
-                : "fpul", "fr0", "fr1"); \
-        __value; })
-
-#define __fcos(x) \
-    ({ float __value, __arg = (x), __scale = 10430.37835f; \
-        __asm__("fmul   %2,%1\n\t" \
-                "ftrc   %1,fpul\n\t" \
-                "fsca   fpul,dr0\n\t" \
-                "fmov   fr1,%0" \
-                : "=f" (__value), "+&f" (__scale) \
-                : "f" (__arg) \
-                : "fpul", "fr0", "fr1"); \
-        __value; })
-
-#define __ftan(x) \
-    ({ float __value, __arg = (x), __scale = 10430.37835f; \
-        __asm__("fmul   %2,%1\n\t" \
-                "ftrc   %1,fpul\n\t" \
-                "fsca   fpul,dr0\n\t" \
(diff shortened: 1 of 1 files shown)


gyrovorbis:
This was absolutely verified in compiler explorer... especially when being called from C++, these totally do produce better codegen. Ian Michael verified they work all the way back to 4.7 better than the inline ASM!

gyrovorbis:
1) I think this is enough of a change to warrant adding you to the author list on the API.

2) On around line `28` near the `\addtogroup`, can we add something like this notice, so people are aware of how to get best performance?
```
\warning
For best performance with this KOS fast-math intrinsics, the GCC compiler flags for enabling fast-math optimizations and the SH4-specific instructions should be enabled within your `environ.sh` file:
    export KOS_CFLAGS="${KOS_CFLAGS} -fbuiltin -ffast-math -ffp-contract=fast -mfsrra -mfsca"
```

gyrovorbis:
> @gyrovorbis shouldn't we better enable -mfsrra / -mfsca unconditionally?

Okay, then, it's decision time. Do we want to change the defaults to enable these fast-math optimizations, despite the fact they're not fully IEEE compliant? In the past, I think Crabby wasn't a fan of doing that for this reason. 

Also about to ask Oleg Endo if there's any compiler define that we can use to check if these things are enabled at compile-time and maybe not use the `__builtin_xx()` versions and instead us the inline ASM ones if not?

pcercuei:
Well, I'm OK either way.

But if fast math isn't enabled, then I guess it's expected that these macros don't use FSCA / FSRRA.

gyrovorbis:
@pcercuei Yo, Oleg is going to make us a little something: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117036.

Because right now, the compiler doing the right thing is going to be totally dependent upon compiler flags... Whether `__builtin_x()` is actually HW accelerated math instructions or not; wheareas Oleg's solution, with the `__builtin_sh_x()` versions will always do the right thing regardless of compiler flags...

Given that's coming should we just hold off on addressing this?  What do you think?

pcercuei:
> @pcercuei Yo, Oleg is going to make us a little something: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117036.
> 
> Because right now, the compiler doing the right thing is going to be totally dependent upon compiler flags... Whether `__builtin_x()` is actually HW accelerated math instructions or not; wheareas Oleg's solution, with the `__builtin_sh_x()` versions will always do the right thing regardless of compiler flags...

I still think we should enable `-ffast-math -mfsrra -mfsca` in the environ.sh by default.

Keep in mind that by default our doubles are floats. If that isn't "fast math" I don't know what is.
 
> Given that's coming should we just hold off on addressing this? What do you think?

Sure, let's hold off until GCC 25.0, when the "oldoldstable" toolchain will be GCC 15.0 :)

pcercuei:
I'll switch this PR back to draft, as there is a bug in GCC that causes the code to be worse than it should when those macros are used outside their functions wrappers.
See also https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116719

PR #756, opened 2024-09-13
Refactored coding style in markdown

andressbarajas:
Our coding style is not current and contains history that I feel detracts from its purpose.  I redid it in markdown since it looks like all the other docs are getting a markdown upgrade.


gyrovorbis (approved):
Everything looks good here, but leaving it open for at least one more maintainer to see, just cuz it's like a global style thing... lol.

PR #757, opened 2024-09-14
Prevent thd_shutdown() from destroying kernel thread

darcagn:
In certain circumstances (e.g. at shutdown time in Rust `std` applications using the experimental `libpthread` branch), `thd_shutdown()` will destroy the kernel thread early, resulting in an `arch: aborting the system` error. This prevents `thd_shutdown()` from destroying the kernel thread.

### DIFF
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -1131,7 +1131,8 @@ void thd_shutdown(void) {
 
     /* Kill remaining live threads */
     LIST_FOREACH_SAFE(cur, &thd_list, t_list, tmp) {
-        thd_destroy(cur);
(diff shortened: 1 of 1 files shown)


QuzarDC (approved):
I'd originally suggested this as a possible fix. It's still not entirely clear what rust is doing behind the scenes to trigger this, but this matches a number of issues I ran into with the threaded controller callbacks where some thread other than the kernel thread attempts to shut down the kernel thread.

Will have to eventually clear these patches out by restructuring the way we do our kernel thread. Perhaps by having the kernel thread with static stack only handle the most initial setup, then having the initialization of threading swap over to an allocated thread until threading is shut down.

PR #762, opened 2024-09-17
Sound improvements and fixes

DC-SWAT:
Ready for review.

- Added `snd_stream_init_ex()` with the ability to reduce consumption of memory for separation interleaved stereo buffer and channels usage.
- Added direct DMA streaming capability for G1Bus-based storage devices.
- Improved sound streams data transfer to the SPU.
- Reduced the number of data requests for sound streams.
- Improved ADPCM channel separation by 56%.
- Added `spu_memload_dma()` as an alternative for `spu_memload_sq()`.
- Added `snd_stream_pan()` to use 2 separate mono streams as one stereo or something.
- Added `snd_sfx_load_ex()` for optimized loading of RAW files with ability to direct transfer from G1Bus-based devices.
- Handling out of SPU memory in `snd_sfx_load()`.
- Added `snd_sfx_load_fd()` the same as  `snd_sfx_load_ex()` but for working with sets.
- Changed SPU DMA trigger to avoid extra stalls since SH4 can frequently accesses AICA registers.
- Reduced the impact and number of mutex triggers on streams. The mutex is there only to protect the DMA and the channel split buffer. I initially placed it poorly.
- Some minor fixes and optimizations.

Sound streams (not direct) tested in DreamShell applications and GTA3. Two streams simultaneously in both cases.
We can save up to 64KB memory for this game (currently 48 KB, because stereo radio are used).
`spu_memload_dma()` works buggy in GTA3 with SFX, I don't know why. Looks like G2 DMA issue, but I've tried use CH3 for this function and get the same behavior, so it is not a conflict of using one DMA channel. That's why I don't use it anywhere yet. Although for libwav from kos-ports it works fine if I use it for streams prefill.

### DIFF
--- a/kernel/arch/dreamcast/hardware/spu.c
+++ b/kernel/arch/dreamcast/hardware/spu.c
@@ -2,13 +2,15 @@
 
    spu.c
    Copyright (C) 2000, 2001 Megan Potter
-   Copyright (C) 2023 Ruslan Rostovtsev
+   Copyright (C) 2023, 2024 Ruslan Rostovtsev
  */
 
+#include <kos/thread.h>
 #include <dc/spu.h>
 #include <dc/g2bus.h>
 #include <dc/sq.h>
 #include <arch/timer.h>
+#include <errno.h>
 
 /*
 
@@ -62,7 +64,7 @@ void spu_memload(uintptr_t dst, void *src_void, size_t length) {
 
 void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     uint8_t *src = (uint8_t *)src_void;
-    int aligned_len;
+    size_t aligned_len;
 
     /* Round up to the nearest multiple of 4 */
     if(length & 3) {
@@ -84,6 +86,41 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     if(length > 0) {
         /* Make sure the destination is in a non-cached area */
         dst |= MEM_AREA_P2_BASE;
+        dst += aligned_len;
+        src += aligned_len;
+        g2_fifo_wait();
+        g2_write_block_32((uint32_t *)src, dst, length >> 2);
+    }
+}
+
+void spu_memload_dma(uintptr_t dst, void *src_void, size_t length) {
+
+    uint8_t *src = (uint8_t *)src_void;
+    size_t aligned_len;
+
+    /* Round up to the nearest multiple of 4 */
+    if(length & 3) {
+        length = (length + 4) & ~3;
+    }
+
+    /* Using DMA (or SQ's on fail) for all that is divisible by 32 */
+    aligned_len = length & ~31;
+    length &= 31;
+
+    do {
+        if (spu_dma_transfer(src_void, dst, aligned_len, 1, NULL, NULL) < 0) {
+            if (errno == EINPROGRESS) {
+                thd_pass();
+                continue;
+            }
+            spu_memload_sq(dst, src_void, aligned_len);
+        }
+        break;
+    } while (1);
+
+    if(length > 0) {
+        /* Make sure the destination is in a non-cached area */
+        dst |= (MEM_AREA_P2_BASE | SPU_RAM_BASE);
         dst += aligned_len;
         src += aligned_len;
         g2_fifo_wait();
--- a/kernel/arch/dreamcast/hardware/spu.c
+++ b/kernel/arch/dreamcast/hardware/spu.c
@@ -94,7 +94,6 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
 }
 
 void spu_memload_dma(uintptr_t dst, void *src_void, size_t length) {
-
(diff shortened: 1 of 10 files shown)


kernel/arch/dreamcast/sound/snd_stream.c
```diff
@@ -363,15 +369,21 @@ void snd_stream_prefill(snd_stream_hnd_t hnd) {
...
-    if(!sep_buffer[0]) {
-        sep_buffer[0] = memalign(32, SND_STREAM_BUFFER_MAX);
-        sep_buffer[1] = sep_buffer[0] + (SND_STREAM_BUFFER_MAX / 8);
+    return snd_stream_init_ex(2, SND_STREAM_BUFFER_MAX);
+}
+
+int snd_stream_init_ex(int channels, size_t buffer_size) {
+    max_channels = channels;
+
+    if (channels == 2) {
```
andressbarajas, line 378:
```suggestion
    if(channels == 2) {
```

kernel/arch/dreamcast/sound/snd_stream.c
```diff
@@ -451,8 +459,12 @@ void snd_stream_destroy(snd_stream_hnd_t hnd) {
         return;
     }
 
+    snd_stream_stop(hnd);
     snd_sfx_chn_free(streams[hnd].ch[0]);
-    snd_sfx_chn_free(streams[hnd].ch[1]);
+
+    if (max_channels == 2) {
```
andressbarajas, line 465:
```suggestion
    if(max_channels == 2) {
```

kernel/arch/dreamcast/sound/snd_stream.c
```diff
@@ -698,6 +708,9 @@ int snd_stream_poll(snd_stream_hnd_t hnd) {
         needed_bytes = (int)stream->buffer_size / stream->channels;
     }
 
+    if (!stream->initted || !stream->get_data) {
```
andressbarajas, line 711:
```suggestion
    if(!stream->initted || !stream->get_data) {
```

kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -91,6 +93,41 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     }
 }
 
+void spu_memload_dma(uintptr_t dst, void *src_void, size_t length) {
+
```
andressbarajas, line 97:
```suggestion
```

kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -91,6 +93,41 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
...
+    aligned_len = length & ~31;
+    length &= 31;
+
+    do {
+        if (spu_dma_transfer(src_void, dst, aligned_len, 1, NULL, NULL) < 0) {
+            if (errno == EINPROGRESS) {
+                thd_pass();
+                continue;
+            }
+            spu_memload_sq(dst, src_void, aligned_len);
```
QuzarDC, line 116: I'm not sure that this kind of fallback should be encoded here. It seems very much like something that would rather be up to the user to take the possible `EINPROGRESS` and decide how to proceed. Especially since the fallback SQ version could *also* be blocked by another SQ transfer already being in process (or cause an unexpected block to pvr DR for instance).

I'll think on this as well, but it may make sense to provide a new version of spu_memload that offers a `type` option (plain G2 write, SQ, or DMA), to have parameters for blocking/nonblocking, etc.

DC-SWAT: Yes, I agree, it doesn't look very good. But returning the EINPROGRESS error here kills the point, since these functions imply replacement with other similar ones, the result of their work is not disputed in the places of their use. What you are talking about here already exists in the form of `spu_dma_transfer()` which is used here. The main point of this function is to guarantee the result, like `spu_memload()` and `spu_memload_sq()`.
I just wanted to replace `spu_memload_sq()` in SFX and stream prefill with this new function. That is, on the contrary, I want to reduce the blocking of SQ's. They will work in extreme cases.

DC-SWAT: Ah sorry, I didn't quite understand you at first.
Yes, I can make a function with a choice of transfer type, but we can't delete the old ones, since backward compatibility is needed, so either add one with types, or one with dma. I chose one with dma, it's the last one anyway, there are simply no other options :)

DC-SWAT: Also I can just remove it, since it's not used yet anywhere.

QuzarDC: Sorry for the long delay, I had missed that there were updates here when checking back. I see now what you mean that this is meant to guarantee the result and sit above the rest with the existing transfer still in place that could be used in the way I describe.

QuzarDC (approved):

PR #763, opened 2024-09-18
isofs: Added cdrom streams and direct DMA support

DC-SWAT:
- Added cdrom DMA streams support. At the sector boundary with aligned buffer, DMA streaming will be started for the entire file and works until file end or cancelled (closed or another file used). Reading from an open stream can be done in chunks of up to a minimum of 32 bytes.
- Added IOCTL function to determine that file system can load directly to SPU and PVR RAM's thought the Root Bus.

Please look at the final code, because the commits here are just a mess now.
These changes almost all need to be done in one commit. It would be possible to separate only `ioctl` and `inline`, but there is no point in redoing the PR for this.


kernel/arch/dreamcast/fs/fs_iso9660.c
```diff
@@ -578,10 +583,13 @@ static struct {
     uint32      size;       /* Length of file in bytes */
     dirent_t    dirent;     /* A static dirent to pass back to clients */
     int     broken;     /* >0 if the CD has been swapped out since open */
+    size_t     stream_part;     /* Stream DMA part of 32 bytes */
+    uint8_t stream_data[32] __attribute__((aligned(32)));
```
gyrovorbis, line 587:
```suggestion
    alignas(32) uint8_t stream_data[32];
```
Since KOS got bumped up to `C17`, we now get alignment as part of the language...

Just be sure to include `<stdalign.h>`.

include/sys/ioctl.h
```diff
@@ -35,6 +36,10 @@ __BEGIN_DECLS
 #define TIOCGETA TCGETS
 #endif
 
+#ifndef IOCTL_FS_ROOTBUS_DMA_READY
```
gyrovorbis, line 39:
```suggestion
/** Brief Description of IOCTL and its argument type + value */
#ifndef IOCTL_FS_ROOTBUS_DMA_READY
```

QuzarDC:
@DC-SWAT As v2.2.0 is out, could you please rebase this? It seems it got conflicted due to other bugfixes unfortunately. Once it's ready I'll be sure to get it merged in promptly.

DC-SWAT:
> @DC-SWAT As v2.2.0 is out, could you please rebase this? It seems it got conflicted due to other bugfixes unfortunately. Once it's ready I'll be sure to get it merged in promptly.

Done.

QuzarDC (approved):
Making sure to note for #1044 the changes to `sys/ioctl.h` here as this may be the kind of thing that we should shift into a more arch independent factoring later, but it could be just as simply to just have a small collection of these as they wouldn't really interfere with one another.

Perhaps as simple as an `arch/ioctl.h` that gets included but that makes more sense only if we have some ioctl defines that would have to be different across different archs for some reason. This may be the case for a full implementation of #693, for example.

Additionally opened #1096 to address the caching issues that this raises. It makes perfect sense for this kind of streaming behavior to simply bypass the cache, and also merits looking at how the cache is being handled generally for iso.

PR #765, opened 2024-09-22
Scoped IRQ disable

pcercuei:
This PR adds `irq_disable_scoped()`, following the work that has been done with `mutex_lock_scoped()`.

This macro will disable interrupts, similarly to irq_disable(), with the difference that the interrupt state will automatically be restored once the execution exits the functional block in which the macro was called.

### DIFF
--- a/kernel/thread/mutex.c
+++ b/kernel/thread/mutex.c
@@ -52,31 +52,28 @@ int mutex_init(mutex_t *m, int mtype) {
 }
 
 int mutex_destroy(mutex_t *m) {
-    int rv = 0, old;
-
-    old = irq_disable();
+    irq_disable_scoped();
 
     if(m->type < MUTEX_TYPE_NORMAL || m->type > MUTEX_TYPE_RECURSIVE) {
         errno = EINVAL;
-        rv = -1;
+        return -1;
     }
-    else if(m->count) {
+
+    if(m->count) {
         /* Send an error if its busy */
         errno = EBUSY;
-        rv = -1;
+        return -1;
     }
-    else {
-        /* Set it to an invalid type of mutex */
-        m->type = -1;
-    }
+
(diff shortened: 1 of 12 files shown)


gyrovorbis (approved):
God *dayum*, I knew this type of C++ RAII-like automatically-releasing IRQ disabler was something we wanted, but I didn't really realize how much the thing would simplify and clean-up control flow... Approved! Very nice!

gyrovorbis:
Tested on GTA3 + C++20 concurrency + every basic/threading example we have in KOS. Feel comfortable merging...

kernel/thread/mutex.c
```diff
@@ -52,31 +52,28 @@ int mutex_init(mutex_t *m, int mtype) {
...
+
     /* If the mutex was created with the deprecated mutex_create(), free it. */
     if(m->dynamic) {
         free(m);
     }
 
-    irq_restore(old);
-    return rv;
+    return 0;
 }
```
QuzarDC, line 77: Just as a note, this has changed the functionality of the function. Previously a dynamic mutex would be freed even if it was still busy/invalid and destroy would return an error indicating that it should not have destroyed the mutex. That mode is deprecated so maybe it's not been tripped before, but totally a bug fix.

pcercuei: Ok, so you are saying that we can keep it like that?

QuzarDC: Yea, it matches the description and the intention of the function better. What was there before was simply a bug that would allow use-after-free. I just wanted to note (that's why I resolved it immediate) this because of @andressbarajas tracking a recently reported bug to this PR, and I didn't have a chance to review it initially.

I can't really even conceive of how someone would have user code where the fix could cause a problem.

PR #770, opened 2024-09-23
Update mat_transform() parameter name and description

voxel-public:
Correct name and description of mat_transform parameter from vecskip to stride, to ensure generated documentation is no longer misleading. 

Currently generated documentation reads 
`vecskip	Number of empty bytes between vectors.  `
which implies a value of 0 for tightly packed vectors, but this is incorrect and needs to be the stride size of `invecs`

### DIFF
--- a/kernel/arch/dreamcast/include/dc/matrix.h
+++ b/kernel/arch/dreamcast/include/dc/matrix.h
@@ -93,9 +93,9 @@ void mat_apply(matrix_t *src);
     \param  invecs          The list of input vectors.
     \param  outvecs         The list of output vectors.
     \param  veccnt          How many vectors are in the list.
-    \param  vecskip         Number of empty bytes between vectors.
+    \param  stride          Number of bytes between vectors.
 */
-void mat_transform(vector_t *invecs, vector_t *outvecs, int veccnt, int vecskip);
+void mat_transform(vector_t *invecs, vector_t *outvecs, int veccnt, int stride);
 
 /** \brief  Transform vectors by the internal matrix into the store queues.
 


kernel/arch/dreamcast/include/dc/matrix.h
```diff
@@ -93,9 +93,9 @@ void mat_apply(matrix_t *src);
     \param  invecs          The list of input vectors.
     \param  outvecs         The list of output vectors.
     \param  veccnt          How many vectors are in the list.
-    \param  vecskip         Number of empty bytes between vectors.
+    \param  stride          Number of bytes between vectors.
 */
-void mat_transform(vector_t *invecs, vector_t *outvecs, int veccnt, int vecskip);
+void mat_transform(vector_t *invecs, vector_t *outvecs, int veccnt, int stride);
```
andressbarajas, line 98:
```suggestion
void mat_transform(vector_t *invecs, vector_t *outvecs, uint32_t veccnt, uint32_t stride);
```
Since the veccnt and stride cant be negative I think it would be good to update the types as well.

gyrovorbis: Well, actually, might even be better to make `stride` a `size_t` here to denote something you'd typically use the `sizeof()` operator on, in byte units.

pcercuei:
No, I disagree with this change. A `stride` would be the size in bytes between two rows, *including* the vectors, and not the number of padding between two vectors. Renaming the parameter to "stide" would be worse than before, as it would be misleading.

voxel-public:
> No, I disagree with this change. A `stride` would be the size in bytes between two rows, _including_ the vectors, and not the number of padding between two vectors. 

We're in agreement - this is exactly how the parameter is used, *if* I'm understanding correctly. I find the sh4 assembly in matrix.s hard to understand so consulted with Falco and UnknownShadow200 in the Simulant discord, who agreed that the parameter is being used as stride ( https://discord.com/channels/488332893324836864/614912396460556298/1287006337850540086 ).

gyrovorbis:
@pcercuei Thoughts? I actually think you might be confused for the exact reason we want to change it. lol.

pcercuei:
Ok.

kernel/arch/dreamcast/include/dc/matrix.h
gyrovorbis: Shit, nvm. If we change these argument types, we probably have to adjust the 'semblies, which is probably outside the scope of this PR.

gyrovorbis (approved):
Thank you kindly for fixing that... disaster of a parameter. ;)

Issue #229, opened 2023-06-01
Maple devices continue last operation (beeping, rumbling, rendering, etc) after driver shuts down


gyrovorbis:
This is a very minor issue that I can get around to eventually, but it's going to be pretty annoying... We already have examples of beeping and rendering to the VMU screens that show off this issue...

If you begin playing a tone on the VMU or if you render some text or presumably if you cause the puru pack to rumble, then exit an example, it will continue playing indefinitely. That is until you power cycle the DC, load another example, or physically take the peripheral out and put it back in.

Basically before the maple peripheral drivers shut down, they need to be sending stop commands to the devices for these kinds of operations; however, if you look at the top level `maple_shutdown()` driver callback, notice it's actually shutting down the maple hardware before even calling shutdown on each peripheral, so this will either have to be swapped or addressed in another manner.

QuzarDC:
Switching the ordering of the peripheral shutdowns should be simple, and is certainly more correct. A big additional problem is that none of the cleanup actions that would need to be taken could happen within an interrupt (sending commands to the vmu or rumble). Since the most common way to exit is arch_exit in the cont_buttons callback, this could be fixed with #209 or would have to have the basic behaviors of the drivers otherwise changed (come up with an int-safe send method just to send the stop).

It could also be that there's some missing low-level signal that should be sent to have all devices cease operation and sending individualized ones isn't the 'correct' way to do it.

QuzarDC:
To not create a lot of needless traffic I was pondering how to only send if a beep/rumble were still in progress on shutdown. It doesn't seem like there's any way to read the current status on those, but hit on the fact that there's a reset command intended to reset the devices back to default which is what we'd want to do on shutdown anyways. (and I just realized I had the hint of being the right thing over a year ago in my last comment here)

Opened #772 to add a facility for sending that reset command to a device, and set it to happen at shutdown. Only problem is that it doesn't seem to have an impact on vmu_beep.

PR #772, opened 2024-09-24
Ensure maple devices are reset back to standard on shutdown

QuzarDC:
In looking for the best way to address #229 , I hit upon the fact that there's a maple device reset command. My understanding is that this should work to stop any ongoing operations and to reset any settings back to default.

It seems to work right with `rumble` and `vmu_lcd` just fine, but `vmu_beep` just keeps beeping. Most likely I'm doing something wrong in enumerating and sending the command as it might need to be one for the LCD and one for the Timer (where the beeping lives).

In the end the cleanups should likely be readded to the examples with notes that it's the polite thing to do. Additionally, it may make sense to have the reset called on startup to clear anything from previous software.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
@@ -14,6 +14,7 @@
 #include <dc/vblank.h>
 #include <kos/thread.h>
 #include <kos/init.h>
+#include <kos/genwait.h>
 
 #include <dc/maple/controller.h>
 #include <dc/maple/keyboard.h>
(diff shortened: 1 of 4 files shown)


kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
```diff
@@ -34,6 +35,55 @@
...
+    /* Reset the frame */
+    maple_frame_init(&dev->frame);
+    dev->frame.cmd = MAPLE_COMMAND_RESET;
+    dev->frame.dst_port = dev->port;
+    dev->frame.dst_unit = dev->unit;
+    dev->frame.length = 0;
+    dev->frame.callback = maple_dev_reset_cb;
+    maple_queue_frame(&dev->frame);
+
+    /* Wait for the device to accept it */
```
QuzarDC, line 65: Just as a note, tried to remove this, as we don't care what the response was after all there's no failure state other than not returning and there's nothing we'd really be able to do about it.

I know I'd discussed before the idea of  removing this or making it optional in 'fire-and-forget' commands (draw, beep, rumble) as it seems mostly useful for something where you need confirmation like the vmu reading and such. The problem is that the wait doesn't *just* ensure we get a return but that the command gets sent at all.

`maple_queue_frame`, as the name implies, just queues up the command to be sent. Normally there's little concern that it'll be sent in a timely fashion, but in this case we're doing it directly before shutting down all the maple hardware so in my testing it usually just failed to send the resets.

The right response though may be to have a test in `maple_hw_shutdown` ensure that the queue gets cleared before shutting down.

QuzarDC:
After further experimentation, my best guess is that it simply doesn't work with the vmu speaker, but that shouldn't hold back the usefulness of this for rumble and lcd.

kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
```diff
@@ -35,6 +36,56 @@
...
+    genwait_wake_all(frame);
+}
+
+static void maple_dev_reset(maple_device_t *dev) {
+
+    assert(dev != NULL);
+
+    /* Lock the frame */
+    while(maple_frame_lock(&dev->frame) < 0)
+        thd_pass();
```
pcercuei, line 55: The behaviour of `maple_frame_lock()` changed since this PR was first opened, the old behaviour is now provided by `maple_frame_trylock()`. You can just call `maple_frame_lock()` here without the while loop.

PR #775, opened 2024-09-25
Filesystem Stat fixes

andressbarajas:
1. Fixed fs_vmu stat() for root directories. 
2. Added stat() to fs_romdisk. 
3. Code style changes for fs_normalize_path and realpath.
4. Fixed browser example to use fs_normalize_path() instead of realpath() because of how realpath works. Internally realpath checks if the file you pass it exists.  It uses lstat() to do that.  If a file system(like romdisk) doesnt have a stat() defined then newlib's lstat() tries a different "hackish" https://github.com/KallistiOS/KallistiOS/blob/f842d4b3c8a391f2793c20e58f8f031f0c316043/kernel/libc/newlib/newlib_stat.c#L28 way of providing info that involves opening and closing files which can mess with realpath result.

### DIFF
--- a/kernel/fs/fs_pty.c
+++ b/kernel/fs/fs_pty.c
@@ -758,7 +758,7 @@ static int pty_stat(vfs_handler_t *vfs, const char *path, struct stat *st,
     memset(st, 0, sizeof(struct stat));
     st->st_dev = (dev_t)('p' | ('t' << 8) | ('y' << 16));
     st->st_mode = S_IFCHR | S_IRUSR | S_IWUSR | S_IRGRP | S_IWGRP | S_IROTH | S_IWOTH;
-    st->st_nlink = 1
+    st->st_nlink = 1;
     st->st_size = ph->cnt;
     st->st_blksize = PTY_BUFFER_SIZE;
 
--- a/kernel/fs/fs_pty.c
+++ b/kernel/fs/fs_pty.c
@@ -706,6 +706,65 @@ static int pty_ioctl(void *h, int cmd, va_list ap) {
     }
 }
 
+static int pty_stat(vfs_handler_t *vfs, const char *path, struct stat *st,
+                    int flag) {
+    ptyhalf_t *ph;
+    int id;
+    int master;
+    size_t len = strlen(path);
+
+    (void)vfs;
+    (void)flag;
+
+    /* Root directory '/pty' */
+    if(len == 0 || (len == 1 && *path == '/')) {
+        memset(st, 0, sizeof(struct stat));
+        st->st_dev = (dev_t)('p' | ('t' << 8) | ('y' << 16));
+        st->st_mode = S_IFDIR | S_IRWXU | S_IRWXG | S_IRWXO;
+        st->st_size = -1;
+        st->st_nlink = 2;
+
+        return 0;
+    }
+
+    /* Handle paths that start directly with '/maXX' or '/slXX' */
+    if(sscanf(path, "/%2c%02x", (char[3]){}, &id) != 2) {
+        errno = ENOENT;
+        return -1;
+    }
+
+    /* Check if it's a master (ma) or slave (sl) */
+    master = (path[0] == 'm' && path[1] == 'a');
+
+    /* Find the corresponding PTY half */
+    mutex_lock(&list_mutex);
+    LIST_FOREACH(ph, &ptys, list) {
+        if(ph->id == id) break;
+    }
+    mutex_unlock(&list_mutex);
+
+    /* If PTY is not found, return error */
+    if(!ph) {
+        errno = ENOENT;
+        return -1;
+    }
+
+    /* If the file requested doesn't match the master/slave role, switch */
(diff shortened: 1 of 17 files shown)


DC-SWAT:
Add `fs_normalize_path` to exports plz.

andressbarajas:
Here is a little program I used to test: 
```c
#include <stdio.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <unistd.h>
#include <dirent.h>
#include <string.h>

void check_lstat(const char *path) {
    struct stat fileStat;

    if (lstat(path, &fileStat) < 0) {
        perror(path);
        return;
    }

    printf("Information for %s:\n", path);
    printf("File Mode: %o\n", fileStat.st_mode);
    printf("File Size: %ld bytes\n", fileStat.st_size);
    printf("Number of Links: %ld\n", fileStat.st_nlink);
    printf("File inode: %ld\n", fileStat.st_ino);
    printf("File Type: ");
    if (S_ISREG(fileStat.st_mode))
        printf("Regular File\n");
    else if (S_ISDIR(fileStat.st_mode))
        printf("Directory\n");
    else if (S_ISLNK(fileStat.st_mode))
        printf("Symbolic Link\n");
    else
        printf("Other\n");

    printf("\n");
}

void traverse_directory(const char *dirpath) {
    DIR *dir;
    struct dirent *entry;
    char fullPath[512];  // Buffer to hold the full path of the file/directory

    // Open the directory
    dir = opendir(dirpath);
    if (dir == NULL) {
        perror(dirpath);
        return;
    }

    // Loop through directory entries
    while ((entry = readdir(dir)) != NULL) {
        // Skip "." and ".." entries
        if (strcmp(entry->d_name, ".") == 0 || strcmp(entry->d_name, "..") == 0)
            continue;

        // Skip "/pc" directory
        if (strcmp(entry->d_name, "pc") == 0 && strcmp(dirpath, "/") == 0)
            continue;

        // Construct the full path to the entry
        // Avoid adding an extra '/' if dirpath already ends with '/'
        if (dirpath[strlen(dirpath) - 1] == '/')
            snprintf(fullPath, sizeof(fullPath), "%s%s", dirpath, entry->d_name);
        else
            snprintf(fullPath, sizeof(fullPath), "%s/%s", dirpath, entry->d_name);

        // Call lstat on the full path
        check_lstat(fullPath);

        // If it's a directory, recursively traverse it
        if (entry->d_type == DT_DIR) {
            traverse_directory(fullPath);
        }
    }

    closedir(dir);
}

int main() {
    // Start traversing from the root directory
    traverse_directory("/");
    return 0;
}
```

andressbarajas:
One thing I wasnt too sure about the size that should be returned for Directories.  I found that we use -1 for size for directories when it comes to fds so I did the same thing with the stat size value.

DC-SWAT (approved):

QuzarDC:
> One thing I wasnt too sure about the size that should be returned for Directories. I found that we use -1 for size for directories when it comes to fds so I did the same thing with the stat size value.

That works fine. By the standard it's undefined for directories. Lots of places seem to use that same -1 convention (or 0) but generally the thing is that you should be doing the `S_ISDIR` test before trying to pull the size and adjust yourself accordingly.

DC-SWAT:
> > One thing I wasnt too sure about the size that should be returned for Directories. I found that we use -1 for size for directories when it comes to fds so I did the same thing with the stat size value.
> 
> That works fine. By the standard it's undefined for directories. Lots of places seem to use that same -1 convention (or 0) but generally the thing is that you should be doing the `S_ISDIR` test before trying to pull the size and adjust yourself accordingly.

-1 is a bit more logical because the file can be of zero size.
But in general, of course, need to check by attributes first.

QuzarDC:
What were the changes in the latest force-push? I was in the middle of reviewing it and github isn't able to tell me which files changed and didn't with the force-push.

andressbarajas:
I rebased with master

kernel/fs/fs_pty.c
```diff
@@ -706,6 +706,65 @@ static int pty_ioctl(void *h, int cmd, va_list ap) {
...
+    }
+
+    /* If the file requested doesn't match the master/slave role, switch */
+    if(master != ph->master) {
+        ph = ph->other;
+    }
+
+    /* Fill in the stat structure */
+    memset(st, 0, sizeof(struct stat));
+    st->st_dev = (dev_t)('p' | ('t' << 8) | ('y' << 16));  // Unique device ID for PTYs
```
andressbarajas, line 759: @QuzarDC Let me know when youre done reviewing. I want to remove these comments

QuzarDC (commented):
Here, just adding in the ones I got to so you can continue the other cleanups you wanted to do.

kernel/fs/fs_pty.c
QuzarDC: Submitted what I had so far.

andressbarajas:
@QuzarDC Done

QuzarDC (approved):

PR #736, opened 2024-09-05
DMA and PIO stream transfer and IRQ support for GD-drive.

DC-SWAT:
Ready for review.

- Complete support for PIO and DMA stream transfers. This allows pre-reading data into the GD-drive's internal cache before next request, dividing the sector into parts when transferring to memory (32-byte in DMA, 2-byte in PIO) and abort last sector transfer on aligned offset. This is the best way to read data from an optical drive. From any G1 drive.
- Ability to use DMA and vblank IRQs to fully free up CPU. Added new modes `CDROM_READ_DMA_IRQ` and `CDROM_READ_PIO_IRQ` that can be used for streams and regular sector readings.
- Ability to transfer data directly to the SPU RAM and PVR RAM.
- Fixed and improved CPU cache management for devices on G1 bus.
- Updated exports.

Tested on GTA3.

PR #778, opened 2024-09-28
CD Syscall Types Update

QuzarDC:
This is in the works to get in the typing and cleanup suggestions from #619 and #736 with a different scope.

The key purpose is to add strongly defined types for the various returns and parameters of the functions for interacting with the GDROM drive, replacing the many sets of defines, generic types, and untyped structs.

Changes:
- [x] Redefine each group of CD defines to an enum, per #619. (The only one left out is the unused group of ATA Status defines @DC-SWAT perhaps you have more info on them since you added them? Are these returned by one of the syscalls? Maybe they're the 4th param of Check Command?)
- [x] Move syscall related define groups to syscalls.h
- [x] Rename some of the define group values to better match our general naming scheme (`CD_ERR_OK` rather than just `ERR_OK`).
- [ ] Cleanups to structure (some due to the typing, some due to the scoped mutex or syscalls updates)
- [ ] Create types for used param structs.
- [ ] Clean up the documentation, ensuring there are good cross-referencces between the data types and functions that they get used by.

Possible further work:
- [ ] Add abort command (per #736) with some refactoring of the command issuance.
- [ ] Separate out gd syscall things into a separate header from the others.
- [ ] Fully type each of the syscall's param structs.

Still need a bit more work for what I was sure to add, but wanted to make sure to make it public now to get feedback/help with the documentation, naming, and just general review for mistakes.

### DIFF
--- a/kernel/arch/dreamcast/hardware/syscalls.c
+++ b/kernel/arch/dreamcast/hardware/syscalls.c
@@ -165,9 +165,9 @@ void syscall_gdrom_reset(void) {
         PARAM_NA, PARAM_NA, SUPER_FUNC_GDROM);
 }
 
-int syscall_gdrom_check_drive(uint32_t status[2]) {
+int syscall_gdrom_check_drive(cd_check_drive_params_t params) {
     MAKE_SYSCALL_INT(VEC_MISC_GDROM, FUNC_GDROM_DRIVE_STATUS, 
-        status, PARAM_NA, SUPER_FUNC_GDROM);
+        params, PARAM_NA, SUPER_FUNC_GDROM);
 }
 
 gdc_cmd_id_t syscall_gdrom_send_command(cd_cmd_code_t cmd, void *params) {
--- a/kernel/arch/dreamcast/hardware/syscalls.c
+++ b/kernel/arch/dreamcast/hardware/syscalls.c
@@ -170,7 +170,7 @@ int syscall_gdrom_check_drive(uint32_t status[2]) {
         status, PARAM_NA, SUPER_FUNC_GDROM);
 }
 
-uint32_t syscall_gdrom_send_command(uint32_t cmd, void *params) {
+gdc_cmd_id_t syscall_gdrom_send_command(cd_cmd_code_t cmd, void *params) {
     uint32_t request_id = 0;
 
     MAKE_SYSCALL_SET(VEC_MISC_GDROM, FUNC_GDROM_SEND_COMMAND, 
@@ -179,7 +179,7 @@ uint32_t syscall_gdrom_send_command(uint32_t cmd, void *params) {
     return request_id;
 }
 
-int syscall_gdrom_check_command(uint32_t id, int32_t status[4]) {
+cmd_cmd_chk_t syscall_gdrom_check_command(gdc_cmd_id_t id, int32_t status[4]) {
     MAKE_SYSCALL_INT(VEC_MISC_GDROM, FUNC_GDROM_CHECK_COMMAND, 
         id, status, SUPER_FUNC_GDROM);
 }
@@ -189,7 +189,7 @@ void syscall_gdrom_exec_server(void) {
         PARAM_NA, PARAM_NA, SUPER_FUNC_GDROM);
 }
 
-int syscall_gdrom_abort_command(uint32_t id) {
+int syscall_gdrom_abort_command(gdc_cmd_id_t id) {
     MAKE_SYSCALL_INT(VEC_MISC_GDROM, FUNC_GDROM_ABORT_COMMAND, 
         id, PARAM_NA, SUPER_FUNC_GDROM);
 }
@@ -204,12 +204,12 @@ void syscall_gdrom_dma_callback(uintptr_t callback, void *param) {
         callback, param, SUPER_FUNC_GDROM);
 }
 
-int syscall_gdrom_dma_transfer(uint32_t id, const int32_t params[2]) {
+int syscall_gdrom_dma_transfer(gdc_cmd_id_t id, const int32_t params[2]) {
     MAKE_SYSCALL_INT(VEC_MISC_GDROM, FUNC_GDROM_DMA_TRANSFER, 
         id, params, SUPER_FUNC_GDROM);
 }
 
-int syscall_gdrom_dma_check(uint32_t id, size_t *size) {
+int syscall_gdrom_dma_check(gdc_cmd_id_t id, size_t *size) {
     MAKE_SYSCALL_INT(VEC_MISC_GDROM, FUNC_GDROM_DMA_CHECK, 
         id, size, SUPER_FUNC_GDROM);
 }
@@ -219,12 +219,12 @@ void syscall_gdrom_pio_callback(uintptr_t callback, void *param) {
         callback, param, SUPER_FUNC_GDROM);
 }
 
-int syscall_gdrom_pio_transfer(uint32_t id, const int32_t params[2]) {
+int syscall_gdrom_pio_transfer(gdc_cmd_id_t id, const int32_t params[2]) {
     MAKE_SYSCALL_INT(VEC_MISC_GDROM, FUNC_GDROM_PIO_TRANSFER, 
         id, params, SUPER_FUNC_GDROM);
 }
 
-int syscall_gdrom_pio_check(uint32_t id, size_t *size) {
+int syscall_gdrom_pio_check(gdc_cmd_id_t id, size_t *size) {
      MAKE_SYSCALL_INT(VEC_MISC_GDROM, FUNC_GDROM_PIO_CHECK, 
         id, size, SUPER_FUNC_GDROM);
 }
--- a/kernel/arch/dreamcast/include/dc/syscalls.h
+++ b/kernel/arch/dreamcast/include/dc/syscalls.h
@@ -280,26 +280,6 @@ typedef enum cd_cmd_code {
     CMD_MAX               = 47,  /**< \brief Max of GD syscall commands */
 } cd_cmd_code_t;
 
-/** \brief      TOC structure returned by the BIOS.
-    \ingroup    gdrom_syscalls
-
-    This is the structure that the CMD_GETTOC2 syscall command will return for
-    the TOC. Note the data is in FAD, not LBA/LSN.
-
-*/
-typedef struct cd_toc {
-    uint32_t  entry[99];          /**< \brief TOC space for 99 tracks */
-    uint32_t  first;              /**< \brief Point A0 information (1st track) */
-    uint32_t  last;               /**< \brief Point A1 information (last track) */
-    uint32_t  leadout_sector;     /**< \brief Point A2 information (leadout) */
-} cd_toc_t;
-
-/* Compat. Not sure if this is a good way to mark it, or if this is useful to do at 
-    all. But 'CDROM_TOC' violates our style guide which has caps reserved for defines
-    or macros.
-*/
-#define CDROM_TOC __depr("Use the type cd_toc_t rather than CDROM_TOC.") cd_toc_t
-
 /** \brief      Responses from GDROM syscalls
     \ingroup    gdrom_syscalls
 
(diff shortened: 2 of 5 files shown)


kernel/arch/dreamcast/include/dc/syscalls.h
```diff
@@ -258,18 +496,20 @@ int syscall_gdrom_check_command(uint32_t id, int32_t status[4]);
 */
 void syscall_gdrom_exec_server(void);
 
-/** \brief   Abort a queued GDROM command.
+/** \brief      Abort a queued GDROM command.
```
DC-SWAT, line 499: There are no queues, you can only request one command.

kernel/arch/dreamcast/include/dc/syscalls.h
```diff
@@ -246,9 +483,10 @@ uint32_t syscall_gdrom_send_command(uint32_t cmd, void *params);
 
     \sa syscall_gdrom_send_command(), syscall_gdrom_exec_server()
 */
-int syscall_gdrom_check_command(uint32_t id, int32_t status[4]);
+cmd_cmd_chk_t syscall_gdrom_check_command(gdc_cmd_id_t id, int32_t status[4]);
 
-/** \brief   Process queued GDROM commands.
+/** \brief      Process queued GDROM commands.
```
DC-SWAT, line 488: There are no queues, you can only request one command.

DC-SWAT (commented):
Are you sure you want to break the API that much?

DC-SWAT:
> This is in the works to get in the typing and cleanup suggestions from #619 and #736 with a different scope.
> 
> The key purpose is to add strongly defined types for the various returns and parameters of the functions for interacting with the GDROM drive, replacing the many sets of defines, generic types, and untyped structs.
> 
> Changes:
> 
> * [x]  Redefine each group of CD defines to an enum, per [Clean up CD-ROM code #619](https://github.com/KallistiOS/KallistiOS/pull/619). (The only one left out is the unused group of ATA Status defines @DC-SWAT perhaps you have more info on them since you added them? Are these returned by one of the syscalls? Maybe they're the 4th param of Check Command?)

Yeah this is it. 4th param in command status.

QuzarDC:
> Are you sure you want to break the API that much?

Could you please elaborate? This shouldn't break much of anything. The renaming of the error codes and `NO_ACTIVE` to `INACTIVE` is possibly the worst, but those weren't mapped to any external use. You might get warnings but shouldn't get errors from passing the old plain types in or assigning the return of a function to an `int` rather than a `cd_cmd_ret_t`. Maybe I'm horribly wrong, or not considering some uses or settings?

Now that you mention it though, I could add a compatibility set for the changed define names to ensure just a warning for those too.

DC-SWAT:
> > Are you sure you want to break the API that much?
> 
> Could you please elaborate? This shouldn't break much of anything. The renaming of the error codes and `NO_ACTIVE` to `INACTIVE` is possibly the worst, but those weren't mapped to any external use. You might get warnings but shouldn't get errors from passing the old plain types in or assigning the return of a function to an `int` rather than a `cd_cmd_ret_t`. Maybe I'm horribly wrong, or not considering some uses or settings?
> 
> Now that you mention it though, I could add a compatibility set for the changed define names to ensure just a warning for those too.

You renamed the error codes that are used externally.
I mean for example ERR_SYS -> CD_ERR_SYS.
I have 2 applications in DreamsShell that use the сdrom api and I can say for sure that they will break. It is not a big problem of course, to rename constants, but nevertheless. I think that all programs that used сdrom api will break.
It just seemed to me that we were trying not to break backward compatibility all time. But here it was done without any particular need.

QuzarDC:
> > > Are you sure you want to break the API that much?
> > 
> > 
> > Could you please elaborate? This shouldn't break much of anything. The renaming of the error codes and `NO_ACTIVE` to `INACTIVE` is possibly the worst, but those weren't mapped to any external use. You might get warnings but shouldn't get errors from passing the old plain types in or assigning the return of a function to an `int` rather than a `cd_cmd_ret_t`. Maybe I'm horribly wrong, or not considering some uses or settings?
> > Now that you mention it though, I could add a compatibility set for the changed define names to ensure just a warning for those too.
> 
> You renamed the error codes that are used externally. I mean for example ERR_SYS -> CD_ERR_SYS. I have 2 applications in DreamsShell that use the сdrom api and I can say for sure that they will break. It is not a big problem of course, to rename constants, but nevertheless. I think that all programs that used сdrom api will break. It just seemed to me that we were trying not to break backward compatibility all time. But here it was done without any particular need. More precisely, of course it’s better to name it as you did, but it breaks.

Ok, that should be the simplest thing to provide a compatibility layer for, as I did with `CDROM_TOC`. Then remove one version out (similar to removing the old `change_dataype` typo now). The rest of the type changes wouldn't break anything for you right? If it's just the err defines then we should be good :)

DC-SWAT:
> > > > Are you sure you want to break the API that much?
> 
> > > 
> 
> > > 
> 
> > > Could you please elaborate? This shouldn't break much of anything. The renaming of the error codes and `NO_ACTIVE` to `INACTIVE` is possibly the worst, but those weren't mapped to any external use. You might get warnings but shouldn't get errors from passing the old plain types in or assigning the return of a function to an `int` rather than a `cd_cmd_ret_t`. Maybe I'm horribly wrong, or not considering some uses or settings?
> 
> > > Now that you mention it though, I could add a compatibility set for the changed define names to ensure just a warning for those too.
> 
> > 
> 
> > You renamed the error codes that are used externally. I mean for example ERR_SYS -> CD_ERR_SYS. I have 2 applications in DreamsShell that use the сdrom api and I can say for sure that they will break. It is not a big problem of course, to rename constants, but nevertheless. I think that all programs that used сdrom api will break. It just seemed to me that we were trying not to break backward compatibility all time. But here it was done without any particular need. More precisely, of course it’s better to name it as you did, but it breaks.
> 
> 
> 
> Ok, that should be the simplest thing to provide a compatibility layer for, as I did with `CDROM_TOC`. Then remove one version out (similar to removing the old `change_dataype` typo now). The rest of the type changes wouldn't break anything for you right? If it's just the err defines then we should be good :) 

Yeah, only error codes and TOC.

QuzarDC:
@DC-SWAT , I've removed the reference to queuing in favor of 'request' as described. I also incorporated the abort command and refactor from #736 to try to reduce the overhead of requesting further changes from you. I know that this will require some rewriting (aside from what is needed for this whole PR) but I hope the changes made sense.

kernel/arch/dreamcast/hardware/syscalls.c
```diff
@@ -204,29 +204,29 @@ void syscall_gdrom_dma_callback(uintptr_t callback, void *param) {
...
+        hnd, size, SUPER_FUNC_GDROM);
 }
 
 void syscall_gdrom_pio_callback(uintptr_t callback, void *param) {
     MAKE_SYSCALL_VOID(VEC_MISC_GDROM, FUNC_GDROM_PIO_CALLBACK, 
         callback, param, SUPER_FUNC_GDROM);
 }
 
-int syscall_gdrom_pio_transfer(uint32_t id, const int32_t params[2]) {
+int syscall_gdrom_pio_transfer(gdc_cmd_hnd_t hnd, const int32_t params[2]) {
```
DC-SWAT, line 222: The same here.

kernel/arch/dreamcast/hardware/syscalls.c
```diff
@@ -204,29 +204,29 @@ void syscall_gdrom_dma_callback(uintptr_t callback, void *param) {
         callback, param, SUPER_FUNC_GDROM);
 }
 
-int syscall_gdrom_dma_transfer(uint32_t id, const int32_t params[2]) {
+int syscall_gdrom_dma_transfer(gdc_cmd_hnd_t hnd, const int32_t params[2]) {
```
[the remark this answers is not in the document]
QuzarDC: Thank you! Updated.

kernel/arch/dreamcast/include/dc/syscalls.h
```diff
@@ -183,42 +182,324 @@ int syscall_flashrom_write(uint32_t pos, const void *src, size_t n);
...
+/** \brief      Params for READ commands.
+    \ingroup    gdrom_syscalls
+
+    These are the parameters for the CMD_PIOREAD and CMD_DMAREAD commands.
+
+*/
+typedef struct cd_read_params {
+    uint32_t    start_sec;  /* Starting sector */
+    size_t      num_sec;    /* Number of sectors */
+    void       *buffer;     /* Output buffer */
```
andressbarajas, line 365:
```suggestion
    void        *buffer;     /* Output buffer */
```

kernel/arch/dreamcast/include/dc/syscalls.h
```diff
@@ -183,42 +182,324 @@ int syscall_flashrom_write(uint32_t pos, const void *src, size_t n);
...
+    CD_SUB_RESERVED       = 4     /**< \brief Reserved */
+} cd_sub_type_t;
+
+/** \brief      Params for GETSCD command
+    \ingroup    gdrom_syscalls
+*/
+typedef struct cd_cmd_getscd_params {
+    cd_sub_type_t   which;
+    size_t          buflen;
+    void           *buffer;
```
andressbarajas, line 445:
```suggestion
    void            *buffer;
```

kernel/arch/dreamcast/include/dc/syscalls.h
```diff
@@ -292,12 +630,23 @@ int syscall_gdrom_sector_mode(uint32_t mode[4]);
...
 void syscall_gdrom_dma_callback(uintptr_t callback, void *param);
 
-/** \brief   Initiates a GDROM DMA transfer.
+/** \brief      Parameters for transfer
+    \ingroup    gdrom_syscalls
+
+    These are parameters passed to the syscall_gdrom_*_transfer syscalls.
+*/
+typedef struct cd_transfer_params {
+    void*   addr;       /**< \brief Destination address of transfer */
```
andressbarajas, line 639:
```suggestion
    void    *addr;       /**< \brief Destination address of transfer */
```

DC-SWAT:
Are you going to finish the last 2 points in this PR?
It needs to be merged. The only request is to return the abortion back to the cdrom driver.

QuzarDC:
I've gone ahead and rebuilt this PR. The commits should now all be quite atomic and cleanly separated. The big thing that I left out was implementing the same for the error codes. Those are going to require a substantial amount of riskier changes as we are currently mixing and matching the different error code types within cdrom.c as well as hijacking some for our own errors. So it seems appropriate to have that as a separate PR that isn't as big.

kernel/arch/dreamcast/include/dc/syscalls.h
```diff
@@ -183,42 +183,244 @@ int syscall_flashrom_write(uint32_t pos, const void *src, size_t n);
...
+    GDROM drive. These were originally provided by maiwe.
+*/
+typedef enum cd_cmd_code {
+    CD_CMD_CHECK_LICENSE     =  2,  /**< \brief Check license */
+    CD_CMD_REQ_SPI_CMD       =  4,  /**< \brief Request to Sega Packet Interface */
+    CD_CMD_PIOREAD           = 16,  /**< \brief Read via PIO */
+    CD_CMD_DMAREAD           = 17,  /**< \brief Read via DMA */
+    CD_CMD_GETTOC            = 18,  /**< \brief Read TOC */
+    CD_CMD_GETTOC2           = 19,  /**< \brief Read TOC */
+    CD_CMD_PLAY              = 20,  /**< \brief Play track */
```
DC-SWAT:
```suggestion
    CD_CMD_PLAY_TRACKS       = 20,  /**< \brief Play track */
```

kernel/arch/dreamcast/include/dc/syscalls.h
```diff
@@ -183,42 +183,244 @@ int syscall_flashrom_write(uint32_t pos, const void *src, size_t n);
...
+    CD_CMD_PLAY2             = 21,  /**< \brief Play sectors */
```
DC-SWAT:
```suggestion
    CD_CMD_PLAY_SECTORS      = 21,  /**< \brief Play sectors */
```

kernel/arch/dreamcast/include/dc/syscalls.h
QuzarDC: Oh! Of course, great idea. Do you know if there's similar kind of info on the difference between GETTOC/GETTOC2?

DC-SWAT: All games use only `GETTOC2`, so I don't even emulate `GETTOC`. One could suggest that there is no high density area in `GETTOC`, but the number of arguments for both is the same, so the difference maybe in something else.

QuzarDC:
Last two force-pushes were simply to ensure there were still no conflicts after other merges and we're up to date. No changes to the commits.

QuzarDC:
Force-pushed to rebase onto master after #1172 .

QuzarDC:
Had to rebase to account for the change in #1217 to use a semaphore rather than mutex for dma.

darcagn:
Seems to be working OK now. This is a big one; hopefully we didn't miss anything...

PR #779, opened 2024-09-28
Add support for Store Queues with MMU and INIT_MMU init flag

pcercuei:
This set of commits add a `INIT_MMU` init flag, and add support for the SQs in programs that use the MMU.

The MMU code will reserve the last TLB entry for translating the SQ memory area using a 1 MiB page.
The `sq_lock()` function now returns the translated pointer (which is equivalent to using `SQ_MASK_DEST()` when MMU is disabled).

The caveat is that the SQ memory window is "only" 1 MiB. But apart from e.g. clearing the whole VRAM (which is still supported as sq_cpy / sq_set32 were updated), I don't think this will cause any issue.


DC-SWAT:
So, you couldn't get rid of the mutex?

pcercuei:
> So, you couldn't get rid of the mutex?

That will be for a future commit, if at all possible.

pcercuei:
Rebased on top of master (and the recursive sq_lock() change).

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -115,38 +122,48 @@ void sq_wait(void) {
...
-            sq_flush(d);
-            d += 8;
+    while (n > 0) {
+        /* Transfer maximum 512 KiB at once. This is because when using the
+         * MMU the SQ area is a 1 MiB page, and the destination address may
+         * not be on a page boundary. */
+        nb = n > 0x4000 ? 0x4000 : n;
+
+        d = sq_lock(curr_dest);
+
```
QuzarDC, line 140: With the changes I suggested for adding a new function to gather this, it shouldn't be necessary to lock/unlock at every 512 loop.

kernel/arch/dreamcast/kernel/mmu.c
```diff
@@ -767,14 +771,13 @@ void mmu_shutdown(void) {
...
+    return *mmucr & 0x1;
 }
 
-void mmu_restore(mmu_token_t token) {
-    *mmucr = (uint32)token;
+void mmu_set_sq_addr(void *addr) {
+    uint32_t ppn = (uint32_t)addr & 0x1ffffc00;
+
+    /* Reset the base target address for the SQs */
+    *(uint32_t *)(0xf7000000 + (0x3f << 8)) = ppn | 0x1fc;
```
QuzarDC, line 782:
```suggestion
void mmu_set_sq_addr(void *addr) {
    uint32_t ppn = (uint32_t)addr & 0x1ffffc00;

    /* Reset the base target address for the SQs */
    *(uint32_t *)(MEM_AREA_UTLB_DATA_ARRAY1_BASE + (0x3f << 8)) = ppn | 0x1fc;
```

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -144,6 +147,8 @@ int  __weak arch_auto_init(void) {
     /* Initialize memory management */
     mm_init();
 
+    KOS_INIT_FLAG_CALL(mmu_init);
```
QuzarDC, line 150: Is there a reason for this other than to save needing the user to call this on their own? It's not like the other subsystems where you'd enable a better ability to garbage collect, since it's never been called by default. And a user would still have to configure the mmu translations they want to use before they could leverage the mode. Though I do get how that would bring the mmu into functioning like the various other subsystems we initialize in `arch_auto_init`

It seems to me like the better use for a compile-time setting for MMU would be to do something like disable it entirely, so that a function like `mmu_enabled()` could be a static `false`.

QuzarDC (changes requested):
Some comments, some requests for change. Overall the notion of reserving a page makes perfect sense, and it should be easily expandable in the future to be dynamic without having to change the sq API.

The biggest changes I'm asking for are mostly for refactoring and to provide simpler compatibility with the use of the sqs outside of sq.c as noted. Then, to a degree I question the `INIT_MMU` flag's usefulness. If anything maybe that could/should be a future PR instead?

kernel/arch/dreamcast/hardware/sq.c
DC-SWAT: Yeah, this change may slow down copying large textures.
But I like the return of the destination, it should have been done right away.

QuzarDC: > Yeah, this change may slow down copying large textures. But I like the return of the destination, it should have been done right away.
> 
> Edit: Although if you look at how I rewrote [convert_YUV420_to_YUV422_texture](https://github.com/KallistiOS/KallistiOS/blob/master/examples/dreamcast/pvr/yuv_converter/YUV420/yuv420.c#L183) in the example, you can understand that the single destination is a rarity :)) snd_pcm16_split_sq is lite case.

Yea, I was originally going to suggest to not return the address and keep it as void adding the macro replacement to get the address. But I see too how it simplifies the most common uses and doesn't interfere with these more complicated ones :)

kernel/arch/dreamcast/include/arch/mmu.h
```diff
@@ -373,19 +370,18 @@ void mmu_shutdown(void);
...
  */
-mmu_token_t mmu_disable(void);
+bool mmu_enabled(void);
 
-/** \brief   Restore MMU address translation.
-    \ingroup mmu
-
-    \param  token           The opaque token obtained from mmu_disable()
- */
-void mmu_restore(mmu_token_t token);
```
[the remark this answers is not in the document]
pcercuei: The disable/restore was added because SQs didn't work with MMU, now they do, so it made sense to drop it. I can keep it around, but it's dead code.

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -67,26 +65,33 @@ typedef struct sq_state {
...
+    mask = with_mmu ? 0x000fffe0 : 0x03ffffe0;
 
-    SET_QACR_REGS_INNER(new_state->dest0, new_state->dest1);
+    if (with_mmu)
+        mmu_set_sq_addr(dest);
+    else
+        SET_QACR_REGS(dest, dest);
+
+    return (uint32_t *)(MEM_AREA_SQ_BASE | ((uintptr_t)dest & mask));
 }
```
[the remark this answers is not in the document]
pcercuei: Why does it *need* to be split though?

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -115,38 +122,48 @@ void sq_wait(void) {
...
-            d[5] = *(s++);
-            d[6] = *(s++);
-            d[7] = *(s++);
-            sq_flush(d);
-            d += 8;
+    while (n > 0) {
+        /* Transfer maximum 512 KiB at once. This is because when using the
+         * MMU the SQ area is a 1 MiB page, and the destination address may
+         * not be on a page boundary. */
+        nb = n > 0x4000 ? 0x4000 : n;
```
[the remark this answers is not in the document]
pcercuei: Well there's a comment right above explaining it :)
The SQ area with MMU is 1MiB, so to avoid getting a segmentation fault, we move the SQ window after each 512 KiB transferred (and not 1 MiB because the destination address may start from an offset within the SQ page).

kernel/arch/dreamcast/kernel/init.c
pcercuei: Consistency? IDK, it just makes sense to me.

People might want to try porting stuff that use `mmap()`, having MMU as an init flag means that they wouldn't have to hack the code to get it to work on KOS.

kernel/arch/dreamcast/hardware/sq.c
pcercuei: > With the changes I suggested for adding a new function to gather this, it shouldn't be necessary to lock/unlock at every 512 loop.

It would still be needed, as you might have to reprogram the MMU SQ window.

kernel/arch/dreamcast/kernel/init.c
gyrovorbis: It is possible for some C++ or Rust or static initializer/constructor function to call `mmap()` or call some nontrivial function during initialization before `main()` even starts that could potentially rely on the MMU.

pcercuei:
Updated.
- Rebased on master.
- INIT_MMU stuff removed, I'll move it to a different PR, and we can discuss it there.
- I now reserve the last two TLB entries and map two 1 MiB pages. This is because the PPN has to be aligned to 1 MiB, and the SQ destination address may be just slightly before a page boundary.

kernel/arch/dreamcast/hardware/sq.c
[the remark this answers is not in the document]
QuzarDC: So the offset cannot be more than 512KiB? My question isn't the reason for limiting to less than the size of the page, which is explained in the comment, but why the specific amount 512KiB is used.

kernel/arch/dreamcast/hardware/sq.c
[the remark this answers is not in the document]
QuzarDC: Isn't 'updating the SQ window' exactly what `mmu_set_sq_addr` does?

kernel/arch/dreamcast/hardware/sq.c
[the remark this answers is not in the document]
pcercuei: I guess I was confused :)
It's 1 MiB now.

QuzarDC (approved):

PR #781, opened 2024-09-30
Blessing GCC14.2.0 as the new "stable" toolchain.

gyrovorbis:
GCC14.2.0 has been previewed by at least half of us KOS devs, plus a substantial proportion of our users ever since it has released. I've only ever seen *fewer* ICEs and issues than with the current "stable" toolchain, plus it has nice new C++ language features. I think the time has come to mark it as stable, as many users are already having to rebuild toolchains manually to get this version anyway...

- Updated "stable" and "13.2.0" toolchain descriptions in Makefile.default.cfg
- Added profile.13.2.0.mk.
- Renamed profile.14.2.0 -> profile.stable.mk
- Updated CHANGELOG.md.


darcagn:
We don't even have a 14.2.0 anymore, so we might as well close this. A new PR can be opened when we do decide to bless a newer version.

PR #782, opened 2024-09-30
CMake: Use kos-cc as the assembler

pcercuei:
CMake does detect the GNU assembler correctly, but does not seem to know how to use it properly; it passes flags it shouldn't (e.g. -with-sysroot or -isystem), or flags with the incorrect arguments.

This is because on most (all?) GNU platforms, CMake is configured to use GCC itself as the assembler. In that case, it passes the correct (and different) flags which makes it able to compile assembler files.

Fix compilation of assembler files in CMake by using kos-cc as the assembler.


gyrovorbis (approved):
Good call.

PR #783, opened 2024-10-02
thread: Add priority boosting system

pcercuei:
Introduce a new 'real_prio' field into the thread structure, which contains the static priority set during the thread creation or with thd_set_prio().

The 'prio' field now represents the dynamic priority. When a low-priority thread A holds a lock, and a high-priority thread B requests it, the dynamic priority of A will be temporarily raised to the dynamic priority of B, until A releases the lock, in which case it falls back to its static priority.

The point of this setup is to avoid starvation, where a low-priority thread holds a resource that is needed by a high-priority thread, and won't release it because it rarely gets preempted.

This is not yet perfect, as we can imagine a setup where a low-priority thread A holds two locks, each one requested by different high-priority tasts B and C; A would get boosted to B's dynamic priority, and when releasing the lock, reset to its original low priority, causing thread C to starve.

An example program is available here: https://gist.github.com/pcercuei/591edca1727aa092ee8aa03b3d35906a
Which shows the starvation problem in the current KOS, and shows it resolved after this change.

### DIFF
--- a/kernel/thread/mutex.c
+++ b/kernel/thread/mutex.c
@@ -133,6 +133,15 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
             deadline = timer_ms_gettime64() + timeout;
 
         for(;;) {
+            if (m->holder->prio >= thd_current->prio) {
+                m->holder->prio = thd_current->prio;
+
+                /* Thread list is sorted by priority, update the position
+                 * of the thread holding the lock */
+                thd_remove_from_runnable(m->holder);
+                thd_add_to_runnable(m->holder, 0);
+            }
+
             rv = genwait_wait(m, timeout ? "mutex_lock_timed" : "mutex_lock",
                               timeout, NULL);
             if(rv < 0) {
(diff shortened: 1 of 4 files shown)


kernel/thread/mutex.c
```diff
@@ -133,6 +133,15 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
             deadline = timer_ms_gettime64() + timeout;
 
         for(;;) {
+            if (m->holder->prio > thd_current->prio) {
```
QuzarDC, line 136: Would it make sense for this to be `>=` ? That wouldn't change the priority level, but what it would do is force it to be rescheduled to the front of the line for its priority level. So if you have 10 threads with the same priority in queue and No.10 in the queue is holding the mutex that No.1 needs, it'll get rescheduled to run before No.2 in the queue.

I don't think that'd be too disruptive or unfair to the scheduling, but I'm not 100% and it might just cause a bit too much extra overhead.

pcercuei: Good idea.

pcercuei:
Updated, with a CHANGELOG entry + the `>=` change suggested by @QuzarDC.

QuzarDC (approved):

kernel/thread/mutex.c
pcercuei: @QuzarDC Now that I see it in master, I'm pretty sure that we don't add it to the front of the line, but to the back of the line. `thd_add_to_runnable`'s last argument is named `front_of_line`, and we set it to 0.

PR #784, opened 2024-10-03
Make the Store Queue mutex recursive

### LINKED ISSUE #780: Some SQ-related functions hangs in a middle of a scene
This happens with `snd_stream_init()` and `pvr_sq_load()` for example

example code:
```C
#include <arch/arch.h>
#include <dc/pvr.h>
#include <dc/sound/stream.h>
#include <kos/dbglog.h>
#include <stdio.h>

int main(int argc, char* argv[]) {
    pvr_init(&(pvr_init_params_t){
        // Enable only translucent polygons with size 16
        .opb_sizes = {PVR_BINSIZE_0, PVR_BINSIZE_0, PVR_BINSIZE_16, PVR_BINSIZE_0, PVR_BINSIZE_0},

        // Vertex buffer size
        .vertex_buf_size = (512 * 1024), // 512KiB

        // No DMA
        .dma_enabled = 0,

        // No FSAA
        .fsaa_enabled = 0,

        // Translucent Autosort disabled.
        .autosort_disabled = 1,

        // Extra OPBs
        .opb_overflow_count = 0
    });

    pvr_scene_begin();
    pvr_list_begin(PVR_LIST_TR_POLY);

    dbglog(DBG_NOTICE, "about to call snd_stream_init()....\n"); fflush(stdout);
    snd_stream_init();

    dbglog(DBG_INFO, "¡sucess!\n"); fflush(stdout);
    arch_menu();
}
````

### PR DESCRIPTION
**Pull request by QuzarDC**
This should reduce the breakages that have happened from the updates to `pvr_list_begin()` and `pvr_prim()` from #740 and fix #780

### DIFF
--- a/kernel/arch/dreamcast/hardware/sq.c
+++ b/kernel/arch/dreamcast/hardware/sq.c
@@ -73,7 +73,7 @@ void sq_lock(void *dest) {
 
     mutex_lock(&sq_mutex);
 
-    assert_msg(sq_mutex.count >= SQ_STATE_CACHE_SIZE, "You've overrun the SQ_STATE_CACHE.");
+    assert_msg(sq_mutex.count < SQ_STATE_CACHE_SIZE, "You've overrun the SQ_STATE_CACHE.");
 
     new_state = &sq_state_cache[sq_mutex.count - 1];
 
--- a/kernel/arch/dreamcast/hardware/sq.c
+++ b/kernel/arch/dreamcast/hardware/sq.c
@@ -7,6 +7,10 @@
    Copyright (C) 2023 Ruslan Rostovtsev
 */
 
+#include <assert.h>
+#include <stdlib.h>
+#include <sys/queue.h>
+
 #include <arch/cache.h>
 #include <arch/mmu.h>
 #include <dc/sq.h>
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by QuzarDC** [comment]
> Maybe for first lock use static memory for sq_state_t?

Yea, at the very least should do this. Two other ways I mentioned on discord last night were:
1) Changing the API so that it's `sq_state_t sq_lock(void* dest)` and `sq_unlock(sq_state_t old)`. Or alternatively to `sq_lock(void* dest, sq_state_t *old)` and no change to sq_unlock.
2) Setting a small static table to store the state. The whole state is actually just 7 bits (3 for each of the QACR regs, one for mmu on/off) and it seems unlikely that anyone would nest more than a few (maybe 8?) after which we throw an error. I hate hardcoding static sizes for variable things, but it seems reasonable here.

**Comment by DC-SWAT** [comment]
> > Maybe for first lock use static memory for sq_state_t?
> 
> Yea, at the very least should do this. Two other ways I mentioned on discord last night were:
> 
> 1. Changing the API so that it's `sq_state_t sq_lock(void* dest)` and `sq_unlock(sq_state_t old)`. Or alternatively to `sq_lock(void* dest, sq_state_t *old)` and no change to sq_unlock.
> 2. Setting a small static table to store the state. The whole state is actually just 7 bits (3 for each of the QACR regs, one for mmu on/off) and it seems unlikely that anyone would nest more than a few (maybe 8?) after which we throw an error. I hate hardcoding static sizes for variable things, but it seems reasonable here.

In my opinion, it is better not to change the API. We can store all states statically, allocate, say, 8x8 bytes, I think it will be enough. In extreme cases it can be warned by debug message.

File: kernel/arch/dreamcast/hardware/sq.c [hunk 1]
```diff
@@ -39,22 +42,55 @@
...
+static mutex_t sq_mutex = RECURSIVE_MUTEX_INITIALIZER;
+
+typedef struct sq_state {
+    void *dest0;
+    void *dest1;
+    mmu_token_t mmu_token;
+} sq_state_t;
 
-static mmu_token_t mmu_token;
+#define SQ_STATE_CACHE_SIZE 8
```
gyrovorbis, line 53:
```suggestion
#ifndef SQ_STATE_CACHE_SIZE
#define SQ_STATE_CACHE_SIZE 8
#endif
```
This is what I'd do, just so that if some crazy person ever shows up needing 20 layers of SQ locks deep, we can tell them to just pass `-DSQ_STATE_CACHE_SIZE=9999` to `make` when they build KOS... lol.

kernel/arch/dreamcast/hardware/sq.c
```diff
@@ -7,6 +7,10 @@
    Copyright (C) 2023 Ruslan Rostovtsev
 */
 
+#include <assert.h>
+#include <stdlib.h>
+#include <sys/queue.h>
```
gyrovorbis:
```suggestion
```
Pretty sure not needed anymore.

gyrovorbis (approved):
Very nice!

PR #785, opened 2024-10-03
environ_base.sh: Remove `-fno-operator-names` flag

glebm:
This flag disabled support for `and/or/not` operator aliases, which are part of the C++ standard since its very first version (C++98, see page 20 at https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf) and are used in some open-source libraries, such as https://github.com/realnc/SDL_audiolib.


gyrovorbis (approved):
He's totally right. I personally never use these operator names, so I never even ran into an issue or noticed that this flag was here, but... We're essentially defaulting everyone's C++ to be noncompliant for probably some legacy reason, when this should be on the offending code to do in their own local project if they actually were unwise enough to use these reserved words as identifiers...

Approving... now let me go make sure we don't have kos-ports that are that foolish real quick... ;)

gyrovorbis:
k, don't see any new KOS-ports bitching at me with the change... it shall be done.

PR #787, opened 2024-10-04
Implement Support for C99 Floating-Point Environment

gyrovorbis:
C99's `<fenv.h>` adds all kinds of useful things for working with the floating-point environment of a FPU, including enabling/disabling exceptions, checking status flags, swapping round modes, etc. Turns out Hitachi went to great lengths to ensure the SH4's FPU mapped extremely well to this standard C API, and it implements almost the entirety of the feature-set.... Just too bad Newlib doesn't have any implementation of this API for most of its embedded targets... and yes, several people working with several different ports have specifically mentioned they need support for this functionality... and tbh, it's just good to have.

TIME TO ROLL OUR OWN!

- Initial implementation of FENV complete
    * GNU extension implementation complete
    * POSIX extension implementation complete
- Doxygen still has work to be done
- Header file/API still needs everything exposed
- Example/Test scenario needs to be written

### DIFF
--- a/include/sys/fenv.h
+++ b/include/sys/fenv.h
@@ -50,6 +50,102 @@ typedef uint8_t fenv_t;
 /** Type representing all FP status flags collectively. */
 typedef uint8_t fexcept_t;
 
+/** Clears the given floating-point exception flags.
+ * 
+ *  \param  excepts     Bitwise OR'd list of FP exception flags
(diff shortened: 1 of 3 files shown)


QuzarDC:
@gyrovorbis - Maybe in 2.3.0? sh4zam could maybe interact with `<fenv.h>` for settings

PR #789, opened 2024-10-05
Remove LTO flags from pvrtex

andressbarajas:
Remove compiling pvrtex with LTO because it is not needed and can cause issues on certain hosts.


gyrovorbis (approved):
Thank you!

PR #790, opened 2024-10-05
Made macros for HZ and THD stack sizes overridable

gyrovorbis:
- Many people are starting to overrun the main kernel thread stack
- We may want to eventually arrive at a better solution for providing a thread size for the kernel stack eventually, but until we get there, I think this will suffice, as it's in-line with our other overridable macros.
- Added `#ifdef`s around `HZ`, `THD_STACK_SIZE`, and `THD_KERNEL_STACK_SIZE` in `arch.h` to only define them if they weren't already provided by the build system or externally.

Here's an example of someone needing such behavior ASAP:
https://github.com/azihassan/devilutionX/issues/3

### DIFF
--- a/kernel/arch/dreamcast/include/arch/arch.h
+++ b/kernel/arch/dreamcast/include/arch/arch.h
@@ -52,7 +52,7 @@ extern uint32 _arch_mem_top;
 /** \brief  Base address of available physical pages. */
 #define page_phys_base  0x8c010000
 
-#ifndef HZ
+#ifndef THD_SCHED_HZ
 /** \brief Scheduler interrupt frequency
 
     Timer interrupt frequency for the KOS thread scheduler.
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/include/arch/arch.h
```diff
@@ -52,6 +52,7 @@ extern uint32 _arch_mem_top;
 /** \brief  Base address of available physical pages. */
 #define page_phys_base  0x8c010000
 
+#ifndef HZ
```
QuzarDC, line 55: Before cementing this to be more public facing, lets finally get it to a name more in line with our style that's less likely to get conflicts. `THD_SCHED_HZ` or something like that? Then toss in a compat def of `HZ` with it that throws a __depr warning and we can get rid of it in the next version or two too, just in case anyone is already relying on it.

gyrovorbis: DONE. But unfortunately it looks as though GCC doesn't really allow you to tag a preprocessor `#define` as deprecated, so the workaround is to just make a constant static global variable called `HZ`, set equal to `THD_SCHED_HZ`, which we can apply the `__depr()` attribute to.

QuzarDC (approved):

kernel/arch/dreamcast/include/arch/arch.h
QuzarDC: Ah yea, I just did it anyways in the cd refresh like this: `#define CDROM_TOC __depr("Use the type cd_toc_t rather than CDROM_TOC.") cd_toc_t` but what it ends up technically doing is returning a kind of mixed message because whatever it's next to might get the actual marking as deprecated rather than the define itself.

Your way's probably a better way to do it. Unless someone was doing `#undef HZ` 🤣

PR #791, opened 2024-10-05
Fixed pvrtex warnings with Clang18 host compiler

gyrovorbis:
- A couple of warnings are manifesting in the pvrtex codebase when bulding with a recent verison of Clang.
- ATOMIC_INIT() was deprecated in later versions of C, which Clang defaults to now
    * forcing project to be built uniformly using GNU17 C standard for everyone
    * got rid of ATOMIC_INIT() usage
- Clang correctly diagnosed a set but unused variable in pvrtex... Simply told the compiler to shut up by marking it (void).

### DIFF
--- a/utils/pvrtex/pvr_texture_encoder.c
+++ b/utils/pvrtex/pvr_texture_encoder.c
@@ -838,6 +838,7 @@ void pteAutoSelectPixelFormat(PvrTexEncoder *pte) {
 	assert(pte->src_img_cnt);
 
 	unsigned clearpix = 0, halfpix = 0, opaquepix = 0;
+	(void)opaquepix;
 
 	for(int j = 0; j < pte->src_img_cnt; j++) {
 		pteImage *img = pte->src_imgs + j;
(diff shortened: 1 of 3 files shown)


utils/pvrtex/pvr_texture_encoder.c
```diff
@@ -838,6 +838,7 @@ void pteAutoSelectPixelFormat(PvrTexEncoder *pte) {
 	assert(pte->src_img_cnt);
 
 	unsigned clearpix = 0, halfpix = 0, opaquepix = 0;
+	(void)opaquepix;
```
QuzarDC, line 841: Why not just remove this entirely? I guess it could be nice for documentation but could just have the inner loop if block have `else if (a != 0xff) halfpix++;` yea?

gyrovorbis: Sure could. I dunno. I was like "maybe one day TapamN wants to use this variable, or he wouldn't have had it?" Dunno. Can just as easily remove it. Will do, np.

QuzarDC (approved):

PR #792, opened 2024-10-05
Add raylib examples taken from raylib4consoles repo with permission

andressbarajas:


QuzarDC:
Since these come from the port repo, we should have them build as part of the kos-ports build process and get dumped into the examples folder there. [My port of PDCurses](https://github.com/KallistiOS/kos-ports/pull/52) was also basically just waiting on trying to work out that process (as we've not done it for any port yet). It makes a lot of sense to do it like that because then we're not maintaining them independently and can benefit from their upstream maintenance and expansion.

The thing I'd not quite worked out is that it seems like we'd rather likely want to also copy the source for examples to be available after a typical `make install` and how to ensure that we could still do something like rebuild individual examples. This would be a great sample case for that, but I'm not sure how much effort would be involved (I'm just garbage with makefiles and environment stuff, so it's hard to tell if it's a simple thing I just have no ideas on or a really complicated one).

If you're not interested, then I can take a bit of a look again and see how bad it might be for the raylib examples, and then likely just come back and approve this, but it would be nice to finally get the kos-ports examples going.

gyrovorbis:
Just to let you guys know, Raysan, author of Raylib, is planning to mention that support has been officially added to kos-ports for Raylib in the notes for the upcoming next big release, which... I honestly think means expect a bunch of new Raylibbers appearing who will need examples. ;)

andressbarajas:
@QuzarDC Good point.  I'll see out some other raylib examples that arent included in the port repo.

PR #793, opened 2024-10-05
Made builtin font used by vmu_printf() public.

gyrovorbis:
- Need a quick font without external dependencies to use with the extended font-drawing methods in the vmufb API? Sweet. We already have one used by vmu_printf()
- Simply remove the static keyword from the font definition and then declare it extern in the vmu_fb.h header file for others to use and profit from.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/vmu_fb.h
+++ b/kernel/arch/dreamcast/include/dc/vmu_fb.h
@@ -146,8 +146,9 @@ void vmufb_print_string_into(vmufb_t *fb,
                             painting the text (or NULL to use the default)
     \param  str             The text to render
  */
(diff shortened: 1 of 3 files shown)


QuzarDC (commented):
Any chance you could go one step further? Expose an API to set/get the font address to be used by `vmu_printf(...)` (which gets initialized to `vmufb_font4x6`)

gyrovorbis:
> Any chance you could go one step further? Expose an API to set/get the font address to be used by `vmu_printf(...)` (which gets initialized to `vmufb_font4x6`)

I think that's a fantastic idea.

pcercuei:
- `builitn` -> `builtin`
- `Needa` -> `Need a`
- `haev` -> `have`

QuzarDC (approved):

PR #794, opened 2024-10-06
Fix: OCRAM mode is initialized with bit 5 of CCR not bit 6

maslevin:
Changed constant ccr_data_ocram in startup.s to 0x091d from 0x092d - previous code was toggling bit 6 of CCR, which is unused, instead of bit 5 which controls OCRAM mode.

See page 79 of the SH4 Manual Here:

https://www.st.com/resource/en/user_manual/cd00147165-sh-4-32-bit-cpu-core-architecture-stmicroelectronics.pdf

### DIFF
--- a/kernel/arch/dreamcast/kernel/startup.s
+++ b/kernel/arch/dreamcast/kernel/startup.s
@@ -236,4 +236,4 @@ ccr_addr:
 ccr_data:
 	.word	0x090d
 ccr_data_ocram:
-	.word	0x092d
+	.word	0x091d


QuzarDC:
I can see the same in the docs: 
image

My only question is how #620 was validated/tested? Because there was much uproar about how that was making a huge difference, but the difference it would have made would be on the OC Index behavior with OCRAM still being disabled.  @gyrovorbis @kilgariff (maybe?) could help comment?

It's important to be able to validate it may be that the value should be set to `0x093d` to keep the index mode set as it was, if that was making a huge difference to performance on its own.

maslevin:
I'm going to back out this pull request because further experimentation has confirmed to me that it's actually bit 6 of CCR that controls the OCRAM mode, not bit 5.  We should probably document this in the cache source code somewhere as well as the wiki.

PR #801, opened 2024-10-08
thread: Improve priority boosting

pcercuei:
When failing to obtain an already locked mutex owned by a thread with the same priority than us, we want the other thread to be bumped to the front of the queue, not to the end of the queue.


gyrovorbis (approved):
It looks as though we seem to have some sort of thread-starvation issue going on that's crashing GTA3 now after the cutscenes end and gameplay starts, which this PR does not fix, but I don't see how this change cannot possibly be the right thing to do and a step in the right direction regardless.

PR #805, opened 2024-10-10
Fix vmu_lcd compile error and extended vmu font API

andressbarajas:
Fix vmu_lcd example compile error by utilizing fontdata_4x6 and vmufb_font4x6 definitions in util/vmu_printf.c


QuzarDC (dismissed):

QuzarDC:
@gyrovorbis - Though this change could still stand, might it make sense to not have the font be extern, even if just to reduce possible name conflicts, and rely on the implemented API to get/set it? Could be the example also gets rewritten to demonstrate that usage.

gyrovorbis:
> @gyrovorbis - Though this change could still stand, might it make sense to not have the font be extern, even if just to reduce possible name conflicts, and rely on the implemented API to get/set it? Could be the example also gets rewritten to demonstrate that usage.

Yeah. Then we should make that thing `static` again, and something to the API for `vmu_get_builtin_font()` which you can use to access it, otherwise, you can just overwrite it when changing the active font and never retrieve it again.

andressbarajas:
@QuzarDC @gyrovorbis Implemented suggestions

QuzarDC:
> > @gyrovorbis - Though this change could still stand, might it make sense to not have the font be extern, even if just to reduce possible name conflicts, and rely on the implemented API to get/set it? Could be the example also gets rewritten to demonstrate that usage.
> 
> Yeah. Then we should make that thing `static` again, and something to the API for `vmu_get_builtin_font()` which you can use to access it, otherwise, you can just overwrite it when changing the active font and never retrieve it again.

Aha, yup, that makes sense!

QuzarDC (approved):

PR #810, opened 2024-10-11
Separate out trap from irq. 

QuzarDC:
This is separated out from #618 to try to reduce the scope of that PR to be more digestible. My intent is to do at least one more that will have the stylistic/type changes from the PR and some of the documentation update.

Since this is almost a review of that PR I made two changes compared to what was in the PR:
1) The \retval documentation for the setter was updated to remove the failure return, as it is no longer possible to trigger or checked at all.
2) Adding of trap.h into kos.h wasn't done. It might be that it's the right way to go by the end of it, but it seems very much to me that including trap.h would *always* require irq.h as they share irq_context. The way it's done here so far is ok, having it be a 'child' header internally to irq.h. What may be needed in the end would be something like an `irq_internal.h` to have `irq_context` (instead of the forward declaration) and perhaps things shared by exceptions then an exception.h as well.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/irq.h
+++ b/kernel/arch/dreamcast/include/arch/irq.h
@@ -27,6 +27,9 @@ __BEGIN_DECLS
 
 #include <arch/types.h>
 
+/* Included for legacy compatibility with these two APIs being one. */
+#include <arch/trap.h>
+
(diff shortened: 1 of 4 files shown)


gyrovorbis (approved):
I approve of this TRAPA cleanery. lol.

PR #812, opened 2024-10-11
matrix.h: Constify and rename some parameters

pcercuei:
Constify parameters where it makes sense, and rename an "out" parameter that's actually an input.


gyrovorbis (approved):
Thank you so much. It has been a ROYAL freaking PITA in C++ to have to `const_cast<>()` constant matrices that we're operating on to be non-const just so we can pass them to this const-incorrect API. 

Very warranted.

PR #813, opened 2024-10-11
perf_monitor: Improve perf_monitor_print()

pcercuei:
- Don't print anything if no perf monitor is present;
- Print the monitors in reverse order, as the first ones in the .monitors section correspond to the last ones in a C file;
- Print the line number after the function name, so that it is easy to differenciate perf monitors used within a single function.


gyrovorbis (approved):
Cool. That is nice.

PR #815, opened 2024-10-12
Refactor and cleanup of IRQ system (Part 2)

QuzarDC:
Second split off of #618 . This one contains all the documentation, rearranging, and cleanup built into it. Tried to keep it as identical to what is in #618 as possible in order to make the diffing in that one as minor as possible.

* Retype interrupt codes to an enum.
* Use stdint and bool types where possible.
* Various minor code cleanups.
* Restructure layout and doxygen in header.
* Clean up register accesses in irq.c to match other drivers.
* Retype global handler into proper irq_cb type.

Changes from #618:
-Left out blocks of comments related to the new functionality in 618.
-Corrected retval documentation for `irq_inside_int`
-Removed some accidental whitespace and a random letter 's'.
-Corrected coherence between params and doc for `irq_set_handler` and `irq_get_handler`.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/irq.h
+++ b/kernel/arch/dreamcast/include/arch/irq.h
@@ -2,10 +2,12 @@
 
    arch/dreamcast/include/irq.h
    Copyright (C) 2000-2001 Megan Potter
+   Copyright (C) 2024 Paul Cercueil
+   Copyright (C) 2024 Falco Girgis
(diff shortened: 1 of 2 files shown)


gyrovorbis (approved):
This looks functionally identical, and I've tested several of the most demanding testbeds I have... everything is working great, so... approving!

PR #816, opened 2024-10-12
Added functionality to biosfont

andressbarajas:
- Added two new functions (bfont_draw_str_vram_vfmt & bfont_draw_str_vram_fmt) that draw to the frame buffer start at a x,y location. 
 - Added support to handle \n and \t. 
 - Fix documentation spacing
 - Update two examples to demo the new API

### DIFF
--- a/examples/dreamcast/video/screenshot/screenshot.c
+++ b/examples/dreamcast/video/screenshot/screenshot.c
@@ -30,7 +30,7 @@
 
 #include <kos/thread.h>
 
-#define SHOW_BLACK_BG 1
+#define SHOW_BLACK_BG  true
 
 /* Keeps track of the amount of screenshots you have taken */
 static int counter = 0;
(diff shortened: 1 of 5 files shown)


QuzarDC:
One high level note, and otherwise am generally fine with it all:
I'd be 100% on-board with just finally removing the deprecated `bfont_set_32bit_mode` if you'd be interested in that. Then the logic you used for choosing bit depth based on video mode can be propagated down into the functions that relied on that. It's been deprecated for 6 years and having auto-detection should be just fine.

I think by doing that your versions could then be simplified down into thinner wrappers around the default behavior instead. My thinking is that it was kind of backwards for us to have had that switch rather than auto-detection in the first place. iirc the use case was *not* someone wanting to draw to a 32-bit texture while having the framebuffer in 16-bit mode.

andressbarajas:
@QuzarDC

QuzarDC (approved):

PR #817, opened 2024-10-12
[RFC] pvr: Add new function pvr_send_to_ta()

pcercuei:
This function can be used to upload a 32-byte payload to the Tile Accelerator.

The difference with the Direct Rendering approach is that the SQs are not used, and therefore can be used for anything else.

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_send_to_ta.s
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_send_to_ta.s
@@ -1,0 +1,43 @@
+! KallistiOS ##version##
+!
+!   arch/dreamcast/hardware/pvr/pvr_send_to_ta.s
+!   Copyright (C) 2024 Paul Cercueil
+!
+! Fast function to upload geometry to the Tile Accelerator without SQs
+!
+
+.globl _pvr_send_to_ta
+
+_pvr_send_to_ta:
+	fschg
+	mov.l	_irq_and,r1
+	mov	#0x78,r2
+	mov.l	_ta_addr,r3
+	stc	sr,r0
+	and	r0,r1
+	fmov	@r4+,dr0
+	shll	r2
+	fmov	@r4+,dr2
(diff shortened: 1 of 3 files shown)


kernel/arch/dreamcast/hardware/pvr/pvr_send_to_ta.s
```diff
@@ -0,0 +1,43 @@
+! KallistiOS ##version##
+!
+!   arch/dreamcast/hardware/pvr/pvr_send_to_ta.s
+!   Copyright (C) 2024 Paul Cercueil
+!
+! Fast function to upload geometry to the Tile Accelerator without SQs
+!
+
+.globl _pvr_send_to_ta
+
```
andressbarajas, line 10:
```suggestion

!
! void pvr_send_to_ta(void *data);
!
! r4: data (should be 8-byte aligned address)
!
```

andressbarajas: Not sure if this adds anything or not but I think it could be helpful to some who are just starting out looking at SH4 assembly. Feel free to disregard.

andressbarajas:
Can this be utilized anywhere in KOS?  In what scenarios do you use it the most?

andressbarajas (approved):

pcercuei:
@andressbarajas I actually do not use it, anywhere. I just thought it was an interesting piece of code, that we *could* have in KOS, *if* we have the use for it somewhere.

gyrovorbis:
This is actually quite interesting. There are a bunch of fairly advanced use-cases for this for like say you're streaming textures using the SQs, you could still be submitting geometry to the TA for rendering... 

I believe Oleg Endo was talking about the need for something like this for some of his advanced stuff. Do think it has the potential to be quite useful on the advanced side.

gyrovorbis (approved):
I approve. This is the advanced kind of stuff I want the PVR driver to be able to do. :)

PR #819, opened 2024-10-14
Loadable library recovery and example

DC-SWAT:
This is an adaptation of what was done in [DreamShell](https://github.com/DC-SWAT/DreamShell).
At the moment in KOS they have long been abandoned and essentially do not work properly.

- Updated Makefile rules and ldscript.
- Fixed library reference counting.
- Added `library_lookup_fn()`.
- Changed logic for opening/closing library to be more usable. Doing it once on loading and unloading to init and shutdown some things inside loadable library.
- Added `export_lookup_path()` and `export_lookup_addr()`.
- Added an example with dependency and dependent, as well as their use with the main program.
- Some cleanup, update gitignore, aligning for elf image.

### DIFF
--- a/examples/dreamcast/library/exports.txt
+++ b/examples/dreamcast/library/exports.txt
@@ -1,0 +1,7 @@
+include kos/exports.h
+include string.h
+
+# Additional newlib functions to export
+# Some part of it already exported by kernel/exports.txt
+strcmp
+strncmp
--- a/kernel/exports/library.c
+++ b/kernel/exports/library.c
@@ -2,6 +2,7 @@
 
    kernel/library.c
    Copyright (C) 2003 Megan Potter
+   Copyright (C) 2024 Ruslan Rostovtsev
 */
 
 #include <assert.h>
@@ -127,7 +128,7 @@ klibrary_t * library_create(int flags) {
             /* Populate the context */
             np->libid = libid;
             np->flags = flags;
-            np->refcnt = 0;
+            np->refcnt = 1;
 
             /* Insert it into the library list */
             LIST_INSERT_HEAD(&library_list, np, list);
@@ -137,8 +138,8 @@ klibrary_t * library_create(int flags) {
     return np;
 }
 
-int library_destroy(klibrary_t * lib) {
-    int     oldirq = 0;
+int library_destroy(klibrary_t *lib) {
+    int oldirq = 0;
 
     oldirq = irq_disable();
 
@@ -160,7 +161,7 @@ int library_destroy(klibrary_t * lib) {
 /*****************************************************************************/
 /* Library attribute functions -- all read-only */
 
-libid_t library_get_libid(klibrary_t * lib) {
+libid_t library_get_libid(klibrary_t *lib) {
     if(lib == NULL) {
         errno = EINVAL;
         return -1;
@@ -169,7 +170,7 @@ libid_t library_get_libid(klibrary_t * lib) {
         return lib->libid;
 }
 
-int library_get_refcnt(klibrary_t * lib) {
+int library_get_refcnt(klibrary_t *lib) {
     if(lib == NULL) {
         errno = EINVAL;
         return -1;
@@ -178,7 +179,7 @@ int library_get_refcnt(klibrary_t * lib) {
         return lib->refcnt;
 }
 
-const char * library_get_name(klibrary_t * lib) {
+const char *library_get_name(klibrary_t *lib) {
     if(lib == NULL || lib->lib_get_name == NULL) {
         errno = EINVAL;
         return NULL;
@@ -187,7 +188,7 @@ const char * library_get_name(klibrary_t * lib) {
         return lib->lib_get_name();
 }
 
-uint32 library_get_version(klibrary_t * lib) {
+uint32_t library_get_version(klibrary_t *lib) {
     if(lib == NULL || lib->lib_get_version == NULL) {
         errno = EINVAL;
         return 0;
@@ -197,8 +198,8 @@ uint32 library_get_version(klibrary_t * lib) {
 }
 
 /*****************************************************************************/
-klibrary_t * library_lookup(const char * name) {
-    klibrary_t  * lib;
+klibrary_t *library_lookup(const char *name) {
+    klibrary_t *lib;
 
     irq_disable_scoped();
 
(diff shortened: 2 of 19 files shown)


andressbarajas:
@DC-SWAT Everything looks good from the parts that I understand.  The changes to loadable/shlelf_dc.xr look kinda scary.  Do those changes to the shlelf_dc.xr file affect all binaries?  If so, did you test them against GTA3?  Any other testing I can do to confirm that these changes wont have bad consequences?

DC-SWAT:
> @DC-SWAT Everything looks good from the parts that I understand.  The changes to loadable/shlelf_dc.xr look kinda scary.  Do those changes to the shlelf_dc.xr file affect all binaries?  If so, did you test them against GTA3?  Any other testing I can do to confirm that these changes wont have bad consequences?

`loadable/shlelf_dc.xr` are used only for linking loadable library. It's not used for regular binaries.

andressbarajas (approved):
You seem to be the only person that flexes this functionality in DreamShell. I trust you.

gyrovorbis:
SO GLAD TO SEE THIS HERE!!! I can volunteer to start reviewing this. Got experience looking at the library loading.

examples/dreamcast/library/exports.txt
```diff
@@ -0,0 +1,7 @@
+include kos/exports.h
+include string.h
+
+# Additional newlib functions to export
+# Some part of it already exported by kernel/exports.txt
+strcmp
```
QuzarDC, line 6: Why would it be that we add exports for an individual library? Shouldn't the exports reflect everything that KOS is guaranteed to offer? It seems to me (though I may misunderstand) that if KOS is not guaranteed to export these, then we should be linking libc to the library to guarantee they are present. If KOS *is* guaranteed to export these then they should be listed in the kernel/exports.

loadable/Makefile.prefab
```diff
@@ -1,30 +1,33 @@
...
+
 # First one checks for missing symbols (and fully links with an offset
 # of zero so we can do tracebacks later), second one makes the real file.
 $(TARGET): $(OBJS)
-	$(KOS_CC) -g -ml -m4-single-only -O2 -g -Wl,-Ttext=0x00000000 -e _start -nostartfiles -nostdlib -o dbg-$(TARGET) $(OBJS) $(KOS_LIB_PATHS) -Wl,--start-group $(LIBS) -lkallisti_exports -lgcc -Wl,--end-group
-	$(KOS_CC) -g -ml -m4-single-only -O2 -Wl,-d -Wl,-r -Wl,-S -Wl,-x -nostartfiles -nostdlib -o $(TARGET) -Wl,-T $(KOS_BASE)/loadable/shlelf_dc.xr $(OBJS) $(KOS_LIB_PATHS) -Wl,--start-group $(LIBS) -lgcc -Wl,--end-group
-	-chmod +x $(TARGET)
+	$(KOS_CC) -g -ml -m4-single-only -O2 -g -Wl,-Ttext=0x00000000 -e _start -nostartfiles -nodefaultlibs -o dbg-$(TARGET) $(OBJS) $(KOS_LIB_PATHS) -Wl,--start-group $(LIBS) $(DBG_LIBS) -lgcc -Wl,--end-group
+	$(KOS_CC) -g -ml -m4-single-only -O2 -Wl,-d -Wl,-r -Wl,-S -Wl,-x -nostartfiles -nodefaultlibs -o $(TARGET) -Wl,-T $(KOS_BASE)/loadable/shlelf_dc.xr $(OBJS) $(KOS_LIB_PATHS) -Wl,--start-group $(LIBS) -lgcc -Wl,--end-group
+	$(KOS_SIZE) $(TARGET)
```
QuzarDC, line 27: Where is `KOS_SIZE` defined? The name seems familiar but I can't find the reference to it with plaintext searches.

QuzarDC (commented):
Mostly just have some questions to help me understand a few of the changes. As others, I have no experience with the linker script and will just trust it.

The one thing that I think requires change is something around the `lib_close`, I think the behavior but possibly just the documentation. I'm relatively open to the API behavior changing as there's no indication that anyone but you is using this.

Otherwise looks great. Thanks for bringing this back out of its rotted state. Hopefully it can pave the way forward for some more expansive software without the constraint of a singular binary.

loadable/Makefile.prefab
DC-SWAT: I've added it in current PR:
https://github.com/KallistiOS/KallistiOS/blob/978c67b4af424bc1329426e6fea064fab6447a3f/environ_base.sh#L50

examples/dreamcast/library/exports.txt
DC-SWAT: KOS is guaranteed to export only some of the libc functions that it uses. You don't need to export everything at once, since they will all be linked into the binary, regardless of whether it is used or not. To minimize memory consumption, you need to select only the necessary functions. And this is individual for each project.
Although if someone is not lazy and writes an export of all libc functions, this can be included as an option. Like exporting the entire libc library. But this is only good for testing, for real projects I think it would be wasteful.

loadable/Makefile.prefab
```diff
@@ -1,30 +1,33 @@
...
+else
+all: rm-elf $(TARGET)
+endif
+
 # First one checks for missing symbols (and fully links with an offset
 # of zero so we can do tracebacks later), second one makes the real file.
 $(TARGET): $(OBJS)
-	$(KOS_CC) -g -ml -m4-single-only -O2 -g -Wl,-Ttext=0x00000000 -e _start -nostartfiles -nostdlib -o dbg-$(TARGET) $(OBJS) $(KOS_LIB_PATHS) -Wl,--start-group $(LIBS) -lkallisti_exports -lgcc -Wl,--end-group
-	$(KOS_CC) -g -ml -m4-single-only -O2 -Wl,-d -Wl,-r -Wl,-S -Wl,-x -nostartfiles -nostdlib -o $(TARGET) -Wl,-T $(KOS_BASE)/loadable/shlelf_dc.xr $(OBJS) $(KOS_LIB_PATHS) -Wl,--start-group $(LIBS) -lgcc -Wl,--end-group
-	-chmod +x $(TARGET)
```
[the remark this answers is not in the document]
DC-SWAT: Initially these libraries were planned as kernel modules like .ko in Linux. Perhaps some changes were planned in the file system to support attributes and that's why it was written here. It's hard to say honestly.
Anyway, I reworked them to ".so" shared libraries instead of ".ko" kernel modules while retaining some of the functionality of the kernel modules, due to the opening and closing functions.

kernel/exports/library.c
```diff
@@ -281,20 +290,24 @@ klibrary_t * library_open(const char * name, const char * fn) {
...
-int library_close(klibrary_t * lib) {
+int library_close(klibrary_t *lib) {
     // Make sure it's valid.
     if(lib == NULL || lib->lib_close == NULL) {
         errno = EINVAL;
         return -1;
     }
 
+    // Check for reference
+    if(--lib->refcnt > 0) {
```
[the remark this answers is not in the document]
DC-SWAT: That's the thing, most often you only have to do something when loading and unloading a library, not every time it's opened and closed. I came to this conclusion through experience. Perhaps it would make sense if these were kernel modules, but these are no longer kernel modules, read below.
Same goes for `refcnt`, there is no reason to do it inside the library, none at all. It only brings confusion to the overall system. I'll fix the docs.

kernel/exports/library.c
```diff
@@ -213,28 +214,36 @@ klibrary_t * library_lookup(const char * name) {
...
+    }
 
-        if(lib) {
-            // Make sure lib_open is valid. Note that since we
-            // _have_ found something, any inconsistencies here
-            // are errors and not a signal to load another copy.
-            if(!lib->lib_open) {
-                errno = EINVAL;
-                return NULL;
-            }
```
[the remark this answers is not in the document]
DC-SWAT: This is an unnecessary condition in general. Because if `lib_open` function is not there, then the elf file itself will not load. I don’t understand how it can disappear if you got loaded library.

loadable/Makefile.prefab
QuzarDC: Wow, thank you. I think because I had clicked it 'viewed' github wasn't showing it when I did a search.

examples/dreamcast/library/exports.txt
QuzarDC: Ah ok, that makes sense. Would that same principle also follow for the rest of the KOS API then? Rather than always exporting as much as we can of the API?

loadable/Makefile.prefab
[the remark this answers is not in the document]
QuzarDC: Ah, that makes some sense, especially since there was an i386 arch for KOS. Thanks.

kernel/exports/library.c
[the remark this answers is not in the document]
QuzarDC: Ok, excellent. It also makes sense that *if* some sort of per-instance work was needed it could still be done without that (and this way simplifies things).

QuzarDC (approved):

PR #821, opened 2024-10-14
Set ramdisk's `stat` to return 0 on successful completion.

QuzarDC:
Was returning -1, which indicates an error (ref https://pubs.opengroup.org/onlinepubs/9799919799/ ). Not sure if this was impacting upstream use.


andressbarajas (approved):
Dont know why I didnt catch this in my PR. Good find!

gyrovorbis (approved):
Good find!

PR #823, opened 2024-10-14
The sys/queue `_FOREACH` functions are not safe to free their elements.

QuzarDC:
Swap usage to the `_FOREACH_SAFE` macro which is the safe version. Without this, each time `qpkt` is freed, the next loop iteration is based on a use-after-free.


andressbarajas (approved):
Looks safe to me!

gyrovorbis (approved):
The master of fixing this kind of un-safety... lol.

PR #827, opened 2024-10-16
Extended snd_sfx functionality

voxel-public:
The KOS AICA sound driver has functionality beyond what is exposed by the current simple snd_sfx functions. My use case is playing short sound effects that loop (like a short engine noise sample), where the pitch and frequency can be changed dynamically at runtime. 

My current options for accessing this functionality are:

- External library from kos-ports (i.e OpenAL) but the partial implementation does not include that functionality.
- Use the audio streaming api. This doesn't provide the convenience functions for loading and playing short sounds like the snd_sfx api does, and as far as I can tell is aimed at much larger sound buffers. It's much more complex and powerful than the current snd_sfx api, to the point where I wasn't able to understand it without examples, of which I couldn't find any that don't depend on external libraries to provide most of the interface.

So finding neither of those options suitable I've added
snd_sfx_play_ex() and snd_sfx_play_chn_ex(), an extended version of snd_sfx_play() and snd_sfx_play_chn() respectively, that expose access to some parameters that the original functions set to default values (frequency, looping, loop start and end points), which enables me to start a looping sound at a given frequency. The original functions call the extended versions with the default parameters they used previously.

I also need to be able to dynamically change the volume, pan, and frequency of playing sound effects, and so have added snd_sfx_volume() snd_sfx_pan(), snd_sfx_freq().

These are fairly simple additions that allow me to forgo the need to use a kos-ports library (if a sufficiently complete one exists) that I think might be useful to others using the simple sfx api but find it slightly lacking.

### DIFF
--- a/kernel/arch/dreamcast/exports-naomi.txt
+++ b/kernel/arch/dreamcast/exports-naomi.txt
@@ -122,6 +122,7 @@ snd_sfx_play_chn
 snd_sfx_play_chn_ex
 snd_sfx_volume
 snd_sfx_pan
+snd_sfx_freq
 snd_sfx_stop
 snd_sfx_chn_alloc
 snd_sfx_chn_free
--- a/kernel/arch/dreamcast/exports-naomi.txt
+++ b/kernel/arch/dreamcast/exports-naomi.txt
@@ -120,6 +120,7 @@ snd_sfx_play_ex
 snd_sfx_stop_all
 snd_sfx_play_chn
 snd_sfx_play_chn_ex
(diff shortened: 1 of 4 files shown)


andressbarajas:
Some overlap with this PR:  https://github.com/KallistiOS/KallistiOS/pull/738

voxel-public:
> Some overlap with this PR: #738

I was too slow! You're right, I think that #738 already includes everything here, making this unneeded

andressbarajas:
@voxel-public Feel free to close this PR.

PR #828, opened 2024-10-16
audio: Fix makefile race condition and arm clean

skmp:
- arm/stream.drv depends on subdirs to generate it, so it can't be a parallel dependency.
- arm/Makefile forgot to delete stream.drv on clean

Fixes make -j 24 for me


andressbarajas (approved):
Looks good to me!

QuzarDC (approved):

PR #832, opened 2024-10-23
Consistency fix in environ_base.sh

cypressru:
Added periods to all commented sentences for consistency.
Also aligns with how comments are written in environ_dreamcast.sh


gyrovorbis (approved):
Hah, I never noticed that inconsistency, but that does bother me now. Thanks. lol.

PR #833, opened 2024-10-25
Fix G2 DMA race conditions.

DC-SWAT:
- Fixed channel variables sets before check DMA is in progress.
- The start/status register can be zero before required IRQ are handled, added variable to fix it.


gyrovorbis (approved):
Wow. Good find. 

This isn't my area, but this is basic concurrency/multithreading issues, which I see have been fixed (along with a couple other corner cases), plus I tested locally. Looks good!

PR #836, opened 2024-10-27
Fix Strided Textures

andressbarajas:
Custom strided textures never worked it seems like.  This PR should fix that.

- Corrected custom strided textures by adjusting PVR_TXRFMT_STRIDE handling. 
- Renamed PVR_TEXTURE_MODULO to PVR_TEXTURE_STRIDE for clarity. 
- Added an example to demonstrate custom strided texture usage.

### DIFF
--- a/examples/dreamcast/pvr/strided_texture/strided_texture.c
+++ b/examples/dreamcast/pvr/strided_texture/strided_texture.c
@@ -1,0 +1,228 @@
+/* KallistiOS ##version##
+
+   strided_texture.c
+   Copyright (C) 2024 Andress Barajas
+*/
+
+/*
+    This example demonstrates rendering a black-and-white chessboard pattern 
+    using a 640x480 texture with 16bpp color depth. In this example, the texture 
+    width and stride are both set to 640. However, because 640 is not a power 
+    of two, we need to use the `PVR_TXRFMT_X32_STRIDE` flag. This flag informs the 
+    PVR that the texture's width (or "stride") is a multiple of 32 rather than 
+    a power of two.
+
+    Steps to configure and render a texture with a 32-pixel multiple 
+    width:
+
+    1. Configure the Polygon Header for Textures:
+    
+       - Use `pvr_poly_cxt_txr()` to set up the polygon context for the texture.
+       - Specify `PVR_TXRFMT_NONTWIDDLED | PVR_TXRFMT_X32_STRIDE` in the format
+         flags.
+       - Provide dimensions in powers of two, which should be larger than the
+         actual texture. For example, if your texture is 640x480, set width and
+         height to `1024` and `512`, respectively.
+
+       Example:
+           ```c
+           pvr_poly_cxt_txr(&cxt, PVR_LIST_OP_POLY,
+                            PVR_TXRFMT_RGB565 | PVR_TXRFMT_NONTWIDDLED |
+                            PVR_TXRFMT_X32_STRIDE, 1024, 512,
+                            texture_pointer, PVR_FILTER_NONE);
+           ```
+
+  2. Set the Global Texture Stride Register:
+
+       - Use `PVR_SET(PVR_TXR_STRIDE_MULT, stride_multi);` to define a custom
+         texture stride width in increments of 32 pixels.
+       - Calculate `stride_multi` by dividing the texture width by 32.
+       - The `PVR_TXR_STRIDE_MULT` register defines how the hardware interprets 
+         each row's width in VRAM. For a 640-pixel wide texture, `stride_multi` 
+         would be:
+        
+           ```c
+           PVR_SET(PVR_TXR_STRIDE_MULT, 640 / 32);
+           ```
+       Important Notes:
+       - Texture widths that are multiples of 32 (but not powers of two) require 
+         the `PVR_TXRFMT_X32_STRIDE` flag.
+       - Palette-based textures are incompatible with the `PVR_TXRFMT_X32_STRIDE` 
+         flag.
+       - `PVR_TXR_STRIDE_MULT` is a global register. All textures using the 
+         `PVR_TXRFMT_X32_STRIDE` flag in the same frame must share the same 
+         stride. Changing `PVR_TXR_STRIDE_MULT` affects all such textures 
+         rendered afterward.
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <malloc.h>
+#include <arch/arch.h>
+
+#include <dc/pvr.h>
+#include <dc/maple.h>
+#include <dc/maple/controller.h>
+
+/* The width of the texture in pixels (must be a multiple of 32) */
+#define TEXTURE_WIDTH    640
+
+/* The height of the texture in pixels */
+#define TEXTURE_HEIGHT   480
+
+/* 
+   Macro to calculate the next power of two for a given integer `x`.
+   
+   Example:
+     NEXT_POWER_OF_TWO(640) -> 1024
+*/
+#define NEXT_POWER_OF_TWO(x)    (1 << (32 - __builtin_clz((x) - 1)))
+
+/* 
+   The power-of-two width to be passed to texture setup functions. 
+   Based on TEXTURE_WIDTH, this should be equal to or greater than 
+   TEXTURE_WIDTH, and will match the hardware requirement for 
+   power-of-two texture dimensions.
+*/
+#define TEXTURE_PADDED_WIDTH    NEXT_POWER_OF_TWO(TEXTURE_WIDTH)
+
+/* 
+   The power-of-two height to be passed to texture setup functions. 
+   Based on TEXTURE_HEIGHT, this should be equal to or greater than 
+   TEXTURE_HEIGHT, following the power-of-two requirement.
+*/
+#define TEXTURE_PADDED_HEIGHT   NEXT_POWER_OF_TWO(TEXTURE_HEIGHT)
+
+/* RGB565 colors for chessboard pattern */
+#define COLOR_BLACK   0x0000
+#define COLOR_WHITE   0xFFFF
+
+static pvr_poly_hdr_t hdr;
+static pvr_vertex_t verts[4];
+
+static pvr_ptr_t board_texture;
+
+static void draw_frame(void) {
+    pvr_wait_ready();
+    pvr_scene_begin();
+
+    pvr_list_begin(PVR_LIST_OP_POLY);
+
+    pvr_prim(&hdr, sizeof(hdr));
+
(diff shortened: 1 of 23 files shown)


pcercuei:
"A strided texture has a padded width"
This is technically true, a "strided texture" would be a texture where the stride (amount of data between two lines, counted in pixels) differs from the line size in pixels, aka. there is padding after each line.

However in the context of the PVR, strided textures (as described above) do not need any specific configuration as long as the stride is a power of two, and do not need the PVR_TXRFMT_STRIDE flag - you adjust the U/V values and that's it.

What the flag does is allow you to use a stride that's a multiple of 32, instead of a power of two.

In your provided example, the chessboard texture is actually *not* a strided texture, because its stride matches its width. However its stride is a multiple of 32, which requires the PVR_TXRFMT_STRIDE to work properly.

andressbarajas:
Thanks for the review! Im a little confused on this part:

> In your provided example, the chessboard texture is actually _not_ a strided texture, because its stride matches its width. However its stride is a multiple of 32, which requires the PVR_TXRFMT_STRIDE to work properly.

Is it because the squares in the chess board are 32x32 pixels?  Can you give me an example of what would be a strided texture?

pcercuei:
> > In your provided example, the chessboard texture is actually _not_ a strided texture, because its stride matches its width. However its stride is a multiple of 32, which requires the PVR_TXRFMT_STRIDE to work properly.
> 
> Is it because the squares in the chess board are 32x32 pixels? Can you give me an example of what would be a strided texture?

No, that's unrelated.

I'll start by clearly defining what "stride" means: it is the length between two rows of an image.
A real "strided texture" would be for instance a 640x480 texture where each line is 1024 pixels wide (or any other number > 640 really). In that case the line width is 640 pixels, but the stride is 1024 pixels, with 1024-640 pixels of padding between two rows.

KallistiOS already supports these kinds of textures, as long as the stride is a multiple of two, without needing the PVR_TXRFMT_STRIDE flag. In my example above, you would just set the texture width to 1024 pixels, and use a maximum U=640/1024.

The PVR_TXRFMT_STRIDE flag is needed when the stride is not a power of two, but a multiple of 32 (due to how the PVR_TEXTURE_STRIDE register works).
If I change my example to be a non-strided 640x480 texture (which means the line size == the stride, there is no gap between two lines), the stride is not a power of two, so we cannot use it directly; in that case, we can use the PVR_TXRFMT_STRIDE to tell the PVR that the stride is a multiple of 32 instead, and set the PVR_TEXTURE_STRIDE register to 640/32.

So in that example, we need PVR_TXRFMT_STRIDE for a non-strided texture, and we can use strided textures without this flag.

If it sounds confusing, it's because the docs (official or non-official) use the term "strided texture" completely incorrectly.
If you consider the PVR_TXRFMT_STRIDE flag as a "stride is multiple of 32" flag, everything makes much more sense.

andressbarajas:
@pcercuei Okay so the issue is not with the code but with the explanation/how things are labeled. You're saying KOS supports stride texture already. The only thing it didn't support was when the stride textures are a multiple of 32 and not a power of 2.

pcercuei:
Yes that's correct, it's a documentation issue. I also think that the PVR_TXRFMT_STRIDE flag should be renamed, because it really doesn't help with the confusion.

andressbarajas:
@pcercuei `PVR_TXRFMT_CUSTOM_STRIDE` sound better?

andressbarajas:
@pcercuei Done. Let me know if the updated documentation and register name work for you.

pcercuei:
@andressbarajas It's definitely better. Still I'd argue that "custom stride" gives you the impression that you can provide any value that you want, what about this:
- `PVR_TXRFMT_STRIDE` -> `PVR_TXRFMT_X32_STRIDE`
- `PVR_TXRFMT_NOSTRIDE` -> `PVR_TXRFMT_POW2_STRIDE`

pcercuei:
@andressbarajas just a nitpick, but the commit message is outdated now. The rest looks good.

andressbarajas:
> @andressbarajas just a nitpick, but the commit message is outdated now. The rest looks good.

Thanks. I'll take care of it tonight.

examples/dreamcast/pvr/strided_texture/strided_texture.c
```diff
@@ -0,0 +1,228 @@
...
+
+    /* Set texture context format for nontwiddled, strided texture */
+    pvr_poly_cxt_txr(&cxt, PVR_LIST_OP_POLY,
+                     PVR_TXRFMT_RGB565 | PVR_TXRFMT_NONTWIDDLED 
+                     | PVR_TXRFMT_X32_STRIDE, TEXTURE_PADDED_WIDTH, 
+                     TEXTURE_PADDED_HEIGHT, board_texture, PVR_FILTER_NONE);
+    pvr_poly_compile(&hdr, &cxt);
+
+    /* Set the STRIDE register to texture width / 32 */
+    PVR_SET(PVR_TXR_STRIDE_MULT, TEXTURE_WIDTH / 32);
```
gyrovorbis, line 157: Am I the only one who feels like stuff like this should be dedicated API functionality rather than us just describing how to go in and directly poke registers manually? I feel like this is too low-level to be user-friendly, and the documentation for these registers will be crappier than a dedicated API, plus people will look at the API/docs and not even see this kind of functionality, plus we've talked about deprecating direct access to registers like this...

...plus, this gives us a way to validate the user's input, PLUS it looks like only the first 4 bits are actually used for stride, while the rest of the bits are reserved or used for other things in that register, so you don't really want to just have the user do a raw set anyway.

I would rather see something like:
```
bool pvr_txr_set_stride(unsigned int texture_width);
unsigned int pvr_txr_get_stride(void);
```
Which can then:
    - do the divide by 32 for the user
    - return an error/assert if the dimensions are invalid
    - only sets the first 4 bits of the register
    - lets you retrieve the current value
    - have API documentation for the group that includes a lot of this great commentary in the example which the user would not get just looking at the header file/register map otherwise.

andressbarajas: Sounds like a great idea! I'll implement it.

andressbarajas: @gyrovorbis Done

pcercuei:
Why do you have to handle `pvr_txr_set_stride(0)` at all?

andressbarajas:
> Why do you have to handle `pvr_txr_set_stride(0)` at all?

Maybe here: https://github.com/KallistiOS/KallistiOS/blob/e1b97e4314fa1a2e933fc3846c3971a8439a0f0b/kernel/arch/dreamcast/hardware/pvr/pvr_init_shutdown.c#L189

pcercuei:
No, you'd still use `PVR_SET()` there, because you want to zero all fields, not just the bits that encode the stride.

andressbarajas:
> No, you'd still use `PVR_SET()` there, because you want to zero all fields, not just the bits that encode the stride.

You're right. Fixed.

pcercuei:
Looks good to me. Consider the PR approved on my side.

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -2270,6 +2271,43 @@ void pvr_poly_cxt_txr_mod(pvr_poly_cxt_t *dst, pvr_list_t list,
...
+    \ingroup pvr_txr_mgmt 
+
+    This function configures the global texture stride register 
+    `PVR_TXR_STRIDE_MULT`, which defines the row width in VRAM for 
+    non-power-of-two textures. The setting applies to all textures 
+    rendered with the `PVR_TXRFMT_X32_STRIDE` flag in the same frame. 
+    Since `PVR_TXR_STRIDE_MULT` is a global register, all textures 
+    using this flag must share the same stride width in each frame.
+
+    **Important Notes:**
```
gyrovorbis, line 2284:
```suggestion
    \warning
```

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -2270,6 +2271,43 @@ void pvr_poly_cxt_txr_mod(pvr_poly_cxt_t *dst, pvr_list_t list,
     Helper functions for handling texture tasks of various kinds.
 */
 
+/** \brief   Set the global stride width for non-power-of-two textures in PVR RAM.
+    \ingroup pvr_txr_mgmt 
+
+    This function configures the global texture stride register
```
gyrovorbis, line 2277: You need to also mention that this stride is not just for any sizes but is only supported for multiples of 32 here.

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -2270,6 +2271,43 @@ void pvr_poly_cxt_txr_mod(pvr_poly_cxt_t *dst, pvr_list_t list,
...
+    using this flag must share the same stride width in each frame.
+
+    **Important Notes:**
+    - Textures that are palette-based cannot use the `PVR_TXRFMT_X32_STRIDE` 
+      flag so the stride set here will not apply to them.
+
+    \param  texture_width   The width of the texture in pixels. Must be a 
+                            multiple of 32 and up to 992 pixels.
+    \retval true            On success.
+    \retval false           On failure.
```
gyrovorbis, line 2291:
```suggestion
    \retval false           On failure.
    
    \sa pvr_txr_get_stride()
```

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -2270,6 +2271,43 @@ void pvr_poly_cxt_txr_mod(pvr_poly_cxt_t *dst, pvr_list_t list,
...
+    This function reads the `PVR_TXR_STRIDE_MULT` register and calculates the 
+    texture stride width in pixels. The value returned is the width in pixels 
+    that has been configured for all textures using the `PVR_TXRFMT_X32_STRIDE` 
+    flag in the same frame. 
+
+    The stride width is computed by taking the current multiplier in 
+    `PVR_TXR_STRIDE_MULT` (which stores the width divided by 32), and 
+    multiplying it back by 32 to return the full width in pixels.
+
+    \return                 The current texture stride width in pixels.
```
gyrovorbis, line 2307:
```suggestion
    \return                 The current texture stride width in pixels.

    \sa pvr_txr_set_stride()
```

gyrovorbis (commented):
Just a couple teensy Doxygen nitpicks, then can merge! Looking good.

andressbarajas:
@gyrovorbis Done

pcercuei:
@QuzarDC Can we still merge this? It's done and looks ready, and some people (including me) are waiting for it.

PR #837, opened 2024-10-27
Speedup and fix AICA cmd iface. Protect SQs on G2 bus.

DC-SWAT:
- Speedup AICA command interface by introducing non-blocking functions for working with the G2 bus (lock it once), getting rid of the semaphore and minor loop/condition optimizations.
- Fixed AICA command interface by alignment for buffer, packet size check and thread-safe for start/stop.
- Speedup AICA channel resetting by using non-blocking functions for working with the G2 bus.
- Added functions to retrieve the channels states `snd_get_pos()` and `snd_is_playing()`.
- Protect the SQs on G2 bus because they cannot be suspended from another thread at PIO access to G2 bus.

Tested on GTA3. Sound controls much stable now.

### DIFF
--- a/kernel/arch/dreamcast/hardware/spu.c
+++ b/kernel/arch/dreamcast/hardware/spu.c
@@ -87,12 +87,12 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     /* Lock the SQs before disabling the interrupts. */
     sq_lock(NULL);
 
-    /* Make sure the FIFOs are empty */
-    g2_fifo_wait();
-
     /* Lock G2 bus because we can't suspend SQs from
      * another thread with PIO access to G2 bus. */
     ctx = g2_lock();
+
+    /* Make sure the FIFOs are empty */
+    g2_fifo_wait();
 
     sq_cpy((void *)dst, src, aligned_len);
 
@@ -229,12 +229,12 @@ void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
     /* Lock the SQs before disabling the interrupts. */
     sq_lock(NULL);
 
-    /* Make sure the FIFOs are empty */
-    g2_fifo_wait();
-
     /* Lock G2 bus because we can't suspend SQs from
      * another thread with PIO access to G2 bus. */
     ctx = g2_lock();
+
+    /* Make sure the FIFOs are empty */
+    g2_fifo_wait();
 
     sq_set32((void *)dst, what, aligned_len);
 
@@ -257,8 +257,9 @@ void spu_reset_chans(void) {
     int i;
     g2_ctx_t ctx;
 
-    g2_fifo_wait();
     ctx = g2_lock();
+    g2_fifo_wait();
+
     g2_write_32_raw(SNDREGADDR(0x2800), 0);
 
     for(i = 0; i < 64; i++) {
--- a/kernel/arch/dreamcast/hardware/spu.c
+++ b/kernel/arch/dreamcast/hardware/spu.c
@@ -84,7 +84,7 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     /* Add in the SPU RAM base (cached area) */
     dst |= SPU_RAM_BASE;
 
-    /* First lock SQs because there is a mutex there. */
+    /* Lock the SQs before disabling the interrupts. */
     sq_lock(NULL);
 
     /* Make sure the FIFOs are empty */
@@ -226,7 +226,7 @@ void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
     /* Add in the SPU RAM base (cached area) */
     dst |= SPU_RAM_BASE;
 
-    /* First lock SQs because there is a mutex there. */
+    /* Lock the SQs before disabling the interrupts. */
     sq_lock(NULL);
 
     /* Make sure the FIFOs are empty */
--- a/kernel/arch/dreamcast/hardware/spu.c
+++ b/kernel/arch/dreamcast/hardware/spu.c
@@ -65,6 +65,7 @@ void spu_memload(uintptr_t dst, void *src_void, size_t length) {
 void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     uint8_t *src = (uint8_t *)src_void;
     size_t aligned_len;
+    g2_ctx_t ctx;
 
     if(length < 32) {
         spu_memload(dst, src_void, length);
@@ -83,10 +84,24 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     /* Add in the SPU RAM base (cached area) */
     dst |= SPU_RAM_BASE;
 
+    /* First lock SQs because there is a mutex there. */
+    sq_lock(NULL);
+
     /* Make sure the FIFOs are empty */
     g2_fifo_wait();
 
+    /* Lock G2 bus because we can't suspend SQs from
+     * another thread with PIO access to G2 bus. */
+    ctx = g2_lock();
+
     sq_cpy((void *)dst, src, aligned_len);
+
+    /* We have some free time here to finish up the SQs work
+       before we unlock G2 and enable IRQ. So we'll unlock it first. */
+    sq_unlock();
+    sq_wait();
+
+    g2_unlock(ctx);
 
     if(length > 0) {
         /* Make sure the destination is in a non-cached area */
@@ -197,6 +212,7 @@ void spu_memset(uintptr_t dst, uint32_t what, size_t length) {
 
 void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
     int aligned_len;
+    g2_ctx_t ctx;
 
     /* Round up to the nearest multiple of 4 */
     if(length & 3) {
@@ -210,10 +226,24 @@ void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
     /* Add in the SPU RAM base (cached area) */
     dst |= SPU_RAM_BASE;
 
+    /* First lock SQs because there is a mutex there. */
+    sq_lock(NULL);
+
     /* Make sure the FIFOs are empty */
     g2_fifo_wait();
 
+    /* Lock G2 bus because we can't suspend SQs from
+     * another thread with PIO access to G2 bus. */
+    ctx = g2_lock();
+
     sq_set32((void *)dst, what, aligned_len);
+
+    /* We have some free time here to finish up the SQs work
+       before we unlock G2 and enable IRQ. So we'll unlock it first. */
+    sq_unlock();
+    sq_wait();
+
+    g2_unlock(ctx);
 
     if(length > 0) {
         /* Make sure the destination is in a non-cached area */
@@ -226,17 +256,19 @@ void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
 void spu_reset_chans(void) {
     int i;
     g2_fifo_wait();
-    g2_write_32(SNDREGADDR(0x2800), 0);
+    g2_ctx_t ctx = g2_lock();
+    g2_write_32_raw(SNDREGADDR(0x2800), 0);
 
     for(i = 0; i < 64; i++) {
-        if(!(i % 4)) g2_fifo_wait();
+        if((i & 3) == 0) g2_fifo_wait();
 
-        g2_write_32(CHNREGADDR(i, 0), 0x8000);
-        g2_write_32(CHNREGADDR(i, 20), 0x1f);
+        g2_write_32_raw(CHNREGADDR(i, 0), 0x8000);
+        g2_write_32_raw(CHNREGADDR(i, 20), 0x1f);
     }
 
     g2_fifo_wait();
-    g2_write_32(SNDREGADDR(0x2800), 0x000f);
+    g2_write_32_raw(SNDREGADDR(0x2800), 0x000f);
+    g2_unlock(ctx);
 }
 
 /* Enable/disable the SPU; note that disable implies reset of the
--- a/kernel/arch/dreamcast/hardware/spu.c
+++ b/kernel/arch/dreamcast/hardware/spu.c
@@ -255,8 +255,10 @@ void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
 /* Reset the AICA channel registers */
 void spu_reset_chans(void) {
     int i;
+    g2_ctx_t ctx;
+
     g2_fifo_wait();
-    g2_ctx_t ctx = g2_lock();
+    ctx = g2_lock();
     g2_write_32_raw(SNDREGADDR(0x2800), 0);
 
     for(i = 0; i < 64; i++) {
--- a/kernel/arch/dreamcast/include/dc/sound/sound.h
+++ b/kernel/arch/dreamcast/include/dc/sound/sound.h
@@ -2,7 +2,7 @@
 
    dc/sound/sound.h
    Copyright (C) 2002 Megan Potter
-   Copyright (C) 2023 Ruslan Rostovtsev
+   Copyright (C) 2023, 2024 Ruslan Rostovtsev
 
 */
 
@@ -27,6 +27,7 @@ __BEGIN_DECLS
 
 #include <arch/types.h>
 #include <stdint.h>
+#include <stdbool.h>
 
 /** \defgroup audio_driver  Driver
     \brief                  Low-level driver for SPU and audio management
@@ -216,7 +217,28 @@ void snd_adpcm_split(uint32_t *data, uint32_t *left, uint32_t *right, size_t siz
 
 /** @} */
 
+/** \brief  Get AICA channel position.
+
+    This function returns actual the channel position
+    that stores in SPU memory and updated by the SPU firmware.
+
+    \param  chn             The channel to retrieve position.
+
+    \return                 Last channel position in samples.
+*/
+uint16_t snd_get_pos(uint32_t ch);
+
+/** \brief  Get AICA channel playback state.
+
+    This function returns actual the channel playback state
+    that stores in AICA registers directly.
+
+    \param  chn             The channel to check.
+
+    \return                 True if the channel is playing.
+*/
+bool snd_is_playing(uint32_t ch);
+
 __END_DECLS
 
 #endif  /* __DC_SOUND_SOUND_H */
-
--- a/kernel/arch/dreamcast/include/dc/sound/sound.h
+++ b/kernel/arch/dreamcast/include/dc/sound/sound.h
@@ -226,7 +226,7 @@ void snd_adpcm_split(uint32_t *data, uint32_t *left, uint32_t *right, size_t siz
 
     \return                 Last channel position in samples.
 */
-uint16_t snd_get_pos(uint32_t ch);
+uint16_t snd_get_pos(unsigned int ch);
 
 /** \brief  Get AICA channel playback state.
 
@@ -237,7 +237,7 @@ uint16_t snd_get_pos(uint32_t ch);
 
     \return                 True if the channel is playing.
 */
-bool snd_is_playing(uint32_t ch);
+bool snd_is_playing(unsigned int ch);
 
 __END_DECLS
 
--- a/kernel/arch/dreamcast/sound/snd_iface.c
+++ b/kernel/arch/dreamcast/sound/snd_iface.c
@@ -1,7 +1,8 @@
 /* KallistiOS ##version##
 
    snd_iface.c
-   (c)2000-2002 Megan Potter
+   Copyright (C) 2000-2002 Megan Potter
+   Copyright (C) 2024 Ruslan Rostovtsev
 
    SH-4 support routines for accessing the AICA via the standard KOS driver
 */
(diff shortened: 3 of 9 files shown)


kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -83,11 +84,25 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
     /* Add in the SPU RAM base (cached area) */
     dst |= SPU_RAM_BASE;
 
+    /* First lock SQs because there is a mutex there. */
+    sq_lock(NULL);
```
gyrovorbis, line 88: This isn't necessary, since `sq_cpy()` locks and unlocks the SQs already, right?

kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -210,11 +226,25 @@ void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
...
```
gyrovorbis, line 230: This isn't necessary because `sq_set32()` already locks the SQs?

gyrovorbis (commented):
Couple very very minor comments... 

Btw is part of the speedup here the fact that you're now locking the G2 BUS only ONCE for each operation, which is then doing MULTIPLE raw access operations with it, before releasing, as opposed to using the non-raw write methods which were locking/unlocking for every transaction causing a perf hit?

kernel/arch/dreamcast/hardware/spu.c
DC-SWAT: Yes and no. The thing is that the G2 lock disables IRQs, and the SQ lock uses a mutex.
If SQ mutex is already locked and we try lock it with disabled IRQs then we get deadlock.
If you have a better solution, I'll be glad, because it looks bad. I think I wrote comments to explain it, but apparently they are bad too haha.

kernel/arch/dreamcast/hardware/spu.c
DC-SWAT: The same.

kernel/arch/dreamcast/sound/snd_iface.c
```diff
@@ -66,57 +61,53 @@ int snd_init(void) {
...
 
 /* Submit a request to the SH4->AICA queue; size is in uint32's */
 int snd_sh4_to_aica(void *packet, uint32_t size) {
-    uint32_t  qa, bot, start, top, *pkt32, cnt;
-    assert_msg(size < 256, "SH4->AICA packets may not be >256 uint32's long");
+    uint32_t qa, bot, start, top, *pkt32, cnt;
+    assert_msg(size < AICA_CMD_MAX_SIZE, "SH4->AICA packets may not be >256 uint32's long");
 
-    sem_wait(&sem_qram);
+    g2_ctx_t ctx = g2_lock();
```
pcercuei, line 80: Try to keep the variables declarations at the top of the function (before the assert_msg). The initialization can stay here.

kernel/arch/dreamcast/sound/snd_iface.c
```diff
@@ -66,57 +61,53 @@ int snd_init(void) {
...
-    start = bot + g2_read_32(qa + offsetof(aica_queue_t, head));
-    pkt32 = (uint32 *)packet;
+    bot = SPU_RAM_UNCACHED_BASE + g2_read_32_raw(qa + offsetof(aica_queue_t, data));
+    top = bot + g2_read_32_raw(qa + offsetof(aica_queue_t, size));
+    start = bot + g2_read_32_raw(qa + offsetof(aica_queue_t, head));
+    pkt32 = (uint32_t *)packet;
     cnt = 0;
 
-    while(size > 0) {
+    while(size-- > 0) {
```
pcercuei, line 92: Unrelated cosmetic change, if you really want this, do it in a separate commit.

kernel/arch/dreamcast/sound/snd_iface.c
```diff
@@ -66,57 +61,53 @@ int snd_init(void) {
...
         /* Fifo wait if necessary */
-        if(!(cnt % 8))
+        if((cnt++ & 7) == 0)
```
pcercuei, line 94: Same here - unrelated cosmetic change.

kernel/arch/dreamcast/sound/snd_iface.c
```diff
@@ -66,57 +61,53 @@ int snd_init(void) {
...
+    uint32_t qa, bot, start, top, *pkt32, cnt;
+    assert_msg(size < AICA_CMD_MAX_SIZE, "SH4->AICA packets may not be >256 uint32's long");
 
-    sem_wait(&sem_qram);
+    g2_ctx_t ctx = g2_lock();
 
     /* Set these up for reference */
     qa = SPU_RAM_UNCACHED_BASE + AICA_MEM_CMD_QUEUE;
-    assert_msg(g2_read_32(qa + offsetof(aica_queue_t, valid)), "Queue is not yet valid");
+    assert_msg(g2_read_32_raw(qa + offsetof(aica_queue_t, valid)), "Queue is not yet valid");
```
pcercuei, line 84: I am not sure that those assert_msg() are safe, since you are inside a critical section now.

kernel/arch/dreamcast/sound/snd_iface.c
```diff
@@ -136,30 +127,29 @@ void snd_sh4_to_aica_stop(void) {
...
     }
 
     /* Check for packet size overflow */
-    size = g2_read_32(start + offsetof(aica_cmd_t, size));
+    size = g2_read_32_raw(start + offsetof(aica_cmd_t, size));
 
-    if(cnt >= AICA_CMD_MAX_SIZE) {
-        sem_signal(&sem_qram);
+    if(size >= AICA_CMD_MAX_SIZE) {
+        g2_unlock(ctx);
```
pcercuei, line 160: This changed from `cnt` to `size`, if that's on purpose, I want to know (in the commit description) why.

kernel/arch/dreamcast/hardware/spu.c
pcercuei: Lock the SQs before disabling the interrupts.

kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -226,17 +256,19 @@ void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
 void spu_reset_chans(void) {
     int i;
     g2_fifo_wait();
-    g2_write_32(SNDREGADDR(0x2800), 0);
+    g2_ctx_t ctx = g2_lock();
+    g2_write_32_raw(SNDREGADDR(0x2800), 0);
 
     for(i = 0; i < 64; i++) {
-        if(!(i % 4)) g2_fifo_wait();
+        if((i & 3) == 0) g2_fifo_wait();
```
pcercuei, line 263: Cosmetic changes to their own commit.

kernel/arch/dreamcast/sound/snd_iface.c
```diff
@@ -27,6 +28,11 @@ static int initted = 0;
 extern uint8_t snd_stream_drv[];
 extern uint8_t snd_stream_drv_end[];
 
+/* The queue processing mutex for snd_sh4_to_aica_start and snd_sh4_to_aica_stop.
+   There are rare case, start stream + sfx control at the same time in separate threads,
+   so we just use fast spinlock for it. */
+static spinlock_t queue_proc_mutex = SPINLOCK_INITIALIZER;
```
pcercuei, line 34: This is not a mutex, it is a spinlock. So name it accordingly.

But it sounds like your brain really wanted you to have a mutex, so why use a spinlock? It sounds like having a mutex would be a much better idea.

kernel/arch/dreamcast/sound/snd_iface.c
```diff
@@ -211,3 +211,11 @@ void snd_poll_resp(void) {
     if(rv < 0)
         dbglog(DBG_ERROR, "snd_poll_resp(): snd_aica_to_sh4 failed, giving up\n");
 }
+
+uint16_t snd_get_pos(uint32_t ch) {
+    return (g2_read_32(SPU_RAM_UNCACHED_BASE + AICA_CHANNEL(ch) + offsetof(aica_channel_t, pos)) & 0xffff);
```
pcercuei, line 216: `return (a & b);` -> `return a & b;`

pcercuei (changes requested):
Overall looks quite OK, but I do want to see a separation of the purely cosmetic changes with the actual functional changes, and much better commit messages that actually tell what's going on.

"Speedup and fix AICA cmd iface", speedup how? Fix what? What was broken and why does it need fixing?

"Protect the SQs on G2 bus", protect from what? "speedup spu_reset_chans", how?

"Added spinlock for start/stop the queue", what for?

"Fixed aligning for AICA cmd channel buffer", why does alignment matter?

Obviously I'm not asking you to write these answers in the PR discussion, but in the commit messages.

kernel/arch/dreamcast/hardware/spu.c
DC-SWAT: This is performance change ;)
Division with remainder takes more CPU time than just AND.

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: I have some doubts here. It's just that this lock is needed extremely rarely and I don't want to pull the mutex for no reason.

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: Yes, I try to do that. But I admit that this one missed it.

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: If the assert initiated, it breaks the entire program anyway.

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: Nope, this is performance! :)

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: Are you sure? GCC can do this loop in another way, by using `dt` for example.
Although he will probably figure it out himself, but still.

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: Because here the packet size is checked, and not the variable `cnt` which is always zero here. Just look at the code a little more closely.

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: OK.

kernel/arch/dreamcast/hardware/spu.c
DC-SWAT: You mean change comment to "Lock the SQs before disabling the interrupts."?

DC-SWAT:
> Couple very very minor comments...
> 
> Btw is part of the speedup here the fact that you're now locking the G2 BUS only ONCE for each operation, which is then doing MULTIPLE raw access operations with it, before releasing, as opposed to using the non-raw write methods which were locking/unlocking for every transaction causing a perf hit?

Exactly. Each g2_lock disables interrupts and suspend the work of 4 DMA channels on the G2 bus and will do more when I get to direct DMAs. Plus I got rid of the semaphore here.

DC-SWAT:
> "Speedup and fix AICA cmd iface", speedup how? Fix what? What was broken and why does it need fixing? speedup spu_reset_chans", how?

I thought it was obvious, but apparently not for everyone.

> "Fixed aligning for AICA cmd channel buffer", why does alignment matter?

I thought it was obvious, since it uses 32-bit access for packets.

> "Protect the SQs on G2 bus", protect from what? "

Comment from code:
    /* Lock G2 bus because we can't suspend SQs from
     * another thread with PIO access to G2 bus. */

> 
> "Added spinlock for start/stop the queue", what for?
> 

Comment from code:

/* The queue processing mutex for snd_sh4_to_aica_start and snd_sh4_to_aica_stop.
   There are rare case, start stream + sfx control at the same time in separate threads,
   so we just use fast spinlock for it. */

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: Although, better really make a mutex, I remembered other cases.

kernel/arch/dreamcast/sound/snd_iface.c
pcercuei: GCC will absolutely know how to optimize that properly.

kernel/arch/dreamcast/sound/snd_iface.c
pcercuei: GCC will absolutely generate the exact same code. So this "performance change" is pointless.

kernel/arch/dreamcast/sound/snd_iface.c
pcercuei: Fair enough.

kernel/arch/dreamcast/hardware/spu.c
pcercuei: Again here GCC will emit the exact same code so the change is pointless.

QuzarDC:
> Commits from PR almost always merge into one commit, which has the description from the PR. I updated the description.

Just to be clear on that last part, that's not some sort of automatic process. I end up doing that squashing when there are multiple commits that can't be untangled easily, then write the specifics based off trying to summarize my understanding of what's happening in combination with the description and any feedback/discussion. When it's in the commit descriptions, github *does* automatically put them all together, which makes things easier (though in those cases I often still edit to remove repeated lines or commit comments like 'Addressed feedback review' which don't serve a purpose in the main repo log).
---
I agree generally with the feedback to remove the small changes to the while loops and alignment checks that shouldn't produce any code difference.
---
From my understanding the biggest portion of the optimization part has to do with favoring maintaining g2 bus locks longer to reduce the overhead of locking/unlocking on each operation (similar to the optimization done for `pvr_prim()`). Has the possible impact of this on networking functionality been observed? That is the only other major use for the G2 bus, and it seems like it should make a difference in some way. 

I could imagine that it might make it less likely for the loads to block net transfers because you aren't spamming the lock/unlock and the total amount of time spent locked is decreased. Alternatively I could see long locks meaning it could be more likely to block due to the amount of continuously blocked time. Or that it might shift based on specific use cases.

I don't think it's really a problem (unless there is a use case that shows that it entirely breaks networking which I highly doubt) but I wanted to see if that was something considered or if anyone had feedback on the subject.

DC-SWAT:
> 
> I agree generally with the feedback to remove the small changes to the while loops and alignment checks that shouldn't produce any code difference.
> 

It's produce ASM generation difference. But the main logic not changed, yes.

> From my understanding the biggest portion of the optimization part has to do with favoring maintaining g2 bus locks longer to reduce the overhead of locking/unlocking on each operation (similar to the optimization done for `pvr_prim()`). Has the possible impact of this on networking functionality been observed? That is the only other major use for the G2 bus, and it seems like it should make a difference in some way.
> 
> I could imagine that it might make it less likely for the loads to block net transfers because you aren't spamming the lock/unlock and the total amount of time spent locked is decreased. Alternatively I could see long locks meaning it could be more likely to block due to the amount of continuously blocked time. Or that it might shift based on specific use cases.
> 
> I don't think it's really a problem (unless there is a use case that shows that it entirely breaks networking which I highly doubt) but I wanted to see if that was something considered or if anyone had feedback on the subject.

The maximum packet size is 256 bytes, it small in general and can't affect anything other badly. Interrupts and DMA will be disabled for a short time.
Previously, for every 4 bytes, interrupts were disabled, SH4 and G2 FIFO were waited for, and 4 G2 DMA channels were suspended. Now this is done once for the entire data packet. The number of operations greatly reduced and AICA FIFO used properly (fully), so it works faster for G2 bus in general.

kernel/arch/dreamcast/hardware/spu.c
DC-SWAT: Okay, I'll know.

kernel/arch/dreamcast/sound/snd_iface.c
DC-SWAT: But you about modern ones, as I understand it.

kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -226,17 +256,19 @@ void spu_memset_sq(uintptr_t dst, uint32_t what, size_t length) {
 void spu_reset_chans(void) {
     int i;
     g2_fifo_wait();
-    g2_write_32(SNDREGADDR(0x2800), 0);
+    g2_ctx_t ctx = g2_lock();
```
pcercuei, line 259: Cosmetic change, but could you declare the `g2_ctx_t ctx;` at the beginning of the function? (can still be locked here)

kernel/arch/dreamcast/include/dc/sound/sound.h
```diff
@@ -216,7 +217,28 @@ void snd_adpcm_split(uint32_t *data, uint32_t *left, uint32_t *right, size_t siz
...
+/** \brief  Get AICA channel position.
+
+    This function returns actual the channel position
+    that stores in SPU memory and updated by the SPU firmware.
+
+    \param  chn             The channel to retrieve position.
+
+    \return                 Last channel position in samples.
+*/
+uint16_t snd_get_pos(uint32_t ch);
```
pcercuei, line 229: The channel doesn't really need to be `uint32_t`, I'd prefer those types to be used when we need a specific size. You could just use "unsigned int" here.

kernel/arch/dreamcast/include/dc/sound/sound.h
```diff
@@ -216,7 +217,28 @@ void snd_adpcm_split(uint32_t *data, uint32_t *left, uint32_t *right, size_t siz
...
+/** \brief  Get AICA channel playback state.
+
+    This function returns actual the channel playback state
+    that stores in AICA registers directly.
+
+    \param  chn             The channel to check.
+
+    \return                 True if the channel is playing.
+*/
+bool snd_is_playing(uint32_t ch);
```
pcercuei, line 240: Same here.

kernel/arch/dreamcast/sound/snd_iface.c
```diff
@@ -215,3 +211,11 @@ void snd_poll_resp(void) {
     if(rv < 0)
         dbglog(DBG_ERROR, "snd_poll_resp(): snd_aica_to_sh4 failed, giving up\n");
 }
+
+uint16_t snd_get_pos(uint32_t ch) {
+    return g2_read_32(SPU_RAM_UNCACHED_BASE + AICA_CHANNEL(ch) + offsetof(aica_channel_t, pos)) & 0xffff;
+}
+
+bool snd_is_playing(uint32_t ch) {
+    return g2_read_32(MEM_AREA_P2_BASE + 0x00700000 + 0x80 * ch) & 0x4000;
```
pcercuei, line 220: Can we have a #define for that bit you are checking, instead of a magic 0x4000 value?

pcercuei (changes requested):
Some minor comments

QuzarDC (dismissed):

kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -83,11 +84,25 @@ void spu_memload_sq(uintptr_t dst, void *src_void, size_t length) {
...
 
+    /* Lock the SQs before disabling the interrupts. */
+    sq_lock(NULL);
+
     /* Make sure the FIFOs are empty */
     g2_fifo_wait();
 
+    /* Lock G2 bus because we can't suspend SQs from
+     * another thread with PIO access to G2 bus. */
+    ctx = g2_lock();
```
pcercuei, line 95: I see a potential problem here. Nothing prevents the main thread from being preempted between the call to `g2_fifo_wait()` and the call to `g2_lock()`. So you could get in a state where you obtain the lock, but the FIFO is not empty.

Ideally all calls to g2_fifo_wait() should be moved *after* calling g2_lock(). Or even better, g2_fifo_wait() could be called inside g2_lock(), so that it is guaranteed that the FIFOs are empty after the function returns.

DC-SWAT: Done.
`g2_lock()` is already waiting for the necessary FIFOs, and this one needs to be waited for only when writing, so not every lock needs this.

pcercuei: Ok, I think I will rework the G2 API a bit in a following PR. But until then this PR looks ready.

pcercuei (approved):

kernel/arch/dreamcast/hardware/spu.c
DC-SWAT: Yeah, here it would probably be better to simply make a mask-argument with the access rights to the bus. In this way all necessary FIFOs will be waiting inside `g2_lock()`.
But the thing is that FIFO for AICA and G2 will have to wait every 8 writes, so this function will still be separate. Although we can keep count inside G2 driver and check with each write whether we need to wait for FIFOs or not... In general, yes, we need to think about how to do it better.

QuzarDC (approved):

PR #839, opened 2024-10-27
Add new regfield.h header, and clean up PVR code

pcercuei:
Add a new `<kos/regfield.h>` header, which will provide the `BIT()`, `GENMASK()`, `FIELD_GET()` and `FIELD_PREP()` macros which are really useful to deal with register fields.

Add  a new `<kos/deprecated.h>` header, to which deprecated stuff can be moved to avoid polluting the main API, while still staying API-compatible until the next major release.

Use the new regfield macros to clean up the PVR code.

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_prim.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_prim.c
@@ -21,148 +21,78 @@
 
 /* Compile a polygon context into a polygon header */
 void pvr_poly_compile(pvr_poly_hdr_t *dst, const pvr_poly_cxt_t *src) {
-    int u, v;
     uint32_t txr_base;
     /* Temporary variables we can read-write-modify, since we cannot do so from
        within the SQs, and we want to be able to compile this header from a PVR
        DR API submission target. */
-    uint32_t cmd, mode[3];
+    uint32_t mode2, mode3;
 
     /* Basically we just take each parameter, clip it, shift it
        into place, and OR it into the final result. */
 
     /* The base values for CMD */
-    cmd = PVR_CMD_POLYHDR;
-
-    if(src->txr.enable == PVR_TEXTURE_ENABLE)
-        cmd |= 8;
-
-    /* Or in the list type, shading type, color and UV formats */
-    cmd |= (src->list_type << PVR_TA_CMD_TYPE_SHIFT) & PVR_TA_CMD_TYPE_MASK;
-    cmd |= (src->fmt.color << PVR_TA_CMD_CLRFMT_SHIFT) & PVR_TA_CMD_CLRFMT_MASK;
-    cmd |= (src->gen.shading << PVR_TA_CMD_SHADE_SHIFT) & PVR_TA_CMD_SHADE_MASK;
-    cmd |= (src->fmt.uv << PVR_TA_CMD_UVFMT_SHIFT) & PVR_TA_CMD_UVFMT_MASK;
-    cmd |= (src->gen.clip_mode << PVR_TA_CMD_USERCLIP_SHIFT) & PVR_TA_CMD_USERCLIP_MASK;
-    cmd |= (src->fmt.modifier << PVR_TA_CMD_MODIFIER_SHIFT) & PVR_TA_CMD_MODIFIER_MASK;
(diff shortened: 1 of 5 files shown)


ljsebald:
Adding a new header that's needed to get deprecated functionality is explicitly changing the API. I do not like that change and request that it be removed...

Not to mention the fact that it shouldn't have been combined with the other changes anyway, since PRs should really relate to one thing and do it well.

pcercuei:
> Adding a new header that's needed to get deprecated functionality is explicitly changing the API. I do not like that change and request that it be removed...

This header is included from <kos.h> though, which is the only one that really has to stay API-compatible, I think?
 
> Not to mention the fact that it shouldn't have been combined with the other changes anyway, since PRs should really relate to one thing and do it well.

Should I just drop the old macros then? Or keep them in the public API, and somebody will eventually remember to remove them?

ljsebald:
> > Adding a new header that's needed to get deprecated functionality is explicitly changing the API. I do not like that change and request that it be removed...
> 
> This header is included from <kos.h> though, which is the only one that really has to stay API-compatible, I think?

No, there is no rule that says only `<kos.h>` has to stay API-compatible. I've been trying to discourage the use of `<kos.h>` for a long time, not encourage its proliferation.

> > Not to mention the fact that it shouldn't have been combined with the other changes anyway, since PRs should really relate to one thing and do it well.
> 
> Should I just drop the old macros then? Or keep them in the public API, and somebody will eventually remember to remove them?

Mark the whole group of them as deprecated in the Doxygen comment, and make a new group for the new macros. That'd be my suggestion.

pcercuei:
@ljsebald ok, updated.

gyrovorbis (approved):
Approving, because I think this is far more readable, concise, pretty, and scalable, and cuz I did review it... buuuuuuuut... Not going to merge until Bossman @ljsebald sees that the API breakage has been fixed and blesses it. :)

pcercuei:
> No, there is no rule that says only `<kos.h>` has to stay API-compatible. I've been trying to discourage the use of `<kos.h>` for a long time, not encourage its proliferation.

@ljsebald should I drop the include of `<kos/regfield.h>` from `<kos.h>` itself then?

ljsebald:
> > No, there is no rule that says only `<kos.h>` has to stay API-compatible. I've been trying to discourage the use of `<kos.h>` for a long time, not encourage its proliferation.
> 
> @ljsebald should I drop the include of `<kos/regfield.h>` from `<kos.h>` itself then?

No, you definitely don’t need to drop it from there. While I’ve been trying to discourage using it, I’m not suggesting we take it away or remove its functionality as the be-all end-all header.

kernel/arch/dreamcast/hardware/pvr/pvr_prim.c
```diff
@@ -338,385 +268,186 @@ void pvr_sprite_cxt_txr(pvr_sprite_cxt_t *dst, pvr_list_t list,
...
-        mode2[1] |= (u << PVR_TA_PM2_USIZE_SHIFT) & PVR_TA_PM2_USIZE_MASK;
-        mode2[1] |= (v << PVR_TA_PM2_VSIZE_SHIFT) & PVR_TA_PM2_VSIZE_MASK;
-
-        dst->mode2_1 = mode2[1];
-
-        /* Polygon mode 3 (within volume) */
-        mode3[1]  = (src->txr2.mipmap << PVR_TA_PM3_MIPMAP_SHIFT) & PVR_TA_PM3_MIPMAP_MASK;
-        mode3[1] |= (src->txr2.format << PVR_TA_PM3_TXRFMT_SHIFT) & PVR_TA_PM3_TXRFMT_MASK;
+        assert_msg(__builtin_popcount(src->txr2.width) == 1
+		   && src->txr2.width <= 1024, "Invalid texture U size");
```
ljsebald, line 425: Just putting it out here that this is really clever (along with the ctz trick below), but it definitely threw me for a minute and made me have to look up what those builtins do... It might be a good idea to eventually add a comment for posterity explaining what in the world is going on there so that others don't get confused if they try to grok the code, but that's for another day...

Ultimately, most people aren't all that familiar with the weirder named gcc builtins.

pcercuei: We can probably wrap those with macro names e.g. `count_bits()` and `count_trailing_zeros()`.

PR #841, opened 2024-11-03
Improve ASM in fmath_base.h

pcercuei:
- Avoid doing multiplications / divisions in ASM, and do them in C instead. The compiler has then much more insight about the possible pipeline stalls incurred.

- Avoid using dedicated registers. Use a complex variable to identify a floating-point register couple that can be used as a target for the fsca instruction, and use the __real__ keyword to identify the first result register, and the __imag__ keyword to identify the second result register. By doing this, the compiler won't have to move the arguments into specific registers or retrieve the results from specific registers, resulting in much better code generation.


pcercuei:
Nevermind, when used with Doom64, GCC will try to emit `fsca fpul,dr5` which is obviously invalid.

PR #843, opened 2024-11-04
Fix versioning macros

andressbarajas:
Recently a PR was reverted for SDL kos-ports(https://github.com/Dreamcast-Projects/kos-ports/commit/852443252b09cc635da3ae6e72a1add37cee2d30)

The fix for the versioning issue is already included in an existing big PR but since this is an emergency I pulled the versioning fixes out into its own PR which makes sense anyway.

### DIFF
--- a/include/kos/version.h
+++ b/include/kos/version.h
@@ -249,7 +249,7 @@
     \returns        Packed version identifier.
 */
 #define KOS_VERSION_MAKE(major, minor, patch) \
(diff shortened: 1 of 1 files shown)


darcagn:
Thanks, this indeed should have been its own PR; and this needs to be pushed to the 2.1.0 branch, with a 2.1.1 tag created.

gyrovorbis (approved):

PR #853, opened 2024-11-17
Correct out-of-memory test in scandir

QuzarDC:
The test after realloc was improperly testing 'list' for NULL rather than '*list'.

### DIFF
--- a/kernel/libc/koslib/scandir.c
+++ b/kernel/libc/koslib/scandir.c
@@ -48,7 +48,7 @@ static int push_back(struct dirent ***list, int *size, int *capacity,
         *list = realloc(*list, *capacity * sizeof(struct dirent*));
 
         /* Handle out-of-memory in case realloc() failed. */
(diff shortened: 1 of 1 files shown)


gyrovorbis (approved):
OMG THANK YOU FOR FINDING THIS!!! I swear I even went through the hell of testing this thing for proper OOM handling, but I guess i didn't run out of memory on that branch in my tests?

QuzarDC:
> OMG THANK YOU FOR FINDING THIS!!! I swear I even went through the hell of testing this thing for proper OOM handling, but I guess i didn't run out of memory on that branch in my tests?

My guess would be something like it freed *enough* memory to continue on through the failure but what would probably be needed to avoid bits like this would be to do a full malloc stats before/after (or some similar sort of test against the exact memory) to ensure that we leave the function with the same memory we entered it. I only found it through static analysis.

PR #854, opened 2024-11-17
Ensure graceful failure for fs_dcload when out of memory

QuzarDC:
As title. These should likely be revised in the future though to reduce the malloc overhead. For instance we might use a methodology where we only reallocate if a new filename is larger than the current, or use a large-enough static buffer.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_dcload.c
+++ b/kernel/arch/dreamcast/fs/fs_dcload.c
@@ -124,12 +124,21 @@ void *dcload_open(vfs_handler_t * vfs, const char *fn, int mode) {
 
             if(fn[strlen(fn) - 1] == '/') {
                 dcload_path = malloc(strlen(fn) + 1);
-                strcpy(dcload_path, fn);
+                if(dcload_path)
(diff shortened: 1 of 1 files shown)


gyrovorbis:
I saw this the other day, actually, and was not very happy to see DC-Load `malloc()`-ing its ass off... Personally, I'm of the opinion it should just be a static buffer, whose size is simply `PATH_MAX`, which *should* be enough to cover everything... `realpath()` and `fs_normalize_path()` both already work like this...

But either way, if you don't want to bother with that, this is certainly still better..

pcercuei:
I agree with @gyrovorbis, it would be much better to have `dcload_path` as a `PATH_MAX` sized static buffer.

QuzarDC:
I got well too deep into seeing how this should be done and it seems that the whole approach was flawed. I've greatly expanded on this in https://github.com/KallistiOS/KallistiOS/pull/893 . Unfortunately it means still malloc-ing on each `open` of a dir, but without doing so `readdir` wasn't working correctly, always only acting on the most recent dir opened rather than the one asked.

Issue #840, opened 2024-11-03
Typedefs in types.h escape into the wider world and conflict


Kazade:
If you try to compile stb_vorbis with KOS, you get the following errors (snipped)

```
stb_vorbis.c:656:21: error: conflicting types for 'int8'; have 'signed char'
  656 | typedef signed char int8;
      |                     ^~~~
In file included from /opt/toolchains/dc/kos/include/sys/_types.h:206,
                 from /opt/toolchains/dc/sh-elf/sh-elf/include/sys/reent.h:16,
                 from /opt/toolchains/dc/sh-elf/sh-elf/include/stdio.h:60,
                 from stb_vorbis.c:84:
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/arch/types.h:39:14: note: previous declaration of 'int8' with type 'int8' {aka 'char'}
   39 | typedef char int8;                  /**< \brief 8-bit signed integer */
      |              ^~~~
stb_vorbis.c:659:22: error: conflicting types for 'uint32'; have 'unsigned int'
  659 | typedef unsigned int uint32;
```

This is because stb_vorbis has the following typedefs: 

```
typedef unsigned char uint8;
typedef signed char int8;
typedef unsigned short uint16;
typedef signed short int16;
typedef unsigned int uint32;
typedef signed int int32;
```

I haven't found an easy fix to this, I have previously ifdef'd out those typedefs, but in GCC 14 I'm now seeing errors due to pointer mismatches (where stb_vorbis uses unsigned int instead of uint32). 

I don't think KOS should be exposing these typedefs through stdio.h? I think that's the issue...

gyrovorbis:
Yes, it's quite a PITA. This is the same issue we've had with Sonic Mania and a few other codebases too. I'll see if I can help look into this as well, because it's also been a total annoyance to me too.

Basically in Mania, I believe we had to rearrange include files and do a bunch of shenanigans... but you don't always own the C/C++ files to do that, either... :(

QuzarDC:
This is caused by this: https://github.com/KallistiOS/KallistiOS/blob/99308c534d03c585d4b4aa14ad278e072081300b/include/sys/_types.h#L201-L202

[...]

PR #855, opened 2024-11-22
Update byteorder to not rely on arch types

QuzarDC:
byteorder.h was including <sys/_types.h> taking advantage of an old workaround that exports our internal types through that header and into stdio.

This was the only file relying on that workaround in the KOS codebase and adjusting it is needed to eventually remove or adjust the workaround for #840 .


gyrovorbis (approved):
Good call.

PR #857, opened 2024-11-25
Fix wide character support.

QuzarDC:
The previous cleanup of bfont.h (#581) accidentally used the size of a thin char to calculate the offsets for all the wide chars.

The impact was visible in the bfont example where instead of printing the start button icon, a JISX-0208 character was printed.


gyrovorbis (approved):
Thank you for fixing this! Apologies for screwing it up. :P

QuzarDC:
> Thank you for fixing this! Apologies for screwing it up. :P

No worries. I get the trout too for missing it when reviewing -_-; Only noticed it when I was messing around with the bfont stuff myself and thought I'd broken it.

PR #859, opened 2024-11-29
Add pvr_wait_render_done() and pvr_get_front_buffer()

pcercuei:
The `pvr_wait_render_done()` differs from `pvr_wait_ready()` as the latter only waits until a new scene can be rendered, which does not mean that the PVR is done rendering.
Having a new `pvr_wait_render_done()` function is useful for when the program needs to swap out textures that may be used by the PVR for rendering, or when it needs to access the front buffer.

The `pvr_get_front_buffer()` returns a pointer to the front buffer, in 64-bit memory space. This allows programs to do some special effects, like re-loading the previous frame for overpainting, or to do blur effects.

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_misc.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_misc.c
@@ -187,7 +187,7 @@ void pvr_begin_queued_render(void) {
     } zclip;
 
     /* Get the appropriate buffer */
-    tbuf = pvr_state.ta_buffers + (pvr_state.ta_target ^ 1);
+    tbuf = pvr_state.ta_buffers + (pvr_state.ta_target ^ pvr_state.vbuf_doublebuf);
     rbuf = pvr_state.frame_buffers + (bufn ^ 1);
 
     /* Calculate background value for below */
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_misc.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_misc.c
@@ -269,3 +269,21 @@ void pvr_blank_polyhdr_buf(int type, pvr_poly_hdr_t * poly) {
     poly->d1 = poly->d2 = poly->d3 = poly->d4 = 0xffffffff;
 
 }
+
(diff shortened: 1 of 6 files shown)


pcercuei:
I updated the `pvr_get_front_buffer()` commit and force-pushed.
I removed the call to `pvr_wait_render_done()`. The front buffer will always be fully rendered by the time the PVR starts to render to the back buffer; for that reason, we actually don't need it to be fully rendered while we use it as a texture.

kernel/arch/dreamcast/hardware/pvr/pvr_misc.c
```diff
@@ -269,3 +269,22 @@ void pvr_blank_polyhdr_buf(int type, pvr_poly_hdr_t * poly) {
     poly->d1 = poly->d2 = poly->d3 = poly->d4 = 0xffffffff;
 
 }
+
+pvr_ptr_t pvr_get_front_buffer(void)
+{
```
gyrovorbis, line 274: Remove newline here.

gyrovorbis:
Looks good, but out of curiosity... do you think there's merit in making `pvr_wait_render_done()` take the number of `msecs` to wait as a timeout value rather than hardcoding it to 100?

gyrovorbis (commented):
Teeny stylistic comment. Looks good otherwise.

pcercuei:
> Looks good, but out of curiosity... do you think there's merit in making `pvr_wait_render_done()` take the number of `msecs` to wait as a timeout value rather than hardcoding it to 100?

Well, I don't know. We don't do that for `pvr_wait_ready()` either.

kernel/arch/dreamcast/hardware/pvr/pvr_misc.c
pcercuei: Fixed, thanks.

pcercuei:
I pushed a new commit to this PR (as it required pvr_wait_render_done()).

This new commit adds an option to disable the double-buffering of the vertex buffer.

QuzarDC (approved):

PR #867, opened 2024-12-09
koslib: Drop our aligned_alloc()

pcercuei:
Newlib already provides this function; which means that sometimes KOS' implementation was picked, and sometimes Newlib's implementation was picked.

This was a problem with LTO in particular, as Newlib's implementation would fail to link against KOS-provided memalign().

Address this issue by dropping KOS' aligned_alloc(), and marking its memalign() implementation as "always used", which fixes the linking when using LTO.


gyrovorbis:
Wonder how the hell this even happened? Did Newlib ALWAYS include `aligned_alloc()` for C11, or was it new? And will this break the toolchain for way older Newlib versions?

...not that I even care, even if it did, though... Gotta draw the line somewhere, this is breaking anything reasonably modern, and it does look like the right thing to do. Just might need to check and make sure we don't keep providing configuration profiles for ancient newlib versions, assuming this symbol was not always part of newlib.

darcagn:
> Wonder how the hell this even happened? Did Newlib ALWAYS include `aligned_alloc()` for C11, or was it new? And will this break the toolchain for way older Newlib versions?
> 
> ...not that I even care, even if it did, though... Gotta draw the line somewhere, this is breaking anything reasonably modern, and it does look like the right thing to do. Just might need to check and make sure we don't keep providing configuration profiles for ancient newlib versions, assuming this symbol was not always part of newlib.
> 
> ...But for all I know, this was added for GCC4.7 and ancient stuff we don't care about anymore, and there's no problem.

KOS (BlueCrab) added aligned_alloc() in Nov 2015 after Newlib did in Oct 2015, but KOS was still on 2.0.0 from 2012 for the 4.7.4 toolchain.

pcercuei:
Something is fishy, it complains about _memalign_r() now:
```
paul@frank:~/dev/dreamcast/bloom/build$ LANG=C nice ninja
[0/2] Re-checking globbed directories...
[1/2] Linking CXX executable bloom.elf
FAILED: bloom.elf 
: && /home/paul/dev/dreamcast/kos/utils/build_wrappers/kos-c++ --sysroot=/home/paul/dev/dreamcast/kos/../kos-ports -O3 -DNDEBUG -Wl,--whole-archive -lromdiskbase -Wl,--no-whole-archive CMakeFiles/bloom.dir/src/cdr.c.obj CMakeFiles/bloom.dir/src/dynload.c.obj CMakeFiles/bloom.dir/src/emu.c.obj CMakeFiles/bloom.dir/src/genmenu.cpp.obj CMakeFiles/bloom.dir/src/mmap.c.obj CMakeFiles/bloom.dir/src/platform.c.obj CMakeFiles/bloom.dir/src/plugins.c.obj CMakeFiles/bloom.dir/src/ide.c.obj CMakeFiles/bloom.dir/romdisk_tmp.cpp.obj -o bloom.elf  /home/paul/dev/dreamcast/kos/../kos-ports/lib/libfastmem.a  liblibpcsxcore.a  libgpu.a  libspu.a  /home/paul/dev/dreamcast/kos/../kos-ports/lib/libtsunami.a  /home/paul/dev/dreamcast/kos/addons/lib/dreamcast/libkosfat.a  deps/pcsx_rearmed/deps/lightrec/liblightrec.a  liblightning.a  /home/paul/dev/dreamcast/kos/../kos-ports/lib/libz.a  /home/paul/dev/dreamcast/kos/../kos-ports/lib/libfastmem.a  /home/paul/dev/dreamcast/kos/../kos-ports/lib/libparallax.a && :
/home/paul/dev/toolchains/sh-elf-lra/bin/../lib/gcc/sh-elf/15.0.0/../../../../sh-elf/bin/ld: /home/paul/dev/toolchains/sh-elf-lra/bin/../lib/gcc/sh-elf/15.0.0/../../../../sh-elf/lib/m4-single-only/libc.a(libc_a-aligned_alloc.o): in function `aligned_alloc':
/home/paul/dev/dreamcast/kos/utils/dc-chain/build-newlib-sh-elf-4.4.0.20231231/sh-elf/m4-single-only/newlib/../../../../newlib-4.4.0.20231231/newlib/libc/stdlib/aligned_alloc.c:36:(.text.aligned_alloc+0x10): undefined reference to `_memalign_r'
ninja: build stopped: subcommand failed.
```

pcercuei:
Updated the PR. The memalign_r() stub was missing, so I added it and now it should link properly.

kernel/libc/koslib/malloc.c
```diff
@@ -1037,6 +1037,7 @@ extern "C" {
 
       Overreliance on memalign is a sure way to fragment space.
     */
+    __attribute__((used))
```
QuzarDC, line 1040: Is there a reason this isn't also `__used` as used everywhere else?

QuzarDC (commented):
Other than the one question, seems fine. 

At some point we should see maybe about revising our interface with newlib. It could well be that there are other unintended overlaps as we've been able to keep current with its updates.

kernel/libc/koslib/malloc.c
pcercuei: None. Fixed and force-pushed (with only this line modified).

QuzarDC (approved):

Issue #868, opened 2024-12-17
Wrong vertex in vec3f_rotd_xy


lerabot:
https://github.com/KallistiOS/KallistiOS/blob/bae7adbde177c04560de31577932ace50988b238/kernel/arch/dreamcast/include/dc/vec3f.h#L333

Both component are using X and **Z** instead of X and **Y**

QuzarDC:
Fixed by #870

PR #870, opened 2024-12-18
Drop vec3f ASM

pcercuei:
Use the vec3f_t type defined by the header, and replace all inline ASM with C code.

Apart from the "fipr" instruction, which we already have macros for, there is absolutely no reason to have these algorithms in ASM. Not only it is unnecessary, it is also harmful as the C compiler cannot re-organize those instructions as it sees fit.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/vec3f.h
+++ b/kernel/arch/dreamcast/include/dc/vec3f.h
@@ -23,6 +23,10 @@
 #include <sys/cdefs.h>
 __BEGIN_DECLS
 
+#include <math.h>
+
+#include <dc/fmath.h>
+
 /** \addtogroup math_matrices
     @{
 */
@@ -37,39 +41,24 @@ typedef struct vec3f {
 #define R_RAD 10430.37835f
 /* \endcond */
 
-/** \brief  Macro to return the scalar dot product of two 3d vectors.
+static inline float vec_fipr(vec3f_t vec) {
+    return fipr_magnitude_sqr(vec.x, vec.y, vec.z, 0.0f);
+}
+
+/** \brief  Function to return the scalar dot product of two 3d vectors.
 
     This macro is an inline assembly operation using the SH4's fast
     (approximate) math instructions, and returns a single-precision
     floating-point value.
 
-    \param  x1               The X coordinate of first vector.
-    \param  y1               The Y coordinate of first vector.
-    \param  z1               The Z coordinate of first vector.
-    \param  x2               The X coordinate of second vector.
-    \param  y2               The Y coordinate of second vector.
-    \param  z2               The Z coordinate of second vector.
-    \param  w                The result of the calculation.
+    \param  vec1             The first vector.
+    \param  vec2             The second vector.
+    \return                  The result of the calculation.
 */
-#define vec3f_dot(x1, y1, z1, x2, y2, z2, w) { \
-        register float __x __asm__("fr0") = (x1); \
-        register float __y __asm__("fr1") = (y1); \
-        register float __z __asm__("fr2") = (z1); \
-        register float __w __asm__("fr3"); \
-        register float __a __asm__("fr4") = (x2); \
-        register float __b __asm__("fr5") = (y2); \
-        register float __c __asm__("fr6") = (z2); \
-        register float __d __asm__("fr7"); \
-        __asm__ __volatile__( \
-                              "fldi0 fr3\n" \
-                              "fldi0 fr7\n" \
-                              "fipr    fv4,fv0" \
-                              : "+f" (__w) \
-                              : "f" (__x), "f" (__y), "f" (__z), "f" (__w), \
-                              "f" (__a), "f" (__b), "f" (__c), "f" (__d) \
-                            ); \
-        w = __w; \
-    }
+static inline float vec_dot(vec3f_t vec1, vec3f_t vec2) {
+    return fipr(vec1.x, vec1.y, vec1.z, 0.0f,
+                vec2.x, vec2.y, vec2.z, 0.0f);
+}
 
 /** \brief  Macro to return scalar Euclidean length of a 3d vector.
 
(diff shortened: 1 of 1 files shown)


darcagn:
Fixes #868.

jnmartin84:
Just wanted to comment. I am using this in Doom 64 with no performance regression. It is working well. I've bundled it into a customized KOS for Doom 64 and look forward to the day it gets merged and I'm one step closer to being able to run on `master` again.

gyrovorbis:
SHIT, HOLD THE PHONE.

This is about to break the shit out of C++, because it doesn't have support for compound literal syntax (`(vec3f){ 0.0f, 1.0f, 2.0f }`).

What I have had to do in `libGimbal` to be able to do temporary struct initialization like this across both C and C++ is to devise a macro to be used for such scenarios, defined differently based on whether it's being compiled as C or C++:

```
#ifdef __cplusplus
#    define STRUCT_INIT(name, ...) (name(__VA_ARGS__))
#else
#    define STRUCT_INIT(name, ...) ((name){ __VA_ARGS__ })
#endif
```

Then all of the `static inline` functions which use compound literals must be changed to use this macro. For example:
```
static inline vec3f_t vec_sub_normalize(vec3f_t vec1, vec3f_t vec2) {
    vec3f_t vec = (vec3f_t){ vec1.x - vec2.x, vec1.y - vec2.y, vec1.z - vec2.z };

    return vec_normalize(vec);
}
```
would become:
```
static inline vec3f_t vec_sub_normalize(vec3f_t vec1, vec3f_t vec2) {
    vec3f_t vec = STRUCT_INIT(vec3f_t, vec1.x - vec2.x, vec1.y - vec2.y, vec1.z - vec2.z);

    return vec_normalize(vec);
}
```

gyrovorbis:
If we're going to do this, we should probably put a big warning in the Doxygen for this header file that it's going to run like ass compared to the inline ASM versions when the proper `-ffast-math -mfsca -mfsrra` flags are not specified--which they are not by default...

gyrovorbis:
If we're going to have a generalized C-based fast vector3f API now... perhaps we should add a cross-product implementation as well for completeness?

pcercuei:
@gyrovorbis time to study your C++, it totally works ;)
https://godbolt.org/z/7hvEGYaEc

> If we're going to do this, we should probably put a big warning in the Doxygen for this header file that it's going to run like ass compared to the inline ASM versions when the proper `-ffast-math -mfsca -mfsrra` flags are not specified--which they are not by default...

I'd prefer to have `-mfsca -mfsrra` unconditionally enabled in the environ.sh, and the `-ffast-math` enabled by default in the environ.sh, with the comment that it can be disabled if needed.

gyrovorbis:
> @gyrovorbis time to study your C++, it totally works ;) https://godbolt.org/z/7hvEGYaEc
False. It's working due to GNU extensions being enabled by default. Add `-pedentic` and enjoy these warnings:
```
<source>: In function 'float vec_distance(vec3f_t, vec3f_t)':
<source>:144:80: warning: ISO C++ forbids compound-literals [-Wpedantic]
  144 |     vec3f_t vec = (vec3f_t){ vec2.x - vec1.x, vec2.y - vec1.y, vec2.z - vec1.z };
      |                                                                                ^
<source>: In function 'vec3f_t vec_normalize(vec3f_t)':
<source>:159:70: warning: ISO C++ forbids compound-literals [-Wpedantic]
  159 |     return (vec3f_t){ vec.x * factor, vec.y * factor, vec.z * factor };
      |                                                                      ^
<source>: In function 'vec3f_t vec_sub_normalize(vec3f_t, vec3f_t)':
<source>:173:80: warning: ISO C++ forbids compound-literals [-Wpedantic]
  173 |     vec3f_t vec = (vec3f_t){ vec1.x - vec2.x, vec1.y - vec2.y, vec1.z - vec2.z };
      |                                                                                ^
<source>: In function 'vec3f_t vec_rotr_xy(vec3f_t, vec3f_t, float)':
<source>:195:5: warning: ISO C++ forbids compound-literals [-Wpedantic]
  195 |     };
      |     ^
 ```
...but... I suppose KOS itself is built around GNU extensions to begin with, so... maybe this isn't such a big deal...

> > If we're going to do this, we should probably put a big warning in the Doxygen for this header file that it's going to run like ass compared to the inline ASM versions when the proper `-ffast-math -mfsca -mfsrra` flags are not specified--which they are not by default...
> 
> I'd prefer to have `-mfsca -mfsrra` unconditionally enabled in the environ.sh, and the `-ffast-math` enabled by default in the environ.sh, with the comment that it can be disabled if needed.
Ah, yeah, that also works for me.

**Comment by pcercuei** [comment]
> > @gyrovorbis time to study your C++, it totally works ;) https://godbolt.org/z/7hvEGYaEc
> 
> False. It's working due to GNU extensions being enabled by default. Add `-pedentic` and enjoy these warnings:

Well, if you really want to go that route... If you add '-pedantic' it will also complain about the `({ })` construct in macros, of which we have plenty in public headers; which means that we indirectly rely on all the apps using KOS accepting "unclean" C/C++.

As nobody seems to have been bothered by that in the past, I think we can continue using GNU C/C++, especially if it does help us writing better and cleaner code.

**Comment by gyrovorbis** [comment]
> > > @gyrovorbis time to study your C++, it totally works ;) https://godbolt.org/z/7hvEGYaEc
> > 
> > 
> > False. It's working due to GNU extensions being enabled by default. Add `-pedentic` and enjoy these warnings:
> 
> Well, if you really want to go that route... If you add '-pedantic' it will also complain about the `({ })` construct in macros, of which we have plenty in public headers; which means that we indirectly rely on all the apps using KOS accepting "unclean" C/C++.
> 
Yep, and that's a good point. Tbh, all that I remembered was that C++ doesn't support compound literals. I actually forgot that they were supported via the GNU extensions we already relied on, so... yeah, it's fine. Point taken. 

> As nobody seems to have been bothered by that in the past, I think we can continue using GNU C/C++, especially if it does help us writing better and cleaner code.

Yeah, fair.

File: kernel/arch/dreamcast/include/dc/vec3f.h [hunk 1]
```diff
@@ -37,408 +41,227 @@ typedef struct vec3f {
...
-                              "fmul  fr3, fr2\n" \
-                              : "=f" (__x), "=f" (__y), "=f" (__z) \
-                              : "0" (__x), "1" (__y), "2" (__z) \
-                              : "fr3" ); \
-        x = __x; y = __y; z = __z; \
-    }
-
-/** \brief  Macro to return the normalized version of a vector minus another
+static inline vec3f_t vec_normalize(vec3f_t vec) {
+    float factor = 1.0f / vec_length(vec);
```
gyrovorbis, line 101:
```suggestion
    float length = vec_length(vec);
    float factor = 1.0f / sqrtf(length * length);
```
If we want this shit to be as fast as possible, shouldn't we be using the fast division approximations?

kernel/arch/dreamcast/include/dc/vec3f.h
```diff
@@ -37,408 +41,227 @@ typedef struct vec3f {
...
-        py = __cy; pz = __cz; \
-    }
+static inline vec3f_t vec_rotd_yz(vec3f_t vec, vec3f_t origin, float angle) {
+    return vec_rotr_yz(vec, origin, angle * R_DEG / R_RAD);
+}
+
+/** \cond */
+/* Compatibility macros */
+#define vec3f_dot(x1, y1, z1, x2, y2, z2, w) \
+    w = vec_dot((vec3f_t){ x1, y1, z1 }, (vec3f_t){ x2, y2, z2 })
```
gyrovorbis, line 223:
```suggestion
    (w = vec_dot((vec3f_t){ x1, y1, z1 }, (vec3f_t){ x2, y2, z2 }))
```
If we add an extra set of parenthesis around it, we can use it as a statement expression in conditionals and shit without side-effects or compiler warning about lack of parenthesis.

kernel/arch/dreamcast/include/dc/vec3f.h
```diff
@@ -37,408 +41,227 @@ typedef struct vec3f {
...
+
+#define vec3f_length(x, y, z, w) \
+    w = vec_length((vec3f_t){ x, y, z })
+
+#define vec3f_distance(x1, y1, z1, x2, y2, z2, w) \
+    w = vec_distance((vec3f_t){ x1, y1, z1 }, (vec3f_t){ x2, y2, z2 })
+
+#define vec3f_normalize(__x, __y, __z) { \
```
gyrovorbis, line 231: All of these multiline macros that don't need to be statement expressions with return values should be in the form of `do { /* ... */ } while(0)` to avoid side effects.

For example, the way ti is now, this is not going to work as intended:
```
if(something) 
   vec3f_normalize(x, y, z);
else
    printf("else");
 ```
 The inclusion of the semicolon after the block-based macros is going to create TWO statements, breaking the ability to use it without side-effects in conditionals and branches in such a form.

kernel/arch/dreamcast/include/dc/vec3f.h
pcercuei: Your suggestion is *not* equivalent.
And with the current code it is already using `fsrra`.

kernel/arch/dreamcast/include/dc/vec3f.h
pcercuei: Except that you woudn't use that macro in any way that hasn't been done before.
Because it's a *deprecated* macro, and you should be using the `vec_dot` function instead.

kernel/arch/dreamcast/include/dc/vec3f.h
pcercuei: The old `vec3f_normalize` was wrapped in braces, the new `vec3f_normalize` is wrapped in braces.

Any new code can use the `vec_normalize` function instead.

kernel/arch/dreamcast/include/dc/vec3f.h
pcercuei: Ok, reading it again I guess it's equivalent.
However, what you're missing is that `vec_length()` already contains a `sqrtf()`, so the whole thing is converted to a FSRRA already.

gyrovorbis: Ahhhh! Yeah, you're right, I guess it should be smart enough to see the `sqrtf()` through the inline call, huh...

kernel/arch/dreamcast/include/dc/vec3f.h
gyrovorbis: I guess. It's true, these macros were written like shit previously, but I figured we might as well fix them to not suck while we were at it... But... I guess it's not a big deal since they're deprecated.  Ah. I see you also aren't even including them in the Doxygen too... So that's good.

gyrovorbis:
Alright, pretty much satisfied here... spent a lot of time obsessing over it. lol.

I guess you should probably comment this line in by default then to finish this PR up, so perf isn't shitty by default: https://github.com/KallistiOS/KallistiOS/blob/515eaeef6a3823b17d50c037fd74099e055dcb81/doc/environ.sh.sample#L184.

You still don't think it's worth a little `\note` under the main header description for the Doxygen that you're screwing yourself using these macros without fast math?

pcercuei:
> Alright, pretty much satisfied here... spent a lot of time obsessing over it. lol.
> 
> I guess you should probably comment this line in by default then to finish this PR up, so perf isn't shitty by default:
> 
> https://github.com/KallistiOS/KallistiOS/blob/515eaeef6a3823b17d50c037fd74099e055dcb81/doc/environ.sh.sample#L184

Yes, I was thinking of doing it in a follow-up PR.

> You still don't think it's worth a little `\note` under the main header description for the Doxygen that you're screwing yourself using these macros without fast math?

I don't know. The way I see it, if `-ffast-math` is the default in the environ.sh file, and we provide a way to disable it with a fat warning above that says "beware, it might make things slow, disable only if really needed", then if people still disable it, that means they're okay with the consequences.

gyrovorbis:
> > Alright, pretty much satisfied here... spent a lot of time obsessing over it. lol.
> > I guess you should probably comment this line in by default then to finish this PR up, so perf isn't shitty by default:
> > https://github.com/KallistiOS/KallistiOS/blob/515eaeef6a3823b17d50c037fd74099e055dcb81/doc/environ.sh.sample#L184
> 
> Yes, I was thinking of doing it in a follow-up PR.
> 
> > You still don't think it's worth a little `\note` under the main header description for the Doxygen that you're screwing yourself using these macros without fast math?
> 
> I don't know. The way I see it, if `-ffast-math` is the default in the environ.sh file, and we provide a way to disable it with a fat warning above that says "beware, it might make things slow, disable only if really needed", then if people still disable it, that means they're okay with the consequences.

k, agreed. I'm satisfied with the warning being in `environ(_base).sh`.

gyrovorbis (approved):
Approved, just don't forget, we REALLY should have this with it: https://github.com/KallistiOS/KallistiOS/pull/887.

PR #871, opened 2024-12-19
DMA Controller API

pcercuei:
Introduce an API to use the DMA Controller of the SH4.
This new API can be used to program DMA transfers between memory and devices, or from memory to memory.

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
@@ -11,9 +11,9 @@
 
 #include <stdio.h>
 #include <errno.h>
+#include <arch/dmac.h>
 #include <dc/pvr.h>
 #include <dc/asic.h>
-#include <dc/dmac.h>
 #include <dc/sq.h>
 #include <kos/thread.h>
 #include <kos/sem.h>
@@ -42,7 +42,7 @@ static void pvr_dma_irq_hnd(uint32_t code, void *data) {
     (void)code;
     (void)data;
 
-    if(DMAC_DMATCR2 != 0)
+    if(dma_transfer_get_remaining(DMA_CHANNEL_2) != 0)
         dbglog(DBG_INFO, "pvr_dma: The dma did not complete successfully\n");
 
     /* Call the callback, if any. */
@@ -95,10 +95,18 @@ static uintptr_t pvr_dest_addr(uintptr_t dest, pvr_dma_type_t type) {
     return dest_addr;
 }
 
+static const dma_config_t pvr_dma_config = {
+    .channel = DMA_CHANNEL_2,
+    .request = DMA_REQUEST_EXTERNAL_MEM_TO_DEV,
+    .unit_size = DMA_UNITSIZE_32BYTE,
+    .src_mode = DMA_ADDRMODE_INCREMENT,
+    .transmit_mode = DMA_TRANSMITMODE_BURST,
+};
+
 int pvr_dma_transfer(const void *src, uintptr_t dest, size_t count,
                      pvr_dma_type_t type, bool block,
                      pvr_dma_callback_t callback, void *cbdata) {
-    uintptr_t src_addr = ((uintptr_t)src);
+    dma_addr_t src_addr = hw_to_dma_addr((uintptr_t)src);
 
     /* Check for 32-byte alignment */
     if(src_addr & 0x1F) {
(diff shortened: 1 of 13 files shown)


kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
```diff
@@ -118,21 +126,8 @@ int pvr_dma_transfer(const void *src, uintptr_t dest, size_t count,
...
-        DMAC_CHCR2 &= ~0x2;
-
-    DMAC_SAR2 = src_addr;
-    DMAC_DMATCR2 = count / 32;
-    DMAC_CHCR2 = 0x12c1;
-
-    if((DMAC_DMAOR & DMAOR_STATUS_MASK) != DMAOR_NORMAL_OPERATION) {
-        dbglog(DBG_ERROR, "pvr_dma: Failed DMAOR check\n");
-        errno = EIO;
+    if(dma_transfer(&pvr_dma_config, 0, src_addr, count, NULL))
```
DC-SWAT, line 129: I guess it is probably better to remove direct access to DMAC here at all and use only the Holly registers for both PVR/TA DMAs as it do for all other DMA transfers on System Bus. And your implementation can be used for those things that the Holly does not allow and for SCIF/SCI or something.

pcercuei: I'm just converting the driver. It used DMAC before, it uses DMAC now.

DC-SWAT: > I'm just converting the driver. It used DMAC before, it uses DMAC now.

Yes, I understand, there are no complaints against you. I'm speaking in general.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -397,6 +402,9 @@ int irq_init(void) {
     /* Set a default FPU exception handler */
     irq_set_handler(EXC_FPU, irq_def_fpu, NULL);
 
+    /* Unmask DMA IRQs */
+    IPRC = 0x300;
```
DC-SWAT, line 406: Mask it back on kernel shutdown.

pcercuei: Done.

kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
pcercuei: I think writing the DMAC registers is actually required here, according to the docs.

From what I gather the DMAC channel 0 is configured in DDT mode, which means it's fully controlled externally by the Holly.

Channel 2 does not use that mode, so software still has to configure it. It is typically configured to use the "external DMA" mode where the DMA start/stop signals come from outside the CPU (from the Holly), and configured for "mem-to-dev" to target what's probably a FIFO in the Holly, which then redirects the stream to wherever it needs (VRAM, TA, palette etc.).

QuzarDC (changes requested):
Just a minor change for one of the sets of comments. Otherwise looks textbook to the sh4 docs.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -20,12 +20,37 @@
...
 #include <kos/library.h>
+#include <kos/regfield.h>
 
 /* Macros for accessing related registers. */
 #define TRA    ( *((volatile uint32_t *)(0xff000020)) ) /* TRAPA Exception Register */
 #define EXPEVT ( *((volatile uint32_t *)(0xff000024)) ) /* Exception Event Register */
 #define INTEVT ( *((volatile uint32_t *)(0xff000028)) ) /* Interrupt Event Register */
 
+#define IPRA   ( *((volatile uint16_t *)(0xffd00004)) ) /* Interrupt priority register A */
+#define IPRB   ( *((volatile uint16_t *)(0xffd00008)) ) /* Interrupt priority register A */
```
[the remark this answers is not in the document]
pcercuei: Ok, I reworked that and pushed (not force-pushed so the previous commits didn't change).

I added a new API to get/set the priority of the various IRQ sources, and updated the code to use that new API.

kernel/arch/dreamcast/include/arch/irq.h
```diff
@@ -486,6 +486,58 @@ static inline void __irq_scoped_cleanup(int *state) {
...
+/** \brief  Interrupt sources
+
+   Interrupt sources at the SH4 level.
+ */
+typedef enum irq_src {
+    IRQ_SRC_RTC,
+    IRQ_SRC_TMU2,
+    IRQ_SRC_TMU1,
+    IRQ_SRC_TMU0,
+    _IRQ_SRC_RESV,
```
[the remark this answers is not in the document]
pcercuei: No. Safety char, it corresponds to a reserved field.

pcercuei:
Updated again, because why not.

- Flush the source buffer in pvr_dma_transfer() and callers. The documentation never stated anywhere that the source buffer had to be fully flushed to RAM, and so the expectation was that manual cache handling wasn't necesary. Now the cache handling is done inside pvr_dma_transfer().
- Add a DMA "speedtest" example.

DC-SWAT:
> * Add a DMA "speedtest" example.

Nice! Could you add a couple of cycles with small transmits to better understand the difference between SQs and DMA.

QuzarDC (approved):

PR #874, opened 2024-12-26
PVR Function Prototypes Cleanup (pvr.h refactor #539 part 1)

QuzarDC:
This implements the changes noted in item 2 of #539 as a set of atomic changes for reviewability.

@gyrovorbis Whenever you can a review of this would be appreciated. Some of the key differences:
1) There were a few places where you changed something to bool in the function prototype but did not follow that through to the uses of the param like `dma_blocking` or the internal functions around tile matrix initialization.
2) In many places you had switched types to `size_t` which didn't seem appropriate, like texture dimensions or the vbl count. To my understanding the point is to measure memory usage/size of objects. I can kind of see how that *might* apply for texture dimensions (width is a size of pixels, height is a size of lines) but it doesn't seem to me to improve clarity, so I've left most of those out.

### DIFF
--- a/examples/dreamcast/pvr/cheap_shadow/shadow.c
+++ b/examples/dreamcast/pvr/cheap_shadow/shadow.c
@@ -96,11 +96,11 @@ int check_start(void) {
 
         if(state->joyy > 64) {
             shadow = CLAMP(0.0f, 1.0f, shadow + 0.01f);
-            pvr_set_shadow_scale(1, shadow);
+            pvr_set_shadow_scale(true, shadow);
         }
         else if(state->joyy < -64) {
             shadow = CLAMP(0.0f, 1.0f, shadow - 0.01f);
-            pvr_set_shadow_scale(1, shadow);
+            pvr_set_shadow_scale(true, shadow);
         }
 
         if(state->buttons & CONT_START)
@@ -212,7 +212,7 @@ int main(int argc, char *argv[]) {
 
     /* Enable Cheap Shadow mode and set affected polygons to be at half of their
        normal intensity. */
-    pvr_set_shadow_scale(1, shadow);
+    pvr_set_shadow_scale(true, shadow);
 
     setup();
 
--- a/examples/dreamcast/pvr/modifier_volume_zclip/example.c
+++ b/examples/dreamcast/pvr/modifier_volume_zclip/example.c
@@ -252,7 +252,7 @@ int main(int argc, char* argv[])
     pvr_init(&params);
     pvr_set_bg_color(0, 0.5f, 1.0f);
     /* Enable cheap shadow */
-    pvr_set_shadow_scale(1, 0.5f);
+    pvr_set_shadow_scale(true, 0.5f);
 
     /* init plane */
     box_tex = pvr_mem_malloc(256 * 256 * 2);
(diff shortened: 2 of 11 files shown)


darcagn:
I believe `size_t` would be the correct thing to use here. To quote from Jens Gustedt's _[Modern C](https://inria.hal.science/hal-02383654v2/file/modernC.pdf)_:

> Use `size_t` for sizes, cardinalities, or ordinal numbers. Use `unsigned` for small quantities that can’t be negative.

To reply to the related comment from the other PR:

> `size_t` to my understanding, isn't typically used for something like 'dimensions' despite that being a kind of size, but rather counts of bytes in a piece of data or elements in an array. I'm not sure it makes more sense here than `uint16_t` (since 32bit is bigger than the biggest size texture we could make).

I would use `uint16_t` or `uint32_t` only if there is something inherently 16-bit or 32-bit about the value. That isn't the case here, therefore, I do believe switching to `size_t` improves clarity. 

> As a completely different comment presuming that the first is wrong and `size_t` is the way to go here ... shouldn't big swaths of the API be adjusted to use it then for representations of texture dimensions?

I believe that would be the proper thing to do.

> Another bit of thought is that I'm not sure that `size_t`s typically get multiplied by each other, do they?

They do, as long as the result value can be expected to fit within `SIZE_MAX`, which it would here. For example, _Modern C_ introduces basic C `+`, `-`, and `*` arithmetic using `size_t` as the type.

pcercuei (approved):

QuzarDC:
> I believe `size_t` would be the correct thing to use here. To quote from Jens Gustedt's _[Modern C](https://inria.hal.science/hal-02383654v2/file/modernC.pdf)_:
> 
> > Use `size_t` for sizes, cardinalities, or ordinal numbers. Use `unsigned` for small quantities that can’t be negative.
> 
> To reply to the related comment from the other PR:
> 
> > `size_t` to my understanding, isn't typically used for something like 'dimensions' despite that being a kind of size, but rather counts of bytes in a piece of data or elements in an array. I'm not sure it makes more sense here than `uint16_t` (since 32bit is bigger than the biggest size texture we could make).
> 
> I would use `uint16_t` or `uint32_t` only if there is something inherently 16-bit or 32-bit about the value. That isn't the case here, therefore, I do believe switching to `size_t` improves clarity.
> 
> > As a completely different comment presuming that the first is wrong and `size_t` is the way to go here ... shouldn't big swaths of the API be adjusted to use it then for representations of texture dimensions?
> 
> I believe that would be the proper thing to do.
> 
> > Another bit of thought is that I'm not sure that `size_t`s typically get multiplied by each other, do they?
> 
> They do, as long as the result value can be expected to fit within `SIZE_MAX`, which it would here. For example, _Modern C_ introduces basic C `+`, `-`, and `*` arithmetic using `size_t` as the type.

I had written a longer response last week but it seems like github just ate it or something. Spent some time with Modern C and it's description of the use is quite compelling and certainly something I think that has merit (despite the author seemingly imbuing `size_t` with 'miraculous' attributes). But that doesn't seem to be the methodology that was being attempted here. There wasn't any effort to update all sorts of other sized uint vars into `size_t`, but just the ones that had to do with a 'size', so I'd stand behind it as a kind of unreasonable stretch of the narrower definition of size_t rather than an embrace of it as the new generic var.

That said, it's also simply not how we have anything else in the codebase, and we're still plodding through the update to stdint types generally. Once we move through that it seems like it would be a great idea to try to work through the codebase distinguishing between values defined by their specific bitsize (esp those that derive from hardware features) and general numbers as just `size_t`.

PS - Thanks for introducing me to 'Modern C', it's quite interesting :)

PR #875, opened 2024-12-31
fs_vmu: Improve handling of VMU file headers

pcercuei:
- Add function fs_vmu_set_header().
  This function can be used to set a specific header (which contains the
  metadata, icons...) to an opened VMU file, replacing the one it previously
  had (if any).

- Add function fs_vmu_set_default_header().
  This function will set a default header, that will be used for new files
  which were not assignated their own header.

The header is now skipped when opening a VMU file (that has a header),
which is a breaking change from the previous behaviour of KallistiOS. It
is however kept internally, which means that a file opened read-write
will reuse its header (unless overwritten by fs_vmu_set_header()).

The point of this change is to hide the header from the application
completely, which can now read/write the VMU files with a proper header
as if they were PC files. It is then much easier for PC ports to
read/write savestates that look good.

A VMU file can still be opened as raw by specifying the O_META flag in the fs_open() call.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/fs_vmu.h
+++ b/kernel/arch/dreamcast/include/dc/fs_vmu.h
@@ -34,6 +34,7 @@
 __BEGIN_DECLS
 
 #include <kos/fs.h>
+#include <dc/vmu_pkg.h>
 
 /** \defgroup vfs_vmu   VMU
     \brief              VFS driver for accessing Visual Memory Unit storage
@@ -46,7 +47,65 @@ __BEGIN_DECLS
 /* Initialization */
 int fs_vmu_init(void);
 int fs_vmu_shutdown(void);
+
+#define IOCTL_VMU_SET_HDR     0x564d5530 /* "VMU0" */
 /* \endcond */
+
+/** \brief  Set a header to an opened VMU file
+
+    This function can be used to set a specific header (which contains the
+    metadata, icons...) to an opened VMU file, replacing the one it previously
+    had (if any).
+
+    Note that the "pkg" pointer as well as the eyecatch/icon data pointers it
+    contain can be freed (if dynamically allocated) as soon as this function
+    return, as the filesystem code will keep an internal copy.
+
+    It is valid to pass NULL as the header pointer, in which case the header
+    for the file will be discarded.
+
+    \param fd               A file descriptor corresponding to the VMU file
+    \param pkg              The header to set to the VMU file
+    \retval 0               On success.
+    \retval -1              On error.
+*/
+static inline int fs_vmu_set_header(file_t fd, const vmu_pkg_t *pkg) {
+    return fs_ioctl(fd, IOCTL_VMU_SET_HDR, pkg);
+}
+
+/** \brief  Set a default header for newly created VMU files
+
+    This function will set a default header, that will be used for new files
+    which were not assignated their own header.
+
+    Note that files which already have a header, either because they were opened
+    read-write or because they were given one, will use their own header instead
+    of the default one.
+
+    Note that the "pkg" pointer as well as the eyecatch/icon data pointers it
+    contain can be freed (if dynamically allocated) as soon as this function
+    return, as the filesystem code will keep an internal copy.
+
+    It is valid to pass NULL as the header pointer, in which case the default
+    header will be discarded.
+
+    \param pkg              The header to set to the VMU file
+    \retval 0               On success.
+    \retval -1              On error.
+*/
+static inline int fs_vmu_set_default_header(const vmu_pkg_t *pkg) {
+    file_t fd;
+    int ret;
+
+    fd = fs_open("/vmu", O_RDONLY | O_DIR);
+    if(!fd)
+        return -1;
+
+    ret = fs_vmu_set_header(fd, pkg);
+    fs_close(fd);
+
+    return ret;
+}
 
 /** @} */
 
--- a/kernel/arch/dreamcast/include/dc/vmu_pkg.h
+++ b/kernel/arch/dreamcast/include/dc/vmu_pkg.h
@@ -115,11 +115,12 @@ int vmu_pkg_build(vmu_pkg_t *src, uint8_t ** dst, int * dst_size);
     files read in.
 
     \param  data            The buffer to parse.
+    \param  data_size       The size of the buffer, in bytes.
     \param  pkg             Where to store the vmu_pkg_t.
     \retval -1              On invalid CRC in the data.
     \retval 0               On success.
 */
-int vmu_pkg_parse(uint8_t *data, vmu_pkg_t *pkg);
+int vmu_pkg_parse(uint8_t *data, size_t data_size, vmu_pkg_t *pkg);
 
 /** \brief   Load a .ico file to use as a VMU file's icon.
     \ingroup vmu_package
--- a/kernel/arch/dreamcast/include/dc/vmu_pkg.h
+++ b/kernel/arch/dreamcast/include/dc/vmu_pkg.h
@@ -53,7 +53,7 @@ typedef struct vmu_pkg {
     int             data_len;           /**< \brief Number of data (payload) bytes */
     uint16_t        icon_pal[16];       /**< \brief Icon palette (ARGB4444) */
     uint8_t         *icon_data;         /**< \brief 512*n bytes of icon data */
-    const uint8_t   *eyecatch_data;     /**< \brief Eyecatch data */
+    uint8_t         *eyecatch_data;     /**< \brief Eyecatch data */
     const uint8_t   *data;              /**< \brief Payload data */
 } vmu_pkg_t;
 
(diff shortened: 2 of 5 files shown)


kernel/arch/dreamcast/include/dc/vmu_pkg.h
```diff
@@ -115,11 +115,12 @@ int vmu_pkg_build(vmu_pkg_t *src, uint8 ** dst, int * dst_size);
     files read in.
 
     \param  data            The buffer to parse.
+    \param  data_size       The size of the buffer, in bytes.
     \param  pkg             Where to store the vmu_pkg_t.
     \retval -1              On invalid CRC in the data.
     \retval 0               On success.
 */
-int vmu_pkg_parse(uint8 *data, vmu_pkg_t *pkg);
+int vmu_pkg_parse(uint8 *data, int data_size, vmu_pkg_t *pkg);
```
gyrovorbis, line 123:
```suggestion
int vmu_pkg_parse(uint8 *data, size_t data_size, vmu_pkg_t *pkg);
```

kernel/arch/dreamcast/util/vmu_pkg.c
```diff
@@ -111,7 +111,7 @@ int vmu_pkg_build(vmu_pkg_t *src, uint8 ** dst, int * dst_size) {
 
 /* Parse an array of uint8's (i.e. a VMU data file) into a
  * vmu_pkg_t package structure. */
-int vmu_pkg_parse(uint8 *data, vmu_pkg_t *pkg) {
+int vmu_pkg_parse(uint8 *data, int data_size, vmu_pkg_t *pkg) {
```
gyrovorbis, line 114:
```suggestion
int vmu_pkg_parse(uint8 *data, size_t data_size, vmu_pkg_t *pkg) {
```

kernel/arch/dreamcast/include/dc/fs_vmu.h
```diff
@@ -46,8 +47,66 @@ __BEGIN_DECLS
...
+    return, as the filesystem code will keep an internal copy.
+
+    It is valid to pass NULL as the header pointer, in which case the default
+    header will be discarded.
+
+    \param pkg              The header to set to the VMU file
+    \retval 0               On success.
+    \retval -1              On error.
+*/
+static inline int fs_vmu_set_default_header(const vmu_pkg_t *pkg) {
```
gyrovorbis, line 96: Do we really want to inline something as nontrivial as this? Gonna be a PITA for people like darc to do bindings to a lot of the API if we start inlining every small function like this, which we are already not doing anywhere else. 

I wouldn't complain, but this isn't exactly a performance-critical function either...

gyrovorbis (commented):
Still need to finish reviewing the gnitty-gritty of the calculations for like start offset and stuff when working with the `vmu_pkg_t`, and I need to make sure no assumptions have been made that will break with `GAME` type files, whose VMS headers are always fixed at block offset 1, but overall, I like it.

Although I really, really wish we didn't hardcode every assumption ever about VMUs, including everything from block size to locations of the FAT and directory... There's a WHOLE lot of blatantly ignoring the maple spec that this filesystem does, which makes it totally incapable of supporting custom VMU volumes which would otherwise totally be possible.

Oh well, it will be a humongous refactor that can come later to fix all of it, and I did already start it on a VMU branch last year that I pushed, using a bunch of code from ElysianVMU.

gyrovorbis:
So isn't the logic that is intentionally ignoring the VMS header and is jumping over it broken? First of all, a VMS header can be located at any arbitrary block offset within the file for `DATA` type files, and the offset into the header is provided as part of the directory entry.

Secondly how will this work with `GAME` files, whose VMS header is hardcoded as always starting at a 1 block offset within the file?

pcercuei:
> So isn't the logic that is intentionally ignoring the VMS header and is jumping over it broken? First of all, a VMS header can be located at any arbitrary block offset within the file for `DATA` type files, and the offset into the header is provided as part of the directory entry.
> 
> Secondly how will this work with `GAME` files, whose VMS header is hardcoded as always starting at a 1 block offset within the file?

I had no idea the header could be located at a different offset. I guess I'll need to install a game on one of my VMUs to test it there.

kernel/arch/dreamcast/include/dc/fs_vmu.h
darcagn: > Looks trivial to me.
> 
> @darcagn are static inlines a problem for binding?

When binding to static inline functions in Rust, there's no symbol to actually bind to, so I have to either
1. Write a C wrapper function that calls the inline C function, and have the cargo build script compile and link it into the bindings crate. e.g.:

`fs_vmu.rs`
```rust
extern "C" {
    #[link_name = "fs_vmu_set_default_header_wrapper"]
    pub fn fs_vmu_set_default_header(pkg: *const vmu_pkt_t) -> c_int;
}
```
`fs_vmu_h.c`
```c
#include <dc/fs_vmu.h>
int fs_vmu_set_default_header_wrapper(const vmu_pkg_t *pkg) {
    return fs_vmu_set_default_header(pkg);
}
```

or

2. Just rewrite the function as an inline fn in Rust, e.g.:

`fs_vmu.rs`
```rust
#[inline]
pub unsafe fn fs_vmu_set_default_header(pkg: *const vmu_pkg_t) -> c_int {
    let fd = fs_open("/vmu", O_RDONLY | O_DIR);
    if fd == 0 {
        return -1;
    }

    let ret = fs_vmu_set_header(fd, pkg);
    fs_close(fd);

    ret
}
```

The first solution incurs a function call penalty due to losing the inlining that the C version has, but I don't have to change anything if the implementation changes in the future; the second solution gets inlined but if the implementation changes, I have to rewrite the function on the Rust side. It doesn't really matter here but if this were a performance critical function I'd of course choose to rewrite it in Rust (unless the function includes inline ASM which I can't do in Rust at the moment since the `rustc` frontend doesn't know what SuperH is yet).

I have purposely avoided making any demands or requests regarding my Rust bindings as I don't want to impose any Rust stuff on a non-Rust project. For my efforts, making a wrapper or rewriting a small function (as inlined functions typically are) is no big deal. I'd much prefer an inlined function over a macro, as inlined functions have concrete inputs and outputs and are easy to reason about, while macros (whether it's a macro constant or a function-like macro) typically require extra effort to figure out the ways it might be used...

gyrovorbis: > I have purposely avoided making any demands or requests regarding my Rust bindings as I don't want to impose any Rust stuff on a non-Rust project. For my efforts, making a wrapper or rewriting a small function (as inlined functions typically are) is no big deal. I'd much prefer an inlined function over a macro, as inlined functions have concrete inputs and outputs and are easy to reason about, while macros (whether it's a macro constant or a function-like macro) typically require extra effort to figure out the ways it might be used...

I don't see caring about this kind of stuff as "imposing" anything on a project. One of C's biggest strengths, especially in a multi-language ecosystem like this is that it's the common language. This exact same issue is there for every other non C-based language which can bind to C as well. Also, without a real symbol, we can't export it for the dynamic libraries to load either, as part of the dynamic symbol table. 

I'm still all for inlining if it's something that's performance-critical that can benefit from it, but I just don't see any real benefit and am only seeing downsides to inlining just due to a function's definition being small.

gyrovorbis:
> > So isn't the logic that is intentionally ignoring the VMS header and is jumping over it broken? First of all, a VMS header can be located at any arbitrary block offset within the file for `DATA` type files, and the offset into the header is provided as part of the directory entry.
> > Secondly how will this work with `GAME` files, whose VMS header is hardcoded as always starting at a 1 block offset within the file?
> 
> I had no idea the header could be located at a different offset. I guess I'll need to install a game on one of my VMUs to test it there.

It's kind of even lamer than that... In my notes and research, there's a "header offset block" field in the Directory Entry structure of the VMU's Directory region, right... It's a `uint32_t` in size, which means in theory it could be at an arbitrary offset, but I have literally never in my entire life seen a `DATA` file that had a header anywhere other than the beginning at offset 0 or a `GAME` file with a header anywhere but offset 1. 

This is one of those things where like... Maple spec makes it sound like that header can be anywhere, but the reality may very well be that it's only ever at block offset 0/1 depending on `GAME`/`DATA` file type due to assumptions that the BIOS is making... Really unsure if it can ACTUALLY be at an arbitrary offset for sure or not.

kernel/arch/dreamcast/include/dc/fs_vmu.h
darcagn: > I don't see caring about this kind of stuff as "imposing" anything on a project.

Fair enough, I just wouldn't want anyone to hold back anything for Rust. 

On a side note, we could do something like what `dc/fmath.h` does with `__FMINLINE` and make it an option whether or not to generate symbols for inlined functions.

gyrovorbis (approved):
Not only does it look great, but funnily enough, I just needed LITERALLY EXACTLY THIS FUNCTIONALITY for adding VMU icons to GTA3 save files.

Think about how important it is to be able to add an icon to an existing standard file handle *at any point in time*. Without that, you have to completely rework how you access files and have to handle saving the header yourself first, manually.

PR #877, opened 2025-01-02
Convert some macros to inline functions and improve inline ASM

pcercuei:
This PR updates arch.h to use inline functions instead of macros, and improves the inline ASM of several functions.


pcercuei:
PR updated, with a commit that improves `dcache_alloc_block()`.

darcagn:
Tossing the macros out makes the API easier to bind to, so thanks. 👍🏼

QuzarDC (approved):

PR #880, opened 2025-01-03
dc-chain Updates

darcagn:
- Newlib 4.5.0 patch added
- Migration to Newlib 4.5.0 and Binutils 2.43.1 for profiles newer than stable
- Offsets adjusted when applicable for development toolchain patches
- Profile successful build dates updated
- GDB updated to 15.2 for all profiles

All toolchain profiles patch & build successfully on my Linux machine.


pcercuei (approved):
LGTM.

I noticed that the newlib patches does set `newlib_cflags`, I wonder if we can build Newlib with LTO there?

darcagn:
> LGTM.
> 
> I noticed that the newlib patches does set `newlib_cflags`, I wonder if we can build Newlib with LTO there?

Someone (Colton, I believe?) tried that but there was an issue with it that prevented us from sticking with it. I don't remember what it was, but it was probably 2023 or so and things have changed since then, so we should probably give it another shot.

darcagn:
I did try it just now: with `-flto` added Newlib and KOS build fine but when building examples there are a tremendous number of link failures. With `-flto -ffat-lto-objects` the majority of linker errors are solved but there are still some. We could probably get this to work out with a little effort.

PR #881, opened 2025-01-04
Remove dependence on minimal termios stub.

QuzarDC:
For #636 a stub of termios functionality was added in order to support isatty. 
This broke libjimtcl and would break other software that detected termios
compatibility as it implemented only the minimum needed for isatty.

This replaces the use of tcgetattr with a check of the kos vfs handler to test if it's pty. 
The stub termios header and tcgetattr is then removed along with the defines in
ioctl.h.

Finally, also a minor doc fix in sys/ioctl.h which was declaring itself as uio.h

This requires the rebuilding of the toolchain/newlib but may be avoided by simply deleting $KOS_CC_BASE/sh-elf/include/sys/termios.h. This corrects the build errors mentioned here: https://github.com/KallistiOS/kos-ports/issues/42#issuecomment-2212414405

The other proposed solution had been to flesh out the termios stub, which could still be undertaken in the future if anyone would want to take up the effort, but having the stub be so minimal was problematic. Tested against some general outputs to printf, the pty example, and testing isatty against different FDs and all seemed to operate as it should.


pcercuei (approved):
LGTM

PR #882, opened 2025-01-05
pvr_prim: Speed up header compilation

pcercuei:
Use the fact that the headers are always aligned to 32 bytes, and allocate a dirty cache line when writing the first 4 bytes. This avoids a possible cache miss that would have loaded the header's 32 bytes from RAM, which would have subsequently been overwritten.


gyrovorbis (approved):
Omg you clever bastard...

PR #883, opened 2025-01-06
IRQ Cleanup Part 3

QuzarDC:
After more of the back and forth with #618 , I found that I could separate out a few more bits of the updates from the larger functional updates. Included here are:
* Some updates to comments and cleanup.
* Swapping to use alignas rather than __attribute
* Updating the handler getters to allow getting the *data
* Adding a prototype in irq.h for irq_get_sr defined in entry.s
* Removing the forced timer clearing within the top level handler

### DIFF
--- a/kernel/arch/dreamcast/include/arch/irq.h
+++ b/kernel/arch/dreamcast/include/arch/irq.h
@@ -417,6 +417,16 @@ void irq_force_return(void);
 */
 typedef void (*irq_handler)(irq_t code, irq_context_t *context, void *data);
 
+
+/** The type of a full callback of an IRQ handler and userdata.
+
+    This type is used to set or get IRQ handlers and their data.
+*/
+typedef struct irq_cb {
+    irq_handler hdl;    /**< A pointer to a procedure to handle an exception. */
+    void       *data;   /**< A pointer that will be passed along to the callback. */
+} irq_cb_t;
+
 /** \defgroup irq_handlers_ind  Individual
     \brief                      API for managing individual IRQ handlers.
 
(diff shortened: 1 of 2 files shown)


pcercuei:
I don't like your functions returning pointers in parameters.

I'd suggest to make `struct irq_cb` public typedef'd to `irq_cb_t`, then your diff would be
```diff
-irq_handler irq_get_handler(irq_t code);
+irq_cb_t irq_get_handler(irq_t code);
```

It makes sense, because on SH4 (unlike on some other archs), scalars and structs that fit in 64 bits are returned in R0 + R1, so it's basically free, and avoids aliasing.

gyrovorbis:
> I don't like your functions returning pointers in parameters.
> 
> I'd suggest to make `struct irq_cb` public typedef'd to `irq_cb_t`, then your diff would be
> 
> ```diff
> -irq_handler irq_get_handler(irq_t code);
> +irq_cb_t irq_get_handler(irq_t code);
> ```
> 
> It makes sense, because on SH4 (unlike on some other archs), scalars and structs that fit in 64 bits are returned in R0 + R1, so it's basically free, and avoids aliasing.

Fair. I dig it. I'm just worried we might have some inconsistencies with how we're handling similar scenarios in like... maybe the TRAPA or ASIC event handler API? Unsure. But either way, good point.

QuzarDC:
Moved to draft to fix merge conflicts and address feedback.

QuzarDC:
Updated the accessors. Wasn't sure if it might also make sense to update the setters to take the same. Might save that for a secondary PR to harmonize trap and asic handler interfaces as well.

pcercuei (approved):
Looks good, thanks.

PR #884, opened 2025-01-07
irq: Shrink IRQ table size

pcercuei:
All exception codes were divisible by 32, except for the two soft exceptions. Since they do not represent any hardware value, modify these code so that they are also divisible by 32, and make them smaller than 0x800.

This means that all exception codes can now be represented in just 6 bits, and the IRQ table can be reduced from 0x100 entries to just 0x40 entries.

This saves 1.5 KiB, but more importantly, the table now fits in just 16 cache lines instead of 64 cache lines.


gyrovorbis (approved):
This is a FANTASTIC change. The sparsity of the IRQ handler table here has long annoyed me, since it's one of the biggest symbols we have left in KOS binaries... I just didn't notice everything other than the last two was divisible by 32. Good work.

QuzarDC (approved):

PR #885, opened 2025-01-07
Add looping support to snd_sfxmgr

nextgeniuspro:
Here is a proposal to add sound loop support to snd_sfxmgr. This allows to play sound effect repeatedly until it stopped manually. Very useful when you simulate sound behaviour like chains, static noise, etc

### DIFF
--- a/kernel/arch/dreamcast/exports-naomi.txt
+++ b/kernel/arch/dreamcast/exports-naomi.txt
@@ -116,10 +116,9 @@ snd_sfx_load
 snd_sfx_load_ex
 snd_sfx_load_fd
 snd_sfx_play
-snd_sfx_play_lp
 snd_sfx_stop_all
 snd_sfx_play_chn
-snd_sfx_play_chn_lp
+snd_sfx_play_ex
 snd_sfx_stop
 snd_sfx_chn_alloc
 snd_sfx_chn_free
--- a/kernel/arch/dreamcast/exports-naomi.txt
+++ b/kernel/arch/dreamcast/exports-naomi.txt
@@ -116,8 +116,10 @@ snd_sfx_load
 snd_sfx_load_ex
 snd_sfx_load_fd
 snd_sfx_play
+snd_sfx_play_lp
 snd_sfx_stop_all
 snd_sfx_play_chn
+snd_sfx_play_chn_lp
 snd_sfx_stop
 snd_sfx_chn_alloc
 snd_sfx_chn_free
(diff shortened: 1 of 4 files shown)


DC-SWAT:
Something tells me that we just need to add an argument instead.

nextgeniuspro:
> Something tells me that we just need to add an argument instead. But backward compatibility...

yes, since C doesn't have default values for function parameters it's not possible to make compatible version with this approach

DC-SWAT:
I would suggest looking a little further and adding a _ex function (or another) to which you can pass a structure with different parameters for channels and the channel itself, in order to replace snd_sfx_play_chn at the same time.
This way we can continue to expand functionality without constantly creating new functions.
And this functionality will most likely be needed. Because here there is still a lack of playback frequency, which sometimes needs to be changed on the fly too (motor sound, for example). It would be nice if you also added frequency to this structure.

nextgeniuspro:
> I would suggest looking a little further and adding a `_ex` function (or another) to which you can pass a structure with different parameters and the channel itself

I updated PR, maybe this version would be better?

DC-SWAT:
> > I would suggest looking a little further and adding a `_ex` function (or another) to which you can pass a structure with different parameters and the channel itself
> 
> I updated PR, maybe this version would be better?

I like it better this way. The only thing is maybe rename sfxplaydata to sfx_play_data and add the frequency right away.

nextgeniuspro:
> > > I would suggest looking a little further and adding a `_ex` function (or another) to which you can pass a structure with different parameters and the channel itself
> > 
> > 
> > I updated PR, maybe this version would be better?
> 
> I like it better this way. The only thing is maybe rename sfxplaydata to sfx_play_data and add the frequency right away.

yep, I had doubts for this too, I just looked how sfxhnd_t named and used the same pattern. Ok, will update soon

nextgeniuspro:
However, no, I can't just update this, I need to discuss one issue first. Currently, frequency is taken from sfxhnd_t which is snd_effect_t structure, from parameter rate. So, when effect loaded from .wav file current fruequency is already known. Do there any other circumstances when frequency need to be overriden?

DC-SWAT:
> However, no, I can't just update this, I need to discuss one issue first. Currently, frequency is taken from sfxhnd_t which is snd_effect_t structure, from parameter rate. So, when effect loaded from .wav file current fruequency is already known. Do there any other circumstances when frequency need to be overriden?

Yes, I already gave an example - this is the sound of the engine of car or plane. It changes depending on the rotation speed. And also it may be needed for slowmo effects (like drunk in GTA3).

nextgeniuspro:
Ah, I see. Ok, so if overriden value is passed as parameter (via structure) then it will use used, otherwise default value from effect, right? Will update PR with these fixes

DC-SWAT (approved):
Looks good to me.

DC-SWAT:
Intersects with #738

QuzarDC (approved):

Issue #244, opened 2023-06-06
Cannot build with `-flto` without `-fno-builtin`. 


gyrovorbis:
We just recently added the option for enabling the GCC builtin stdlib replacement functions from the user's `environ.sh`. Cursory testing all seemed fine. @andressbarajas has already cleared out the warnings that arose when building like that...

However, it seems @darcagn has just discovered that we get unresolved Newlib symbol errors when we enable `-flto` when commenting out `-fno-builtin` from within `environ.sh`. I was able to reproduce the issue, and here's but a sample from the first example failing to build:

```
/bin/ld: /opt/toolchains/dc/sh-elf-13.1.0/bin/../lib/gcc/sh-elf/13.1.0/../../../../sh-elf/lib/libc.a(libc_a-exit.o): in function `exit':
/opt/toolchains/dc/kos/utils/dc-chain/build-newlib-sh-elf-4.3.0.20230120/sh-elf/newlib/../../../newlib-4.3.0.20230120/newlib/libc/stdlib/exit.c:65: undefined reference to `__exit'
```

Probably not a very high priority, as we clearly state that the path isn't as well tested/supported and everything else seems to work fine, but just wanted to capture it in an issue.

PSVita-Dev:
Hello. I am currently porting the godot engine to the dreamcast. The same issue occurs at the linking step. Here are the flags/libs:

```
	env["CC"]="sh-elf-gcc"
	env["CXX"]="sh-elf-g++"
	env["LD"]="sh-elf-g++"
	env["AR"]="sh-elf-gcc-ar"
	env["RANLIB"]="sh-elf-gcc-ranlib"

	env.Append(CCFLAGS=['-Wl,','-fno-builtin'])
	
	kos_path = os.environ["KOS_BASE"]
	kos_arch = os.environ["KOS_ARCH"]
	
	env.Append(CPPPATH=[kos_path+"/../sh-elf/sh-elf/include/"])
	env.Append(LIBPATH=[kos_path+"/../sh-elf/sh-elf/lib", kos_path+"/lib/"+kos_arch])

	env.Append(CPPFLAGS=['-DSERVER_ENABLED', '-std=c++98', '-DKOS', '-DNO_THREADS', '-DNEED_LONG_INT', '-DLIBC_FILEIO_ENABLED', '-DNO_STATVFS'])
	env.Append(LDFLAGS=['-Wl,--start-group', '-Wl,--end-group'])
[...]

PR #887, opened 2025-01-08
environ.sh: Disable -fno-builtin, enable -ffast-math

pcercuei:
- Enable `-mfsrra` and `-mfsca` unconditionally in environ_dreamcast.sh, as those instructions will only be used alongside -ffast-math anyway;
- Disable -fno-builtin, as I do not see any reason to have it enabled by default;
- Enable -ffast-math by default, since everybody seems to be using it anyway.

### DIFF
--- a/doc/environ.sh.sample
+++ b/doc/environ.sh.sample
@@ -165,13 +165,9 @@ export KOS_CFLAGS="${KOS_CFLAGS} -fomit-frame-pointer"
 
 # GCC Builtin Functions
 #
-# Comment out this line to enable GCC to use its own builtin implementations of
-# certain standard library functions. Under certain conditions, this can allow
-# compiler-optimized implementations to replace standard function invocations.
-# The downside of this is that it COULD interfere with Newlib or KOS implementations
-# of these functions, and it has not been tested thoroughly to ensure compatibility.
-#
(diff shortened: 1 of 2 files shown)


QuzarDC:
So I'm 100% onboard with `-mfsrra` and `-mfsca`, and mostly for the `-ffast-math` but the builtins are a bit trickier.

I haven't seen significant discussion of the issues with using them since: https://github.com/KallistiOS/KallistiOS/issues/244 but it was mostly around all sorts of non-math stuff. Has that been retested against at least examples/ports? My understanding was that enabling them was messing with lto in strange ways.

pcercuei:
@QuzarDC I can confirm that all the KOS examples that do actually build on master also build just fine here, with or without LTO, and with `-fno-builtin` removed.

QuzarDC:
> @QuzarDC I can confirm that all the KOS examples that do actually build on master also build just fine here, with or without LTO, and with `-fno-builtin` removed.

Yea, and from the time of that review I have been working with lto and `no-builtins` removed to try to verify. I was able to trigger some new warnings but that's mostly because gcc now just has more info by swapping in its builtins for things (warnings like 'possible NULL for non-NULL param'). But everything generally looks fine :)

QuzarDC (approved):

PR #890, opened 2025-01-10
pvr: Fix dma_next_list() incorrectly unlocking mutex

pcercuei:
Only the first call of dma_next_list() is done by the calling thread; the function is registered as the DMA complete callback, so the following calls are done in interrupt context. In that case, the mutex should be unlocked differently by passing a pointer to the thread that locked it.

This was detected by defaulting all mutexes to error-checking.


QuzarDC (approved):
Good idea with the defaulting to error checking.

PR #891, opened 2025-01-10
mmu: Add support for creating static memory maps

pcercuei:
This PR cleans up a bit the MMU code, and adds two new functions:
- `mmu_init_basic()` which sets up the bare minimum to create static memory maps, without initializing the full dynamic mapping system;
- `mmu_page_map_static()` which can create static memory mappings.

This basically moves the static memory mapping facility I have in Bloom back into KOS.

### DIFF
--- a/kernel/arch/dreamcast/kernel/mmu.c
+++ b/kernel/arch/dreamcast/kernel/mmu.c
@@ -28,20 +28,11 @@ static volatile uint32 * const ttb = (uint32 *)(SH4_REG_MMU_TTB);
 static volatile uint32 * const tea = (uint32 *)(SH4_REG_MMU_TEA);
 static volatile uint32 * const mmucr = (uint32 *)(SH4_REG_MMU_CR);
 
-#define SET_PTEH(VA, ASID) \
-    do { *pteh = ((VA) & 0xfffffc00) | ((ASID) & 0xff); } while(0)
-
 #define BUILD_PTEH(VA, ASID) \
     ( ((VA) & 0xfffffc00) | ((ASID) & 0xff) )
 
-#define SET_PTEL(PA, V, SZ, PR, C, D, SH, WT) \
-    do { *ptel = ((PA) & 0x1ffffc00) | ((V) << 8) \
-                     | ( ((SZ) & 2) << 6 ) | ( ((SZ) & 1) << 4 ) \
-                     | ( (PR) << 5 ) \
-                     | ( (C) << 3 ) \
-                     | ( (D) << 2 ) \
-                     | ( (SH) << 1 ) \
-                     | ( (WT) << 0 ); } while(0)
+#define SET_PTEH(VA, ASID) \
+    do { *pteh = BUILD_PTEH(VA, ASID); } while(0)
 
 #define BUILD_PTEL(PA, V, SZ, PR, C, D, SH, WT) \
     ( ((PA) & 0x1ffffc00) | ((V) << 8) \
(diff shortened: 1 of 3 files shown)


kernel/arch/dreamcast/kernel/mmu.c
```diff
@@ -713,9 +713,49 @@ static void initial_page_write(irq_t source, irq_context_t *context, void *data)
...
+    SET_MMUCR(head, head, 1, 0, 0, 1);
+    mmu_ldtlb(0, virt, phys, page_size, page_prot, cached, 1, 0, 0);
+    SET_MMUCR(head - 1, 0, 1, 0, 0, 1);
+
+    tlb_nb_static++;
+
+    return 0;
+}
+
+void mmu_init_basic(void) {
```
QuzarDC, line 741: Could/should there be a matching `mmu_shutdown_basic` or perhaps some other sort of exposed way to do disable the mmu? Otherwise when using the basic functionality there's no way that I see to return to the previous state. Or is the notion that it just doesn't matter if the irq handlers get cleared even though they're not set and the regular `mmu_shutdown` should be used?

kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
```diff
@@ -28,12 +28,10 @@
 
 // Find the next list to DMA out. If we have none left to do, then do
 // nothing. Otherwise, start the DMA and chain back to us upon completion.
-static void dma_next_list(void *data) {
```
QuzarDC, line 31: Looks like this commit slipped in from a different PR.

kernel/arch/dreamcast/kernel/mmu.c
pcercuei: So my idea is that you would just call `mmu_shutdown`. But I can add a `mmu_shutdown_basic`, if it's not really useful right now, it might be useful later should we need to update the regular shutdown function in an incompatible way.

pcercuei:
Rebased (which automatically dropped the unrelated change in pvr_irq.c), and added a new commit that introduces `pvr_shutdown_basic()`.

kernel/arch/dreamcast/kernel/mmu.c
```diff
@@ -738,27 +778,17 @@ int mmu_init(void) {
...
-    /* Clear the ITLB */
-    mmu_reset_itlb();
+    mmu_init_basic();
+}
 
-    /* All done */
-    return 0;
+void mmu_shutdown_basic(void) {
+    /* Turn off MMU */
+    *mmucr = 0x00000204;
```
DC-SWAT, line 786: Maybe it's better to save the old value in `mmu_init()`?

pcercuei: No, because then you can't guarantee that the MMU has been disabled.

DC-SWAT: I ask because not all hardware states are turned off during a shutdown, some should return to the previous state, since there may be another bootloader above. But in this case, it is probably better to turn it off.

pcercuei: I mean I'm not against restoring a previous state, I can see the arguments for both cases. As you stated that it's probably better to turn it off, I'll leave it like that for now, and we can revisit it later if we need to.

pcercuei:
Rebased once again and added a doc  entry (other commits untouched).

QuzarDC (approved):

PR #893, opened 2025-01-11
Unhack fs dcload

QuzarDC:
Minor overhaul to `fs_dcload`. I tried to be very atomic with the commits and their messages. The core change here is to track opened dirs to remove a hack that was in place to try to detect dirs by their FD value. In the past, dcload was set up so that the FD being returned for dirs was a raw copy of the `DIR *` opened on the pc side. Unfortunately when moving to 64-bit systems that broke functionality and had to be changed to pass an arbitrary value (as noted in https://github.com/KallistiOS/dcload-serial/pull/27/ ). Using that kind of arbitrary value above 100 made it more likely for a clash of fds between a file and dir.

Now with the tracking of opened dirs we can also cache the path and dirent values which is important for ensuring the correct behavior of `readdir` which was previously always only returning based on the most recently opened dir rather than the one being passed to it, and which could only be done on one dir at a time due to a static `dirent`.

While at it, cleaned up some obsolete portions and added `rewinddir` which was supported in both dcloads for quite some time.

Some additional TODOs from this:
- With the `fd>100` hack removed, we should consider eventually adjusting the host-side to track all opened files to ensure there can't be a random conflict where the fd of a file == the fd of a dir + 1337.
- I think the dcload syscall can be adjusted slightly to not rely on the separate assembly and VA_ARGS macro to work. Not sure if there'd be too much benefit to that though.
- `dcl_dir_t` could likely be optimized better to not use as much space (or require fewer mallocs). I considered using a static buffer, but the size benefits seemed to be a wash unless you are opening an absolutely massive number of dirs which wouldn't be supported by a static buffer anyways.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_dcload.c
+++ b/kernel/arch/dreamcast/fs/fs_dcload.c
@@ -4,6 +4,7 @@
    Copyright (C) 2002 Andrew Kieschnick
    Copyright (C) 2004 Megan Potter
    Copyright (C) 2012 Lawrence Sebald
+   Copyright (C) 2025 Donald Haase
 
 */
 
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/fs/fs_dcload.c
```diff
@@ -146,21 +147,26 @@ void *dcload_open(vfs_handler_t * vfs, const char *fn, int mode) {
             return (void *)NULL;
         }
 
-        entry->hnd = hnd;
-        LIST_INSERT_HEAD(&dir_head, entry, fhlist);
-
         if(dcload_path)
             free(dcload_path);
```
pcercuei: You can change this one too while you're at it - no need to NULL-check the pointer, free(NULL) is properly defined.

pcercuei: Nevermind, I see you remove it in a following commit.

pcercuei (approved):
Reviewed and LGTM.

PR #894, opened 2025-01-13
Byteswap builtins

QuzarDC:
Replacing a a small amount of asm with gcc builtins. My understanding is that the worst case will be identical generation. When testing with just a constant value the builtin managed to just emit the transformed constant, which seemed quite a bit smarter.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/byteorder.h
+++ b/kernel/arch/dreamcast/include/arch/byteorder.h
@@ -49,11 +49,9 @@ __BEGIN_DECLS
                         or equivalent.
     \return             The swapped value.
 */
-#define arch_swap16(x) ({ \
-    uint16_t __x = (x); \
-    __asm__ __volatile__("swap.b %0, %0" : "=r" (__x) : "0" (__x)); \
-    __x; \
-})
+static inline uint16_t arch_swap16(uint16_t x) {
+    return __builtin_bswap16(x);
+}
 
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/include/arch/byteorder.h
```diff
@@ -49,11 +49,7 @@ __BEGIN_DECLS
                         or equivalent.
     \return             The swapped value.
 */
-#define arch_swap16(x) ({ \
-    uint16_t __x = (x); \
-    __asm__ __volatile__("swap.b %0, %0" : "=r" (__x) : "0" (__x)); \
-    __x; \
-})
+#define arch_swap16(x) __builtin_bswap16(x)
```
pcercuei, line 52: While you're at it... static inline? :)

pcercuei:
Not sure why you use `const uint16_t x` as the parameter, sounds like a C++ thing. `x` is not a pointer, what does `const` bring here?

Also, I don't see why you used `__always_inline` instead of a regular "inline".

QuzarDC:
> Not sure why you use `const uint16_t x` as the parameter, sounds like a C++ thing. `x` is not a pointer, what does `const` bring here?
> 
> Also, I don't see why you used `__always_inline` instead of a regular "inline".

Just a horrific show of slapdash cargo culting from me.

pcercuei (approved):

PR #895, opened 2025-01-13
Update changelogs and AUTHORS

darcagn:
As usual, I forgot to update the changelogs for my dc-chain updates 😛

Also, I updated the `AUTHORS` file with the years for contributors who have authored commits from 2025.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -27,7 +27,7 @@ Sam Steele: 2004
 Gil Megidish: 2002
 Florian Schulze: 2002
 Walter van Niftrik: 2005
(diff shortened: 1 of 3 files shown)


QuzarDC (approved):
I actually think the changelog and authors needs to shift into being updated this way, rather than in the individual PRs (in large part to avoid merge conflicts). Hadn't made a fuss about it, but hopefully with more frequent releases and a bit more discipline in commit messages it's less important to track things via CHANGELOG and the git log can be the source of these with only periodic updates if we want.

PR #896, opened 2025-01-14
Fix and improve performance counters

pcercuei:
The `perf_cntr_count()` function had a (very unlikely) bug, where a value read could be off by about 21.5 seconds. This would happen when the high part of the counter was read before and the low part of the counter was read after a 32-bit overflow.

This can be addressed easily by reading the high part twice and comparing the two reads. The overhead should be minimal as it only adds one extra register read and a comparison + a rarely taken branch, while removing the superfluous 16-bit mask.

The `perf_cntr_timer_ns()` was also updated, so that the counter value is not read before the `perf_cntr_timer_enabled()` call. As written in the commit's description, the argument used was bogus, and reading the counter as late as possible slightly reduces the latency incurred.

### DIFF
--- a/kernel/arch/dreamcast/kernel/perfctr.c
+++ b/kernel/arch/dreamcast/kernel/perfctr.c
@@ -106,12 +106,9 @@ void perf_cntr_timer_disable(void) {
 }
 
 uint64_t perf_cntr_timer_ns(void) {
-    /* Grab value first, before checking, to not record overhead. */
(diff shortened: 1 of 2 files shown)


gyrovorbis (approved):

QuzarDC (approved):
This looks good. The only thing I'd caution on is that since we don't have documentation we may find that those upper bits of `PMCTR_HIGH` get some stuff in them. But that would only really be discernible via exploration and testing so we'll find out soon enough :)

PR #897, opened 2025-01-15
Volatility fix for PVR DR API.

gyrovorbis:
1) Made the return-type for pvr_dr_target() a pointer to volatile
   memory, so that the compiler will be forced to actually do the
intended writes, rather than keeping values within registers, and so that it cannot simply optimize away all writes due to not seeing the values ever get read and used later in the program. 2) Made the return-type for pvr_dr_target() void*, since you are free to
   submit ANY 32-byte type through it, including polygon headers, pvr
sprites, and modifier volumes.
3) Improved documentation on pvr_dr_target() and pvr_dr_commit(). 4) Fixed warnings that were introduced due to not storing the return
   value from pvr_dr_target() within a pointer to volatile memory in
pvrmark_strips_direct example.
5) Quick update to pvrmark_strips_direct's T&L loop to not do an
   integer->float conversion for z coordinat every iteration (thanks
darc).
6) Fixed parameter type to pvr_send_to_ta() to be const-correct. 7) Updated AUTHORS with this year for my entry.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -43,7 +43,7 @@ Joe Fenton: 2016
 Stefan Galowicz: 2016, 2017
 Luke Benstead: 2020, 2021, 2022, 2023
 Eric Fradella: 2023, 2024, 2025
-Falco Girgis: 2023, 2024
+Falco Girgis: 2023, 2024, 2025
 Ruslan Rostovtsev: 2014, 2016, 2023, 2024
 Colton Pawielski: 2023
 Andy Barajas: 2023, 2024
--- a/examples/dreamcast/pvr/pvrmark_strips_direct/pvrmark_strips_direct.c
+++ b/examples/dreamcast/pvr/pvrmark_strips_direct/pvrmark_strips_direct.c
@@ -89,8 +89,9 @@ inline static void get_vert(int *seed, int *x, int *y, int *col) {
 }
 
 static void do_frame(void) {
-    pvr_vertex_t *vert;
-    int x=0, y=0, z=0, col=0;
+    volatile pvr_vertex_t *vert;
+    int x=0, y=0, col=0;
+    float z = 0.0f;
     int i;
     static int oldseed = 0xdeadbeef;
     int seed = oldseed;
--- a/kernel/arch/dreamcast/include/dc/pvr.h
+++ b/kernel/arch/dreamcast/include/dc/pvr.h
@@ -2023,6 +2023,32 @@ void pvr_dr_init(pvr_dr_state_t *vtx_buf_ptr);
 
 /** \brief   Obtain the target address for Direct Rendering.
 
+    This function returns a pointer which maps directly into the alternate
+    32-byte \ref store_queues "store queue" from the one which was mapped from
+    the previous call to pvr_dr_target(). This is where vertex information
+    should be written into for it later be submitted to the TA via
+    pvr_dr_commit().
+
+    The return value is of `volatile void*` type, so it should be casted to the
+    desired vertex structure or whatever type is meant to be written.
+
+    \warning
+    The store queues are WRITE-ONLY. Any attempt to read from this address will
+    return all 0s.
+
+    \note
+    Since each store queue is only 32-bytes large, while certain PVR formats
+    such as pvr_sprite_t are 64-bytes in size, they should be written in pairs,
+    with two calls to `pvr_dr_target()` and `pvr_dr_commit()`, one for each
+    half of the vertex data.
+
+    \warning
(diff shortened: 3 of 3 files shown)


QuzarDC:
Adding the example provided by in discord by Bruce
[PR-897 Test.zip](https://github.com/user-attachments/files/18429637/PR-897.Test.zip)
When building this with gcc 14.2.0 and with kos and it built with `-O3` optimization the result is either a blank screen or blue triangle. The intent is a blue quad.

From what I can tell this is fallout from this change: https://github.com/KallistiOS/KallistiOS/commit/19ce9ecad9136f36c1f3d170a0eedde22218eeb7

As in the special case of triggering an sq writeback I think should be signaling that memory is being changed by it.

Not 100% sure that's all correct, but I would think it would allow for optimization to be performed on the var otherwise (unlike volatile) but let the compiler understand that at the point where it gets to the `sq_flush`, we expect changes and that part can't be optimized away.

Changing this in `cache.h` fixed it for me:
```
static __always_inline void dcache_wback_sq(const void *src) {
    __asm__ __volatile__("pref @%0\n"
                         : /* No outputs */
                         : "r" (src)
                         : "memory"
    );
}

gyrovorbis:
> The various "compile" functions should probably be updated to take a volatile pointer as well, otherwise we'll have to cast the pointers to non-volatile every time.

Good call.



> Adding the example provided by in discord by Bruce [PR-897 Test.zip](https://github.com/user-attachments/files/18429637/PR-897.Test.zip) When building this with gcc 14.2.0 and with kos and it built with `-O3` optimization the result is either a blank screen or blue triangle. The intent is a blue quad.
> 
> From what I can tell this is fallout from this change: [19ce9ec](https://github.com/KallistiOS/KallistiOS/commit/19ce9ecad9136f36c1f3d170a0eedde22218eeb7)
> 
> As in the special case of triggering an sq writeback I think should be signaling that memory is being changed by it.
> 
> Not 100% sure that's all correct, but I would think it would allow for optimization to be performed on the var otherwise (unlike volatile) but let the compiler understand that at the point where it gets to the `sq_flush`, we expect changes and that part can't be optimized away.
> 
> Changing this in `cache.h` fixed it for me:
> 
> ```
> static __always_inline void dcache_wback_sq(const void *src) {
>     __asm__ __volatile__("pref @%0\n"
>                          : /* No outputs */
>                          : "r" (src)
>                          : "memory"
>     );
> }
> ```

NICE FIND! You are right, I guess not clobbering memory on a prefetch, when that prefetch is issued for an SQ flush, is the wrong thing to do... and the right thing is to have two prefetch versions, one that does clobber memory, one that doesn't, as you've done.

@pcercuei mentioned how to limit the clobber range of memory to just the SQ destination addresses, which we should do to be more specific.

But also... I fear you've found ANOTHER issue here. I don't see how my addition of `volatile` to the SQ-mapped return pointer cannot be warranted. What you fixed here, was ensuring the compiler understands that flushing a store queue might modify memory out from under it. What I have fixed is ensuring the compiler actually writes to the SQ to begin with... WIthout mine, the compiler could still keep values destined for SQs within registers and never actually bother commiting to memory to fill the SQs. OR it could see that there's zero reason to write to the SQ since you're never reading the value from it.

pcercuei:
I think if you mark the 32 bytes following the address pointed to by "src" as read and written by the ASM, the compiler will know that anything written to the SQ prior to that call needs to happen before the ASM is inlined (as the ASM would "read" it), and anything written before the ASM cannot be re-used after the call (as the ASM would "write" it). So it *should* work without using volatiles everywhere, without resorting on clobbering the whole memory which has performance implications.

QuzarDC:
> But also... I fear you've found ANOTHER issue here. I don't see how my addition of `volatile` to the SQ-mapped return pointer cannot be warranted. What you fixed here, was ensuring the compiler understands that flushing a store queue might modify memory out from under it. What I have fixed is ensuring the compiler actually writes to the SQ to begin with... WIthout mine, the compiler could still keep values destined for SQs within registers and never actually bother commiting to memory to fill the SQs. OR it could see that there's zero reason to write to the SQ since you're never reading the value from it.

Hrm. I get what you're saying, but there's no indication of that happening. If that were the case too, then it would seem that the way to go would be something like have `MEM_AREA_SQ_BASE` be typed as volatile or something like that in order to ensure that any usage of them is treated as such. I would think though that this would have been a long standing issue with SQs if that were the case. What it very much seems like though was that it was a regression from a recent PR.

gyrovorbis:
This solution was definitely correct, but... the alternative isn't API breaking, so we should do it like that.

PR #898, opened 2025-01-17
vmufb: Add vmufb_paint_xbm()

pcercuei:
This function can be used to display a X11 XBM image onto the VMU framebuffer.


StrikerTheHedgefox:
Works as it should, from what I can tell! (Using this in Doom64-DC)

gyrovorbis (approved):
Yeah, this is good to have, since the high-level, direct-drawing VMU API supports drawing these things, so existing tooling is out there for it... Makes sense to have it supported on our better VMU FB API, where you have more control. Approved.

QuzarDC (approved):

Issue #858, opened 2024-11-26
RFC: A common 1-bit font interface


QuzarDC:
It has occurred to me that we now have 3 different 1-bit font drawing systems (bfont, minifont, and vmufont). vmufont draws the 1-bit font to a 1-bit screen buffer, which is not supported by bfont or minifont, the other two use fixed-size fonts as their backing but repeat their drawing methods with different levels of overlap in functionality.

I think that it would make sense to move to a shared KOS-level (rather than arch-level) interface where we define a datatype for a 1-bit font based around minifont and vmufont's data, supporting subsets of ISO-8859-1 (as they currently do), and arbitrary character sizes. Then our use of the basic drawing functionalities that we show for the current 'bfont' would still just show how to draw with the 'default' font (bios font for DC, minifont for NAOMI). This would basically mean extracting the logic from `bfont_draw_one_row` and `bfont_draw_ex` and defining required spec data for a font.

It should be possible to do this with no API breakage by having the old font systems simply stay as wrappers around the new functionality which force the use of their specific type of font (and to leave in bfont, for now at least, the JISX-0201/JISX-0208 characters and the dreamcast-specific ones). This should also help give a good infrastructure to have good and simple solutions for #829 and #710 as well as immediately granting the minifont system the benefits of not being locked to a fixed color and only 16-bit output.

I am currently working on a PR to help address #710 using the basic concept of redirecting bfont to use minifont when on NAOMI and intend to do so in a way that should make the above easier if we decide to do it.

PR #902, opened 2025-01-20
Harmonizing minifont and bfont

QuzarDC:
It may look like a jumble of unrelated changes but this is meant to work towards #858 where we have a common 1-bit interface for fonts.

To do this it's taking steps to move minifont, vmufont, and bfont towards being more interoperable. The biggest gap between them is that minifont and vmufont both presume only 7-bit ASCII and bfont currently doesn't support this (in the sense of 'to the exclusion of further characters).

Putting this in mostly to not lose the work that I almost accidentally cleared out of my local repo.

### DIFF
--- a/examples/dreamcast/video/bfont/bfont.c
+++ b/examples/dreamcast/video/bfont/bfont.c
@@ -23,18 +23,29 @@ int main(int argc, char **argv) {
        top of the screen and two widths from the left */
     o = (640 * BFONT_HEIGHT) + (BFONT_THIN_WIDTH * 2);
 
+    /* Test with ASCII encoding */
+    bfont_set_encoding(BFONT_CODE_ASCII);
+    bfont_draw_str(vram_s + o, 640, 1, "Test of basic ASCII");
+    /* After each string, we'll increment the offset down by one row */
+    o += 640 * BFONT_HEIGHT;
+    bfont_draw_str(vram_s + o, 640, 1, "The brackets below should be empty:");
+    o += 640 * BFONT_HEIGHT;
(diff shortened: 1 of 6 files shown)


andressbarajas:
@QuzarDC Mind if I take a stab at this one? I have some experience in my fork which is how I got many (20+ I think) examples that require biosfont working for NAOMI which lacks bios font. Should also help with porting KOS to other archs that may lack biosfont or access to biosfont is very inconvenient.

QuzarDC:
> @QuzarDC Mind if I take a stab at this one? I have some experience in my fork which is how I got many (20+ I think) examples that require biosfont working for NAOMI which lacks bios font. Should also help with porting KOS to other archs that may lack biosfont or access to biosfont is very inconvenient.

Sure. I had seen the method you'd used but what I was aiming for was more as I'd described here: https://github.com/KallistiOS/KallistiOS/issues/858 where at the kos API level there would be a standard 1-bit font printing geared around allowing the Dreamcast's font to be used but on other platforms to leverage the minifont or any other sort of built-in font. Sort of the inverse of what you did of having bfont get backed by minifont when the bios font wasn't available. The start here was to try to get the current bfont/minifont into a place where they could at least both easily support standard 7-bit ASCII as a precursor to having that be the first supported character set of a `kosfont`.

Once that's in place then examples could use the arch-argnostic API and most of them that otherwise are not hardware dependent would be more easily cross-platform. In the new API I'd also figured to move away from the clunky color changing methodology into escape code processing. Getting complex escape codes set up would also allow us to finally have a simple way to display the platform-specific symbols by setting aside bespoke escape codes for them.

PR #903, opened 2025-01-20
Adjust SFX properties on the fly

QuzarDC:
Based on #738 and #885 . I'm not entirely sure about keeping the data structure. It seems like it should be shared, at least in part, with `aica_channel_t` . 

Additionally I'm not sure if the usage makes much sense. The idea would be that you retain your `sfx_play_data_t` and can adjust the vol/pan/freq after using `snd_sfx_play_ex` (even if you set `chn = -1`, as it writes the chosen channel back) then send it over to `snd_sfx_update` with the new values.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/sound/sfxmgr.h
+++ b/kernel/arch/dreamcast/include/dc/sound/sfxmgr.h
@@ -229,6 +229,17 @@ int snd_sfx_play_chn(int chn, sfxhnd_t idx, int vol, int pan);
 */
 int snd_sfx_play_ex(sfx_play_data_t *data);
 
+/** \brief  Update the parameters of a sound channel.
+
+    Update a specified channel's vol, pan, or freq.
+
+    \param  data            The data structure containing the information needed
+                            to play the sound effect.
+
+    \return                 -1 if data or data->chn were invalid, 0 otherwise.
+*/
+int snd_sfx_update(sfx_play_data_t *data);
+
 /** \brief  Stop a single channel of sound.
 
     This function stops the specified channel of sound from playing. It does no
--- a/kernel/arch/dreamcast/exports-naomi.txt
+++ b/kernel/arch/dreamcast/exports-naomi.txt
@@ -119,6 +119,7 @@ snd_sfx_play
 snd_sfx_stop_all
 snd_sfx_play_chn
 snd_sfx_play_ex
+snd_sfx_update
 snd_sfx_stop
 snd_sfx_chn_alloc
 snd_sfx_chn_free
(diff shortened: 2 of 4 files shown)


kernel/arch/dreamcast/include/dc/sound/sfxmgr.h
```diff
@@ -60,17 +60,17 @@ typedef uint32_t sfxhnd_t;
     playback functions.
 */
 typedef struct sfx_play_data {
-    int chn;        /**< \brief The channel to play on. If chn == -1, the next
+    int chn;            /**< \brief The channel to play on. If chn == -1, the next
```
DC-SWAT, line 63: We completely forgot that there are stereo samples on two channels...
And this issue is also applies to `snd_sfx_update()`.

pcercuei: Ideally we should have a `chn2` field as well. It is not a good idea to expect the following channel to be used.

kernel/arch/dreamcast/include/dc/sound/sfxmgr.h
```diff
@@ -60,17 +60,17 @@ typedef uint32_t sfxhnd_t;
...
                             available channel will be used automatically. */
-    sfxhnd_t idx;   /**< \brief The handle to the sound effect to play. */
-    int vol;        /**< \brief The volume to play at (between 0 and 255). */
-    int pan;        /**< \brief The panning value of the sound effect. 0 is all
+    sfxhnd_t idx;       /**< \brief The handle to the sound effect to play. */
+    int vol;            /**< \brief The volume to play at (between 0 and 255). */
+    int pan;            /**< \brief The panning value of the sound effect. 0 is all
                             the way to the left, 128 is center, 255 is all the way
                             to the right. */
```
pcercuei, line 69: These two should probably be `uint8_t`.

kernel/arch/dreamcast/include/dc/sound/sfxmgr.h
```diff
@@ -60,17 +60,17 @@ typedef uint32_t sfxhnd_t;
...
-    int loop;       /**< \brief Whether to loop the sound effect or not. */
-    int freq;       /**< \brief Frequency */
-    unsigned int loopstart;  /**< \brief Loop start index (in samples). */
-    unsigned int loopend;    /**< \brief Loop end index (in samples). If loopend == 0, 
+    int loop;           /**< \brief Whether to loop the sound effect or not. */
+    uint32_t freq;      /**< \brief Frequency */
+    uint32_t loopstart; /**< \brief Loop start index (in samples). */
+    uint32_t loopend;   /**< \brief Loop end index (in samples). If loopend == 0,
```
pcercuei, line 73: These two should be `uint16_t`.

pcercuei:
I think `snd_sfx_update` is a good idea. But FYI I will completely rework the implementation in my "aicaos" PRs.

Issue #786, opened 2024-10-04
'Send-and-forget' Maple Operations


QuzarDC:
By default, the process for issuing a command to a maple device involves queuing the command then waiting for a callback to signal that the command has actually been completed or for the command to time out (with an arbitrarily chosen timeout time). An example:

https://github.com/KallistiOS/KallistiOS/blob/495e77fd60d5b09a1ad52a26cd4a7e73cd0d9d51/kernel/arch/dreamcast/hardware/maple/purupuru.c#L27-L61

I think that `vmu_beep` `vmu_draw_lcd` `purupuru_rumble_raw` could all be refactored to have versions that ignore this step. Unlike things like memory card read/writes, there's not really much to be done other than note that there was some hardware error (and perhaps that kind of checking could be added to the default callback). This would have the positive impact of significantly reducing the amount of time each of those functions block for without having to manage them in threads to avoid the impact.

QuzarDC:
This was effectively implemented via #904 . The only change at the API level is that there is no longer the possibility for the functions to return an `MAPLE_ETIMEOUT` return code. If some usecase is found in the future that would necessitate the old behavior, we can see about providing API for that then.

PR #904, opened 2025-01-20
Don't block for rumble and vmu beep/screen commands

QuzarDC:
There simply isn't anyone that is relying on this blocking behavior and all users end up working around it. The nature of the maple queue makes it so that the blocking time is significant and there really isn't much of anything that an end user could do in the case of a failure.

Though I had originally considered (in #786) adding new functions to select blocking/non-blocking there doesn't seem to be any reason for it. If some use cases are discovered in the future, then the API expansion should be to add intentionally blocking versions.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/purupuru.c
+++ b/kernel/arch/dreamcast/hardware/maple/purupuru.c
@@ -14,16 +14,6 @@
 /* Be warned, not all purus are created equal, in fact, most of
    them act different for just about everything you feed to them. */
 
-static void purupuru_rumble_cb(maple_state_t *st, maple_frame_t *frame) {
-    (void)st;
-
-    /* Unlock the frame */
-    maple_frame_unlock(frame);
-
-    /* Wake up! */
-    genwait_wake_all(frame);
-}
-
 int purupuru_rumble_raw(maple_device_t *dev, uint32 effect) {
     uint32 *send_buf;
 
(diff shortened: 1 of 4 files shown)


pcercuei (approved):
I had to verify, the frame is correctly unlocked in `maple_dma_irq_hnd` when no callback is registered.
So, all looks good.

QuzarDC:
> I had to verify, the frame is correctly unlocked in `maple_dma_irq_hnd` when no callback is registered. So, all looks good.

Yea, a number of the maple commands work without callbacks.

Was going to leave this open to see about discussing a bit more either here or on the discord about the possible repercussions. You mentioned the fact that the retry condition would be more common and possibly switching the opening trylocks into locks would be a good idea. I like it compared to blocking at the end as a means of ensuring the command is actually sent but the problem to my taste is just that it requires more overhead (spin up a thread that can hold the lock) for the user to turn the lock behavior into a trylock than the other way around (loop over the command until it returns OK).

@gyrovorbis also raised a good use case for needing to be certain a command was issued which is in sending a stop for beep. It would be quite annoying if that got lost as the last beep would go on indefinitely. Certainly, as well, if you wanted to do any kind of complex tones with it, you would not want to drop any piece of the sequence. The same is true as well for rumble if not setting any decay.

megavolt85:
this is a really important change that needs to be adopted.
as @QuzarDC already said, no one relies on the response of these functions, but waiting for the response reduces the performance of the games

PR #908, opened 2025-01-24
Deleted memcpy2/4() and memset2/4() routines.

gyrovorbis:
- These routines are slower than just using the builtin compiler routines for memcpy() and memset().
- Worse yet these routines are fundamentally broken. Due to the fact that they violate C and C++'s strict aliasing rules, the compiler can and will optimize out calls to these functions from not uint16_t/uint32_t types when type-punning.
     * This caused major issues with GTA3, as trying to do a memcpy4()
       with a matrix_t gets optimized away, since float * gets aliased
as uint32_t *.
     * This has caused intermittent issues with background color not
       being cleared properly, since it aliases a pvr structure as a
uint32_t * which can also get optimized away.

EDIT: The ONLY KOS port impacted is `GLdc`, which I have already opened a teensy PR to fix: https://gitlab.com/simulant/GLdc/-/merge_requests/139.


QuzarDC:
These basically should never have been used where they were. They're a relic of the gba arch because of how much difference it made there.

gyrovorbis:
Superseded by: https://github.com/KallistiOS/KallistiOS/pull/913

PR #911, opened 2025-01-25
Ansi cleanups

QuzarDC:
I started this trying to match a cleanup done in 2014 for behavior with `-ansi`. I think though since we've updated our language standards the right thing to do with the forced forward declares is likely to remove the other ones (rather than add a new one) as we now have `alignas` and `alloca` which require gnu extensions so we just don't support `c11`/`c17`.

Edit: I had forgotten that the key reason to keep this cleaned up is to more easily allow user software or other libs that link against us to do so safely (as we saw was an issue in #919 and #920 ).

### DIFF
--- a/kernel/arch/dreamcast/include/dc/video.h
+++ b/kernel/arch/dreamcast/include/dc/video.h
@@ -107,15 +107,15 @@ typedef enum vid_display_mode_generic {
 */
 #define DM_MULTIBUFFER  0x2000
 
-//-----------------------------------------------------------------------------
-// More specific modes (and actual indices into the mode table)
+/* ------------------------------------------------------------------------- */
+/* More specific modes (and actual indices into the mode table) */
 
 /** \brief   Specific Display Modes 
     \ingroup video_modes_display
 */
 typedef enum vid_display_mode {
     DM_INVALID = 0,                 /**< \brief Invalid display mode */
-    // Valid modes below
+    /* Valid modes below */
     DM_320x240_VGA = 1,             /**< \brief 320x240 VGA 60Hz */
     DM_320x240_NTSC,                /**< \brief 320x240 NTSC 60Hz */
     DM_640x480_VGA,                 /**< \brief 640x480 VGA 60Hz */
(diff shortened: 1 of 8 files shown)


kernel/libc/koslib/readdir.c
```diff
@@ -10,6 +10,13 @@
 #include <string.h>
 #include <kos/fs.h>
 
+#ifdef __STRICT_ANSI__
+/* Newlib doesn't prototype this function in strict standards compliant mode, so
+   we'll do it here. It is still provided either way, but it isn't prototyped if
+   we use -std=c17 (or any other non-gnuXX value). */
+size_t strnlen(const char *, size_t);
```
gyrovorbis, line 17:
```suggestion
extern size_t strnlen(const char *, size_t);
```
Okay, so this is a Falco-ism, but maybe I can convince you... The default is already `extern` linkage, BUT, I typically explicitly mark the declarations of anything within KOS that it is getting linked from outside--like Newlib, libc, etc--as `extern`, so that someone looking at the coedz knows not to go hunting for the symbol in the KOS codebase and that it's from the outside.

QuzarDC: I do like the idea as functional commentary *but* that is the role of the giant comment above it, and we kinda wanna have that so that it's easily findable for "newlib", and this is a direct copy+paste of similar ones for other gnu extensions.

I think the thing though might really be to say that we do rely on the gnu extensions and not pretend that we should work in strict standards mode. It's effectively an abuse of newlib to blindly link back into these when we are telling the compiler that they're not allowed. Just not sure if that'd have any implication other than warnings with `-ansi`.

kernel/arch/dreamcast/include/dc/video.h
```diff
@@ -191,12 +190,13 @@ extern vid_mode_t vid_builtin[DM_MODE_COUNT];
 */
 extern vid_mode_t *vid_mode;
 
-// These point to the current drawing area. If you're not using a multi-buffered
-// mode, that means they do what KOS always used to do (they'll point at the
-// start of VRAM). If you're using a multi-buffered mode, they'll point at the
-// next framebuffer to be displayed. You must use vid_flip for this to work
-// though (if you use vid_set_start, they'll point at the display base, for
-// compatibility's sake).
+/** These point to the current drawing area. If you're not using a multi-buffered
```
ljsebald, line 193:
```suggestion
/*  These point to the current drawing area. If you're not using a multi-buffered
```
I don't think this is meant to be a doxygen comment...

QuzarDC: Fixed in force push. Whoops, had forgotten that's reserved for doxygen!

kernel/libc/koslib/readdir.c
ljsebald: I'm going to agree with @QuzarDC here that the comment is good enough. The extern isn't necessary here and without commentary it's actually a bit confusing to see (at least I'd be taken a bit aback by it just browsing the code).

ljsebald (approved):

PR #913, opened 2025-01-25
Deprecated string.h and removed usages.

gyrovorbis:
<kos/string.h> is full of routines which completely break at high optimization levels (they get optimized out), from invoking undefined behavior in C and C++ due to breaking strict aliasing rules. This has caused a myriad of subtle issues at high optimization levels over the years, due to these calls getting optimized away, and now we finally know why. Avoid them at all costs.

- Deprecated the heck out of kos/string.h with Doxygen documentation, tags, and attributes explaining why.
- Removed all usages of these functions from within KOS.
- Removed all usages of these functions from KOS-ports.


QuzarDC:
Any thoughts on this though? https://github.com/KallistiOS/KallistiOS/pull/549#issuecomment-2104022139 Perhaps low optimization, or different toolchain?

gyrovorbis:
> Any thoughts on this though? [#549 (comment)](https://github.com/KallistiOS/KallistiOS/pull/549#issuecomment-2104022139) Perhaps low optimization, or different toolchain?

To summarize Discord discussion, it's breaking strict aliasing rules, it can't be certain of the 4-byte alignment necessarily required there, and then finally, in my own testing `memcpy()` still beat an improved `memcpy4()`.

QuzarDC (approved):

PR #915, opened 2025-01-25
VMU Button Bitflag Fix, Plus Maintaining Previous Button States.

gyrovorbis:
- Fixed grievous mistake with VMU button bitflag definitions being... retardedly incorrectly.
- Added caching of the previous frame's button states so that the user can easily determine whether a button was JUST pressed or released, in-line with the keyboard functionality.
- Moved an internal maple frame raw structure from the public API to the internal driver.


QuzarDC:
We don't have an example for it, and the fact that the defines were broken means that it never would have worked anyways, but just wanted to make sure that was intentional. Is it meant to be that you can do `state->buttons & VMU_A` to match against current? Or should the end user always be using `state->button.current`?

kernel/arch/dreamcast/include/dc/maple/vmu.h
```diff
@@ -537,14 +537,14 @@ int vmu_get_datetime(maple_device_t *dev, time_t *unix);
...
-#define VMU_MODE       (6<<1)   /**< \brief Mode button on the VMU */
-#define VMU_SLEEP      (7<<1)   /**< \brief Sleep button on the VMU */
+#define VMU_DPAD_UP    (1 << 0)   /**< Up Dpad button on the VMU */
+#define VMU_DPAD_DOWN  (1 << 1)   /**< Down Dpad button on the VMU */
+#define VMU_DPAD_LEFT  (1 << 2)   /**< Left Dpad button on the VMU */
+#define VMU_DPAD_RIGHT (1 << 3)   /**< Right Dpad button on the VMU */
+#define VMU_A          (1 << 4)   /**< 'A' button on the VMU */
+#define VMU_B          (1 << 5)   /**< 'B' button on the VMU */
+#define VMU_MODE       (1 << 6)   /**< Mode button on the VMU */
+#define VMU_SLEEP      (1 << 7)   /**< Sleep button on the VMU */
```
pcercuei, line 548: While you're at it, please use the BIT() macro from `<kos/regfield.h>`.

gyrovorbis:
> We don't have an example for it, and the fact that the defines were broken means that it never would have worked anyways, but just wanted to make sure that was intentional. 

Yeah, so that's something that I REALLY need to get off my ass and do, but I just haven't had time for... rounding out our very strong set of examples that fully cover the VMU APIs with one for utilizing these buttons and another for manipulating date/time... 

I have tested the shit out of the PR though, before it ever made it into KOS and have personally used it multiple times in apps, but... this was accessing single-button states through the *bitfields* which were right, not the broken-ass bitflags that I `#define`d wrong. 

>Is it meant to be that you can do `state->buttons & VMU_A` to match against current? Or should the end user always be using `state->button.current`?

Since `state->buttons` is now a struct rather than a single value, you have to do either:
`state->buttons.current.raw & VMU_A` or `state->buttons.current.a` for current OR
`state->buttons.previous.raw & VMU_A` or `state->buttons.previous.a` for previous states.

kernel/arch/dreamcast/include/dc/maple/vmu.h
gyrovorbis: Yeah, good call... actually would've saved me from screwing this up in the first place. trolo.

pcercuei: Oh wow, I had not noticed the screwup :)

QuzarDC (approved):

PR #920, opened 2025-01-26
Fix type clash in vmu example.

QuzarDC:
A strange side-effect from #870 . Now math.h is being included via kos.h which caused the static y1 in vmu.c to clash with the defined `double y1(double)` function from math.h.

Corrected this by both restricting the headers being included and renaming the variable.

Not quite sure how this doesn't cause problems elsewhere...

image


gyrovorbis (approved):
LOL OMG I just saw this one the other day and was like WTF!? 

...apparently this is used for some weird Bessel function stuff? ...lolk, great name, C!

anyway, approved!

PR #923, opened 2025-01-26
Various minor fixes found via `-fanalyzer`

QuzarDC:
Some changes were similar in multiple files:
- Ensuring pointer math is being done with the correct types - init.c, snd_iface.c
- Simple allocation failure checks - fs_dcload.c, thread.c

- exec.c updates should functionally be only to use the right types for pointer math, but as the file was not using stdint types, updated it to do so. Additionally fixed a comment that was naming the wrong file, used a memory area define on top of the default start of exec, and added a new variable casting parity/cleanup.

- screenshot.c's update regards a leak with `malloc(0)` which I figure is fine to test against such bad behavior in a util (as opposed to a critical path).

- scandir.c was testing a value for null *after* already derefing it. While I was there I removed some pieces that seemed unnecessary. I also added a null check that is a bit redundant but calmed a warning. @gyrovorbis would be great if you could give this a good looking at.

EDIT: I've integrated the changes that were here for thread.c into #956

### DIFF
--- a/kernel/arch/dreamcast/kernel/exec.c
+++ b/kernel/arch/dreamcast/kernel/exec.c
@@ -4,36 +4,37 @@
    Copyright (C) 2002 Megan Potter
 */
 
+#include <assert.h>
+
 #include <arch/arch.h>
+#include <arch/cache.h>
 #include <arch/exec.h>
 #include <arch/irq.h>
-#include <arch/cache.h>
-#include <assert.h>
-#include <string.h>
-#include <stdio.h>
+#include <arch/memory.h>
 
-/* Pull the shutdown function in from main.c */
+/* Pull the shutdown function in from init.c */
 void arch_shutdown();
 
 /* Pull these in from execasm.s */
-extern uint32 _arch_exec_template[];
-extern uint32 _arch_exec_template_values[];
-extern uint32 _arch_exec_template_end[];
+extern uint32_t _arch_exec_template[];
+extern uint32_t _arch_exec_template_values[];
+extern uint32_t _arch_exec_template_end[];
 
 /* Pull this in from startup.s */
-extern uint32 _arch_old_sr, _arch_old_vbr, _arch_old_stack, _arch_old_fpscr;
+extern uint32_t _arch_old_sr, _arch_old_vbr, _arch_old_stack, _arch_old_fpscr;
 
 /* Replace the currently running image with whatever is at
    the pointer; note that this call will never return. */
-void arch_exec_at(const void *image, uint32 length, uint32 address) {
+void arch_exec_at(const void *image, uint32_t length, uint32_t address) {
     /* Find the start/end of the trampoline template and make a stack
        buffer of that size */
-    uint32  tstart = (uint32)_arch_exec_template,
-            tend = (uint32)_arch_exec_template_end;
-    uint32  tcount = (tend - tstart) / 4;
-    uint32  buffer[tcount];
-    uint32  * values;
-    uint32 i;
+    uintptr_t   tstart = (uintptr_t)_arch_exec_template,
+                tend   = (uintptr_t)_arch_exec_template_end,
+                tvals  = (uintptr_t)_arch_exec_template_values;
+    size_t      tcount = (tend - tstart) / 4;
+    uint32_t    buffer[tcount];
+    uint32_t    *values = buffer + (tvals - tstart);
+    size_t      i;
 
     assert((tend - tstart) % 4 == 0);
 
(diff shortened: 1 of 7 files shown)


kernel/thread/thread.c
```diff
@@ -1093,29 +1096,24 @@ int thd_init(void) {
...
     /* Initialize GBR register for Main Thread */
     __builtin_set_thread_pointer((void*)kern->context.gbr);
 
-    /* De-scehdule the thread (it's STATE_RUNNING) */
-    thd_remove_from_runnable(kern);
-
     /* Setup an idle task that is always ready to run, in case everyone
        else is blocked on something. */
     thd_idle_thd = thd_create_ex(&idle_attr, thd_idle_task, NULL);
-    thd_idle_thd->state = STATE_READY;
```
QuzarDC, line 1107: This was already being set in `thd_create_ex`

kernel/arch/dreamcast/kernel/exec.c
```diff
@@ -48,7 +48,7 @@ void arch_exec_at(const void *image, uint32 length, uint32 address) {
         buffer[i] = _arch_exec_template[i];
 
     /* Plug in values */
-    values = buffer + (_arch_exec_template_values - _arch_exec_template);
+    values = buffer + ((uint32)_arch_exec_template_values - tstart);
```
gyrovorbis, line 51:
```suggestion
    values = buffer + ((uintptr_t)_arch_exec_template_values - tstart);
```
When casting a pointer to an integral type, the proper C type to use is `uintptr_t`, so that it's not architecture-specific. I know it is here, but I'm of the opinion it offers more clarity into the intention that the integer is an address.

kernel/thread/thread.c
```diff
@@ -750,7 +750,10 @@ void thd_schedule_next(kthread_t *thd) {
         return;
 
     /* Unfortunately we have to take care of this here */
-    if(thd_current->state == STATE_ZOMBIE) {
+    if(!thd_current) {
+        /* During init */
```
gyrovorbis, line 754: I feel like rather than just having an empty block here which doesn't match the styling anywhere else, you're better off doing this:

```
if(thd_current) {
    if(thd_current->state == STATE_ZOMBIE) {
        sem_signal(&thd_reap_sem);
    }
    else if(thd_current->state == STATE_RUNNING) {
        thd_current->state = STATE_READY;
        thd_add_to_runnable(thd_current, 0);
     }
 }
 ```

gyrovorbis: But wait, is even that correct? Can you remove a `NULL` thread from runnable and call `thd_update_cpu_time()` on it? Should probably enclose the next couple calls before `thd_current = thd;` in the outer conditional too?

QuzarDC: > But wait, is even that correct? Can you remove a `NULL` thread from runnable and call `thd_update_cpu_time()` on it? Should probably enclose the next couple calls before `thd_current = thd;` in the outer conditional too?

No, you can't remove a NULL thread from runnable or call `thd_update_cpu_time` on it, but those are called against `kern` that is passed in as `thd`, not against `thd_current`. The check against the current state is as the way to check for an exiting thread so we can clean up after it OR moving the current thread to runnable (from running). Then the rest of the ops happen against the new thread. 

So, the if condition of 'no current thread' is the third option for the condition of the thread passing off control (either it's dying, going back to the queue, or didn't exist yet). I structured it that way to have a place to put the comment for the one condition without removing the other, and because I had it organized in the way I had described as 'these are the three different possible states of `thd_current`' rather than 'only do these if there is a current thread'.

I'll ponder if there's a different way, cause you're right that's not really a structure we use anywhere.

kernel/arch/dreamcast/kernel/exec.c
QuzarDC: > When casting a pointer to an integral type, the proper C type to use is `uintptr_t`, so that it's not architecture-specific. I know it is here, but I'm of the opinion it offers more clarity into the intention that the integer is an address.

Yea, I did that elsewhere (for instance init.c) the reason for it here was that the file isn't currently using stdint types and is doing the same elsewhere for casting, so was just trying to solve that issue without cleaning up the whole thing. Will take it back and just clean up the whole thing, it's due.

pcercuei:
First commit looks completely bogus.

QuzarDC:
> First commit looks completely bogus.

Hadn't meant to make this confusing, but had been coincidentally already fixing it through the morning and had finally gotten it ready just now. I presume you were referring to the same one as in the discord, for bss values in init.c where I had mistakenly missed that the extern'd values weren't already the addresses to their respective things. The current version should be right.

kernel/thread/thread.c
QuzarDC: So, ended up finding a better way to do this. Resolving this one and you can take a look in #956

pcercuei (approved):

PR #925, opened 2025-01-27
pvr: random cleanups and IRQ race fix in dma_next_list()

pcercuei:
See the descriptions of the commits.

Last one is a fix for a race that I noticed while studying the code, so it is unclear whether the race ever happened - but I didn't want to wait until it bit one of us.


pcercuei:
Added a fix for a race in `pvr_dma_transfer()`.

QuzarDC (approved):

PR #927, opened 2025-01-28
Remove duplicated copy of the raytris example.

QuzarDC:


gyrovorbis (approved):
OMG SORRY! I didn't even see this directory got basically kept from the previous rename with how big the PR already was.

PR #929, opened 2025-01-28
cache: Fix non-working dcache_{flush,purge}_all()

### LINKED ISSUE #928: `dcache_flush_all` corrupts memory
After months of agonizing debugging (where I hesitated to blame KOS until I exhausted all possibilities I could think of), I came up with a reliable way to reproduce a particularly nasty crash in Doom 64.

I added heap consistency checks around every function called in the main game loop and found that it had corrupted state after many calls to `pvr_scene_finish`, always after `pvr_scene_finish`. 

Working extensively with @pcercuei , the root cause of the issue appears to be bug in the implementation of `dcache_flush_all` , called by way of `dcache_flush_range` when range is `> 64kb` , called by way of `dma_next_list` , called by `pvr_scene_finish` .

I am filing this issue so it can be tracked and referenced in the future for PR or historical reasons.

@pcercuei pinpointed an issue with the value of `dca_addr` in `cache.S`.

According to CPU doc,

`For non-associative addresses (bit 3 of address cleared) there is no write-back. For associative addresses (bit 3 of address set) there is a write-back.`

Changing the value of `dca_addr` from `0xf4000000` to `0xf4000008` would force the writeback.

In testing, this resolves my heap corruption issue and Doom 64 is extremely stable, more stable than it has ever been. 

Cannot reproduce this particular issue after this change.

pcercuei:
The algorithm of dcache_flush_all() and dcache_purge_all() were
sometimes exhibiting problems under unidentified conditions, in which
case some cache lines were not written back properly to RAM.

This caused all kind of memory corruption problems, when the PVR DMA was
used for transfers with a size above the threshold at which the
"dcache_flush_all()" function is called.

Fix these functions by accessing the cache address array using
association. Since the underlying problem is not fully understood, the
fix is not fully understood either, but it was reported as fixing the
memory corruption in Doom64.

Fixes #928.

### DIFF
--- a/kernel/arch/dreamcast/kernel/cache.s
+++ b/kernel/arch/dreamcast/kernel/cache.s
@@ -262,7 +262,7 @@ ifr_addr:
 
 ! D-cache (Data cache)
 dca_addr:
-    .long    0xf4000000    ! dcache array address
+    .long    0xf4000008    ! dcache array address
 dc_ubit_mask:
     .long    0xfffffffd    ! Mask to zero out U bit
 


gyrovorbis (approved):
Yep, you literally showed us the exact page of the SH4 manual that proves this is the correct thing to do... thanks for saving me time. ;)

...should we go check if the Linux kernel is doing this right if we think maybe that's where this could've come from?

pcercuei:
> Yep, you literally showed us the exact page of the SH4 manual that proves this is the correct thing to do... thanks for saving me time. ;)

The referenced doc:
image

The point 2.) does not explicitly say that a write to the offset in the OC address array corresponding to a cache line triggers a write-back. So my first thought was (because of the buggy behaviour of the current implementation) that it wasn't and the algorithms were never tested.

However I tried to get dcache_flush_all() to fail, and it always seemed to trigger a write-back. I just could not reproduce the issue in a small program. So the stance is now that the old dcache_flush_all() *mostly* worked, except under certain conditions that we did not pinpoint yet, and switching to using association for flushing the cache lines make it work everywhere.

jnmartin84:
I just wanted to comment that this seemingly insignificant change has made Doom 64 bulletproof. I had mystery crashes that I had grouped into a handful of categories previously due to where crash dumps were pointing. They all must have stemmed from inconsistent cache flushing because it is stable now, solid. None of the previous crashes have reappeared with many hours of play time between myself and two other testers.

QuzarDC:
So the one thing I question is what impact (if any) the setting of every entry as associative would have on the operation of the cache. The whole point of the functions was to have a faster path than the standard purge/flush. Since association adds extra checks (and presumably time) is there a notion that this is still a fast path? Could reverting to the old behavior without the magic number limits on when to perform the `_all()` versions be a better fix (at the very least we wouldn't be relying on 'don't know why the other wasn't working randomly')?

I couldn't see any notes in the manuals about the distinction in the modes and performance. What version is that of the manual by the way? I'm using Rev6 from '02 and it's slightly different (the same page has the same text with slightly different formatting and has the figure as 4.10 rather than 4.8).

pcercuei:
Could reverting to the old behavior (#342) without the magic number limits on when to perform the `_all()` versions be a better fix (at the very least we wouldn't be relying on 'don't know why the other wasn't working randomly')?

It wouldn't be a fix, since `dcache_{flush,purge}_all()` would still be broken.
> 
> I couldn't see any notes in the manuals about the distinction in the modes and performance. What version is that of the manual by the way? I'm using Rev6 from '02 and it's slightly different (the same page has the same text with slightly different formatting and has the figure as 4.10 rather than 4.8).

SH7550 manual v1.2.1.

QuzarDC:
> Could reverting to the old behavior (#342) without the magic number limits on when to perform the `_all()` versions be a better fix (at the very least we wouldn't be relying on 'don't know why the other wasn't working randomly')?
> 
> It wouldn't be a fix, since `dcache_{flush,purge}_all()` would still be broken.
> 
Yes, my mistake for not following through explicitly. Since the `_all` versions were set up to be special cases for an optimized method where the range was the full cache (or larger than the magic numbers), reverting to the old behavior would have that performed by simply running the existing range functions over the full cache (or range larger than the magic numbers).

These were shortcutted optimizations that are only currently used internally. It would seem to be a bit backwards to have the fast-path lead the standard one. Especially if it could mean slowing the more typical usage.

QuzarDC:
Ok. Well, still unconvinced this won't be a performance regression, but I'll leave that to the 'gainz' crew to prove and fix if so.

PR #930, opened 2025-01-29
Revert PR #887: environ.sh: Disable -fno-builtin, enable -ffast-math

darcagn:
Due to the ICE issues with particular kos-ports on toolchains 12 - 15, we are reverting the new default flag set.


QuzarDC (approved):

QuzarDC:
Thanks, yea, we can revisit it and see if #870 's performance can be improved by more narrow means. 

https://github.com/KallistiOS/kos-ports/issues/97 is where the issues caused have been reported.

PR #934, opened 2025-02-05
pvr.h cleanup II

QuzarDC:
Second split from #539 . Direct copies out of pvr.h into their sub-headers with the additional changes noted (all just related to comments).


kernel/arch/dreamcast/include/dc/pvr/pvr_regs.h
```diff
@@ -0,0 +1,213 @@
...
+
+    These values are written to the PVR_RESET register in order to reset the
+    system or to take it out of reset.
+
+    @{
+*/
+#define PVR_RESET_ALL       0xffffffff  /**< \brief Reset the whole PVR */
+#define PVR_RESET_NONE      0x00000000  /**< \brief Cancel reset state */
+#define PVR_RESET_TA        0x00000001  /**< \brief Reset only the TA */
+#define PVR_RESET_ISPTSP    0x00000002  /**< \brief Reset only the ISP/TSP */
```
pcercuei, line 197: I would guess those are bits, ideally they should use the BIT() macro.
As this is just moving code, not a blocker.

pcercuei (approved):

kernel/arch/dreamcast/include/dc/pvr/pvr_regs.h
QuzarDC: Yea, that's what I had figured on. Was trying to make this as clearly just a move of the code as possible.

PR #935, opened 2025-02-06
Improved control of protection system on the G1 bus for ATA/ATAPI devices.

DC-SWAT:
- Added defines for the protection system on the G1 bus for ATA/ATAPI devices.
- Initiating BIOS checking only if needed.
- Unlocking all memory on the Root Bus for CDROM DMA transfers.

Ready for review. Someone else also needs to check this.
Information about the BIOS check status register is taken from here:
http://www.dc-swat.ru/forum/thread-2150-post-32005.html#pid32005

### DIFF
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -5,7 +5,7 @@
    Copyright (C) 2000 Megan Potter
    Copyright (C) 2014 Lawrence Sebald
    Copyright (C) 2014 Donald Haase
-   Copyright (C) 2023, 2024 Ruslan Rostovtsev
+   Copyright (C) 2023, 2024, 2025 Ruslan Rostovtsev
    Copyright (C) 2024 Andy Barajas
 
  */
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -408,35 +408,69 @@ int cdrom_spin_down(void) {
...
+    uint32_t end_sc = start_sc + size_sc;
+    uintptr_t cur_loc;
+    int count = 0;
+
+    for(cur_loc = start_sc; cur_loc <= end_sc; cur_loc += sizeof(uint32_t)) {
+        if(*(uint32_t *)cur_loc == (uint32_t)G1_ATA_DMA_UNLOCK_SYSMEM) {
+            *(uint32_t *)cur_loc = G1_ATA_DMA_UNLOCK_ALLMEM;
+            ++count;
+        }
+    }
```
pcercuei, line 429: It looks like you search for the protection value in the first 16 KiB of RAM then patch it.
- Don't we know the exact offset at which it is located?
- Isn't there a risk that it would overwrite valid and unrelated code/data?

pcercuei (approved):
Cautious ACK, with some questions inline.

kernel/arch/dreamcast/hardware/cdrom.c
QuzarDC: > It looks like you search for the protection value in the first 16 KiB of RAM then patch it.
> 
>     * Don't we know the exact offset at which it is located?
> 
>     * Isn't there a risk that it would overwrite valid and unrelated code/data?

My understanding for (1) is that because it's part of the syscalls that are copied to ram, the exact location would be dependent on the specific BIOS version as there are differences in them. I do believe though that it should be possible to narrow it down because I agree that (2) seems like it could be an issue. 

For instance, is `count` actually variable? I would think that this only appears in the gd syscalls that allow setting up DMA reads (perhaps *just* `CMD_DMAREAD_STREAM_EX`). Is the fact that there's a branch for `count` being 0 meant to reflect custom bioses where this has already been patched out or just a lack of certainty that the value will be found generally?

kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -408,35 +408,69 @@ int cdrom_spin_down(void) {
...
 
+/*
+    Unlocks G1 ATA DMA access to all memory on the root bus, not just system memory.
+    Patches syscall region where the DMA protection register is set,
+    ensuring it allows broader memory access, and updates the register accordingly.
+ */
+static void unlock_dma_memory(void) {
+    volatile uint32_t *prot_reg = (uint32_t *)(G1_ATA_DMA_PROTECTION | MEM_AREA_P2_BASE);
+    const size_t size_sc = 16 << 10;
+    uintptr_t start_sc = (0x0c000000 | MEM_AREA_P2_BASE);
```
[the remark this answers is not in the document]
DC-SWAT: By the way, yes, I have encountered this more than once and each time I forget to add it somewhere.
I think this needs to be fixed later everywhere by one PR.

kernel/arch/dreamcast/hardware/cdrom.c
DC-SWAT: It works on my BIOS'es, so I ask to check on some other ones.
Yes, these 16 KB contain system calls and they may differ from the BIOS version. Yes, they may be rewritten by the DreamShell ISO Loader firmware, for example. And yes, I'm afraid that this will hinder me someday, but I haven't figured out how to do it better yet. Although I think it's still possible to try to reduce the search area to minimize this.
This is for any cdrom DMA transfers, not only CMD_DMAREAD_STREAM_EX.

DC-SWAT: Offsets in my BIOS'es: `0xac001c20` and `0xac0023fc`

DC-SWAT: Another option is to make a set with specific offsets. Maybe you can look at your BIOS and we will add to the list if necessary.

DC-SWAT: OK, I checked all retail BIOSes and all of them have the same two offsets. They differ only in pre-retail (below 1.0) versions, but they are not used by anyone in fact. So I reworked the algorithm, now only these two addresses are patched when necessary.

kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -408,35 +408,66 @@ int cdrom_spin_down(void) {
...
+    uintptr_t patch_addr[2] = {0x0c001c20, 0x0c0023fc};
+
+    for(i = 0; i < sizeof(patch_addr) / sizeof(uintptr_t); ++i) {
+        if(*(uint32_t *)(patch_addr[i] | MEM_AREA_P2_BASE) == (uint32_t)G1_ATA_DMA_UNLOCK_SYSMEM) {
+            *(uint32_t *)(patch_addr[i] | MEM_AREA_P2_BASE) = G1_ATA_DMA_UNLOCK_ALLMEM;
+            ++patched;
+        }
+    }
+    if(patched) {
+        icache_flush_range(patch_addr[0] | MEM_AREA_P1_BASE, patch_addr[1] - patch_addr[0]);
```
pcercuei, line 428: This won't flush the `patch_addr[1]` address, as it will stop just before - is that a problem? Maybe you should add a +4 there?

QuzarDC (approved):

PR #936, opened 2025-02-06
isofs: Short-circuit the cache and use multi-sector reads if possible.

DC-SWAT:
Also fixed check for cdrom_read_sectors_ex() result.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_iso9660.c
+++ b/kernel/arch/dreamcast/fs/fs_iso9660.c
@@ -273,7 +273,7 @@ static int bread_cache(cache_block_t **cache, uint32 sector) {
     /* Load the requested block */
     j = cdrom_read_sectors_ex(cache[i]->data, sector + 150, 1, CDROM_READ_DMA);
 
-    if(j < 0) {
+    if(j != ERR_OK) {
         //dbglog(DBG_ERROR, "fs_iso9660: can't read_sectors for %d: %d\n",
         //  sector+150, j);
         if(j == ERR_DISC_CHG || j == ERR_NO_DISC) {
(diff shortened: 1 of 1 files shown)


pcercuei (approved):

QuzarDC (approved):
The only real concern I have here has to do with the interaction (or lack thereof) with the cache. It would seem to me that it would make more sense for it to at least check if the first and last blocks in the case that it had already been read (and cache it if not).

It also sort points out the weakness of the caching system in general. We should perhaps always be short-circuiting the cache on the reading of files that couldn't fit in it. If we're reading a 32kb file it doesn't make sense that that would blow away the whole cache for data that is unlikely to be requested again (aside from that last sector). 

Perhaps it might make more sense to do something like caching the data on the file level. Extract small files from read sectors into `/ram` or something like that.

DC-SWAT:
> The only real concern I have here has to do with the interaction (or lack thereof) with the cache. It would seem to me that it would make more sense for it to at least check if the first and last blocks in the case that it had already been read (and cache it if not).
> 
> 
> 
> It also sort points out the weakness of the caching system in general. We should perhaps always be short-circuiting the cache on the reading of files that couldn't fit in it. If we're reading a 32kb file it doesn't make sense that that would blow away the whole cache for data that is unlikely to be requested again (aside from that last sector). 
> 
> 
> 
> Perhaps it might make more sense to do something like caching the data on the file level. Extract small files from read sectors into `/ram` or something like that.

There is no point in caching data at all, this cache is most often not reused. Only the file table makes sense to cache. Here it will be enough to cache one sector for each file, so that it can be split into parts. That's all that is needed. If you need caching data, then it is in stdlib for FILE, just use it.

PR #907, opened 2025-01-23
Vmu pkg cleaning

QuzarDC:
Validate inputs to prevent from writing to garbage, check an OOM, and some minor cleanups.

PR #943, opened 2025-02-16
Correct erroneous resetting of output buffer in vmu_pkg_build.

### LINKED ISSUE #942: `vmu_pkg_build` overwrites `*dst` with `NULL`
This particular line of `vmu_pkg_build` is erroneous.

https://github.com/KallistiOS/KallistiOS/blob/8bf3d94c8360e5e7cdabd3595860dac56036142f/kernel/arch/dreamcast/util/vmu_pkg.c#L140

This is over-writing `out` with `NULL` or whatever your `uint8_t **dst` has for a value when calling `vmu_pkg_build`. 

It just shouldn't be there at all, looks like an oversight.

The correct and expected operation is happening a few lines later with:
`*dst = out;`

If you remove `out = *dst;` it works as expected.

This is an unpleasant bug also because `vmu_pkg_build` returns success value when it happens.

QuzarDC:
Regression from #907. Resolves #942

### DIFF
--- a/kernel/arch/dreamcast/util/vmu_pkg.c
+++ b/kernel/arch/dreamcast/util/vmu_pkg.c
@@ -108,6 +108,10 @@ int vmu_pkg_build(vmu_pkg_t *src, uint8_t **dst, int *dst_size) {
 
     if(!out) return -1;
 
+    /* We're done so write outputs */
+    *dst_size = out_size;
+    *dst = out;
+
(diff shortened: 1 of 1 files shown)


pcercuei:
I don't know Chief, the code looks buggy to me.

`assert((out - *dst) == out_size);`

How can this ever be valid?

PR #945, opened 2025-02-18
Pvr h cleanup iii

QuzarDC:
These seem to be the last two that can be done easily and seamlessly. The other three files have heavy integration between the move and a large retyping and deprecating of defines.


gyrovorbis (approved):
Sorry for the late review! Looks fantastic!! Gonna already be so much better!

PR #946, opened 2025-02-20
Simplify toolchain profiles by deprecating older toolchain components

darcagn:
Remove profiles and patches for older toolchains (`9.3.0`, `10.5.0`, `11.5.0`, `12.4.0`) and Rust dev toolchains. Support for Newlib versions prior to 4.x now deprecated.

As discussed with Quzar a few weeks ago.

This also means no more profiles that build anything earlier than Newlib 4.3.0. Any code in the rest of KOS existing to support Newlib 3.3.0 and earlier can be deprecated.

Justifications:
- `9.3.0-legacy` has been marked as legacy since before the KOS 2.1.0 release. It was once `stable` configuration, but not for that long, and does not generate good code in comparison to newer compilers. I don't know of anyone prefering it for any reason. DreamSDK will be removing it with the upcoming R4 release as well. `9.5.0-winxp` remains and can be used by anyone wanting to use a GCC 9.x configuration.
- `10.5.0` was never really supported, just offered as an alt option. It was never used by anyone and is now EOL upstream.
- `11.5.0` was never really supported, just offered as an alt option. It was never used by anyone and will very soon be EOL upstream.
- `12.4.0` was never really supported, just offered as an alt option. `12.2.0` was briefly supported as a `testing` compiler option, but never widely used, and was supplanted by `13.1.0` and onward.
- `gccrs` is still not capable of building `core` and there's nothing unique about its patch compared to other GCC versions. We don't support doing anything with its flavor of Rust at this time, so we should remove it as a maintenance burden. The option to build Rust support with GCC remains in the scripts so it remains fairly trivial to build this for anyone wanting to do so. 
- `rustc` will be maintained by myself in the downstream `rust-for-dreamcast` repo. There are times where I have to keep particular components in sync, which necessitates pinning GCC to a particular commit ID, and it's much more practical to keep that maintenance outside of KOS itself rather than coordinate merging changes for particular components at the same time.


sizious:
I think there is an historical directory with all patches, maybe it’s time to clean up it as well?

sizious:
Yeah indeed. Probably this directory can be deleted as well! WDYT?

PR #947, opened 2025-02-23
Fix Naomi 23 feb 2025 (broken since 6 sept 2024)

Tchan0:
Fix for 2 PRs that broke the Naomi build:
- 6 sept 2024:
	Refactor sysinfo_icon/id (https://github.com/KallistiOS/KallistiOS/pull/734)
	Calling DC syscalls is a no-go on Naomi
	
- 8 oct 2024: 
	maple: Rework device detection, allocation and status buffers (https://github.com/KallistiOS/KallistiOS/pull/552)
	The Naomi has some different hardware connected to the Maple Bus (ie, the MIE), eg it will respond with a MAPLE_RESPONSE_BADCMD to a MAPLE_COMMAND_DEVINFO to A0.
	So the 'catch all' else in vbl_autodet_callback() must set scan_ready_mask, otherwise maple_wait_scan() in arch_auto_init() will wait endlessly

With these 2 fixes, the current master branch for Naomi (23 feb 2025) creates useable programs again

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_irq.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_irq.c
@@ -182,6 +182,7 @@ static void vbl_autodet_callback(maple_state_t *state, maple_frame_t *frm) {
     else {
         /* dbglog(DBG_KDEBUG, "maple: unknown response %d on device %c%c\n",
             resp->response, 'A'+p, '0'+u); */
+        state->scan_ready_mask |= 1 << p;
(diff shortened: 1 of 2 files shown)


QuzarDC (approved):
Thank you. Both these make a lot of sense. Also saw your suggestion of locking syscalls.h out for naomi and it makes a lot of sense. Hopefully I can get mine set up sometime soon. 

Last I tried (~ a year ago) I simply couldn't get any results to work. Probably had made some silly mistake that I just couldn't see.

PR #951, opened 2025-03-02
AICAOS step2

pcercuei:
AICAOS Step 2. This PR can be merged before or after step 1, they are independent and do not conflict between each other.

Introduce `g2_lock_scoped()` as a scoped variant of `g2_lock().`
Add a new `<dc/aram.h>` API to simplify transfers from/to the sound RAM.
Add a firmware header structure, that right now contains hardcoded pointers to the memory offsets that are currently used in the firmware for the message queue, channels array and sample buffer. In the future, this header will be provided by the firmware.
Update the interface code to use the new ARAM API as well as the header structure.

### DIFF
--- a/kernel/arch/dreamcast/hardware/Makefile
+++ b/kernel/arch/dreamcast/hardware/Makefile
@@ -21,6 +21,9 @@ OBJS += g2dma.o
 # Sound
 OBJS += spu.o
 
+# Sound RAM
+OBJS += aram.o
+
 # Bus support
 OBJS += asic.o g2bus.o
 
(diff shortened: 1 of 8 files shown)


QuzarDC:
So I think the reason to have the `aica_cmd_iface.h` that was so strangely included was to prevent it from being part of the public API (one of the few places where we had any kind of non-public header). I'm not specifically opposed to moving it into the public-facing include folder, but that didn't seem to be your intent with the change.

pcercuei:
@QuzarDC force-pushed with the commit that moved the header removed.

I can always move it back in a future PR if there is a valid reason for it to be in the public API.

pcercuei:
Closing as this will be replaced by a proper KOS port to the ARM.

PR #111, opened 2023-02-22
Enable TLS Support for C/C++

gyrovorbis:
This is the work required to support static TLS for KOS and in the toolchains per the SH TLS ABI is variant I.

- [X] Remove disabled TLS in the toolchain build scripts
- [X] Add "tcbhead" intrusive structure to KOS's thread structure, which is what's stored within GBR
- [X] Allocating extra chunk of memory for "tcbhead" followed by .tdata and .tbss segements
- [X] Set static TLS block data based on .tdata values for initialized data, 0 for .tbss data
- [X] Added TLS-specific section exports to the linker script for access within KOS
- [X] Created example flexing TLS using thread-local variables requiring both .tbss and .tdata initialization
- [X] Tested with GCC12.2.0
- [X] Tested with GCC9.3.0
- [X] Tested with GCC4.7.0
- [X] Tested with GCC13.1.0 
- [X] TLS variables working properly for the main thread
- [X] Manually aligned TLS variables working properly 

This implementation still provides a void *dtv field which is NULL initialized for now, which can later be used to add support for non-static dynamic module TLS in the future as well. 

This PR also adds an example of using TLS storage that also doubles as a test suite to flex the hell out of this implementation and return EXIT_SUCCES/EXIT_FAILURE based on the results (which we can then use in CI/automation once we get return values in DC-Tool).

PR #956, opened 2025-03-22
Move compiler-based static tls arch-specifc code

QuzarDC:
This moves the code that was implemented in #111 out of the main platform-independent thread.c into new arch file.

The only change to behavior on DC should be that the allocation of the TLS memory can now fail gracefully to be recovered from in the same way thread stack allocation can rather than merely asserting.

The third commit in this PR is a cleanup based on one that was intended to be in #923 . I found a better methodology to do it than in there and moved it to here in order both for the two open PRs not to conflict and in order to allow overlapped testing.


kernel/arch/dreamcast/kernel/tls_static.c
```diff
@@ -0,0 +1,133 @@
...
+*/
+typedef struct tcbhead {
+    void *dtv;               /* Dynamic TLS vector (unused) */
+    uintptr_t pointer_guard; /* Pointer guard (unused) */
+} tcbhead_t;
+
+void arch_tls_init(void) {
+    /* Initialize GBR register for Main Thread */
+    __builtin_set_thread_pointer((void*)(thd_get_current()->context.gbr));
+    return;
```
gyrovorbis, line 39: Any particular reason for the explicit return here?

gyrovorbis (approved):
Yeah, this looks like a fine way to make static TLS into a link-time polymorphic architecture-specific construct. Approved.

kernel/arch/dreamcast/kernel/tls_static.c
QuzarDC: Whoops. My personal style bleeding in. Went ahead and undid that and the other pointless one in `arch_tls_destroy_data`.

Unfortunately will need re-approval after that.

kernel/arch/dreamcast/kernel/tls_static.c
```diff
@@ -0,0 +1,131 @@
...
+
+/* TLS Section ELF data - exported from linker script. */
+extern int _tdata_start, _tdata_size;
+extern int _tbss_size;
+extern long _tdata_align, _tbss_align;
+
+/* Utility function for aligning an address or offset. */
+static inline size_t align_to(size_t address, size_t alignment) {
+    return (address + (alignment - 1)) & ~(alignment - 1);
+}
```
pcercuei, line 23: This should probably be moved to a separate header.

pcercuei (approved):
LGTM.

One question though, would it be worthwhile to have the option to disable TLS with an init flag? Or it's not really worth it?

kernel/arch/dreamcast/kernel/tls_static.c
QuzarDC: Yea, you're right. Will try to hunt down where this pattern appears elsewhere along with the array size macro from #966 .

QuzarDC:
> LGTM.
> 
> One question though, would it be worthwhile to have the option to disable TLS with an init flag? Or it's not really worth it?

Sorry, forgot to answer this. I was under the impression that it basically used no memory unless you opted into building with tls defined. Would there be a good reason to have it disable-able? If there is some to save we could see about detecting if it's been enabled by the compiler.

PR #957, opened 2025-03-22
Fix CDDA on GDEMU clones

kilgariff:
With a real GDROM drive, sending `CMD_INIT` shortly after making a `FUNC_GDROM_RESET` syscall works with no issues.

With GDEMU clones running 5.15b (and possibly other ODEs or firmware versions), the drive is left in a bad state when this is done. Specifically, attempting to play CDDA tracks doesn't work and the TOC returned from `cdrom_read_toc` is invalid.

This issue wasn't observed before [this commit](https://github.com/kilgariff/KallistiOS/commit/a246cbfe5f2756e526da17f3d67c2910a4765cf7) which adds a reset syscall.

The fix polls the drive by making a `FUNC_GDROM_DRIVE_STATUS` syscall in a busy loop until the drive becomes available, making sure to pass control to other threads between iterations.

This takes ~33us on real hardware and ~150ms on GDEMU clones.

> [!NOTE]  
> The best way to repro this issue is to use the `basic_cdda` KOS example as the very first thing it does is play a CDDA track.

It was suggested that `thd_sleep(10)` could be used to avoid a blocking busy-wait loop, but I opted  for `thd_pass` as this is what other wait loops in the same file use and this way we reduce unnecessary wait time as much as possible. If there is a clear reason to use `thd_sleep(10)` instead then this can be changed.

I have also tested this on Flycast and LXDream Nitro. Flycast already did the same thing as a physical drive. LXDream Nitro would always report `CD_STATUS_BUSY`, so I [submitted an MR with a fix](https://gitlab.com/simulant/community/lxdream-nitro/-/merge_requests/45).

I am not able to test other ODEs such as:
- Original GDEMU
- GDEMU clone running 5.20+
- MODE
- USB-GDROM

> [!NOTE]  
> If an application reads a file from the CD-ROM (ISO9660) filesystem before trying to play CDDA audio, `cdrom_reinit()` is called as part of per-disc initialization and this issue is not encountered (because enough time has passed between the reset and drive re-init).

### DIFF
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -439,6 +439,7 @@ void cdrom_init(void) {
     volatile uint32_t *react = (uint32_t *)(G1_ATA_BUS_PROTECTION | MEM_AREA_P2_BASE);
     volatile uint32_t *state = (uint32_t *)(G1_ATA_BUS_PROTECTION_STATUS | MEM_AREA_P2_BASE);
     volatile uint32_t *bios = (uint32_t *)MEM_AREA_P2_BASE;
+    int status, disc_type;
 
     mutex_lock(&_g1_ata_mutex);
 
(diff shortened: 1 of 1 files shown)


kilgariff:
We found a simpler fix for this: https://github.com/KallistiOS/KallistiOS/pull/959

PR #960, opened 2025-03-28
math: Add mat_multiply()

pcercuei:
This function is similar to mat_apply(), except that it does not write the result matrix back to the internal matrix, but to a destination pointer in memory.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/matrix.h
+++ b/kernel/arch/dreamcast/include/dc/matrix.h
@@ -85,6 +85,22 @@ void mat_identity(void);
 */
 void mat_apply(const matrix_t *src);
 
+/** \brief  Multiply a matrix.
+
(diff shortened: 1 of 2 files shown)


gyrovorbis:
https://github.com/user-attachments/assets/a6ca91fa-af4f-4cef-a1cd-4bdaee5b6739

Changed the main mesh world-space transformation logic to use this new routine for all 3D geometry in GTA3, just to test it out, and it works great! There would be nothing onscreen or everything screwed to hell if it didn't work. lol.

gyrovorbis (approved):
Extremely useful routine for somewhat novel/niche situations where you want to perform a matrix multiplication WITHOUT overriding the banked matrix. 

This is really useful for skinning, among other things in graphics and collision. Ship it!

QuzarDC (approved):

PR #961, opened 2025-03-29
vmufb: Add function vmufb_screen_shot()

pcercuei:
This function can be used to save a screenshot of a VMU framebuffer into a PBM image.


gyrovorbis (approved):
Oh shit, what a good idea.

Just had to look up wtf a "PBM" was though, and read about the P4 variant... pretty sweet this is just exactly the format we're using for the FB natively...

QuzarDC (approved):

PR #963, opened 2025-04-03
pvr: Fix race between driver and TA

pcercuei:
The driver was calling pvr_start_ta_rendering() at two different spots:
when a non-DMA list was opened, or in the pvr_scene_finish() function
before uploading DMA lists. This function would wait for the TA to be
ready (if not previously done), then mark it as "busy" again.

However, in the specific case where the DMA feature was enabled but no
lists were actually configured for DMA transfers, and for a simple
enough scene, it was possible for the TA to complete before the second
call of pvr_start_ta_rendering(), causing it to reset the "busy" flag
once again, which would then never be cleared by the IRQ handler.

This caused any following call to pvr_wait_ready() to timeout, dropping
the framerate to 5 fps or lower.

Fix this race by only setting the "busy" flag once, as we only need to
set it in the first call (whichever it is) of pvr_start_ta_rendering().


QuzarDC (dismissed):

QuzarDC:
Approved for content, but needs to be conflict resolved first due to #962

pcercuei:
@QuzarDC force-pushed with the conflicts resolved.

pcercuei:
Reopened, I think I accidentally deleted the branch for this PR and it auto-closed it. It was not on purpose.

PR #964, opened 2025-04-05
Make the fs_iso9660 file table dynamically sized.

QuzarDC:
Move the file table out of being a static array and to being in dynamically allocated memory. Start at the previous default size of 8 and expand by doubling in size with the same fail condition if we cannot expand in that way.

The `FS_CD_MAX_FILES` define now has the same basic meaning but its usage is reversed. The reason to set it would be to limit the size of the table, rather than to expand the allowed size. Accordingly the default has been set to 0 which indicates 'no max limit'.

Need to write some test for it, as I wrote it without any but hopefully could get some real-world validation from anyone who has had to use `FS_CD_MAX_FILES` in the past to expand the table would be able to give it a go.

### DIFF
--- a/include/kos/opts.h
+++ b/include/kos/opts.h
@@ -114,9 +114,9 @@ __BEGIN_DECLS
 #define VMUFS_DEBUG 1
 #endif
 
-/** \brief  The maximum number of cd files that can be open at a time. */
+/** \brief  The maximum number of cd files that can be open at a time. 0 no max. */
 #ifndef FS_CD_MAX_FILES
-#define FS_CD_MAX_FILES 8
+#define FS_CD_MAX_FILES 0
 #endif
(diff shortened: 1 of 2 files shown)


pcercuei:
I don't really understand why we need a dynamically sized table here. Or a table at all, for that matter.

Why do we use an integer (`file_t`) as our file token, cast to `void *`, and use it as the index into a shared table that we eventually need to make dynamic, instead of just allocating a `fh` struct and use it as the file token?

The only thing that gives it *some* sense, is that `iso_break_all()` needs to "break" the file descriptors. But I think it would be a much better idea to just keep a single-linked list of the `fh` then.

Thoughts?

QuzarDC:
Obsolete due to #973

PR #966, opened 2025-04-07
Kbd update i

QuzarDC:
Split from #503 . Like the splits from the irq and pvr overhaul, the aim is to bring the changes in via easily readable atomic commits. Implements a majority of the non-breaking changes from #503 . The full set of changes are detailed in the individual commit comments.

The biggest change here is the removal of various pieces of private data for the keyboard queue out of the public `kbd_state_t` struct and into an internal driver struct. This should not be breaking though as the struct members were both noted to be internal and of no real use outside of the driver.

Edit: had some more time, so was able to find further changes to integrate and tried to fix up the doxygen as well. No further changes expected unless some thing is found wrong.
Edit2: Finally got my keyboard out to test. Had to correct a few errors, but no new content.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/keyboard.c
+++ b/kernel/arch/dreamcast/hardware/maple/keyboard.c
@@ -31,7 +31,17 @@ repeat handling.
     It seems unreasonable that one might want different repeat
     timings set on each keyboard.
     The values are arbitrary based off a survey of common values. */
-uint16_t kbd_repeat_start = 600, kbd_repeat_interval = 20;
+static struct {
+    uint16_t start;
+    uint16_t interval;
+} repeat_timing = {
+    600, 20
+};
+
+void kbd_set_repeat_timing(uint16_t start, uint16_t interval) {
+    repeat_timing.start    = start;
+    repeat_timing.interval = interval;
+}
 
 /* Built-in keymaps. */
 #define KBD_NUM_KEYMAPS 8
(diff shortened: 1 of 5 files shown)


kernel/arch/dreamcast/hardware/maple/keyboard.c
```diff
@@ -44,8 +60,8 @@ void kbd_set_repeat_timing(uint16_t start, uint16_t interval) {
 }
 
 /* Built-in keymaps. */
-#define KBD_NUM_KEYMAPS 8
-static kbd_keymap_t keymaps[KBD_NUM_KEYMAPS] = {
+#define KBD_NUM_KEYMAPS (sizeof(keymaps) / sizeof(keymaps[0]))
```
pcercuei, line 83: Small suggestion, we could have an `ARRAY_SIZE()` macro in e.g. <kos/cdefs.h>, since it's a very useful macro to have.

pcercuei (approved):
Thanks a lot for this!
I had a look through it, commit by commit, and everything looked OK to me.

kernel/arch/dreamcast/hardware/maple/keyboard.c
QuzarDC: That does sound good. I'll see about putting in a followup PR for it. Seems like there's at least 6 other places in the codebase where we do the same thing (though sometimes in more cryptic ways like hardcoding the type rather than 0-index).

PR #968, opened 2025-04-12
Macros expansion

QuzarDC:
Wanted to toss this up quickly to get some feedback on the naming and goal before putting them in place. The names should possibly instead be something like `__ARRAY_SIZE` or `KOS_ARRAY_SIZE`. It seemed inappropriate to use the plainest `ARRAY_SIZE` due to how easily that could clash with external software.

The macros were taken from public domain implementations from here: https://ccodearchive.net/info/array_size.html and https://ccodearchive.net/info/build_assert.html . 

This PR is going to be the response to this PR comment: https://github.com/KallistiOS/KallistiOS/pull/966#discussion_r2034001652

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_utils.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_utils.c
@@ -93,7 +93,6 @@ static const char *maple_cap_names[] = {
     "Mouse",
     "JumpPack"
 };
-#define maple_cap_name_cnt (sizeof(maple_cap_names)/sizeof(char *))
 
 /* Print the capabilities of a given driver to dbglog; NOT THREAD SAFE */
 static char caps_buffer[64];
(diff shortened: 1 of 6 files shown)


kernel/arch/dreamcast/hardware/maple/maple_utils.c
```diff
@@ -133,7 +133,7 @@ static const char *maple_resp_names[] = {
     "OK",
     "DATATRF"
 };
-#define maple_resp_name_cnt ((int)(sizeof(maple_resp_names)/sizeof(char *)))
+#define maple_resp_name_cnt ((int)(__array_size(maple_resp_names)))
```
pcercuei, line 136: Not sure why the cast is needed? You can also drop the parentheses around `__array_size()` here and in the other two instances above.

QuzarDC: `size_t` is unsigned and `int` is not. Without the cast, its subsequent use generates a warning. I went ahead though and collapsed it down and performed the cast to size_t which should be a bit more readable. Since the two defines were only used immediately after their definition, I also took the step of just removing them.

As for the parens: 1) cause they were already there 2) habit of not trusting the safety of defines when used in statements.

gyrovorbis (approved):
Clever shit going on here... by necessity, I see, as it's not like you can `Static_assert()` in this file which gets included in all the toolchain crap which gets built with ancient C, so...

Very cool, looks good. Doxygen improvements much appreciated.

PR #970, opened 2025-04-16
g1ata: Fixed DMA protection and cleanup.

DC-SWAT:
- Fixed DMA protection register memory area.
- Introduced constants for maximum sector counts (LBA28 and LBA48).
- Refactored sector count handling in various functions to use these constants.
- Improved DMA chaining logic and error handling in the DMA IRQ handler.
- Updated type usage for sector counts to size_t for better compatibility.
- Using macros instead of hardcoded values.

### DIFF
--- a/kernel/arch/dreamcast/hardware/g1ata.c
+++ b/kernel/arch/dreamcast/hardware/g1ata.c
@@ -150,6 +150,9 @@ typedef struct ata_devdata {
 #define ATA_TRANSFER_WDMA(x)        0x20 | ((x) & 0x07)
 #define ATA_TRANSFER_UDMA(x)        0x40 | ((x) & 0x07)
 
+#define ATA_MAX_SECTORS_LBA28       256
+#define ATA_MAX_SECTORS_LBA48       65536
+
 /* Access timing data. */
 #define G1_ACCESS_WDMA_MODE2        0x00001001
 #define G1_ACCESS_PIO_DEFAULT       0x00000222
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/hardware/g1ata.c
```diff
@@ -218,9 +221,9 @@ inline int g1_ata_mutex_unlock(void) {
     return mutex_unlock(&_g1_ata_mutex);
 }
 
-static void g1_ata_set_sector_and_count(uint64_t sector, uint32_t count, int lba28) {
+static void g1_ata_set_sector_and_count(uint64_t sector, size_t count, int lba28) {
     if(!lba28) {
-        OUT8(G1_ATA_SECTOR_COUNT, (uint8_t)(count >> 8));
+        OUT8(G1_ATA_SECTOR_COUNT, (uint8_t)((count >> 8) & 0xFF));
```
ljsebald, line 226: This `& 0xFF`, as well as the one farther down that were added should not be necessary due to the cast to `uint8_t`.

ljsebald (dismissed):
Looks good, with the exception of the seemingly unnecessary masking operations.

kernel/arch/dreamcast/hardware/g1ata.c
DC-SWAT: Yeah sorry for that, I did this when I was trying to figure out what was causing the problem and was stuck on the sector count. I'll remove those unnecessary changes.

ljsebald (approved):

PR #971, opened 2025-04-16
maple: Support multiple hotplug callbacks

pcercuei:
Previously, maple_attach_callback() would register a callback function to be called when a device featuring one or more of the specified functions was plugged.

The problem is that there could only be a single callback function registered at any given time; it was therefore impossible to register a callback e.g. for a controller and a different one for a VMU.

Address this issue by adding callback pointers inside the maple_driver structure itself, and update maple_attach_callback and maple_detach_callback to update the fields of the drivers that match the specified functions.

With this new setup, there can be one attach / detach callback couple registered for each single function, or any combination of them, that KallistiOS supports.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_driver.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_driver.c
@@ -8,20 +8,38 @@
 #include <stdlib.h>
 #include <dc/maple.h>
 
-static maple_attach_callback_t attach_callback = NULL;
-static uint32 attach_callback_functions = 0;
-
-static maple_detach_callback_t detach_callback = NULL;
-static uint32 detach_callback_functions = 0;
-
 void maple_attach_callback(uint32 functions, maple_attach_callback_t cb) {
-    attach_callback_functions = functions;
-    attach_callback = cb;
+    maple_driver_t *i;
+
+    if(!functions)
+        functions = 0xffffffff;
+
+    LIST_FOREACH(i, &maple_state.driver_list, drv_list) {
+        if(i->functions & functions) {
+            i->user_attach = cb;
+            functions &= ~i->functions;
+
+            if(!functions)
+                break;
+        }
+    }
 }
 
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/hardware/maple/maple_driver.c
```diff
@@ -129,17 +144,15 @@ int maple_driver_detach(int p, int u) {
 
     dev->valid = false;
 
-    if(dev->drv && dev->drv->detach)
-        dev->drv->detach(dev->drv, dev);
+    if(dev->drv) {
+        if(dev->drv->detach)
+            dev->drv->detach(dev->drv, dev);
+        if(dev->drv->user_detach)
+            dev->drv->user_detach(dev);
```
QuzarDC, line 151: It's kind of a moot point at the moment since we don't currently have any maple drivers with detach functions, but shouldn't `user_detach` happen first to mirror `user_attach` being called after `attach`?

pcercuei: Sure, I guess. I will swap them.

pcercuei: @QuzarDC Fixed and force-pushed.

gyrovorbis:
Guys, while we're at it, would you be okay with fixing the most grievous mistake in the initial design...

...without those callbacks taking a `void*` userdata, they're worthless to C++ and their applicability becomes drastically diminished.

pcercuei:
@gyrovorbis and where would the argument come from?

ljsebald (approved):

ljsebald:
> Guys, while we're at it, would you be okay with fixing the most grievous mistake in the initial design...
> 
> ...without those callbacks taking a `void*` userdata, they're worthless to C++ and their applicability becomes drastically diminished.

This can be discussed later. I don't think it's necessary here to muck with any API breakage.

PR #972, opened 2025-04-17
cdrom: Introduced IRQ handling for DMA and command polling. 

DC-SWAT:
This is a part of code from a larger PR #736, slightly modified and improved.
Here is introduced the processing of interrupts of DMA and polling of commands in vblank. This is what original games do, but I improve this algorithm and adapted it for modern devices with flash memory and fast access.
The next PR will contain the streaming API.

### DIFF
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -133,8 +133,7 @@ int cdrom_exec_cmd_timed(int cmd, void *param, uint32_t timeout) {
         /* If the command is still in progress, it timed out. */
         if(cmd_in_progress) {
             cmd_in_progress = false;
-            syscall_gdrom_reset();
-            syscall_gdrom_init();
+            cdrom_abort_cmd(1000, true);
             return ERR_TIMEOUT;
         }
     }
@@ -160,6 +159,58 @@ int cdrom_exec_cmd_timed(int cmd, void *param, uint32_t timeout) {
     }
 
     return ERR_SYS;
+}
+
+int cdrom_abort_cmd(uint32_t timeout, bool abort_dma) {
+    int rv = ERR_OK;
+    uint64_t begin;
+    int old = irq_disable();
+
+    if(cmd_hnd <= 0) {
+        irq_restore(old);
+        return ERR_NO_ACTIVE;
+    }
+    cmd_in_progress = false;
+
+    if(abort_dma && dma_in_progress) {
+        dma_in_progress = false;
+        dma_blocking = false;
+        dma_thd = NULL;
+        /* G1 ATA mutex already locked */
+    }
+    else {
+        mutex_lock(&_g1_ata_mutex);
+    }
+
+    irq_restore(old);
+    syscall_gdrom_abort_command(cmd_hnd);
+
+    if(timeout) {
+        begin = timer_ms_gettime64();
+    }
+    do {
+        syscall_gdrom_exec_server();
+        cmd_response = syscall_gdrom_check_command(cmd_hnd, cmd_status);
+
+        if(cmd_response == NO_ACTIVE || cmd_response == COMPLETED) {
+            break;
+        }
+        if(timeout) {
+            if((timer_ms_gettime64() - begin) >= timeout) {
+                dbglog(DBG_ERROR, "cdrom_abort_cmd: Timeout exceeded, resetting.\n");
+                rv = ERR_TIMEOUT;
+                syscall_gdrom_reset();
+                syscall_gdrom_init();
+                break;
+            }
+        }
+        thd_pass();
+    } while(1);
+
+    cmd_hnd = 0;
+
+    mutex_unlock(&_g1_ata_mutex);
+    return rv;
 }
 
 /* Return the status of the drive as two integers (see constants) */
@@ -368,11 +419,6 @@ int cdrom_read_sectors_ex(void *buffer, int sector, int cnt, int mode) {
     params.num = cnt;       /* Number of sectors */
     params.is_test = 0;     /* Enable test mode */
 
-    /* The DMA mode blocks the thread it is called in by the way we execute
-       gd syscalls. It does however allow for other threads to run. */
-    /* XXX: DMA Mode may conflict with using a second G1ATA device. More 
-       testing is needed from someone with such a device.
-    */
     if(mode == CDROM_READ_DMA) {
         if(buf_addr & 0x1f) {
             dbglog(DBG_ERROR, "cdrom_read_sectors_ex: Unaligned memory for DMA (32-byte).\n");
@@ -591,6 +637,10 @@ void cdrom_init(void) {
     volatile uint32_t *react = (uint32_t *)(G1_ATA_BUS_PROTECTION | MEM_AREA_P2_BASE);
     volatile uint32_t *state = (uint32_t *)(G1_ATA_BUS_PROTECTION_STATUS | MEM_AREA_P2_BASE);
     volatile uint32_t *bios = (uint32_t *)MEM_AREA_P2_BASE;
+
+    if(inited) {
+        return;
+    }
 
     mutex_lock(&_g1_ata_mutex);
 
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -14,13 +14,17 @@
 #include <arch/cache.h>
 #include <arch/timer.h>
 #include <arch/memory.h>
+#include <arch/irq.h>
 
+#include <dc/asic.h>
 #include <dc/cdrom.h>
 #include <dc/g1ata.h>
 #include <dc/syscalls.h>
+#include <dc/vblank.h>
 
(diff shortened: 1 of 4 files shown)


pcercuei:
I don't really like how you unhook the CDROM IRQ handler, to replace it with the G1ATA handler that eventually calls the old handler...

I would prefer to have a top-level function registered as the interrupt handler, that would then call the G1ATA handler if provided, and the CDROM handler if provided.

DC-SWAT:
> I don't really like how you unhook the CDROM IRQ handler, to replace it with the G1ATA handler that eventually calls the old handler...
> 
> 
> 
> I would prefer to have a top-level function registered as the interrupt handler, that would then call the G1ATA handler if provided, and the CDROM handler if provided.

You mean something like vblank handlers but for G1?
But they will always be shared only on two devices...
I also don't really like the current interruption mechanism, it's not possible to set multiple handlers for one interruption. I think that's what needs to be fixed.

pcercuei:
@DC-SWAT Nooo, nothing that bad. I mean something like this:
```c
void g1_dma_irq_handler(uint32 code, void *data) {
   if(g1_ata_irq_hdl)
     g1_ata_irq_hdl(code, data);
   else if (cdrom_irq_hdl)
     cdrom_irq_hdl(code, data);
}
```

Then only the top-level function needs to be registered as IRQ handler.

DC-SWAT:
> @DC-SWAT Nooo, nothing that bad. I mean something like this:
> 
> ```c
> void g1_dma_irq_handler(uint32 code, void *data) {
>    if(g1_ata_irq_hdl)
>      g1_ata_irq_hdl(code, data);
>    else if (cdrom_irq_hdl)
>      cdrom_irq_hdl(code, data);
> }
> ```
> 
> Then only the top-level function needs to be registered as IRQ handler.

OK, where? hardware_init()? In theory these drivers can be used separately and together.

pcercuei:
Another idea: setting a ASIC handler is a cheap operation, why don't you set the handler / enable the IRQ right before starting DMA, and disable the IRQ in the IRQ handler right before you unlock the G1 mutex? Since the G1ATA and CDROM code cannot do DMA in parallel (because of the G1 mutex), this should work just fine.

ljsebald:
> I'm torn, I'll let @QuzarDC or other maintainers decide.

I'll defer to @QuzarDC on this one.

QuzarDC:
I think this solution:

```
void g1_dma_irq_handler(uint32 code, void *data) {
   if(g1_ata_irq_hdl)
     g1_ata_irq_hdl(code, data);
   else if (cdrom_irq_hdl)
     cdrom_irq_hdl(code, data);
}
```
Should work just fine, and `hardware_init()` does seem like the right place for it. It would just need to be excluded for NAOMI. That should be an easier and less risky structure without the overhead from swapping them in and out.

DC-SWAT:
> I think this solution:
> 
> ```
> void g1_dma_irq_handler(uint32 code, void *data) {
>    if(g1_ata_irq_hdl)
>      g1_ata_irq_hdl(code, data);
>    else if (cdrom_irq_hdl)
>      cdrom_irq_hdl(code, data);
> }
> ```
> 
> Should work just fine, and `hardware_init()` does seem like the right place for it. It would just need to be excluded for NAOMI. That should be an easier and less risky structure without the overhead from swapping them in and out.

Do not need exclude it for NAOMI, because any ATA IDE devices also can be used on it.
And still doesn't clear how to do that. OK, `g1_ata_irq_hdl` and `cdrom_irq_hdl` is extern variables that can be set. But it means I should always enable IRQ's for G1 ATA even if no one devices are used...
Are you sure that external variables and always-on interrupts are a better solution than what we have now? I'll do it, no problem, I just want to clarify.
Yes, the current solution is not the best, the problem here is in the interrupt API, but what you propose will either add some handlers like vblank or will look like a hack. Or I don't understand something correctly.

kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -408,6 +548,65 @@ int cdrom_spin_down(void) {
...
+    }
+    else if(cmd_timeout && (timer_ms_gettime64() - cmd_begin_time) >= cmd_timeout) {
+        sem_signal(&cmd_done);
+        thd_schedule(1, 0);
+    }
+}
+
+static void g1_dma_irq_hnd(uint32_t code, void *data) {
+    (void)data;
+
```
QuzarDC, line 586: Shouldn't `code` be checked here for overrun/illaddr? Or is that not needed due to relying on the output of `cmd_response` from the syscall to tell that an error happened?

kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -33,109 +37,176 @@ gets everything situated. After that it will read raw sectors from
...
+    if(abort_dma && dma_in_progress) {
+        dma_in_progress = false;
+        dma_blocking = false;
+        dma_thd = NULL;
+        /* G1 ATA mutex already locked */
+    }
+    else {
+        mutex_lock(&_g1_ata_mutex);
+    }
+    syscall_gdrom_abort_command(cmd_hnd);
```
QuzarDC, line 182: I think there's a race here with the irq handlers. If you issue an abort and one of the irqs triggers between the check for cmd and the disabling of `cmd_in_progress` (vblank and g1dma) or `dma_in_progress` (g1dma) then the rest could proceed and even send the abort without the mutex locked.

Racing against the vblank handler seems like it would be fixed by having `cmd_in_progress = false;` be the first thing in the function. If there's no active command, then it will have no effect, but it would prevent the vblank handler from acting on a command that is in the middle of being aborted.

For the dma handler I'm not as sure. disabling irqs would do it, but there's likely a better way. An interrelated issue is that if you call with `abort_dma == true` and dma is not in progress it will proceed to abort whatever command might be running and if `abort_dma == false` and dma *is* in progress then it will wait on the mutex lock until the dma is finished and send a probably useless/unintentional abort.

QuzarDC (changes requested):
A small comment/question on the irq handling, but also seems like the abort command would race against the handlers.

Other than that, the rest looks great.

QuzarDC:
> > I think this solution:
> > ```
> > void g1_dma_irq_handler(uint32 code, void *data) {
> >    if(g1_ata_irq_hdl)
> >      g1_ata_irq_hdl(code, data);
> >    else if (cdrom_irq_hdl)
> >      cdrom_irq_hdl(code, data);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should work just fine, and `hardware_init()` does seem like the right place for it. It would just need to be excluded for NAOMI. That should be an easier and less risky structure without the overhead from swapping them in and out.
> 
> Do not need exclude it for NAOMI, because any ATA IDE devices also can be used on it. And still doesn't clear how to do that. OK, `g1_ata_irq_hdl` and `cdrom_irq_hdl` is extern variables that can be set. But it means I should always enable IRQ's for G1 ATA even if no one devices are used... Are you sure that external variables and always-on interrupts are a better solution than what we have now? I'll do it, no problem, I just want to clarify. Yes, the current solution is not the best, the problem here is in the interrupt API, but what you propose will either add some handlers like vblank or will look like a hack. Or I don't understand something correctly.

Hrm. Ok. I was thinking that since the dma init is not optional in `cdrom_init`, then the interrupts would be on all the time anyways. But you're absolutely right that the case of the no-cdrom no-ata shouldn't need to have the extra interrupts set. The original solution should be fine, and since it's all entirely internal we can see about how to refactor in the future. I do agree @pcercuei that it's not a good idea to have that nesting but even the simple alternatives seem like they'd have to be messier in other ways.

I think what it would probably need would be a new `g1bus.c` or similar to hold the pieces shared by g1ata and cdrom so it's not just forced in `hardware.c`. OR a more simple/standard interface for adding multiple handlers for a single asic interrupt (perhaps migrating/enhancing the vblank dispatcher up and making the current vblank handler system backed by a generic one for asic).

I finished my review of it otherwise and it seems great and ready aside from a possible set of race conditions in the manual aborting of commands.

kernel/arch/dreamcast/hardware/cdrom.c
DC-SWAT: We do not process DMA errors in G1ATA or here yet. They basically only happen when the protection system is configured incorrectly. And in general, they only happen when the protection system is configured incorrectly.

kernel/arch/dreamcast/hardware/cdrom.c
DC-SWAT: OK, I fixed it I guess.

DC-SWAT:
> I think what it would probably need would be a new `g1bus.c` or similar to hold the pieces shared by g1ata and cdrom so it's not just forced in `hardware.c`. OR a more simple/standard interface for adding multiple handlers for a single asic interrupt (perhaps migrating/enhancing the vblank dispatcher up and making the current vblank handler system backed by a generic one for asic).
> 

Yeah, I've thought about g1bus.c few times already and it's definitely time to do it. But for now I want to get the rest of the changes out first.

QuzarDC (approved):

PR #973, opened 2025-04-17
fs_iso9660: Get rid of global FD table

pcercuei:
Instead of using one big FD table with a fixed size, and using an index into that table as the token passed among the backend functions, use a pointer to the file handle structure as the token itself.

The FD table can then be removed, and replaced by a single-linked list of the file handle structs, so that we can still keep track of what's opened in case we need to invalidate all the handles at once.

This change removes the arbitrary limitation on the number of files that can be opened with the iso9660 filesystem.

This is my alternative to #964.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_iso9660.c
+++ b/kernel/arch/dreamcast/fs/fs_iso9660.c
@@ -42,6 +42,7 @@ ISO9660 systems, as these were used as references as well.
 #include <ctype.h>
 #include <string.h>
 #include <strings.h>
+#include <sys/queue.h>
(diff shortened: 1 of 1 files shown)


QuzarDC (approved):
It occurs to me that here, and elsewhere, it might be useful to have a standardized sort of debug assert that a pointer being passed in is one that resulted from a malloc. With the static table there was the test against `FS_CD_MAX_FILES` which played the same role of ensuring that the handle being passed in was plausibly valid.

Anyways, just some ramblings inspired by this. All looks good, well better than the minimal change in #964

PR #976, opened 2025-04-25
Rework linker scripts + add code subsections

pcercuei:
This PR adds a way to override the start address from the shlelf.xc linker script, which makes it usable for the Naomi. Therefore the shlelf-naomi.xc (which was already lagging behind) can be dropped.

The last commit introduce icache-aligned code subsections. These can be used to enforce some specific functions to be sitting close together in the binary, to make sure that they won't flush one another outside the icache.


DC-SWAT:
It would be great to be able to set it directly to a variable in the Makefile. Since for example I compile for Naomi under the same address as for Dreamcast.

pcercuei:
> It would be great to be able to set it directly to a variable in the Makefile. Since for example I compile for Naomi under the same address as for Dreamcast.

You can, just pass `-Wl,--defsym=LOAD_OFFSET=0x8c020000` to the linker line.

utils/ldscripts/shlelf.xc
```diff
@@ -4,16 +4,51 @@ OUTPUT_FORMAT("elf32-shl", "elf32-shl",
 OUTPUT_ARCH(sh)
 ENTRY(start)
 SEARCH_DIR("/usr/local/dc-new/sh-elf/sh-elf/lib");
```
QuzarDC, line 6: Could you have this get replaced by the NAOMI version which went to the actual default toolchain install location? Or even better, have it be based on `KOS_CC_BASE`? Alternatively if it has no use here (it seems unlikely to considering it's been this junk path since it was added), just drop it.

pcercuei: I'll just drop it then.

QuzarDC (approved):

PR #977, opened 2025-04-25
sound: Fix loading of ARM firmware

pcercuei:
The offending commit 1d507167d ("Use proper typing for pointer math.") had the following change:

-extern uint8_t snd_stream_drv[];
-extern uint8_t snd_stream_drv_end[];
+extern uintptr_t snd_stream_drv;
+extern uintptr_t snd_stream_drv_end;

and was later computing the size of the firmware by doing the difference (snd_stream_drv_end - snd_stream_drv).

However, the behaviour was changed; before, it computed the difference between two addresses, and now computed the difference between the two uintptr_t values located at these addresses.
Therefore the difference should have been changed to (&snd_stream_drv_end - &snd_stream_drv).

Instead of just reverting the commit, update the code to use a blob generated by bin2c instead of bin2o, since we already moved to bin2c for romdisks.

The source blob generated by bin2c can be included directly, which means that the symbols can be used directly without having to use "extern".

### DIFF
--- a/kernel/arch/dreamcast/sound/Makefile
+++ b/kernel/arch/dreamcast/sound/Makefile
@@ -20,5 +20,10 @@ include $(KOS_BASE)/Makefile.prefab
 
 arm/stream.drv: subdirs
 
-snd_stream_drv.o: arm/stream.drv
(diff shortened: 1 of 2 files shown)


QuzarDC (approved):
Thanks for the catch. Looks good.

PR #978, opened 2025-04-26
Implemented driver for SCI interface.

DC-SWAT:
As part of the development of Naomi, I decided to write a driver for the SCI interface of SH4. It is brought out to the CN1 connector there along with the SCIF. Although it can also be brought out on the Dreamcast by soldering several wires to the motherboard. This interface turned out to be very interesting, since its speed is higher than I previously thought.

Tested on both Dreamcast and Naomi.

Both UART and SPI modes are supported at all possible speeds and protocols, PIO and DMA.
-   Maximum with internal clock in asynchronous mode (UART) : ~1,562,500 bps
-   Maximum with internal clock in synchronous mode (SPI): ~12.5 Mbps

**Reliability tests:**
- In UART mode I have up to 256,000 bitrate without errors with my CP2102 adapter.
- In SPI mode the maximum bitrate works without errors with all tested SD cards in full-duplex mode.

**SD card tests:**
- SPI PIO mode: ~1300 KB/s on per-sector writing and ~735 KB/s on per-sector reading without CRC check.
- SPI DMA mode: ~1000 KB/s on per-sector writing (not best case for DMAC), read can't be implemented.
- SPI Hybrid (DMA + PIO) mode: ~1450 KB/s on per-sector reading.

Unfortunately, I found an unpleasant flaw in the SCI with sync (SPI like) mode. It cannot work in the receive-only mode and must always transmit a byte to have a clock signal. Without transmission the clock signal is completely unclear, especially at high speeds. At the same time, it can work in the write-only mode, since it is the write that initiates the clock signals. If there were two free DMA channels, it would be possible to organize full-duplex DMA, but unfortunately we only have one channel at our disposal.

**Key differences between SCI SPI and Standard SPI**

Bit Order:
- SCI SPI: Fixed LSB-first transmission, requires software bit reversal
- Standard SPI: Configurable bit order, typically supports MSB-first natively

Hardware Implementation:
- SCI SPI: Repurposed UART hardware with limited SPI capabilities
- Standard SPI: Dedicated controller with full SPI feature set

Chip Select Management:
- SCI SPI: No dedicated CS pin, uses GPIO or borrowed SCIF RTS
- Standard SPI: Integrated CS/SS pin handling

Clock Generation:
- SCI SPI: Must transmit dummy bytes to generate clock for receiving (undocumented)
- Standard SPI: Can generate clock independently of transmission data

Protocol Modes:
- SCI SPI: Fixed clock polarity/phase, not all devices are suitable
- Standard SPI: Full support for all CPOL/CPHA combinations (modes 0-3)

**Guide for SCI mod on the Sega Dreamcast and NAOMI**
http://www.dc-swat.ru/blog/hardware/1143.html


kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -565,6 +637,74 @@ sci_result_t sci_read_data(uint8_t *data, size_t len) {
...
+    /* Configure DMA */
+    dma_config_t config = sci_dma_tx_config;
+    config.callback = callback ? callback : sci_dma_dummy_callback;
+
+    /* Prepare DMA source and destination */
+    dma_addr_t src = dma_map_src(data, len);
+    dma_addr_t dst = hw_to_dma_addr(SCTDR1_ADDR);
+
+    /* Flush data cache to ensure data is written to memory before DMA reads it */
+    dcache_flush_range((uintptr_t)data, len);
```
pcercuei, line 658: This is already ensured by `dma_map_src`, so you can skip the explicit flush.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,1160 @@
...
+    uint16_t sptr;
+    uint16_t fcr;
+
+    if(cs_mode == SCI_SPI_CS_GPIO) {
+        PCTRA = PCTRA & ~SCI_SPI_CS_PIN_MASK;
+    }
+    else if(cs_mode == SCI_SPI_CS_RTS) {
+        fcr = SCFCR2;
+        fcr &= ~SCFCR_MCE;
+        SCFCR2 &= ~SCFCR_MCE;
```
QuzarDC, line 425: Should this be `SCFCR2 = fcr;` ? 

Also, does the modem control option interfere with regular usage of the SCIF? I couldn't quite tell from the manual. It seems as though even if it doesn't in the hardware, there would need to be adjustments in the SCIF driver here:

https://github.com/KallistiOS/KallistiOS/blob/2762ceff10c06ff491290630faeee1584b7c96a3/kernel/arch/dreamcast/hardware/scif.c#L65-L76

and elsewhere to not just wipe out `SCSPTR2` and the same for `SCFCR2`. The same would be true for the scif-spi driver as well.

I presume the idea was for them to be mutually exclusive.

DC-SWAT: I can use SCIF for logging as before, no issues. This is the one of main goal for me, keep logging on SCIF and use SCI for SD card.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,900 @@
...
+
+    return SCI_OK;
+}
+
+static inline void spi_delay(int count) {
+    int i;
+    for(i = 0; i < count; i++) {
+        __asm__("nop");
+    }
+}
```
pcercuei, line 164: Don't make your own loop, please use `timer_spin_delay_ns()` instead.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,900 @@
...
+
+    if(initialized) {
+        sci_shutdown();
+    }
+
+    /* Enable SCI module (disable standby) */
+    if(STBCR & STBCR_SCI_STP) {
+        STBCR &= ~STBCR_SCI_STP;
+        /* Delay to allow the module to power up */
+        spi_delay(10000);
```
pcercuei, line 294: Ugh, no. Don't busy-wait for such big times. Use `thd_sleep()` instead.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,900 @@
...
+        result = check_sci_errors();
+        if(result != SCI_OK) {
+            return result;
+        }
+
+        if(++timeout_cnt > SCI_MAX_WAIT_CYCLES) {
+            dbglog(DBG_ERROR, "SCI: Timeout waiting for TDRE\n");
+            return SCI_ERR_TIMEOUT;
+        }
+    } while(!(SCSSR1 & TDRE));
```
pcercuei, line 375: This code (the "wait for bit in SCSSR1") is repeated multiple times, you can factorize that.

I also dislike the busy-waiting, you should probably yield the thread in the loop.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,900 @@
...
+    }
+    else {
+        smr = *scsmr1;
+    }
+
+    smr |= config;
+
+    if(scsmr1 == NULL) {
+        SCSMR1 = smr;
+        spi_delay(10000);
```
pcercuei, line 398: `thd_sleep()`

pcercuei (changes requested):

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -565,6 +628,68 @@ sci_result_t sci_read_data(uint8_t *data, size_t len) {
...
+        return SCI_ERR_NOT_INITIALIZED;
+    }
+
+    if(data == NULL || len == 0) {
+        return SCI_ERR_PARAM;
+    }
+
+    /* Configure DMA */
+    dma_config_t config = sci_dma_tx_config;
+    config.callback = callback ? callback : sci_dma_dummy_callback;
```
pcercuei, line 642: `config.callback = callback;`
And you can trash `sci_dma_dummy_callback`.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -91,8 +99,37 @@ static sci_mode_t sci_mode = SCI_MODE_NONE;
...
+    .callback = NULL
+};
+
+static dma_config_t sci_dma_rx_config = {
+    .channel = DMA_CHANNEL_1,
+    .request = DMA_REQUEST_SCI_RECEIVE,
+    .unit_size = DMA_UNITSIZE_8BIT,
+    .src_mode = DMA_ADDRMODE_FIXED,
+    .dst_mode = DMA_ADDRMODE_INCREMENT,
+    .transmit_mode = DMA_TRANSMITMODE_CYCLE_STEAL,
```
pcercuei, line 122: TX in burst and RX in cycle steal?

kernel/arch/dreamcast/hardware/sci.c
DC-SWAT: No, without callback you using thd_pass() instead of genwait.

kernel/arch/dreamcast/hardware/sci.c
DC-SWAT: Yes, because when writing, the CPU does not access the SCI, but when reading, I use hybrid mode, where DMA and CPU access the SCI at the same time.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,900 @@
...
+    else {
+        dbglog(DBG_ERROR, "SCI: Invalid mode\n");
+        return SCI_ERR_PARAM;
+    }
+
+    SCSMR1 = scsmr1;
+    SCBRR1 = scbrr1;
+
+    /* Allow time for the changes to take effect */
+    spi_delay(10000);
```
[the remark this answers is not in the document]
DC-SWAT: Yeah I'll replace all spi_delay() with thd_sleep()

kernel/arch/dreamcast/hardware/sci.c
DC-SWAT: If you pass a thread there, the speed will drop a lot. These cycles are short.
Also not sure about separate function, it also can drop the speed, but I'll check it.

pcercuei: The code will be smaller, it will fit better in the icache ;) It can go both ways.  I honestly don't think it will have any effect on speed whatsoever.

kernel/arch/dreamcast/hardware/sci.c
DC-SWAT: 10k nop's in a loop it's 0.3ms :)

kernel/arch/dreamcast/hardware/sci.c
DC-SWAT: Each loop is 20-30 ns, so here not more than 0.3ms.

kernel/arch/dreamcast/hardware/sci.c
megavolt85: this will reduce the reading speed by 5 times

kernel/arch/dreamcast/hardware/sci.c
megavolt85: using cycle steal for TX reduces performance.
using burst for RX results in a hang.

megavolt85: the problem is that the SCI port does not have a classic SPI mode, reading without writing is impossible, so a hybrid mode is used, data is read via DMA, and writing occurs programmatically, which is why the DMA mode is different

pcercuei: You don't need to convince me - document that in the code.

kernel/arch/dreamcast/hardware/sci.c
pcercuei: What? Calling a function drops the reading speed by a factor 5?

kernel/arch/dreamcast/hardware/sci.c
pcercuei: Then use `timer_spin_delay_us(300)`;

kernel/arch/dreamcast/hardware/sci.c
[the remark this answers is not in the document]
pcercuei: If it's only called in the init, we don't really care if it's a bit slower.

kernel/arch/dreamcast/hardware/sci.c
megavolt85: no, i'm talk about thd_pass() in this cycle

pcercuei: Yes, that's fine without it. It would still be better to factorize the code.

megavolt85: in fact, hundreds of tests have been conducted, even replacing while(len--) with for(i=0; i<len; i++) reduces the reading speed by 7-10 KB/s.
at the moment, the absolute maximum has been squeezed out of the SCI interface
I agree that macros and delay functions can be replaced in some places, but nothing can be changed in the read/write functions, any change leads to a regression in speed

pcercuei: The code should be factorized either way. Then you can mark the function "always_inline" if it really makes the code faster.

kernel/arch/dreamcast/hardware/sci.c
DC-SWAT: > You don't need to convince me - document that in the code.

I don't think any additional explanations are needed here, because you already wrote them yourself in the DMAC header in the DMA modes. This is exactly the principle by which the modes were chosen.

pcercuei: @DC-SWAT the fact that I asked why RX is cycle steal means it wasn't obvious to me; especially since it's a different mode than TX. I would not expect it to hang actually.

pcercuei: Fine, I guess it makes sense, no need to comment.

DC-SWAT: No, I already wrote in detail :D

kernel/arch/dreamcast/hardware/sci.c
DC-SWAT: > Should this be `SCFCR2 = fcr;` ?

Yeah!

kernel/arch/dreamcast/hardware/sci.c
pcercuei: Feel free to fix it in DMAC then. Use `&channels_cfg[channel]` as the genwait object instead of the callback, and then it works even when no callback is set.

kernel/arch/dreamcast/hardware/sci.c
DC-SWAT: OK, I tried moving these loops into always_inline functions, but this led to 15% performance loss. No matter how you look at it, this generates unnecessary instructions in the code that aren't needed at all.
The only thing I could extract is the TEND flag, which has almost no impact, but it's not even worth making a commit for this.

kernel/arch/dreamcast/hardware/dmac.c
```diff
@@ -114,13 +114,11 @@ static bool dma_is_running(dma_channel_t channel) {
...
-            if(channels_cfg[channel]->callback)
-                genwait_wait(channels_cfg[channel]->callback, "DMA complete wait", 0, NULL);
-            else
-                thd_pass();
-        }
+    if(dma_is_running(channel) && !irq_inside_int()) {
+        if(channels_cfg[channel]->callback)
+            genwait_wait(channels_cfg[channel]->callback, "DMA complete wait", 0, NULL);
+        else
+            genwait_wait(&channels_cfg[channel], "DMA complete wait", 0, NULL);
```
pcercuei, line 121: This looks completely broken.

kernel/arch/dreamcast/hardware/reverse_bits.s
```diff
@@ -0,0 +1,35 @@
+! KallistiOS ##version##
+!
+! arch/dreamcast/hardware/reverse_bits.s
+! Copyright (C) 2025 Ruslan Rostovtsev
+!
+! Optimized SH4 assembler function for reversing bits of a byte.
```
pcercuei, line 6: We already have `bit_reverse()`, did you try that?

kernel/arch/dreamcast/hardware/sci.c
pcercuei: Sorry, but I call that BS. The SH4 being what it is, every time you touch the code somewhere the resulting program can work much faster or much slower.

Have a look at the diff, between the inlined code and a "static inline always_inline", the generated code is exactly the same: https://godbolt.org/z/dYr9Ydx3T

DC-SWAT: I don't know, It's just slower in fact and that it.

kernel/arch/dreamcast/hardware/dmac.c
DC-SWAT: Why? You says me use &channels_cfg[channel] as object for genwait. For what here while cycle without thd_pass()?

kernel/arch/dreamcast/hardware/reverse_bits.s
DC-SWAT: Hmm, didn't see this.
It's do the same, but with cmp and bf, so it's slower.
Can I just replace it with my own version or?

pcercuei: You can rename it to `bit_reverse8` for instance, and move it to `kernel/arch/dreamcast/math/math.s` alongside `bit_reverse`.

kernel/arch/dreamcast/hardware/dmac.c
pcercuei: And what's singaling `&channels_cfg[channel]` here?

Did you even test that code?

kernel/arch/dreamcast/hardware/reverse_bits.s
DC-SWAT: I didn't immediately notice that you were flipping 32 bits.

kernel/arch/dreamcast/hardware/dmac.c
DC-SWAT: > Did you even test that code?

Not yet, just followed your instructions.
To be honest, I also found it strange. But I got carried away with something else.

DC-SWAT: Ahh forgot about wake_all...

DC-SWAT: Fixed

kernel/arch/dreamcast/hardware/reverse_bits.s
DC-SWAT: Done

kernel/arch/dreamcast/hardware/dmac.c
pcercuei: You don't need to conditionally genwait on the callback or on the channels_cfg[channel] object, just unconditionally use the latter as the genwait object (that you wait on, and that you signal).

You should also keep this structure in `dma_wait_complete()`:
```c
while(dma_is_running(channel)) {
    if(!irq_inside_int())
        genwait_wait(&channels_cfg[channel], "DMA complete wait", 0, NULL);
}
```

Because it should busy-wait in the case where it was called inside an interrupt (it really shouldn't, but we're handling the case anyway), and otherwise it avoids spurious wakeups (which shouldn't happen either, but better be safe than sorry).

DC-SWAT: Got it. Done.

ljsebald:
As I asked in the other PR (#989), do you have some details on how one would actually perform the modification to actually use this on a regular console? It might be a good idea if we had some documentation of it available for anyone who actually wants to play around with this (including myself).

As I can't actually test this myself, I'll trust the results you've provided in the PR and approve it.

ljsebald (approved):

QuzarDC (changes requested):
Overall looks fine. Some minor things I think could be cleaned up and questions to be sure about the behavior.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,1179 @@
...
+    .dst_mode = DMA_ADDRMODE_INCREMENT,
+    .transmit_mode = DMA_TRANSMITMODE_CYCLE_STEAL,
+    .callback = NULL
+};
+
+static __always_inline void clear_sci_errors() {
+    SCSSR1 &= ~(ORER | FER | PER);
+}
+
+static __always_inline sci_result_t check_sci_errors() {
```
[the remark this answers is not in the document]
megavolt85: for asynchronous mode, clearing these flags is not required, but it does not contradict their clearing in any way, you can of course add a separate function, but this is just an unnecessary waste of memory

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,1179 @@
...
+        reg &= ~(TE | RE);
+        SCSCR1 = reg;
+    }
+    else if(transfer_mode != mode) {
+        reg = SCSCR1;
+
+        if(transfer_mode) {
+            reg &= ~(TE | RE);
+            SCSCR1 = reg;
+            timer_spin_delay_ns(1500);
```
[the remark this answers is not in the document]
megavolt85: Снимок экрана от 2025-05-10 00-03-50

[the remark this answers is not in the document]
QuzarDC: Thank you! I figured it might be there somewhere, but I was just missing it.

kernel/arch/dreamcast/hardware/sci.c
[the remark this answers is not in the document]
DC-SWAT: This is a common interface-wide cleaning function, there is no point in dividing it, so its naming is fine. It's just not used everywhere, where the SPI mode cycle is very sensitive, even inline functions are not used, because for some reason they slow down. So in the end this function is only used in asynchronous mode and in initialization.

kernel/arch/dreamcast/hardware/sci.c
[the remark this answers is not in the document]
DC-SWAT: > Thank you! I figured it might be there somewhere, but I was just missing it.

By the way, the documentation doesn't say that a small pause is needed when switching modes. Without it, things start to glitch. But I'm not sure how much of a pause is needed, it's very small, so here I've made it with some reserve. Switching from mode to mode does not happen very often.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,1179 @@
...
+    if(!initialized || sci_mode != SCI_MODE_UART) {
+        return SCI_ERR_NOT_INITIALIZED;
+    }
+
+    if(data == NULL) {
+        return SCI_ERR_PARAM;
+    }
+
+    /* Wait until data is available */
+    timeout_cnt = 0;
```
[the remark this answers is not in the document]
DC-SWAT: Look on 559 line.

kernel/arch/dreamcast/hardware/sci.c
```diff
@@ -0,0 +1,1179 @@
...
+            dbglog(DBG_ERROR, "SCI: Timeout waiting for TDRE in SPI rw byte\n");
+            return SCI_ERR_TIMEOUT;
+        }
+    }
+
+    /* Send a byte and clear TDRE flag */
+    SCTDR1 = byte;
+    SCSSR1 &= ~TDRE;
+
+    timeout_cnt = 0;
```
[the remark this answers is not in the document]
DC-SWAT: Yeah need remove it.

kernel/arch/dreamcast/hardware/sci.c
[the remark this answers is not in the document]
DC-SWAT: OK, I found in another place second timeout

kernel/arch/dreamcast/hardware/sci.c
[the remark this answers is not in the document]
QuzarDC: Right, sorry, I meant 'that these are errors only for sync mode/SPI mode' or 'not for async mode', I had just inverted async/sync.

I wasn't advocating a separate function, just something to note why check_sci_errors isn't always being used (which makes sense, to save performance). I guess it makes sense to just leave it to the hardware documentation that notes only ORER is will trigger in other conditions.

kernel/arch/dreamcast/hardware/sci.c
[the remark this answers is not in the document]
QuzarDC: > > Thank you! I figured it might be there somewhere, but I was just missing it.
> 
> By the way, the documentation doesn't say that a small pause is needed when switching modes. Without it, things start to glitch. But I'm not sure how much of a pause is needed, it's very small, so here I've made it with some reserve. Switching from mode to mode does not happen very often.

Ah ok, that's good to know in case conditions are found where it needs to be made slightly bigger in the future.

kernel/arch/dreamcast/hardware/sci.c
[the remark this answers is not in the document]
QuzarDC: Ah ok, it looks like I attached this to the wrong extra timeout clear.

PR #981, opened 2025-04-30
pvr: Move + improve PVR poly headers and types into their own file

pcercuei:
Create enums and types for the various macros relative to the PVR headers and polygon contexts.
Note the change with PVR_FILTER_* values - which does not break API since the register field was moved as well.

Create types for the cmd, mode1, mode2, mode3 fields of a PVR header. Redefine pvr_poly_hdr_t to use these new types, in a fully backwards-compatible manner.

This allows applications to create poly headers directly, without having to go through creating a poly context that is later "compiled". It also greatly improves documentation, if only because of the type system.


pcercuei:
Force-pushed with a tiny documentation fix.

ljsebald (approved):
Nice job. I appreciate the clear documentation and painstaking work to keep it all backwards compatible.

PR #983, opened 2025-05-03
dmac: Drop old DMAC interface and header

pcercuei:
And move the code initializing DMAC on Naomi to the DMAC driver.


ljsebald (approved):
Seems reasonable to me.

I don't mind removing these macros from public view because there aren't particularly compelling reasons for users to be messing around with the DMA registers, tbh.

pcercuei:
Updated the PR to avoid calling `dma_transfer()` on channel 2, even if it was convenient to avoid some (slight) code duplication, it did more than set the registers that were needed to be set. This seemed to work fine, but I've had some strange DMA issues with it, so I reverted to a safer option.

For that reason I tagged this for v2.3.0 just to be sure.

QuzarDC (approved):

PR #984, opened 2025-05-04
Keyboard Cleanup Part 2 + Event Driven API

QuzarDC:
Second split from #503 . Recreates the changes with more atomic commits while integrating additional changes and backwards compatibility.

Includes:
- `kbd_state_t'`s "key matrix" has been modified to store both the current frame and previous frame's key states, which allows for the driver to check for state changes and emit events accordingly. This added zero extra space overhead, since we were storing key states within a byte anyway, and I'm only using two bits of it. This also optimized the key state tracking by setting it up to be a mere bitshift.
- Added an event-driven callback mechanism for registering a callback to be invoked whenever a key changes states (similar to how Qt, GTk, and high-level widgeting frameworks process key events).
- Adjusted libDream's keyboard example to show off event-driven handler mechanism.
- Rename `MAX_PRESSED_KEYS` to be in line with our naming scheme, but leaving deprecated compatibility helper.
- Clean up some internal typing and remove a duplicated set of keymaps for the global queue.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/keyboard.c
+++ b/kernel/arch/dreamcast/hardware/maple/keyboard.c
@@ -428,6 +428,16 @@ static const kbd_keymap_internal_t keymaps[] = {
     }
 };
 
+char kbd_key_to_ascii(kbd_key_t key, kbd_region_t region, kbd_mods_t mods, kbd_leds_t leds) {
+
+    if(mods.ralt || (mods.lctrl && mods.lalt))
+        return keymaps[region - 1].alt[key];
+    else if((mods.raw & KBD_MOD_SHIFT) || leds.caps_lock)
+        return keymaps[region - 1].shifted[key];
+    else
+        return keymaps[region - 1].base[key];
+}
+
 /* The keyboard queue (global for now) */
 static volatile int kbd_queue_active = 1;
 static volatile int kbd_queue_tail = 0, kbd_queue_head = 0;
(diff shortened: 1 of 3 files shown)


pcercuei:
I am not really fond of the event-driven callback mechanism, for the reason that it would run within IRQ context.

If we run the whole maple interrupt handling in a threaded interrupt handler, then I'm fine with it.

QuzarDC:
> I am not really fond of the event-driven callback mechanism, for the reason that it would run within IRQ context.
> 
> If we run the whole maple interrupt handling in a threaded interrupt handler, then I'm fine with it.

So, I agree generally. But that's a big change to slip in here that will likely require non-trivial refactoring as the maple operations are meant to be during the vblank.

pcercuei:
> > I am not really fond of the event-driven callback mechanism, for the reason that it would run within IRQ context.
> > If we run the whole maple interrupt handling in a threaded interrupt handler, then I'm fine with it.
> 
> So, I agree generally. But that's a big change to slip in here that will likely require non-trivial refactoring as the maple operations are meant to be during the vblank.

Yes and no.

The driver's `.periodic` callback is called on vblank. The event API callbacks are called from the "frame transfer done" IRQ.

QuzarDC:
> > > I am not really fond of the event-driven callback mechanism, for the reason that it would run within IRQ context.
> > > If we run the whole maple interrupt handling in a threaded interrupt handler, then I'm fine with it.
> > 
> > 
> > So, I agree generally. But that's a big change to slip in here that will likely require non-trivial refactoring as the maple operations are meant to be during the vblank.
> 
> Yes and no.
> 
> The driver's `.periodic` callback is called on vblank. The event API callbacks are called from the "frame transfer done" IRQ.

Ah ok, yes understood. Those should be much safer to transition to threaded handlers, but I still believe even that would be bigger in scope than being slipped into this. I agree that it should happen, but would end up being paired with reworking the controller callbacks (no need for the handler to spawn threads anymore) and would probably need some refactoring to split into the parts that should stay in the irq and the rest.

kernel/arch/dreamcast/hardware/maple/keyboard.c
```diff
@@ -41,12 +41,14 @@ typedef struct kbd_state_private {
...
-    volatile int queue_len;             /**< \brief Current length of queue. */
-
-    kbd_key_t kbd_repeat_key;           /**< \brief Key that is repeating. */
-    uint64_t kbd_repeat_timer;          /**< \brief Time that the next repeat will trigger. */
+    size_t queue_tail;                     /**< \brief Key queue tail. */
+    size_t queue_head;                     /**< \brief Key queue head. */
+    volatile size_t queue_len;             /**< \brief Current length of queue. */
+
+    struct {
+    kbd_key_t key;          /**< \brief Key that is repeating. */
```
gyrovorbis, line 49:
```suggestion
        kbd_key_t key;          /**< \brief Key that is repeating. */
```

kernel/arch/dreamcast/hardware/maple/keyboard.c
```diff
@@ -41,12 +41,14 @@ typedef struct kbd_state_private {
...
+    uint64_t timeout;       /**< \brief Time that the next repeat will trigger. */
```
gyrovorbis, line 50:
```suggestion
        uint64_t timeout;       /**< \brief Time that the next repeat will trigger. */
```

gyrovorbis (approved):
Just two teensy spacing issues.

PR #991, opened 2025-05-09
New pthreads implementation: Rebased version

darcagn:
Now that BlueCrab has moved #283 from draft status and indicated he's ready to merge, here is a rebased version of that branch without the dozens of merge commits.

This was created by checking out `master` and `libpthread`, sequentially cherry-picking all of the non-merge commits from `libpthread` into `master` and manually resolving conflicts, `diff`ing the cherry-pick version with the merge version, and finally creating one last commit with the few changes that were included in the merge commits. 

Do note that while this compiles fine using the stable toolchain, I have not yet run it to be sure, but I'm opening the PR now to make it available for testing.


ljsebald (approved):
Fine by me to merge this instead of the other one. Approving with the same caveat I mentioned in the other thread.

darcagn:
@Bruceleeto caught an error with `pthread_attr_getname_np()`, so I added a fix.

```
Fix string handling error in pthread_attr_getname_np()
- Fixed erroneous comparison of buflen and thread label sizes
- Added check if buflen is 0 to avoid memory corruption issues
```

PR #992, opened 2025-05-10
Correct and apply the `__pure` attribute

QuzarDC:
Switching the `__pure` define to use [the `pure` attribute](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-pure-function-attribute) to use  rather than [`__const__`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-const-function-attribute) . This seems more in-line with the documentation and should generally be more useful. It'll also be backwards compatible as `__const__` is just a stricter `pure`.

Both generally allow flagging gcc to know that a function's output is entirely dependent on its input and will never vary if the input is the same. This is meant to allow it to better optimize cases where there's multiple calls with the same inputs.

Then went and applied it to everything in the main codebase that seemed to fit the attribute. As a side effect, searching for `__pure` will lead you to functions that are likely targets for further optimization (if they haven't already). A few seemed like they could take advantage of bitop builtins.

EDIT: Rather than fixing our definition of `__pure` I've just dumped it entirely. Its existence predates our newlib integration, which provide this (as well as `__pure2`) in its own `cdefs.h`.

### DIFF
--- a/include/kos/cdefs.h
+++ b/include/kos/cdefs.h
@@ -47,12 +47,6 @@
 #define __noreturn  __attribute__((__noreturn__))
 #endif
 
-#ifndef __pure
-/** \brief  Identify a function that has no side effects other than its return,
-            and only uses its arguments for any work. */
-#define __pure      __attribute__((__const__))
-#endif
-
 #ifndef __unused
 /** \brief  Identify a function or variable that may be unused. */
 #define __unused    __attribute__((__unused__))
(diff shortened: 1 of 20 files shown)


QuzarDC:
Resolved a conflict from a types update, and squashed most of the commits down.

QuzarDC:
Force pushed to resolve conflicts caused by #1102 .

gyrovorbis (approved):
Just went through each function one-by-one, and they all do look like they are `pure` to me!

gyrovorbis (approved):

PR #993, opened 2025-05-11
Add LRA patches for GCC 15.1.1 and GCC 16.0.0-dev

pcercuei:
Import the GCC patches from Oleg's devel/sh-lra tree. Use them for the GCC 15.1.1 stable and 16.0.0 development profiles.

### DIFF
--- a/utils/dc-chain/patches/gcc-15.1.1-0001-SH-Tighten-memory-predicates-and-constraints.diff
+++ b/utils/dc-chain/patches/gcc-15.1.1-0001-SH-Tighten-memory-predicates-and-constraints.diff
@@ -1,0 +1,206 @@
+From c3de2295f0f7959adb0d39a0c96f02606faddb4d Mon Sep 17 00:00:00 2001
+From: Oleg Endo <olegendo@gcc.gnu.org>
+Date: Sun, 29 Sep 2024 21:33:29 +0900
+Subject: [PATCH 01/13] SH: Tighten memory predicates and constraints
+
+In particular, reject invalid hard-regs for memory address registers when using
+LRA.  Unfortunately we need to distingquish between old reload and LRA behaviors
+for the transitional period.  LRA seems to require stricter predicates and
+constraints.
+
+gcc/ChangeLog:
+
+	PR target/55212
+	* config/sh/predicates.md (simple_mem_operand):
+	Use 'satisfies_constraint_Sra'.
+	(post_inc_mem, pre_dec_mem): Use 'satisfies_constraint_Rab'.
+	* config/sh/constraints.md (Rab, Rai, Sgb): New constraints.
+	(Sua, Sdd, Snd, Ssd, Sbv, Sra, Ara, Add): Use Rab and Rai constraints.
+	* config/sh/sync.md (atomic_mem_operand_0, atomic_mem_operand_1): Reject
+	GBR addresses when hard-llcs atomic mode is enabled.
+---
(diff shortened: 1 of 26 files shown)


olegendo:
Thanks for checking it. Those patches will eventually be merged into GCC 16. I'm not sure whether in the current form or in a modified form. There are still some missing patches that I haven't added to my branch.

pcercuei:
> Thanks for checking it. Those patches will eventually be merged into GCC 16. I'm not sure whether in the current form or in a modified form. There are still some missing patches that I haven't added to my branch.

FYI I picked the patches (except the one that makes -mlra the default) from upstream GCC branch `devel/sh-lra`, which seemed to have the same patches as in your last branch.

Given that 16-0.0-dev is a *development* profile, we could actually enable it by default there. At least it would get some decent testing.

olegendo:
> FYI I picked the patches (except the one that makes -mlra the default) from upstream GCC branch devel/sh-lra, which seemed to have the same patches as in your last branch.

The devel/sh-lra branch is stale. I started it and then realized that it doesn't allow forced pushes and rebases for rewriting a patchset. So I've moved on to github early on with this

https://github.com/olegendo/gcc/commits/devel/sh-lra/

darcagn:
`15.1.1` is the development branch for the future `15.2.0` release (expected soon in June). These patches won't be in the `15.2.0` or any 15-series release, so I wouldn't include them unless they were somehow optional. 

Since they are likely to be merged in the 16-series, I'm good with that, but if we do this we need to stay on top of changes, i.e. if Oleg alters the patches in his repo we should update these to make sure we aren't hosting stale patches, or if they are merged upstream we need to remove them here to avoid conflicts.

olegendo:
> Since they are likely to be merged in the 16-series, I'm good with that, but if we do this we need to stay on top of changes, i.e. if Oleg alters the patches in his repo we should update these to make sure we aren't hosting stale patches, or if they are merged upstream we need to remove them here to avoid conflicts.

The LRA patches will definitely go in, but I'm not sure yet in which form. I will be rewriting the patchset (rebase + force-push) while working on them on the github branch before merging it upstream.

darcagn:
> The LRA patches will definitely go in, but I'm not sure yet in which form. I will be rewriting the patchset (rebase + force-push) while working on them on the github branch before merging it upstream.

If you're going to be actively managing a repo/branch then maybe it's better that we don't manage the patches here. To make things easier we can either 
- create a toolchain profile in dc-chain for your repo
- have people use the `16.0.0-dev` profile, but build with `make sh_gcc_git_repo=https://github.com/olegendo/gcc.git sh_gcc_git_branch=devel/sh-lra` which would pull your repo/branch but apply the KOS patches for `16.0.0` to it

olegendo:
Yes, one of those sounds sensible.

pcercuei:
We can have a profile for Oleg's tree, as e.g. `16.0.0-dev-lra`, but this should point to a fixed git hash and not a moving branch (and we can add a version number to the profile name, e.g. `16.0.0-1-dev-lra` if we want to bump to a new commit hash).
As long as we agree on that, I'm fine with a separate profile.

pcercuei:
> > , but this should point to a fixed git hash and not a moving branch
> 
> This won't work for me with rebase & force-push, which I'm using to work on the patch set.

That's why we need to use a fixed git hash in the toolchain profile. Then you can rebase and force-push and it won't affect the toolchain profile.

pcercuei:
Closing as the LRA toolchain is added by #1097.

PR #994, opened 2025-05-11
Add missing include in sci

QuzarDC:
The new pthreads implementation that was merged
directly before the sci work changed some
dependencies so it's needed to explicitly include
thread.h now. Otherwise it's a warning in gcc13
and an error in newer.


gyrovorbis (approved):
Sweet!

darcagn (approved):

PR #997, opened 2025-05-12
Fix mismatched types for ubc callback.

QuzarDC:
`ubc.h` is using the irq_context type but for
some reason wasn't using the typedef'd name
for it, and wasn't including `irq.h` either.

This fixes the use of it in gcc14+.

Threw in the void cast to clean up unused param warning in the example as well.

### DIFF
--- a/examples/dreamcast/basic/breaking/breaking.c
+++ b/examples/dreamcast/basic/breaking/breaking.c
@@ -48,6 +48,9 @@ static atomic_bool handled = false;
 static bool on_break(const ubc_breakpoint_t *bp,
                      const irq_context_t *ctx,
                      void *ud) {
+    /* Don't warn about unused bp */
+    (void)bp;
+
     /* Signal to the outside that the breakpoint has been handled. */
     handled = true;
 
--- a/kernel/arch/dreamcast/include/dc/ubc.h
+++ b/kernel/arch/dreamcast/include/dc/ubc.h
@@ -26,6 +26,7 @@ __BEGIN_DECLS
 
 #include <stdbool.h>
 #include <stdint.h>
+#include <arch/irq.h>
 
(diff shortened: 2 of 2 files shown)


pcercuei:
Alternatively keep the .h as-is and include irq.h in the .c.

QuzarDC:
> Alternatively keep the .h as-is and include irq.h in the .c.

I don't see how that would solve anything. The alternative, I believe, would be to update the example to define the callback it sends as `struct irq_context` rather than the typedef'd `irq_context_t`. Otherwise that would then mean that anyone using ubc would get these errors until they include both `ubc.h` and `irq.h` and get the (quite weird) errors that it's getting until they do.

gyrovorbis (approved):
Yeah, it's either keep the forward declaration in the header file like it was and then include it in the `.c` file OR ditch the forward declaration in the header and include the IRQ header there.

They should be type-compatible, `struct irq_context*` and `irq_context_t*`.

Upside is faster compile times with forward declarations like that, but I'm not picky, cuz I don't think we really do that anywhere else in KOS, so approving. It's fine to me either way.

QuzarDC:
> Yeah, it's either keep the forward declaration in the header file like it was and then include it in the `.c` file OR ditch the forward declaration in the header and include the IRQ header there.
> 
> They should be type-compatible, `struct irq_context*` and `irq_context_t*`.
> 
> Upside is faster compile times with forward declarations like that, but I'm not picky, cuz I don't think we really do that anywhere else in KOS, so approving. It's fine to me either way.

Yea, it's not a bad methodology to have, and might trim some things down, but is just not in line with the way we really do anything else around the API. I could even see it if it was the case that just naming as `struct irq_context*` in the example would have it work fine. But then in order to do anything with the breakpoint you still need to bring in `irq.h` to look into the context.

PR #1000, opened 2025-05-12
Stop dbglog from being included in stdio

QuzarDC:
Following #995 I found that dbglog.h was being exposed in `<stdio.h>` and so many internal headers were being accidentally automatically provided indirectly through dbglog.h out to anything needing stdio. So this PR ensures all uses of dbglog include the header explicitly, then removes it from stdio.h. In it's place, two of the removed includes from #995 in order to reduce the breakage to `kos-ports` (and likely user software) as it seems some types and internal functions were only being exposed through that.

### DIFF
--- a/addons/libkosext2fs/ext2internal.h
+++ b/addons/libkosext2fs/ext2internal.h
@@ -45,13 +45,15 @@ struct ext2fs_struct {
 #define EXT2_FS_FLAG_SB_DIRTY   1
 
 #ifdef EXT2_NOT_IN_KOS
-#include <stdio.h>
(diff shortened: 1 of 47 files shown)


darcagn (commented):
To make a note here, this PR is blocked by:
- https://github.com/KallistiOS/libtsunami/pull/2
- https://github.com/KallistiOS/libparallax/pull/3

darcagn (approved):
Approving this PR, assuming the above 2 fixes get synced to the sourceforge repos. ;)

PR #761, opened 2024-09-17
Introduce CI

drpaneas:
* CI for Ubuntu, MacOS and FreeBSD
 * Running per PR, on master branch

An example of how this looks like, can be found https://github.com/drpaneas/KallistiOS/actions/runs/10901703715

There are some modification in the scripts, so they know if this is BSD or Linux system, which is important for executing shell scripts properly. Then, some modifications to the BSD installation packages were made, since there are some changes there and certain libraries were needed. Eventually, for Mac and BSD, the CI is building against gcc-14.2.0 (the previous ones do not build).

PR #1003, opened 2025-05-12
CI: Add autobuilder

pcercuei:
Build the default toolchain + KOS using default environ.sh.

This takes about ~40min to complete, and will build the master branch (when it's updated) and PRs.

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -1,0 +1,23 @@
+name: Linux build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: install-dependencies
+      run: |
+        sudo apt install -y libjpeg-dev libpng-dev
+
+    - name: build
+      run: |
+        cp $GITHUB_WORKSPACE/utils/dc-chain/Makefile.default.cfg $GITHUB_WORKSPACE/utils/dc-chain/Makefile.cfg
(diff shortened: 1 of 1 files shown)


pcercuei:
For some reason the KallistiOS repo does not have "actions"?

QuzarDC:
Wanted to make sure to link here: #761 and #85 . Both have a extra info from the GH docs and may provide useful strategies. iirc @cepawiel 's implementation was eventually set up so as to have the toolchain environments pre-built and have the action rebuild KOS+ports+examples over multiple environments. Saving a lot of time as certainly the toolchain setup is the most expensive part of the process. Additionally drpaneas provided usage stats that showed that even in some of our faster times it would be well under limits: https://github.com/KallistiOS/KallistiOS/pull/761#issuecomment-2406085160

I've got no problem with starting nightly, other than the general reasoning for it over per-merge. I don't see why it would be a negative thing to be out of free time in a month unless it started billing (which it doesn't, you just get cut off). Certainly there's a mechanism by which that could be managed (for instance, don't run automatically if we're over 90% or such) where we could reserve time for fewer more important runs. 

And even if not, the negative is ... being where we are now.

pcercuei:
Ok, I have another suggestion.
We build *all* toolchain profiles into a prebuilt docker image, that is used as the base for building KOS in CI.

This would allow building KOS against all toolchains we have a profile for, and since building KOS is fast, it would probably take 10min max.

The docker image can be built manually at first - then we can automate that later, so that it's updated any time something in dc-chain changed.

PR #1004, opened 2025-05-13
dmac: Added dma_is_running() and dma_transfer_abort()

DC-SWAT:
Also use it in SCI driver.

### DIFF
--- a/kernel/arch/dreamcast/hardware/dmac.c
+++ b/kernel/arch/dreamcast/hardware/dmac.c
@@ -108,7 +108,7 @@ static void dma_irq_handler(irq_t code, irq_context_t *context, void *d) {
     }
 }
 
-static bool dma_is_running(dma_channel_t channel) {
+bool dma_is_running(dma_channel_t channel) {
     uint32_t chcr = dmac_read(channel, DMA_REG_CHCR);
 
     return (chcr & (REG_CHCR_TRANSFER_END | REG_CHCR_DMAC_EN)) == REG_CHCR_DMAC_EN;
@@ -165,4 +165,11 @@ size_t dma_transfer_get_remaining(dma_channel_t channel) {
     uint32_t tcr = dmac_read(channel, DMA_REG_TCR);
 
     return tcr * dma_unit_size[unit_size];
+}
+
+void dma_transfer_abort(dma_channel_t channel) {
+    irq_disable_scoped();
+
+    dmac_write(channel, DMA_REG_CHCR, 0);
+    genwait_wake_all((void *)&channels_cfg[channel]);
 }
--- a/kernel/arch/dreamcast/hardware/sci.c
+++ b/kernel/arch/dreamcast/hardware/sci.c
@@ -1021,7 +1021,7 @@ sci_result_t sci_spi_read_data(uint8_t *rx_data, size_t len) {
             if(status & ORER) {
                 SCSSR1 &= ~ORER;
                 sci_set_transfer_mode(0);
-                dbglog(DBG_ERROR, "SCI: Overrun error\n");
+                dbglog(DBG_ERROR, "SCI: Overrun error in SPI read data\n");
                 return SCI_ERR_OVERRUN;
             }
         } while(!(status & RDRF));
@@ -1082,7 +1082,7 @@ sci_result_t sci_spi_dma_write_data(const uint8_t *data, size_t len, dma_callbac
         while(!(SCSSR1 & TEND)) {
             if(++timeout_cnt > SCI_MAX_WAIT_CYCLES) {
                 sci_set_transfer_mode(0);
-                dbglog(DBG_ERROR, "SCI: Timeout waiting for TEND in SPI read data\n");
+                dbglog(DBG_ERROR, "SCI: Timeout waiting for TEND in SPI DMA write\n");
                 return SCI_ERR_TIMEOUT;
             }
         }
@@ -1109,7 +1109,6 @@ sci_result_t sci_spi_dma_read_data(uint8_t *data, size_t len, dma_callback_t cal
 
     /* Configure DMA */
     dma_config_t config = sci_dma_rx_config;
-    config.callback = callback;
 
     /* Prepare DMA */
     dma_addr_t src = hw_to_dma_addr(SCRDR1_ADDR);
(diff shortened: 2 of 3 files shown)


pcercuei:
Looks good.

pcercuei (approved):

kernel/arch/dreamcast/hardware/dmac.c
```diff
@@ -166,3 +166,7 @@ size_t dma_transfer_get_remaining(dma_channel_t channel) {
 
     return tcr * dma_unit_size[unit_size];
 }
+
+void dma_transfer_abort(dma_channel_t channel) {
+    dmac_write(channel, DMA_REG_CHCR, 0);
```
QuzarDC, line 171: Will this raise an irq? If not, it might make sense to also wake anything waiting on the channel like the handler does.

pcercuei: I disagree. IMHO it is up to the caller code to handle that should it decide to abort a DMA transfer.

QuzarDC: > I disagree. IMHO it is up to the caller code to handle that should it decide to abort a DMA transfer.

Is there a way to do so? That would require access directly to `channels_cfg`. Or do you mean they'd just generally be responsible for cleaning up after and not explicitly waking threads up?

DC-SWAT: > > I disagree. IMHO it is up to the caller code to handle that should it decide to abort a DMA transfer.
> 
> Is there a way to do so? That would require access directly to `channels_cfg`. Or do you mean they'd just generally be responsible for cleaning up after and not specifically waking threads up?

What is the purpose of access to channel configuration? It is not necessary.

QuzarDC: > > > I disagree. IMHO it is up to the caller code to handle that should it decide to abort a DMA transfer.
> > 
> > 
> > Is there a way to do so? That would require access directly to `channels_cfg`. Or do you mean they'd just generally be responsible for cleaning up after and not specifically waking threads up?
> 
> What is the purpose of accessing the channel configuration? It is not necessary to clear them, they are always overwritten.

As @pcercuei noted the user would be responsible for waking up threads waiting on the dma transfer and they're `gen_wait`ing on `channels_cfg` objects.

I'm thinking the idea might be not that the exact behavior is able to be replicated, but rather some other means would be used to recover the waiting threads, but I'm not sure what it'd be.

DC-SWAT: > > > > I disagree. IMHO it is up to the caller code to handle that should it decide to abort a DMA transfer.
> > > 
> > > 
> > > Is there a way to do so? That would require access directly to `channels_cfg`. Or do you mean they'd just generally be responsible for cleaning up after and not specifically waking threads up?
> > 
> > 
> > What is the purpose of accessing the channel configuration? It is not necessary to clear them, they are always overwritten.
> 
> As @pcercuei noted the user would be responsible for waking up threads waiting on the dma transfer and they're `gen_wait`ing on `channels_cfg` objects.
> 
> I'm thinking the idea might be not that the exact behavior is able to be replicated, but rather some other means would be used to recover the waiting threads, but I'm not sure what it'd be.

Hmm, well then we can simply cancel this genwait in abort(), without calling a callback, if there is one at all.

QuzarDC: > > > > > I disagree. IMHO it is up to the caller code to handle that should it decide to abort a DMA transfer.
> > > > 
> > > > 
> > > > Is there a way to do so? That would require access directly to `channels_cfg`. Or do you mean they'd just generally be responsible for cleaning up after and not specifically waking threads up?
> > > 
> > > 
> > > What is the purpose of accessing the channel configuration? It is not necessary to clear them, they are always overwritten.
> > 
> > 
> > As @pcercuei noted the user would be responsible for waking up threads waiting on the dma transfer and they're `gen_wait`ing on `channels_cfg` objects.
> > I'm thinking the idea might be not that the exact behavior is able to be replicated, but rather some other means would be used to recover the waiting threads, but I'm not sure what it'd be.
> 
> Hmm, well then we can simply cancel this genwait in abort(), without calling a callback, if there is one at all.

That is what I'd imagined there. I see now though that there's a secondary concurrency issue. If another dma transfer is started on the same channel before the waiting thread is able to check `is_running` then it will go right back into waiting:

https://github.com/KallistiOS/KallistiOS/blob/332612268bde1da60a5d59c3d9b4ae697075cd09/kernel/arch/dreamcast/hardware/dmac.c#L120-L125

This seems like the kind of thing that might be usual as a common response to having to abort a timeout is to retry. 

Is there a reason to wrap the genwait in a while loop? If the thread returns from the genwait then that should signal that the transfer has been completed and I don't believe there's a circumstance where it'd be expected to wait again. Perhaps here's where the interface for the user to interact would be? `genwait_wake_all_err` could be used and `dma_wait_complete` could return an error if it stopped but wasn't completed.

DC-SWAT: The loop here is for the case when the call is made from an interrupt. This shouldn't happen, but the author decided to make sure of it this way. We've already discussed this @pcercuei in PR with SCI.
`genwait_wake_all_err()` is not needed here, since the DMA did not fail, we simply cancelled it at our own request.

DC-SWAT: > > Hmm, well then we can simply cancel this genwait in abort(), without calling a callback, if there is one at all.
> 
> That is what I'd imagined there. I see now though that there's a secondary concurrency issue. If another dma transfer is started on the same channel before the waiting thread is able to check `is_running` then it will go right back into waiting:
> 
> https://github.com/KallistiOS/KallistiOS/blob/332612268bde1da60a5d59c3d9b4ae697075cd09/kernel/arch/dreamcast/hardware/dmac.c#L120-L125
> 
> This seems like the kind of thing that might be usual as a common response to having to abort a timeout is to retry.
> 
> Is there a reason to wrap the genwait in a while loop? If the thread returns from the genwait then that should signal that the transfer has been completed and I don't believe there's a circumstance where it'd be expected to wait again. Perhaps here's where the interface for the user to interact would be? `genwait_wake_all_err` could be used and `dma_wait_complete` could return an error if it stopped but wasn't completed.

The loop here is for the case when the call is made from an interrupt. This shouldn't happen, but the author decided to make sure of it this way. We've already discussed this with @pcercuei in the SCI PR.
genwait_wake_all_err() is not needed here, since the DMA did not fail, we simply cancelled it at our own request.

QuzarDC (approved):
Will put in a separate bug for the race condition I described as it's really not tied to this functionality.

PR #1005, opened 2025-05-13
Roll back bad pointer math update for arch_exec_at.

QuzarDC:
From #923. I clearly need to take a remedial course in these casting operations, as I made the same mistake with the aica driver's.


pcercuei (approved):

DC-SWAT:
Brilliant work guys, but the fix had to be like this :D
`uint32_t    *values = &buffer[(tvals - tstart) / 4];`

PR #346, opened 2023-11-09
Sound streams fixes and improvements.

DC-SWAT:
- Fixed multistream support.
- Fixed non-16-bit streams support.
- Fixed stream prefill.
- Used single channel for mono.
- Clean playback.
- Some cleanup and refactoring.

PR #1006, opened 2025-05-13
Deprecate `snd_stream_prefill`

QuzarDC:
`snd_stream_fill` (and thus prefill) is currently not able to be run before the first call to `snd_stream_start` as the number of channels is undefined at the time. This behavior worked partially prior to #346 as the logic for most of the code was `if(channels == 2) ... else{}` so channels being zero didn't cause issues.

What I've done is deprecate `snd_stream_prefill` leaving a warning stub in its place and folding its operation into `snd_stream_start`. Additionally added minor error checking to prevent potential crashes

Found while debugging [vcdc](https://gitlab.completext.com/Dreamcast/vcdc) which did:
`snd_stream_init();
shnd = snd_stream_alloc(mpv_callback, sbsize);
snd_stream_prefill(shnd);
`

### DIFF
--- a/kernel/arch/dreamcast/sound/snd_stream.c
+++ b/kernel/arch/dreamcast/sound/snd_stream.c
@@ -315,27 +315,6 @@ void snd_pcm16_split_sq(uint32_t *data, uintptr_t left, uintptr_t right, size_t 
     g2_unlock(ctx);
 }
 
-
-/* Prefill buffers -- implicitly called by snd_stream_start() */
-void snd_stream_prefill(snd_stream_hnd_t hnd) {
-    strchan_t *stream;
-
-    CHECK_HND(hnd);
-    stream = &streams[hnd];
-
-    /* No way to request data, or not yet allocated. */
-    if((!stream->get_data && !stream->req_data) ||
-                            (!stream->channels)) {
-        return;
-    }
-
-    snd_stream_fill(hnd, 0, stream->buffer_size / 2);
-    snd_stream_fill(hnd, stream->buffer_size / 2, stream->buffer_size / 2);
-
-    /* Start playing from the beginning */
-    stream->last_write_pos = 0;
-}
-
 /* Initialize stream system */
 int snd_stream_init(void) {
     return snd_stream_init_ex(2, SND_STREAM_BUFFER_MAX);
@@ -536,8 +515,12 @@ static void snd_stream_start_type(snd_stream_hnd_t hnd, uint32_t type, uint32_t 
         }
     }
 
-    /* Prefill buffers */
-    snd_stream_prefill(hnd);
+    /* As long as there's a way to get/request data, prefill buffers */
+    snd_stream_fill(hnd, 0, streams[hnd].buffer_size / 2);
+    snd_stream_fill(hnd, streams[hnd].buffer_size / 2, streams[hnd].buffer_size / 2);
+
+    /* Start playing from the beginning */
+    streams[hnd].last_write_pos = 0;
 
     /* Make sure these are sync'd (and/or delayed) */
     snd_sh4_to_aica_stop();
--- a/kernel/arch/dreamcast/sound/snd_stream.c
+++ b/kernel/arch/dreamcast/sound/snd_stream.c
@@ -323,7 +323,9 @@ void snd_stream_prefill(snd_stream_hnd_t hnd) {
     CHECK_HND(hnd);
     stream = &streams[hnd];
 
-    if(!stream->get_data && !stream->req_data) {
+    /* No way to request data, or not yet allocated. */
+    if((!stream->get_data && !stream->req_data) ||
+                            (!stream->channels)) {
         return;
     }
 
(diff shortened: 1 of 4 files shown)


DC-SWAT:
I think this is all wrong.
Firstly, you can't call a fill before the start, you don't know how many channels will play, you can specify 1 or 2. The variable max_channels is responsible for the maximum possible, but not for the current number of channels in the stream.
And vcdc works incorrectly.

QuzarDC:
> I think this is all wrong. Firstly, you can't call a fill before the start, you don't know how many channels will play, you can specify 1 or 2. The variable max_channels is responsible for the maximum possible, but not for the current number of channels in the stream. And vcdc works incorrectly.

I don't doubt that vcdc could be working incorrectly, but it did manage to do it incorrectly for many years without crashing. I updated it anyways. But it was trying to work based on what prefill says it does: `This function prefills the stream buffers before starting it.` It seems what may be the case is that prefill should have been nop'd and made internal if you're saying it's not possible to use externally? Or when is it meant to be used if not before? Maybe just the description needs to change?

In either case, it shouldn't mean allowing a crash. We could leave the checks in place without the part that updates the channels in alloc, and at least avoid crashing if prefill/fill get called ahead of start.

DC-SWAT:
> > I think this is all wrong. Firstly, you can't call a fill before the start, you don't know how many channels will play, you can specify 1 or 2. The variable max_channels is responsible for the maximum possible, but not for the current number of channels in the stream. And vcdc works incorrectly.
> 
> I don't doubt that vcdc could be working incorrectly, but it did manage to do it incorrectly for many years without crashing. I updated it anyways. But it was trying to work based on what prefill says it does: `This function prefills the stream buffers before starting it.` It seems what may be the case is that prefill should have been nop'd and made internal if you're saying it's not possible to use externally? Or when is it meant to be used if not before? Maybe just the description needs to change?
> 
> In either case, it shouldn't mean allowing a crash. We could leave the checks in place without the part that updates the channels in alloc, and at least avoid crashing if prefill/fill get called ahead of start.

I think the issue is that the stream_prefill function was never meant to be public in the first place, but for some reason, it was made so. Its documentation explicitly says it’s not recommended to call it manually, yet vcdc (and others) do it anyway, going against that recommendation.

What’s more, there’s no real point in doing so, since stream_start calls prefill by itself — there’s no need to do it manually. In fact, it shouldn’t be called before the stream is started at all, because the number of channels is not yet defined.

Whoever did this was relying on the current implementation details — that it would work simply because of how it’s written — not because it’s actually safe or intended to be used that way.

QuzarDC:
> > > I think this is all wrong. Firstly, you can't call a fill before the start, you don't know how many channels will play, you can specify 1 or 2. The variable max_channels is responsible for the maximum possible, but not for the current number of channels in the stream. And vcdc works incorrectly.
> > 
> > 
> > I don't doubt that vcdc could be working incorrectly, but it did manage to do it incorrectly for many years without crashing. I updated it anyways. But it was trying to work based on what prefill says it does: `This function prefills the stream buffers before starting it.` It seems what may be the case is that prefill should have been nop'd and made internal if you're saying it's not possible to use externally? Or when is it meant to be used if not before? Maybe just the description needs to change?
> > In either case, it shouldn't mean allowing a crash. We could leave the checks in place without the part that updates the channels in alloc, and at least avoid crashing if prefill/fill get called ahead of start.
> 
> I think the issue is that the stream_prefill function was never meant to be public in the first place, but for some reason, it was made so. Its documentation explicitly says it’s not recommended to call it manually, yet vcdc (and others) do it anyway, going against that recommendation.
> 
> What’s more, there’s no real point in doing so, since stream_start calls prefill by itself — there’s no need to do it manually. In fact, it shouldn’t be called before the stream is started at all, because the number of channels is not yet defined.
> 
> Whoever did this was relying on the current implementation details — that it would work simply because of how it’s written — not because it’s actually safe or intended to be used that way.

Those are great points. I'll take this back to just deprecate the public `stream_prefill` with a stub (so that it does nothing rather than doing useless things or potentially crashing). Then the initialization isn't needed, though I think it'd still be a good idea to have the error checks since it's nicer to have the error than potentially crash.

kernel/arch/dreamcast/sound/snd_stream.c
```diff
@@ -534,8 +515,14 @@ static void snd_stream_start_type(snd_stream_hnd_t hnd, uint32_t type, uint32_t
         }
     }
 
-    /* Prefill buffers */
-    snd_stream_prefill(hnd);
+    /* As long as there's a way to get/request data, prefill buffers */
+    if(streams[hnd].get_data || streams[hnd].req_data) {
```
DC-SWAT, line 519: This is already checked at the beginning of the function.

QuzarDC: Thanks, fixed.

DC-SWAT:
Looks good now. Virtual approval :D

QuzarDC:
> Looks good now. Virtual approval :D

An additional small question, since you're the main user of the exports: is it better to remove the export at the time of marking deprecated, or when the deprecated function is actually removed later? In this case it shouldn't matter as there shouldn't be anything actually needing it, but for a general process.

DC-SWAT:
> > Looks good now. Virtual approval :D
> 
> An additional small question, since you're the main user of the exports: is it better to remove the export at the time of marking deprecated, or when the deprecated function is actually removed later? In this case it shouldn't matter as there shouldn't be anything actually needing it, but for a general process.

Well, it seems to me that it would be more correct to delete it only after the function itself is completely deleted.
But on the other hand, this way we will get rid of them faster in the binary file (with library support). After all, it doesn't matter whether they are used or not, if they are in the export - they are linked into the binary.

darcagn (approved):

PR #1007, opened 2025-05-15
Fix for executing binary files.

DC-SWAT:
Fix for that "fix": #1005
Reordered back fs_pty_shutdown() after: #680

### DIFF
--- a/kernel/arch/dreamcast/kernel/exec.c
+++ b/kernel/arch/dreamcast/kernel/exec.c
@@ -33,7 +33,7 @@ void arch_exec_at(const void *image, uint32_t length, uint32_t address) {
                 tvals  = (uintptr_t)_arch_exec_template_values;
     size_t      tcount = (tend - tstart) / 4;
     uint32_t    buffer[tcount];
-    uint32_t    *values = buffer + (_arch_exec_template_values - _arch_exec_template);
+    uint32_t    *values = &buffer[(tvals - tstart) / 4];
     size_t      i;
 
     assert((tend - tstart) % 4 == 0);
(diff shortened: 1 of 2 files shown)


gyrovorbis (dismissed):
I'm not sure wtf the deal is with PTY, but the pointer arithmetic fix for `arch_exec()` looks good!

QuzarDC (changes requested):
To fix fs_pty, could you please try instead to move the `fs_shutdown` in `arch_auto_shutdown` to be where it was before directly between shutting down rnd and fs_ramdisk?

#680 moved it to be the last thing called after the shutdown of the individual vfses. The top-level fs-shutdown is meant to happen first, to clear references to files and request that the vfses each close their files, putting it at the end means that there will be problems in shutting down as fs shutdown will request closing of the different files after the vfses have been shut down entirely.

The old ordering was as:
```
    KOS_INIT_FLAG_CALL(fs_dcload_shutdown);
    KOS_INIT_FLAG_CALL(vmu_fs_shutdown);
    if (!KOS_PLATFORM_IS_NAOMI)
        KOS_INIT_FLAG_CALL(fs_iso9660_shutdown);
#if defined(__NEWLIB__) && !(__NEWLIB__ < 2 && __NEWLIB_MINOR__ < 4)
    fs_rnd_shutdown();
#endif
    fs_shutdown();
    fs_ramdisk_shutdown();
    KOS_INIT_FLAG_CALL(fs_romdisk_shutdown);
    fs_pty_shutdown();
    fs_null_shutdown();
    fs_dev_shutdown();
    thd_shutdown();
    rtc_shutdown();
```

This likely also could cause problems with shutting down with open romdisk or ramdisk files, not just pty.

DC-SWAT:
OK, I reorder shutdown only for fs_pty and it seems work now.

QuzarDC (approved):

PR #1008, opened 2025-05-15
Include arch/types.h in thread.h to fix toolchain building

darcagn:
Otherwise the toolchain fails to build not knowing what `tid_t` or `prio_t` are.


gyrovorbis (approved):
Yep, needs to be there.

PR #1009, opened 2025-05-16
Fix type errors in dc/fifo.h and dc/g2bus.h

darcagn:
I'm getting compilation errors when building the Rust examples due to missing types. Sure enough, these particular headers need `arch/types.h` included. Rather than making a PR to add the `arch/types.h` include, however, I just switched the types in these headers to standard types.


gyrovorbis (approved):
At least we're fixing da build with good changes to make. lol.

QuzarDC (approved):

PR #1011, opened 2025-05-16
Eliminate non-standard fixed-width integer types from examples

darcagn:
As it will be a goal to remove the non-standard fixed width integer types from the repository in KOS 2.3.0, I am starting off by removing them from the `examples` tree so as to not mislead anyone into using them in new code or future examples. 

This is **NOT** an automated search-and-replace job. I used the following `grep` to find instances and scanned through the files _manually_ so I could avoid any unforeseen issues that may pop up when doing a blanket search-and-replace:

`grep -r -l -E '(vuint(64|32|16|8)|vint(64|32|16|8)|uint(64|32|16|8)([^_]|$)|int(64|32|16|8)([^_]|$))' .`

This eliminates roughly 25% of the total files using non-standard fixed width integer types in the KOS repo.


QuzarDC:
Pushed to resolve the conflicts caused by some warning cleanups that were done in the meantime to help this along.

QuzarDC (approved):

gyrovorbis (approved):
Not seeing any issues scanning through everything... nice.

PR #1016, opened 2025-05-17
scif: disable MCE by default

buhman:
Enabling MCE is not a reasonable library default because the typical USB-UARTs that are connected to the Dreamcast have large ~1024 byte TX/RX buffers and the maximum Dreamcast SCIF baud rate is slow compared to the capabilities of these chips.

Given the use-case for scif.c appears to be to allow for synchronously-blocking debug prints over scif, from a library/policy perspective MCE should be considered a "specialized" configuration.

For KallistiOS users, the behavior prior to this commit was:

- serial transmission may be silently permanently disabled due to the SCIF CTS pin being disconnected or tied high

- it is not possible to use RTS/CTS as GPIO pins while also using scif.c

### DIFF
--- a/kernel/arch/dreamcast/hardware/scif.c
+++ b/kernel/arch/dreamcast/hardware/scif.c
@@ -242,8 +242,8 @@ int scif_init(void) {
     for(i = 0; i < 800000; i++)
         __asm__("nop");
 
-    /* Unreset, enable hardware flow control, triggers on 8 bytes */
-    SCFCR2 = 0x48;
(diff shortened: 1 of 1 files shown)


DC-SWAT:
I can confirm that without this mode everything works fine, but the CTS still needs to be grounded, otherwise it won’t work. Only RTS can be used as GPIO if you want keep SCIF working for debugging.

QuzarDC (approved):

PR #1023, opened 2025-05-20
Deprecate custom `ptr_t` type.

QuzarDC:
`ptr_t` is one of our custom types from <arch/types> and is obsoleted by the standard `uintptr_t` which has the same definition and usage.

Replace its use where found with `uintptr_t`, ensure each use has <stdint> included, and do one extra minor stdint cleanup along the way in genwait.

Working towards #840 

edit: force-pushed to fix a mistake in the commit message.

### DIFF
--- a/addons/libkosext2fs/fs_ext2.c
+++ b/addons/libkosext2fs/fs_ext2.c
@@ -1703,7 +1703,7 @@ int fs_ext2_stat(vfs_handler_t *vfs, const char *path, struct stat *st,
     /* Root directory ext2 device */
     if(len == 0 || (len == 1 && *path == '/')) {
         memset(st, 0, sizeof(struct stat));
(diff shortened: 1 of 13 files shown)


addons/libkosext2fs/fs_ext2.c
darcagn, line 1: On line 1726 in this file:
`    st->st_dev = (dev_t)((ptr_t)vfs);`
must become
`    st->st_dev = (dev_t)((uintptr_t)vfs);`

QuzarDC: Wow. I slipped up due to #1021 breaking building of it and not thinking to rebuild without fanalyzer or -Werror to get it figured out.

Force-pushed with this added.

pcercuei (approved):

darcagn (approved):

PR #1024, opened 2025-05-20
Move gitignore declarations to sub-project dirs

darcagn:
In several places in the KOS tree, sub-projects are hosting their own `.gitignore` file, but we also have a massive `.gitignore` file in the root.

To address this discrepancy, this PR moves all `.gitignore` declarations to files in the root of each sub-project's directory.

Additionally, some people (obi_wan_hes_me.jpg) put the kos-ports tree in `$KOS_BASE/ports` so that they can couple together separate KOS installations with their own kos-ports tree (for building them together with different flags, or building them together to review PRs that affect both). So this PR also adds `/ports/` to the root `.gitignore` file for those people.


QuzarDC:
> Additionally, some people (obi_wan_hes_me.jpg) put the kos-ports tree in `$KOS_BASE/ports` so that they can couple together separate KOS installations with their own kos-ports tree (for building them together with different flags, or building them together to review PRs that affect both). So this PR also adds `/ports/` to the root `.gitignore` file for those people.

Any way to have that be conditional to the defined ports dir? That way, no matter what the name is or where it is in the tree it's ignored. Additionally/alternatively, would you want to have that become the new standard in the setup guide/environ sample?

darcagn:
> > Additionally, some people (obi_wan_hes_me.jpg) put the kos-ports tree in `$KOS_BASE/ports` so that they can couple together separate KOS installations with their own kos-ports tree (for building them together with different flags, or building them together to review PRs that affect both). So this PR also adds `/ports/` to the root `.gitignore` file for those people.
> 
> Any way to have that be conditional to the defined ports dir? That way, no matter what the name is or where it is in the tree it's ignored. Additionally/alternatively, would you want to have that become the new standard in the setup guide/environ sample?

I don't think there's any way to make it conditional on the environment, but even if there were, it wouldn't be very helpful because I'd still see an untracked file in my terminal whenever I enter a KOS dir that I don't have sourced. For example, the `?1` here:

Screenshot_20250520_193051

As for wanting `$KOS_BASE/ports` to be the default, I do think it'd be better that way, and it would integrate better with how we have `make kos-ports_all`, `make kos-ports_clean`, `make kos-ports_distclean` targets that can be run from the KOS root. We could even make it a git submodule so it's pulled in at the same time as KOS. The only downside I can think of immediately is if someone wanted to delete their KOS dir without deleting kos-ports, but they should probably rebuild anyway if they do so that's not a big deal.

QuzarDC (approved):
Great idea. This should also prevent any issues that arise with possible name conflicts or if things get moved around/renamed and needing to update the global ignore. Would also make for utils and examples to be friendlier when copied out of the project.

PR #1026, opened 2025-05-20
CI: Add autobuilder

pcercuei:
Build KallistiOS with all toolchains we have a profile for (minus the WinXP one).

The toolchains themselves are not built in Github actions; they are manually built using the Docker script provided in dc-chain and pushed to Docker Hub.

This replaces #1003.

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -1,0 +1,34 @@
+name: Linux build
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        toolchain_profile:
+          - stable
+          - 13.3.0
+          - 14.2.0
+          - 15.1.0
+          - 13.3.1-dev
+          - 14.2.1-dev
+          - 15.1.1-dev
+          - 16.0.0-dev
+
+    runs-on: ubuntu-latest
+    container: pcercuei/kallistios-toolchain:${{ matrix.toolchain_profile }}
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: install-dependencies
+      run: |
+        apk --update add --no-cache coreutils
+
+    - name: build
+      run: |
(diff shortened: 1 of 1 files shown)


pcercuei:
I updated the scripts.

Now, the toolchains are also built in CI, after a *manual* trigger (click on the corresponding button on Github's UI). All the toolchains we have a profile for will be built as Docker images, and pushed to Docker Hub. Those images can then be used in the CI for building KallistiOS itself.

The autobuilder script for KallistiOS was updated as well to drop the hardcoded list of toolchain profiles; those are now automatically read from the dc-chain folder.

pcercuei:
Updated once again, addressing a rather important issue.

Previously, upon changes to the files in CI or in dc-chain, the CI script would re-build the toolchain, but only update the copy on Docker Hub when the code was merged to the master branch. This means that in these circumstances, KallistiOS was not built with the new version of the toolchain, but with the cached one on Docker Hub.

Address this issue by always using the local Docker image that is built by the script, if available, and fall back to downloading it from Docker Hub if we're just building KallistiOS without the toolchain.

pcercuei:
Rebased to latest master, and the CI script now correctly passes the "platform" variable to the Docker script.

Some profiles seem to fail because of failed downloads.

pcercuei:
Rebased on top of master, no changes.

pcercuei:
@QuzarDC Github does not count CI time for public repos, see: https://docs.github.com/en/billing/concepts/product-billing/github-actions#how-use-of-github-actions-is-measured

Disabling verbose logging is certainly possible, but we'd lose to option to easily see what's going on on build error ("easily" being slightly relative as the toolchain builds is multi-threaded so the output can be mangled). I also don't think it would affect much; Github's UI is not a terminal, it's very possible that internally it redirects to file already. But I can hardcode it for a quick test and see how it performs.

I updated the CI script with just an option set in the file filter to ignore .md files.

pcercuei:
I just checked, building the toolchain on Github's CI with `verbose=0` is not faster in any way, it is even slightly worse for some reason.

Issue #1010, opened 2025-05-16
Pthread support for pthread_getschedparam and pthread_setschedparam missing?


GPF:
Does the new pthread.a not support these functions ?

[ 19%] Building C object CMakeFiles/SDL3-static.dir/src/thread/pthread/SDL_systhread.c.obj
/home/gpf/code/dreamcast/SDL3/src/thread/pthread/SDL_systhread.c: In function ‘SDL_SYS_SetThreadPriority’:
/home/gpf/code/dreamcast/SDL3/src/thread/pthread/SDL_systhread.c:199:9: error: implicit declaration of function ‘pthread_getschedparam’; did you mean ‘pthread_setschedprio’? [-Wimplicit-function-declaration]
  199 |     if (pthread_getschedparam(thread, &policy, &sched) != 0) {
      |         ^~~~~~~~~~~~~~~~~~~~~
      |         pthread_setschedprio
/home/gpf/code/dreamcast/SDL3/src/thread/pthread/SDL_systhread.c:253:32: error: implicit declaration of function ‘sched_get_priority_min’ [-Wimplicit-function-declaration]
  253 |         sched.sched_priority = sched_get_priority_min(policy);
      |                                ^~~~~~~~~~~~~~~~~~~~~~
/home/gpf/code/dreamcast/SDL3/src/thread/pthread/SDL_systhread.c:255:32: error: implicit declaration of function ‘sched_get_priority_max’ [-Wimplicit-function-declaration]
  255 |         sched.sched_priority = sched_get_priority_max(policy);
      |                                ^~~~~~~~~~~~~~~~~~~~~~
/home/gpf/code/dreamcast/SDL3/src/thread/pthread/SDL_systhread.c:277:9: error: implicit declaration of function ‘pthread_setschedparam’; did you mean ‘pthread_setschedprio’? [-Wimplicit-function-declaration]
  277 |     if (pthread_setschedparam(thread, policy, &sched) != 0) {
      |         ^~~~~~~~~~~~~~~~~~~~~
      |         pthread_setschedprio
make[2]: *** [CMakeFiles/SDL3-static.dir/build.make:2347: CMakeFiles/SDL3-static.dir/src/thread/pthread/SDL_systhread.c.obj] Error 1

darcagn:
It does not, but it wouldn't be difficult for us to add it.

DC-SWAT:
[...]

PR #1033, opened 2025-05-21
libpthread: Add support for pthread_getschedparam() and pthread_setschedparam()

darcagn:
This only supports round robin policy getting/setting right now, but it would help things compile that try to use this function (see #1010).

### DIFF
--- a/addons/libpthread/pthread_getschedparam.c
+++ b/addons/libpthread/pthread_getschedparam.c
@@ -1,0 +1,31 @@
+/* KallistiOS ##version##
+
+   pthread_getschedparam.c
+   Copyright (C) 2025 Eric Fradella
+*/
+
+#include "pthread-internal.h"
+#include <pthread.h>
+#include <sched.h>
+#include <errno.h>
+#include <kos/thread.h>
+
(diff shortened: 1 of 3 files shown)


addons/libpthread/pthread_getschedparam.c
```diff
@@ -0,0 +1,31 @@
...
+    if(!thd)
+        return EINVAL;
+
+    if(!policy)
+        return EINVAL;
+
+    if(!param)
+        return EFAULT;
+
+    *policy = SCHED_RR;
```
gyrovorbis, line 26: Should be `SCHED_FIFO`, pretty sure. Our scheduler is just first in, first out, when it comes to threads of the same priority.

darcagn: I wasn't entirely sure, but i decided to go with `SCHED_RR` based on this comment:
<https://github.com/KallistiOS/KallistiOS/blob/40acafbb21bfaf6e75fa06df75aaf762079d2dd0/kernel/thread/thread.c#L611>

gyrovorbis (approved):
Looking gud!

PR #1035, opened 2025-05-22
Added CMake installation include paths for KOS-ports to toolchain.

gyrovorbis:
Modified dreamcast.toolchain.cmake to automatically configure CMake's two environment variables controlling where files get copied to when installing (CMAKE_INSTALL_INCLUDEDIR, CMAKE_INSTALL_LIBDIR) to automatically point to the proper locations within kos-ports.

This is the most reasonable default, imho, for a user wishing to "install" their project into the KOS repo. For example, I'm using this with libSH4ZAM to give it a rule to easily add it as a kos-port so people can use/test before it becomes official.

### DIFF
--- a/utils/cmake/dreamcast.toolchain.cmake
+++ b/utils/cmake/dreamcast.toolchain.cmake
@@ -1,6 +1,6 @@
 # CMake Toolchain file for targeting the Dreamcast or NAOMI with CMake.
 #  Copyright (C) 2023 Luke Benstead
-#  Copyright (C) 2023, 2024 Falco Girgis
+#  Copyright (C) 2023, 2024, 2025 Falco Girgis
 #  Copyright (C) 2024 Donald Haase
 #  Copyright (C) 2024 Paul Cercueil
 #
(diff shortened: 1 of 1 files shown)


pcercuei:
I think you should set `CMAKE_INSTALL_PREFIX` instead.

Also, from a quick look here I'm not sure it sets the default, but forces it to what you hardcode without letting the user set a different path in ccmake later.

gyrovorbis:
> I think you should set `CMAKE_INSTALL_PREFIX` instead.
> 
> Also, from a quick look here I'm not sure it sets the default, but forces it to what you hardcode without letting the user set a different path in ccmake later.

I didn't realize `CMAKE_INSTALL_PREFIX` was a thing, but with that, wouldn't we need to also still set the path for the actual includes to go into `include` and archives to go into `lib` subdirectories?

Also this is how I have to use `install()` in my `CMakeLists.txt` (to not require too new of a version):
```cmake
install(TARGETS sh4zam
        INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}")
 ```
 The fact that I still have to specify that variable by hand tells me I can override it within here? Or do you mean if I were to set that manually by passing it to CMake directly, the toolchain would override it now and it wouldn't work?

darcagn:
> a user wishing to "install" their project into the KOS repo

That's what `addons` is for: https://github.com/KallistiOS/KallistiOS/tree/master/addons#installing-addons

We really should be utilizing it more and advertising it. External projects can put headers in `addons/include` and install built libraries to `addons/lib/$KOS_ARCH`. You can also clone a repo directly into `addons` and have it build/clean whenever you build/clean KOS.

pcercuei:
Nevermind `CMAKE_INSTALL_PREFIX`, I think you need to set `CMAKE_SYSTEM_PREFIX_PATH`: https://cmake.org/cmake/help/latest/variable/CMAKE_SYSTEM_PREFIX_PATH.html#variable:CMAKE_SYSTEM_PREFIX_PATH

This should be copied to the project's default `CMAKE_INSTALL_PREFIX`, which is the base used for `CMAKE_INSTALL_INCLUDEDIR` and `CMAKE_INSTALL_LIBDIR` (which are relative to the prefix path anyway).

QuzarDC (approved):

QuzarDC:
@gyrovorbis This looks ready as-is, but did you want to see about @pcercuei 's further feedback on setting arch?

QuzarDC:
Seems no further reviewing was forthcoming, so just getting this in as-is.

PR #1039, opened 2025-05-22
Update Changelog and Relnotes for v2.2.0

QuzarDC:
Looking for feedback and especially if anyone had items they wanted added/removed/edited/rephrased/clarified.

Had to remake this since the previous one's branch name triggered protection rules for version branches, which would prevent force-pushing.


gyrovorbis (approved):
Sorry it took so long. Read every word of it. Well written, and great job covering everything.

gyrovorbis (approved):

Issue #1031, opened 2025-05-21
Cleanup warnings in KOS documentation


darcagn:
```
/opt/toolchains/dc/kos/include/kos/cdefs.h:34: warning: missing title after \defgroup system_attributes
/opt/toolchains/dc/kos/include/kos/cdefs.h:176: warning: missing title after \defgroup system_compat
/opt/toolchains/dc/kos/include/kos/cdefs.h:209: warning: missing title after \defgroup system_helpers
/opt/toolchains/dc/kos/include/pthread.h:251: warning: More #endif's than #if's found.
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/arch/irq.h:60: warning: missing title after \defgroup Context
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/perf_monitor.h:120: warning: end of file while inside a group
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/pvr/pvr_header.h:364: warning: end of file while inside a group
/opt/toolchains/dc/kos/include/kos/regfield.h:9: warning: Found non-existing group 'kernel' for the command '@ingroup', ignoring command
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/maple/keyboard.h:251: warning: End of list marker found without any preceding list items
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/maple/keyboard.h:251: warning: End of list marker found without any preceding list items
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/sound/sound.h:222: warning: argument 'chn' of command @param is not found in the argument list of snd_get_pos(unsigned int ch)
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/sound/sound.h:222: warning: The following parameter of snd_get_pos(unsigned int ch) is not documented:
  parameter 'ch'
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/sound/sound.h:233: warning: argument 'chn' of command @param is not found in the argument list of snd_is_playing(unsigned int ch)
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/sound/sound.h:233: warning: The following parameter of snd_is_playing(unsigned int ch) is not documented:
  parameter 'ch'
/opt/toolchains/dc/kos/kernel/arch/dreamcast/include/dc/maple/keyboard.h:251: warning: End of list marker found without any preceding list items
[...]

PR #1043, opened 2025-05-22
Correct various doxygen warnings/issues.

QuzarDC:
In a few files group end markers were missing, in a few special characters needed escaping, and others had incorrect names in the doc vs the actual code.

Partially addresses #1031 , but still leaves a few that I'm not 100% sure on how to correct. Marked as bug since some of the warnings did reflect things that was breaking parts of the documentation pages.


include/kos/cdefs.h
```diff
@@ -31,7 +31,7 @@
 #   warning Your GCC is too old. This will probably not work right.
 #endif
 
-/** \defgroup system_attributes
+/** \defgroup system_attributes System Attribute Defines
```
gyrovorbis, line 34:
```suggestion
/** \defgroup system_attributes Attribute Defines
```

include/kos/cdefs.h
```diff
@@ -173,7 +173,7 @@
 
 /** @} */
 
-/** \defgroup system_compat
+/** \defgroup system_compat System Compatibility Defines
```
gyrovorbis, line 176:
```suggestion
/** \defgroup system_compat Compatibility Defines
```

include/kos/cdefs.h
```diff
@@ -206,7 +206,7 @@
 
 /** @} */
 
-/** \defgroup system_helpers
+/** \defgroup system_helpers System Helper Macros
```
gyrovorbis, line 209:
```suggestion
/** \defgroup system_helpers Helper Macros
```

gyrovorbis (commented):
k, so these might seem like weird requests, but it's because fully qualifying the subtopic names really bloats the nice tree view navigation, by duplicating "System" at each layer:
image

QuzarDC:
> k, so these might seem like weird requests, but it's because fully qualifying the subtopic names really bloats the nice tree view navigation, by duplicating "System" at each layer: image

So that makes perfect sense, but perhaps also looks like might not really belong in the same kind of 'system' group as the rest, since these are kind of more like 'environment' or something? I certainly wouldn't think to look in the same place as IRQs for '__inline' or whatnot.

Alternatively, it does seem like it fits with Name Manager and Library. Maybe a separation is needed here for arch specific stuff?

gyrovorbis (approved):
Yeah, that all totally works for me!

PR #1045, opened 2025-05-22
Added flag for disabling TLS for individual threads.

gyrovorbis:
I thought of a potential optimization the other day for how we handle TLS that could lead to some RAM savings... Say that we KNOW a thread is just some little simple task that is never accessing any TLS variables. If we were able to tell the kernel this is the case, we could avoid allocating a TLS segment that will never be used for that particular thread.

...and we already have two very good use-cases immediately within the kernel: the reaper and idle threads are not going to be accessing any TLS variables, yet they're getting their own copies of them allocated upon their creation.

1) Added THD_DISABLE_TLS to kthread_flags_t for signaling to the kernel
   that we're not using TLS for a particular thread.
2) Added kthread_attr_t::disable_tls flag for telling the kernel we
   don't want TLS when spawning a thread and passing its attributes.
    - notice it's the last member of the struct, so I don't break backwards compat, also it's the NEGATIVE so a 0 default value keeps the old functionality
3) Added checks to thread.c when creating and destroying threads to see
   whether THD_DISABLE_TLS is used, and if so, skip TLS segment
processing.
4) Made the [reaper] and [idle] threads, by default, disable TLS.

I validated these changes by rerunning my TLS test located in: examples/dreamcast/basic/threading/compiler_tls. Everything looks good!

### DIFF
--- a/include/kos/thread.h
+++ b/include/kos/thread.h
@@ -4,7 +4,7 @@
    Copyright (C) 2000, 2001, 2002, 2003 Megan Potter
    Copyright (C) 2009, 2010, 2016, 2023 Lawrence Sebald
    Copyright (C) 2023 Colton Pawielski
-   Copyright (C) 2023, 2024 Falco Girgis
+   Copyright (C) 2023, 2024, 2025 Falco Girgis
 
 */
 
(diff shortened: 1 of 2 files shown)


gyrovorbis:
Ugh. I thought I responded to this... sorry for the delay.

Unfortunately, there's no way to actually signal from the compiler to the kernel that something actually is indeed using TLS or to go the other way around and inform the compiler that a particular thread might not need it. This information really isn't known until run-time, when a particular thread executes some GBR-relative load or store.

Yes, actually, part of the motivation here is that GBR can be repurposed in such a thread within an app that otherwise uses static TLS. You'd be able to set GBR to whatever you wanted it to be within such a thread, and the context switching would simply always keep it associated with the particular thread, and since there's no TLS accesses from the compiler for such a thread, you're free to do what you want with GBR as an extra register.

I would want this to honestly NOT interact with "OS-level" TLS. The way that stuff is allocated is kind of "lazily" so that it only gets created for threads which need it already. This is pretty desirable, while compiler TLS is all-or-nothing. I could even see a scenario where you could purposely use OS-level TLS rather than compiler TLS so that you're free to clobber GBR.

I actually have a little wrapper that maps OS-level TLS to compiler-level TLS semantics for platforms that don't implement our compiler TLS. I could imagine using something like this to force usage of OS-level TLS to free GBR:
https://github.com/gyrovorbis/libgimbal/blob/e4eac5f0240d26df671b63ae09320e21cd56c110/lib/api/gimbal/core/gimbal_tls.h#L40

QuzarDC (approved):

PR #1047, opened 2025-05-23
Primitives cleanup

QuzarDC:
A few sets of cleanups happening here:
1) Removing deprecated dynamic creation options for primitives. These were all deprecated in 2.0.0, so it seems like we're safe to finally remove them in 2.3.0 (13 years later).
2) Update primitive functions to use the `__nonnull` macro. This is provided by newlib's cdefs (no need for us to redefine it in ours) and applies the corresponding [gcc attribute](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-nonnull-function-attribute) which allows the generation of a warning if null is passed to the function and can help optimize their use.
3) Correct minor naming issue in sem.c/h to keep them consistent with each other.
4) Mark arguments for test/accessor functions as `const`.
5) Use an unsigned int for mutex type to simplify bounds checks.

@gyrovorbis - I think this might need or benefit from some doxygen specifics on how to handle the attribute. Is there a way so that the attribute being on a function would mark it with some text? "Pointer parameters to this function cannot be null" or such?


kernel/thread/rwsem.c
```diff
@@ -340,11 +316,11 @@ int rwsem_read_tryupgrade(rw_semaphore_t *s) {
 }
 
 /* Return the current reader count */
-int rwsem_read_count(rw_semaphore_t *s) {
+int rwsem_read_count(const rw_semaphore_t *s) {
```
gyrovorbis, line 319: OMG THANK YOU.

kernel/thread/sem.c
```diff
@@ -183,7 +151,7 @@ int sem_signal(semaphore_t *sm) {
 }
 
 /* Return the semaphore count */
-int sem_count(semaphore_t *sm) {
+int sem_count(const semaphore_t *sm) {
```
gyrovorbis, line 154: YES, GOOD CATCH!

include/kos/mutex.h
```diff
@@ -84,34 +83,20 @@ typedef struct kos_mutex {
 #define MUTEX_TYPE_OLDNORMAL    1   /**< \brief Alias for MUTEX_TYPE_NORMAL */
 #define MUTEX_TYPE_ERRORCHECK   2   /**< \brief Error-checking mutex type */
 #define MUTEX_TYPE_RECURSIVE    3   /**< \brief Recursive mutex type */
+#define MUTEX_TYPE_DESTROYED    4   /**< \brief Mutex that has been destroyed */
```
gyrovorbis, line 86: The only thing that I was doing in my cleanups that you haven't also done is just making these into an enumeration. I think it would be in-line with the rest of the codebase and would be type-safe.

QuzarDC: So I described the tradeoff on your PR for the same : https://github.com/KallistiOS/KallistiOS/pull/1066#discussion_r2155235979

What I did here was flip it around a bit so that it would be an unsigned int and a single check for validity. Making it an enum, I believe forces it to be a signed int which would undo that. I know it's possible to force it otherwise but I figured that'd run into issues with toolchain usage (since typed enums are C23). I guess it's not a big difference, but it seemed like a fine tradeoff to me because the only place the bespoke type would be used is in the `mutex_init` which we sort of discourage the use of in place of the initializer defines while the reduced checks help on every lock.

include/kos/cond.h
```diff
@@ -60,48 +60,30 @@ __BEGIN_DECLS
...
 */
-int cond_init(condvar_t *cv);
+int cond_init(condvar_t *cv) __nonnull_all;
 
 /** \brief  Free a condition variable.
 
-    This function frees a condition variable, releasing all memory associated
-    with it (but not with the mutex that is associated with it). This will also
-    wake all threads waiting on the condition.
+    This function destroys a condition variable (but not with the mutex
```
ljsebald:
```suggestion
    This function destroys a condition variable (but not the mutex
```
The with here is definitely not needed, since that was referring to the freeing of memory.

include/kos/mutex.h
gyrovorbis: OH. Okay, that's legit. I see what you're saying.

QuzarDC:
Fixed the cond comment, rebased and corrected conflict with another PR, and removed the cont qualifier for the obj in genwait wait as the thread message object would discard it.

Issue #931, opened 2025-02-01
Cannot compile toolchain - '-m4-single' is not supported


Xeroxxx:
```
make[4]: Entering directory `/opt/toolchains/dc/kos/kernel/arch/dreamcast/fs'
kos-cc   -c fs_iso9660.c -o fs_iso9660.o
sh-elf-gcc.exe: error: command-line option '-m4-single' is not supported by this configuration
sh-elf-gcc.exe: error: command-line option '-m4-single' is not supported by this configuration
sh-elf-gcc.exe: error: command-line option '-m4-single' is not supported by this configuration
sh-elf-gcc.exe: error: command-line option '-m4-single' is not supported by this configuration

```

Is this related to https://github.com/KallistiOS/KallistiOS/pull/834 ?

gyrovorbis:
Sure is... the time has come to update that antique toolchain of yours to enjoy C23, C++26, and the glory and power that is `double`-precision floating point variables. 

...just go into `dc-chain` and make a new, updated toolchain. PITA? Maybe, but GCC has new features and better perf than what you're on, I'm guessing. Also, I'd recommend changing the profile to `14.2.0`, as it's about to be the new stable. 

Will keep the thread open  until you're done in case you run into issues.

darcagn:
This looks like DreamSDK?
We added support for `m4-single` in all toolchains in #526 in April, but unfortunately @sizious hadn't updated his prebuilt toolchains in DreamSDK yet before we merged #834.

You can either
1) Alter your `environ.sh` configuration file to switch back to the old setting, `m4-single-only`.
2) Rebuild your toolchain manually now -- which will automatically pick up the new support.
3) Wait for DreamSDK R4, which will have newly built toolchains. I talked to @sizious  about it this week, and he is releasing it very soon.

Xeroxxx:
So yes I'm using DreamSDK, I expected this to work with latest master or allowing me th choose the Toolchain on UI.

However you cannot compile 14.2.0 or earlier with DreamSDK on DreamShell as it verifys the downloads with sha512sum or sh256sum.

darcagn:
DreamShell isn't using `m4-single` yet anyway, so you simply need to alter DreamSDK's `environ.sh` to change to `m4-single-only`. It's just a simple text change, no need to rebuild any toolchains. Then everything should work as expected under DreamSDK.

Xeroxxx:
[...]

PR #1049, opened 2025-05-23
environ_base.sh: Source environ_dreamcast.sh later

darcagn:
The `$KOS_SH4_PRECISION` check is dependent on `$KOS_CC` being established, so move the sourcing of `environ_dreamcast.sh` after compiler paths are set.


QuzarDC (approved):

QuzarDC:
Hopefully with this #931 doesn't come up again.

PR #1053, opened 2025-05-30
aica: Fixed and improved checks for stream init

DC-SWAT:
We must not fall apart over trifles.


QuzarDC (approved):
This also manages to prevent the bad behavior that would result if someone passed `channels` as `0` which would still allocate at least one buffer before.

PR #1056, opened 2025-05-30
Use THD_STACK_ALIGNMENT to adjust arch's stack alignment requirements

darcagn:
SH4 only requires 4-byte alignment for the stack, but other platforms (like PowerPC) may require alignment >4 bytes.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/arch.h
+++ b/kernel/arch/dreamcast/include/arch/arch.h
@@ -79,6 +79,11 @@ extern char _etext;
 static const
 unsigned HZ __depr("Please use the new THD_SCHED_HZ macro.") = THD_SCHED_HZ;
 
+#ifndef THD_STACK_ALIGNMENT
+/** \brief  Required alignment for stack. */
+#define THD_STACK_ALIGNMENT 4
+#endif
+
 #ifndef THD_STACK_SIZE
 /** \brief  Default thread stack size. */
 #define THD_STACK_SIZE  32768
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -404,7 +404,8 @@ kthread_t *thd_create_ex(const kthread_attr_t *restrict attr,
 
             /* Create a new thread stack */
             if(!real_attr.stack_ptr) {
-                nt->stack = (uint32_t*)malloc(real_attr.stack_size);
+                nt->stack = (uint32_t*)aligned_alloc(THD_STACK_ALIGNMENT,
+                                                     real_attr.stack_size);
 
                 if(!nt->stack) {
                     free(nt);


ljsebald (approved):

darcagn:
Moving this to draft to prevent a merge as there's at least one other adjustment that needs to be made here.

darcagn:
The other changes I need to make are a little bit different of scope so I'm opening this one back up for review/merging.

I did make another change though, changing
```c
static alignas(8) uint8_t thd_reaper_stack[512];
static alignas(8) uint8_t thd_idle_stack[64];
```
to
```c
static alignas(THD_STACK_ALIGNMENT) uint8_t thd_reaper_stack[512];
static alignas(THD_STACK_ALIGNMENT) uint8_t thd_idle_stack[64];
```

However, since `THD_STACK_ALIGNMENT` is `4` for SH4 this reduces the alignment requirement. This should be OK for SH4 though. I haven't yet tested though, not at my Dreamcast setup at the moment.

DC-SWAT:
> However, since `THD_STACK_ALIGNMENT` is `4` for SH4 this reduces the alignment requirement. This should be OK for SH4 though. I haven't yet tested though, not at my Dreamcast setup at the moment.

For Dreamcast it is worth making 8-byte alignment for threads using FPU. And for threads without using FPU 4-byte alignment is enough.

QuzarDC:
> > However, since `THD_STACK_ALIGNMENT` is `4` for SH4 this reduces the alignment requirement. This should be OK for SH4 though. I haven't yet tested though, not at my Dreamcast setup at the moment.
> 
> For Dreamcast it is worth making 8-byte alignment for threads using FPU. And for threads without using FPU 4-byte alignment is enough.

Why is there the distinction? Is stack usage different between them in some way?

DC-SWAT:
> > > However, since `THD_STACK_ALIGNMENT` is `4` for SH4 this reduces the alignment requirement. This should be OK for SH4 though. I haven't yet tested though, not at my Dreamcast setup at the moment.
> > 
> > 
> > For Dreamcast it is worth making 8-byte alignment for threads using FPU. And for threads without using FPU 4-byte alignment is enough.
> 
> Why is there the distinction? Is stack usage different between them in some way?

Because only FPU has 64-bit instructions. But we don't have a division into types of threads that can handle different sets of instructions and save only needed regs on thread switch.

darcagn:
I've adjusted the stack alignment to `8`, matching what it was for reaper/idle already anyway.

DC-SWAT:
> > > > > However, since `THD_STACK_ALIGNMENT` is `4` for SH4 this reduces the alignment requirement. This should be OK for SH4 though. I haven't yet tested though, not at my Dreamcast setup at the moment.
> > > > 
> > > > 
> > > > For Dreamcast it is worth making 8-byte alignment for threads using FPU. And for threads without using FPU 4-byte alignment is enough.
> > > 
> > > 
> > > Why is there the distinction? Is stack usage different between them in some way?
> > 
> > 
> > Because only FPU has 64-bit instructions. But we don't have a division into types of threads that can handle different sets of instructions and save only needed regs on thread switch.
> 
> Wouldn't that have to do with the alignment of the thread context rather than its stack? We do already align the thread object, which keeps the context at the front of alignment for faster context switch. I just don't see how that would be of impact for the start of the stack, unless the current SP inherits the alignment of the stack start in some way.

I thought it was about both. Can't a 64-bit variable be allocated on the stack for the FPU? That's why the Dreamcast has an alignment of 8 for thread stack.

PR #1058, opened 2025-05-31
Fix various warnings in examples.

QuzarDC:
Uninitialized variables, unused args, and signedness mismatches.nehe26 got a bit deeper of a cleanup in order to address unchecked allocations.


gyrovorbis (approved):
VERY NICE. Especially the standard type cleanups like integers and booleans, I'm a huge fan of. lol.

Issue #1059, opened 2025-05-31
Failed examples testing for v2.2.0 part 1


QuzarDC:
These were built with the stable toolchain and standard frame pointer options
- [ ] In `basic/asserthnd` it seems like the stack tracing is broken:
```
Our assert handler got called!
  file = asserthnd.c
  line = 41
  expr = a != 5
  msg = This is a test message!
  func = func2
Our assert handler got called!
  file = asserthnd.c
  line = 42
  expr = a != 5
  msg = (null)
  func = func2

*** ASSERTION FAILURE ***
Assertion "a != 5" failed at asserthnd.c:41 in `func2': This is a test message!

-------- Stack Trace (innermost first) ---------
   (invalid frame pointer)
-------------- End Stack Trace -----------------
arch: aborting the system
```
- [ ] `basic/mmu/nullptr` - hangs on start.

- [ ] `basic/stacktrace` similar to asserthnd:
```
-------- Stack Trace (innermost first) ---------
   (invalid frame pointer)
-------------- End Stack Trace -----------------
-------- Stack Trace (innermost first) ---------
   (invalid frame pointer)
-------------- End Stack Trace -----------------

arch: exit return code 0
arch: shutting down kernel
maple: final stats -- device count = 22, vbl_cntr = 8, dma_cntr = 8
vid_set_mode: 640x480 VGA with 1 framebuffers.
```
- [ ] `basic/threading/barrier` - runs for a long time then hangs:
```
Initializing Watchdog timer...
Creating 5 barriers...
Spawning 14 threads...
Executing logic from main thread...
Thread[0]: Before barrier[0]!
Thread[1]: Before barrier[0]!
Thread[2]: Before barrier[0]!
Thread[3]: Before barrier[0]!
Thread[4]: Before barrier[0]!
Thread[5]: Before barrier[0]!
Thread[6]: Before barrier[0]!
Thread[7]: Before barrier[0]!
Thread[8]: Before barrier[0]!
Thread[9]: Before barrier[0]!
Thread[10]: Before barrier[0]!
Thread[11]: Before barrier[0]!
Thread[12]: Before barrier[0]!
Thread[13]: Before barrier[0]!
Thread[14]: Before barrier[0]!
Thread[14]: After barrier[0]: SERIAL!
Thread[14]: Before barrier[1]!
Thread[0]: After barrier[0]: NONSERIAL!
Thread[0]: Before barrier[1]!
[...]

PR #1061, opened 2025-06-07
Fix stack tracing

QuzarDC:
There's three sets of fixes here that work together to fix stack tracing, at least in the `stacktrace` example broken as noted in #1059 :
- #600 had a mistake where it was causing the frame pointer value to be dereferenced twice, which meant that after the first one, it would always break. It was overriding the known fp with the return address from `arch_fptr_ret_addr`.

- #877 converted `arch_get_ret_addr` and `arch_get_fptr` into `static inline`s rather than `#define`s as a cleanup and for typing. The problem is that the functionality of these is wholly dependent on being inline to the calling function. Have marked them with the [`__always_inline` function attribute](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-noinline-function-attribute) to avoid that. This would have caused stack tracing to fail randomly whenever gcc happened to not inline them. It may be that what we should do is set up the stack tracing differently to take advantage of [gcc's builtins for these](https://gcc.gnu.org/onlinedocs/gcc/Return-Address.html#index-_005f_005fbuiltin_005fframe_005faddress) .

- Added the [`__noinline` function attribute](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-noinline-function-attribute) to a number of functions. In the examples, this was added for the examples that demonstrate stack tracing as they rely on the functions to be functions. This was also added for the default assert handler, and stack trace functions as each of them makes assumptions about their frame pointer level.

Unfortunately these changes still don't seem to fix the `asserthnd` example for me. Perhaps that's an environment issue or something there are further things to fix for it. Would preferably want to be able to figure out why that's also broken and incorporate the fixes here as they're likely interrelated, but the changes here should be good on their own as well

### DIFF
--- a/examples/dreamcast/basic/asserthnd/asserthnd.c
+++ b/examples/dreamcast/basic/asserthnd/asserthnd.c
@@ -35,14 +35,18 @@ assuming you compiled your program with -g:
 
 */
 
-void func2(void) {
+/*  These are marked as __noinline to ensure the compiler
+    doesn't try to get smart and inline them which would
+    defeat the purpose of the example.
+*/
+__noinline void func2(void) {
     int a = 5;
 
     assert_msg(a != 5, "This is a test message!");
     assert(a != 5);
 }
 
-void func1(void) {
+__noinline void func1(void) {
     func2();
 }
 
--- a/examples/dreamcast/basic/stacktrace/stacktrace.c
+++ b/examples/dreamcast/basic/stacktrace/stacktrace.c
@@ -6,7 +6,11 @@
 
 #include <kos.h>
 
-void func(void) {
+/*  This is marked as __noinline to ensure the compiler
+    doesn't try to get smart and inline it which would
+    defeat the purpose of the example.
+*/
+__noinline void func(void) {
     arch_stk_trace(0);
 }
 
--- a/kernel/arch/dreamcast/kernel/stack.c
+++ b/kernel/arch/dreamcast/kernel/stack.c
@@ -25,7 +25,7 @@ extern uintptr_t arch_stack_32m __attribute__((weak,alias("arch_stack_32m_dft"))
 
 /* Do a stack trace from the current function; leave off the first n frames
    (i.e., in assert()). */
-void arch_stk_trace(int n) {
+__noinline void arch_stk_trace(int n) {
     arch_stk_trace_at(arch_get_fptr(), n + 1);
 }
 
@@ -46,7 +46,7 @@ void arch_stk_trace_at(uint32_t fp, size_t n) {
             dbgio_printf("   %08lx   (invalid frame pointer)\n", fp);
             break;
         }
-        
+
         if(n == 0) {
             /* Get the return address from the function pointer */
             ret_addr = arch_fptr_ret_addr(fp);
--- a/kernel/libc/koslib/assert.c
+++ b/kernel/libc/koslib/assert.c
@@ -16,7 +16,7 @@
 #include <arch/stack.h>
 
 /* The default assert handler */
-static void __noreturn assert_handler_default(const char *file, int line,
+__noinline static void __noreturn assert_handler_default(const char *file, int line,
                             const char *expr, const char *msg, const char *func) {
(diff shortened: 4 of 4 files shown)


pcercuei:
Having debugged it for a bit, the whole stack tracing algorithm is buggy. It is based on the assumption that the function return register is stored at the address pointed to by the frame pointer, followed by the previous value of the frame pointer. However this assumption is not valid, as GCC will happily skip the return register store if it can, which means that the previous frame pointer is not stored into the current frame at a fixed offset.

For that reason, there is no simple fix for that issue, and I don't think it is worth spending time on it before the 2.2.0 release.

QuzarDC:
> Having debugged it for a bit, the whole stack tracing algorithm is buggy. It is based on the assumption that the function return register is stored at the address pointed to by the frame pointer, followed by the previous value of the frame pointer. However this assumption is not valid, as GCC will happily skip the return register store if it can, which means that the previous frame pointer is not stored into the current frame at a fixed offset.
> 
> For that reason, there is no simple fix for that issue, and I don't think it is worth spending time on it before the 2.2.0 release.

Yes, but that's *always* been the case. The functionality has always been susceptible to that kind of break down, but would still work most of the time (with the suggested cflags of `-Og` and `-fno-omit-frame-pointer`). Two of the fixes here explicitly fix changes that were made recently that were incorrect and make it *never* work correctly, with the third strengthening the examples to help ensure they are more likely to work properly.

So yes, it's still buggy, but these changes take it back to its customary state of bugginess rather than the current completely broken state. I take your point though for not holding up the v2.2.0 release, and unless the solution to the further `asserthnd` issues are forthcoming during the time of checking through the rest of the examples, I wouldn't hold it up just for that.

darcagn:
Applying these fixes to `arch_stk_trace()` and `arch_stk_trace_at()` so it doesn't double-deref, I'm getting good stack traces on PowerPC (obviously, using my own implementations for the `arch.h` functions). It does make things easier that PowerPC's frame structure is consistent and doesn't need `-fno-omit-frame-pointers -DFRAME_POINTERS` to walk the stack, though.

However, I did have to add `-fno-optimize-sibling-calls` to the `$KOS_CFLAGS` in the `Makefile` for `stacktrace`, otherwise GCC can _still_ chose to inline certain functions despite being `__noinline`!

On the Dreamcast side, I played with a few optimization flags and still couldn't seem to get it to work properly though. I also tried playing with the builtin stack functions and `-Wframe-address` and didn't get anything different.

Because this is an improvement and bug fix nonetheless, I'd approve, but one thing: `arch_stk_trace_at()` doesn't actually need `__noinline`? Since it's working off of a fixed pointer. Or am I missing it?

QuzarDC:
> Applying these fixes to `arch_stk_trace()` and `arch_stk_trace_at()` so it doesn't double-deref, I'm getting good stack traces on PowerPC (obviously, using my own implementations for the `arch.h` functions). It does make things easier that PowerPC's frame structure is consistent and doesn't need `-fno-omit-frame-pointers -DFRAME_POINTERS` to walk the stack, though.
> 
> However, I did have to add `-fno-optimize-sibling-calls` to the `$KOS_CFLAGS` in the `Makefile` for `stacktrace`, otherwise GCC can _still_ chose to inline certain functions despite being `__noinline`!
> 
> On the Dreamcast side, I played with a few optimization flags and still couldn't seem to get it to work properly though. I also tried playing with the builtin stack functions and `-Wframe-address` and didn't get anything different.

Yea, I looked at using the builtin functions but they just work fundamentally differently and I don't see how we'd replicate the current implementation.

I'd guess you started with just the plain environ sample and set frame pointers from there?
> Because this is an improvement and bug fix nonetheless, I'd approve, but one thing: `arch_stk_trace_at()` doesn't actually need `__noinline`? Since it's working off of a fixed pointer. Or am I missing it?

Ah absolutely. Went overboard there. Removed that one and force-pushed. And yea, if it weren't for the fact that the first two changes were addressing specific regressions I think I would have moved instead to just mark it deprecated and see about how we might replace the functionality instead.

PR #1065, opened 2025-06-09
Faster IRQ enable/disable/restore + SR fetch.

gyrovorbis:
Despite being tiny little routines that clobber either just one register or nothing at all, the following routines were being implemented in pure out-of-line ASM, which meant the compiler could never inline them even with LTO:
    a. irq_get_sr()
    b. irq_restore()
    c. irq_disable()
    d. irq_enable()

This PR moves them into inline functions as inline ASM where the compiler should definitely be inlining them at even low optimization levels for pretty high perf gainz, since these are very hot throughout the codebase.

1) Removed all out-of-line implementations of theses routines from
  entry.s.
2) Rewrote them as inlined functions in inline ASM within irq.h 3) Had to tweak the following ASM routines to no longer call into the
   removed routines, since ASM cannot call inline C functions:
    a. irq_save_regs() [entry.s]
    b. thd_block_now() [thdswitch.s]

### DIFF
--- a/kernel/arch/dreamcast/include/arch/irq.h
+++ b/kernel/arch/dreamcast/include/arch/irq.h
@@ -1,9 +1,9 @@
 /* KallistiOS ##version##
 
    arch/dreamcast/include/arch/irq.h
-   Copyright (C) 2000-2001 Megan Potter
+   Copyright (C) 2000, 2001 Megan Potter
    Copyright (C) 2024 Paul Cercueil
-   Copyright (C) 2024 Falco Girgis
+   Copyright (C) 2024, 2025 Falco Girgis
 
 */
 
@@ -329,7 +329,25 @@ typedef uint32_t irq_mask_t;
     \retval                 Status register word
     \sa irq_disable()
 */
-irq_mask_t irq_get_sr(void);
+static inline irq_mask_t irq_get_sr(void) {
+    irq_mask_t value;
+    __asm__ volatile("stc sr, %0" : "=r" (value));
+    return value;
+}
+
+/** Restore IRQ state.
+
+    This function will restore the interrupt state to the value specified. This
+    should correspond to a value returned by irq_disable().
+
+    \param  v               The IRQ state to restore. This should be a value
+                            returned by irq_disable().
+
+    \sa irq_disable()
+*/
+inline static void irq_restore(irq_mask_t old) {
+    __asm__ volatile("ldc %0, sr" : : "r" (old));
+}
 
 /** Disable interrupts.
 
@@ -342,7 +360,11 @@ irq_mask_t irq_get_sr(void);
 
     \sa irq_restore(), irq_enable()
 */
-irq_mask_t irq_disable(void);
+inline static irq_mask_t irq_disable(void) {
+    uint32_t mask = (uint32_t)irq_get_sr();
+    irq_restore((mask & 0xefffff0f) | 0x000000f0);
+    return mask;
+}
 
 /** Enable all interrupts.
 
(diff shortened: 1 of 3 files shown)


gyrovorbis:
So obviously this change ain't gonna make it in this release... and it shouldn't, but I do think it's definitely going to offer quite a perf boost... going to see if I can find time to quantify it with the perf counters...

WHAT I WANTED TO ASK, THOUGH, is... notice how I've now including inline ASM within `irq.h` which is specific to DC's SH4... Is that header intended to be arch-independent? Is this going to screw up da Lamecube?

If so... one pattern I've used and have seen used a decent amount when faced with a similar situation is to keep `irq.h` a vanilla, cross-platform API, but at the bottom of the file, do something like this:

```c
#ifdef __DREAMCAST__
#    include <dc/irq_impl.h>
#elif defined(__LAMECUBE__)
#    include <gc/irq_impl.h>
#endif
```

This way, you keep the header file clean and cross-platform, but a particular architecture still gets the chance to inline hot routines for performance...

QuzarDC:
> So obviously this change ain't gonna make it in this release... and it shouldn't, but I do think it's definitely going to offer quite a perf boost... going to see if I can find time to quantify it with the perf counters...
> 
> WHAT I WANTED TO ASK, THOUGH, is... notice how I've now included inline ASM within `irq.h` which is specific to DC's SH4... Is that header intended to be arch-independent? Is this going to screw up da Lamecube?
> 
> If so... one pattern I've used and have seen used a decent amount when faced with a similar situation is to keep `irq.h` a vanilla, cross-platform API, but at the bottom of the file, do something like this:
> 
> ```c
> #ifdef __DREAMCAST__
> #    include <dc/irq_impl.h>
> #elif defined(__LAMECUBE__)
> #    include <gc/irq_impl.h>
> #endif
> ```
> 
> This way, you keep the header file clean and cross-platform, but a particular architecture still gets the chance to inline hot routines for performance...

No no, since the header is in `arch/dreamcast` it's meant to be platform specific. 

That is something that we likely need to review some though because while the implementation should be arch-specific I'm not sure why that should have to do with the headers which should be defining the common API. It may make more sense to have something like `kos/irq.h` that includes `arch/irq.h` to supply arch-specific functionality with the API we need to have implemented for KOS to work should be defined at the KOS level. Or just have some commentary separation in the headers to show common vs arch-specific.

kernel/arch/dreamcast/include/arch/irq.h
```diff
@@ -329,7 +329,25 @@ typedef uint32_t irq_mask_t;
...
+
+    This function will restore the interrupt state to the value specified. This
+    should correspond to a value returned by irq_disable().
+
+    \param  v               The IRQ state to restore. This should be a value
+                            returned by irq_disable().
+
+    \sa irq_disable()
+*/
+inline static void irq_restore(irq_mask_t old) {
```
QuzarDC: Is there a distinction in these between static inline and inline static?

kernel/arch/dreamcast/kernel/entry.s
```diff
@@ -111,8 +112,12 @@ _irq_save_regs:
...
 	! Before we enter the main C code again, re-enable exceptions
 	! (but not interrupts) so we can still debug inside handlers.
-	bsr		_irq_disable
-	nop
+	mov.l	irqd_and,r1
+	mov.l	irqd_or,r2
+	stc	sr,r0
+	and	r0,r1
+	or	r2,r1
+	ldc	r1,sr
```
pcercuei: LGTM overall but the alignment is not the same anymore (in all your ASM additions).

QuzarDC (approved):

kernel/arch/dreamcast/kernel/entry.s
gyrovorbis: ...he's right. Bringing it up in Discord now to see how he wants to handle it.

PR #1066, opened 2025-06-09
Bugfix for invalid mutex destruction + API cleanup

gyrovorbis:
A mutex whose mutex_init() call failed would have had an uninitialized "type" member, which would've made any subsequent usage of the "invalid" mutex result in undefined behavior.

1) Added "MUTEX_TYPE_INVALID" as a way to tag such a mutex, allowing for subsequent usages to gracefully error out.
2) Added internal util function for checking wither an enum is valid.

While I was at it, I took the liberty of also doing as much API cleanup as I could without breaking the ABI (which would require rebuilding toolchains).

3) Enum-ified "mutex_type_t."
4) Constified mutex pointer to mutex_is_locked().
5) Reordered the scoped mutex code so that the public API comes first within the header file, followed by private internals at the end.

### DIFF
--- a/include/kos/mutex.h
+++ b/include/kos/mutex.h
@@ -3,6 +3,7 @@
    include/kos/mutex.h
    Copyright (C) 2001, 2003 Megan Potter
    Copyright (C) 2012, 2015 Lawrence Sebald
+   Copyright (C) 2025 Falco Girgis
 
 */
 
@@ -42,6 +43,10 @@
     recursive_lock_t type that was available in KallistiOS for a while (before
     it was basically merged back into a normal mutex).
 
+    An invalid mutex (MUTEX_TYPE_INVALID) is simply used to denote a mutex
+    which failed to initialize properly. Any subsequent operations on such a
+    mutex will fail.
+
     There is a fourth type of mutex defined (MUTEX_TYPE_DEFAULT), which maps to
     the MUTEX_TYPE_NORMAL type. This is simply for alignment with POSIX.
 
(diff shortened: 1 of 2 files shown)


gyrovorbis:
I don't know if I follow how it's not a bug... I've not added any additional checks or logic to the lock/unlock routines, I've simply cleaned them up and have moved them to a utility function. We have always checked whether the type was valid before doing anything with the mutex, we just failed to *set* the type to be invalid upon initialization should it fail.

I see our existing checks for the validity of the mutex to be totally pointless or worthless if we aren't actually invalidating the mutex should it fail to initialize. So basically any subsequent call with a mutex that failed to initialize is currently UB, as its "type" could be any value, including valid values.

QuzarDC:
> I see our existing checks for the validity of the mutex to be totally pointless or worthless if we aren't actually invalidating the mutex should it fail to initialize. So basically any subsequent call with a mutex that failed to initialize is currently UB, as its "type" could be any value, including valid values.

See, what I was understanding their value as was in preventing 'destroyed' mutexes from having operations called on them, and that all uninitialized was just a separate unknown/untestable state apart from a destroyed mutex. But your thinking on this makes a lot more sense, especially in context with the behavior in sem.

kernel/thread/mutex.c
```diff
@@ -224,6 +230,8 @@ int mutex_trylock(mutex_t *m) {
 
             ++m->count;
             break;
+        default:
```
QuzarDC, line 233: assert here maybe? Mostly to make it clear the default isn't a catch-all but impossible.

include/kos/mutex.h
```diff
@@ -58,37 +63,36 @@ __BEGIN_DECLS
...
+    The values defined in here are the various types of mutexes that KallistiOS
+    supports.
+*/
+typedef enum mutex_type {
+    MUTEX_TYPE_INVALID    = -1, /**< \brief Invalid mutex type */
+    MUTEX_TYPE_NORMAL     = 0,  /**< \brief Normal mutex type */
+    MUTEX_TYPE_OLDNORMAL  = 1,  /**< \brief Alias for MUTEX_TYPE_NORMAL */
+    MUTEX_TYPE_ERRORCHECK = 2,  /**< \brief Error-checking mutex type */
+    MUTEX_TYPE_RECURSIVE  = 3   /**< \brief Recursive mutex type */
+} mutex_type_t;
```
QuzarDC, line 77: So the rest of the changes seem fine, but this doesn't do much since the type should be percolated out to make it have value for enforcing the type. Also, it would then conflict with this tiny optimization I know you'd want: https://github.com/KallistiOS/KallistiOS/pull/1047/commits/f0da64498ee9a9c1f4dc8fc1619368cee92c3c1f

Note there too how I named it `DESTROYED` due to my understanding that a destroyed mutex was what was being tested against. Could easily though have both 'destroyed' and 'bad init' in case it ever matters for troubleshooting. But it may be worthwhile to have it be unsigned (and a c23 unsigned enum would likely choke the toolchain like `alignas` did).

But it's really a trivial difference so I wouldn't mind scrapping that one if you think this is better.

QuzarDC (commented):
Just some notes. No changes required, but some might make sense.

gyrovorbis:
This has totally been superseded by Quzar's fine primitive cleanup PR here: https://github.com/KallistiOS/KallistiOS/pull/1047

PR #1069, opened 2025-06-12
cdrom: Fixed confusion between sessions and areas.

DC-SWAT:
GD-ROM drive does not differentiate sessions at all. This TOC is always divided into two, for the low-density area and the high-density area, which is only on the GD-ROM disc.
And in fact, this is one and the same TOC, it’s just that for non-requested area the tracks are erased (and first/last changed) in this common TOC, because the tracks numbering remains the same.
There may be multiple sessions in one area, but there is no division between them.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_iso9660.c
+++ b/kernel/arch/dreamcast/fs/fs_iso9660.c
@@ -343,7 +343,7 @@ static int init_percd(void) {
         return -1;
     }
 
-    if((i = cdrom_read_toc(&toc, 0)) != 0)
+    if((i = cdrom_read_toc(&toc, false)) != 0)
         return i;
 
     if(!(session_base = cdrom_locate_data_track(&toc)))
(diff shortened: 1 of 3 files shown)


pcercuei:
Looks OK, but I wonder if it would be better to have a `bool high_density` parameter, instead of a custom `cdrom_area_t` type.
Because then it means there are only two accepted values (true and false), which means the check you added to `cdrom_read_toc()` can go away.

pcercuei:
> GD-ROM drive does not differentiate sessions at all. This TOC is always divided into two, for the low-density area and the high-density area, which is only on the GD-ROM disc. And in fact, this is one and the same TOC, it’s just that for non-requested area the tracks are erased (and first/last changed) in this common TOC, because the tracks numbering remains the same. There may be multiple sessions in one area, but there is no division between them.

Can you reuse this text as the commit message? That's useful information.

DC-SWAT:
> > GD-ROM drive does not differentiate sessions at all. This TOC is always divided into two, for the low-density area and the high-density area, which is only on the GD-ROM disc. And in fact, this is one and the same TOC, it’s just that for non-requested area the tracks are erased (and first/last changed) in this common TOC, because the tracks numbering remains the same. There may be multiple sessions in one area, but there is no division between them.
> 
> Can you reuse this text as the commit message? That's useful information.

Sure, done.

PR #1070, opened 2025-06-14
Fix toolchain compilation on Darwin 24.4.0+ using Apple Clang 17

ppxxcc:
## Background

Starting with Xcode 16.3 (Darwin 24.4.0, Sequoia 15.4) the compiler was updated to Apple Clang 17. 
Same goes for Xcode 16.4 (Darwin 24.5.0, Sequoia 15.5)

## Problem
Old versions of zlib break under Apple Clang 17 because of some lines in `zutil.h` redefining `fdopen` to `NULL` which causes compilation to error out when it reaches zlib when building binutils (and gcc).

The latest zlib versions [seem to have fixed this ](https://github.com/madler/zlib/blob/develop/zutil.h#L144) some time ago by simply removing those lines redefining the symbol.

However, binutils and gcc bundle in their own copy of copy of zlib under `binutils-x.y/zlib/` and `gcc-x.y.z/zlib/` which appears to be old enough that it doesn't have the latest upstream fixes for macOS.

## Solution
I backported the zlib changes as patches in `aarch64-apple-darwin24.4.0/` and `aarch64-apple-darwin24.5.0/` under `dc-chain/patches/`

The patches are for all the stable versions of the toolchain, being:
- binutils-2.43
- binutils-2.44
- gcc-13.2.0
- gcc-13.4.0
- gcc-14.3.0
- gcc-15.1.0

## Conclusion
With these patches the toolchain builds without issue. 🚀

## Other
Googling around this seems to be an overall issue with Clang 17 which also affects Intel platforms as people elsewhere were encountering issues with zlib on their Intel Macs, so patches should probably exist for the Intel target as well, but I only have an Apple Silicon Mac so I can't confirm. 

Though considering those lines were removed from upstream zlib entirely, the patches could probably just be copy pasted into a directory for the Intel target too.

This should probably be added to the KOS 2.2.x branch if possible if they are to be the new stable version going forward.

### DIFF
--- a/utils/dc-chain/patches/aarch64-apple-darwin24.4.0/binutils-2.43-zlibfix.diff
+++ b/utils/dc-chain/patches/aarch64-apple-darwin24.4.0/binutils-2.43-zlibfix.diff
@@ -1,0 +1,24 @@
+diff --color -ruN binutils-2.43/zlib/zutil.h binutils-2.43/zlib/zutil-apple-clang17-fix.h
+--- binutils-2.43/zlib/zutil.h	2024-08-04 08:00:00
++++ binutils-2.43/zlib/zutil-apple-clang17-fix.h	2025-06-14 21:19:11
+@@ -139,15 +139,11 @@
+ 
+ #if defined(MACOS) || defined(TARGET_OS_MAC)
+ #  define OS_CODE  7
+-#  ifndef Z_SOLO
+-#    if defined(__MWERKS__) && __dest_os != __be_os && __dest_os != __win32_os
+-#      include <unix.h> /* for fdopen */
+-#    else
+-#      ifndef fdopen
+-#        define fdopen(fd,mode) NULL /* No fdopen() */
+-#      endif
+-#    endif
+-#  endif
++/*
++ * Backport fix from zlib 1.3.1 where these lines
++ * redefining fdopen were removed. In Apple Clang 17
++ * these will cause errors when building zlib.
++ */
+ #endif
+ 
+ #ifdef __acorn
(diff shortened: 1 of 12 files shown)


ppxxcc:
This is tangentially related and not so much an "issue" but just an observation, but I see the `patch.mk` file responsible for apply patches does the following:
```
$(patch_binutils): diff_patches := $(wildcard $(patches)/$(src_dir)*.diff)
$(patch_binutils): diff_patches += $(wildcard $(patches)/$(host_triplet)/$(src_dir)*.diff)
```
where `host_triplet` is deduced in `host_detect.mk` with
```
host_triplet=$(shell ./config.guess)
```

In the case of macOS where `config.guess` will return different triples depending on the architecture and kernel version, this can lead to a lot of visual redundancy for architecture/kernel agnostic patches.

In the case of this PR, if you wanted to fix it for all current platforms, you would have to have 4 patch directories:
1. aarch64-apple-darwin24.4.0
2. aarch64-apple-darwin24.4.0
3. x86_64-apple-darwin24.4.0
4. x86_64-apple-darwin24.4.0

I'm not sure if this is an issue per se, but as macOS gets updates going forward and gets versions `24.5.0` and `24.6.0` and so on, additional patch directories will need to be made for those as well. I think this can get out of hand quickly leading to a lot of visual clutter. Maybe it isn't a huge deal, but seems like there should be a more intelligent way around this.

ppxxcc:
In retrospect this is probably a sloppy way to do it so I'll come up with a better way.

QuzarDC:
> In retrospect this is probably a sloppy way to do it so I'll come up with a better way.

If it was the same underlying fix, then just changing the outermost `#ifdef` would do it with fewer lines at least.

PR #1073, opened 2025-06-17
thread: Speed up polling by bypassing thd_pass()

pcercuei:
When the current thread is the only runnable thread (apart from the idle thread), there is no need to go through the process of doing a full context switch, since we know that the scheduler will pick the current thread once again.

Note that it speeds up *in theory* (it just makes sense) but I did not benchmark it.

### DIFF
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -762,6 +762,18 @@ void thd_sleep(unsigned int ms) {
     genwait_wait((void *)0xffffffff, "thd_sleep", ms, NULL);
 }
 
+static bool thd_should_pass(void) {
+    kthread_t *thd;
+
+    /* Look up the queue to find if there are more threads waiting to run */
+    TAILQ_FOREACH(thd, &run_queue, thdq) {
+        if(thd != thd_current && thd->state == STATE_READY)
+            break;
+    }
+
+    return thd != thd_idle_thd;
+}
+
 /* Manually cause a re-schedule */
 __used
 void thd_pass(void) {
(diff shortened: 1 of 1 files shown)


pcercuei:
Actually one issue I see now, is that `genwait_check_timeouts()` is not called, so the threads that timed out waiting on a genwait object aren't reset to the "ready" state anymore.

gyrovorbis:
I feel like this is a pretty damn useful PR that SURELY is going to have some performance implications...

Is it possible to call `genwait_check_timeouts()` just before checking if `thd_should_pass()` in `thd_pass()`? That way we would give genwait a chance to update the run queue before deciding if there's anything else to pass to?

Maybe there's some side-effects or other implications I'm missing though?

pcercuei:
@QuzarDC I don't see why you need to change the `thd_block_now` at all, thd_pass could simply `irq_disable_scoped()` and it would work just as well.

pcercuei:
Alternatively, one idea I had (but didn't really investigate), was to have a `int thd_poll(bool (*cb)(void *), void *data)` function that takes a callback as a parameter. It is run once, and if returns false, the thread is put to sleep. Then the scheduler itself will periodically run the callback function, and if it returns true, the thread is marked as "ready" again.

The conditional `thd_block_now` idea speeds up polling if there's only one thread, on the other hand this would reduce the overhead of a polling thread by not going back and forth between the thread and the scheduler, leaving CPU time for the other threads. If the polling thread is the only one active, we could also keep polling into the scheduler until it returns true or until something in genwait's queue wakes up.

PR #1075, opened 2025-06-17
Updated environ.sh.sample CFLAGs.

gyrovorbis:
1) Link-Time Optimization
    - Included -ffat-lto-objects by default, which is required for atomics and many KOS ports, with a note about this. 2) Additional Optimizations
    - No longer is about LTO
    - Has a list of random flags which have been found to also positively benefit perf at least SOMETIMES but are potentially situational.
3) Position Independent Code
    - Disabled by default with explanation of when to enable (dynamic linking) 4) Frame Pointers
    - Added info on using an extra register for frame pointer 5) GCC Builtin Functions
    - enabled by default now for performance reasons, as it has not been found to cause any issues (and our compiler doesn't do much with it)
    - Modified warning to be less harsh, as at this point anything broken with this flag is a new bug, most likely. 6) Fast Math Instructions
    - Added comment advising people to still try to enable these flags on individual compilation units rather than globally, sure enabling globally cause any issues.


pcercuei:
@gyrovorbis NAK on the `-fno-PIC -fno-PIE`. It is a good idea, just not a day before 2.2.0. So I'd suggest to fork it to a new PR that can be merged after 2.2.0 is out.

darcagn:
I find the phrasing around LTO confusing. What would the user think "fatter object files" implies? It's also reasonable to conclude when reading "atomics REQUIRE fat LTO objects to work" that atomics require LTO.

I would leave in a warning about LTO. It doesn't have to be the exact phrasing `Certain KOS-ports and examples do not work properly with "-flto=auto"!` but it's well known to cause build issues and uncover subtle bugs on any platform, not just Dreamcast (Ubuntu, for example, maintains a list of packages that are not to be compiled with LTO), so users should be aware of reasons not to enable it (and if it were the case that there are none, then we'd be enabling it by default). It also increases compile times and memory utilization although that is much less of a concern when targeting Dreamcast.

The "Additional Optimizations" section should have each flag on its own line and not refer to it as a "set." The optimizations are independent of one another and should be easily toggled on and off individually, unlike most other options in the environ where the flags go together.

gyrovorbis:
What do you guys think now?

pcercuei:
@QuzarDC Using `sinf()` and `cosf()` with builtins enabled will resolve to `__builtin_sinf()` and `__builtin_cosf()`; without `-ffast-math` they won't use the `FSCA` opcode, but resolve to a call to a C library function.

With builtins disabled, they will directly resolve to calling the same C library functions.

Therefore disabling builtins by default without a valid reason is a bad idea and will make the code *potentially* slower, and not faster.

doc/environ.sh.sample
```diff
@@ -172,21 +191,21 @@ export KOS_CFLAGS="${KOS_CFLAGS} -fomit-frame-pointer"
 
 # GCC Builtin Functions
 #
-# Comment out this line to enable GCC to use its own builtin implementations of
-# certain standard library functions. Under certain conditions, this can allow
-# compiler-optimized implementations to replace standard function invocations.
-# The downside of this is that it COULD interfere with Newlib or KOS implementations
-# of these functions, and it has not been tested thoroughly to ensure compatibility.
+# Comment in this line to prevent GCC from using its own builtin implementations of
```
darcagn:
```suggestion
# Uncomment this line to prevent GCC from using its own builtin implementations of
```

QuzarDC:
> Also, on the subject of builtins, I concur with @pcercuei. You definitely want them enabled by default because they can prevent full function invocations to Newlib for things like special memcpy() specializations (which we have almost none of in GCC), or for replacing sinf()/cosf() with FSCA.
> 
> I've only ever built with builtins enabled for many, many months and have never had a problem.

My question was specifically because you raised the issue of not having builtins turned on *unless* `-ffast-math` was also turned on: https://discord.com/channels/488332893324836864/614912396460556298/1333936876200919133 back in January when we tried turning them both on and fast-math (not builtins) caused issues with raylib and something else. At the time we were going to just disable fast-math but you'd raised that the math builtins have worse performance without fast-math than without builtins at all. 

If I misunderstood, or it's changed since then that's fine. If not, would just be a good thing to note that they should/need to be used together. I know we haven't had any compatibility issues in ages (iirc back when it was occasional linking errors and stdio breakage), I just remember there being this other downside.

gyrovorbis:
> > Also, on the subject of builtins, I concur with @pcercuei. You definitely want them enabled by default because they can prevent full function invocations to Newlib for things like special memcpy() specializations (which we have almost none of in GCC), or for replacing sinf()/cosf() with FSCA.
> > I've only ever built with builtins enabled for many, many months and have never had a problem.
> 
> My question was specifically because you raised the issue of not having builtins turned on _unless_ `-ffast-math` was also turned on: https://discord.com/channels/488332893324836864/614912396460556298/1333936876200919133 back in January when we tried turning them both on and fast-math (not builtins) caused issues with raylib and something else. At the time we were going to just disable fast-math but you'd raised that the math builtins have worse performance without fast-math than without builtins at all.
> 
> If I misunderstood, or it's changed since then that's fine. If not, would just be a good thing to note that they should/need to be used together. I know we haven't had any compatibility issues in ages (iirc back when it was occasional linking errors and stdio breakage), I just remember there being this other downside.

oh, I see. That conversation was basically me being shocked that `__builtin_sinf()` and `__builtin_cosf()` would not generate the `FSCA` instructions alone without `-ffast-math` and `-mfsca`. It wasn't really that builtins were worse for trig than not using builtins, it was just that using builtins *alone* was not sufficient to emit the `FSCA` fast trig instructions.

That's why the "fast math" line in the `environ.sh.sample` file looks like so:
`#export KOS_CFLAGS="${KOS_CFLAGS} -fbuiltin -ffast-math -ffp-contract=fast"`

Where you can see it also EXPLICITLY enables builtins, because you must combine both builtins *and* fast math to get the desired outcome.

There IS an issue currently with Raylib ICE-ing the compiler on a particular file with `-ffast-math` and builtins enabled, but I have opened a PR to address that already, which hopefully gets merged any day now.

gyrovorbis:
Since I fucked up the rebasery, this PR has been replaced by: https://github.com/KallistiOS/KallistiOS/pull/1127

PR #1077, opened 2025-06-18
cdrom: Factorize and fix commands taking way too much time

pcercuei:
Previously, `cdrom_exec_cmd_timed()` would send a command to the GD block, then wait on a semaphore until the command was executed. The problem was that the execution status was done in a VBLANK handler, which meant that each command would take up to 16 milliseconds to complete.

This caused the performance of Bloom to tank when reading original PSX discs, from 12 fps down to 1 fps.

Factorize the CD-ROM code by introducing a `cdrom_poll()` function that is then reused all across the cdrom.c code. Use it in `cdrom_exec_cmd_timed()`, and poll until the command is executed (aka. not in `PROCESSING` state), instead of waiting on a semaphore. This probably has a slight CPU overhead, but is much faster, which should still be a net gain. The VBLANK interrupt handler is now also much simpler which is a benefit in itself.

### DIFF
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -139,6 +139,11 @@ static int cdrom_check_cmd_done(void *d) {
     return cmd_response != BUSY && cmd_response != PROCESSING;
 }
 
+static int cdrom_check_drive_ready(void *d) {
+    int rv = syscall_gdrom_check_drive(d);
+    return rv != BUSY;
+}
+
 static int cdrom_check_abort_done(void *d) {
     syscall_gdrom_exec_server();
 
@@ -236,8 +241,8 @@ int cdrom_abort_cmd(uint32_t timeout, bool abort_dma) {
 
 /* Return the status of the drive as two integers (see constants) */
 int cdrom_get_status(int *status, int *disc_type) {
-    int rv = ERR_OK;
     uint32_t params[2];
+    int rv;
 
     /* We might be called in an interrupt to check for ISO cache
        flushing, so make sure we're not interrupting something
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -236,18 +273,13 @@ int cdrom_get_status(int *status, int *disc_type) {
...
 
-    do {
-        rv = syscall_gdrom_check_drive(params);
-
-        if(rv != BUSY) {
-            break;
-        }
-        thd_pass();
-    } while(1);
+    rv = cdrom_poll(params, 0, cdrom_check_drive_ready);
```
DC-SWAT, line 290: `syscall_gdrom_check_drive()` is not a command, so doesn't need call `syscall_gdrom_exec_server()` (`cdrom_poll()`) here.

kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -591,55 +588,21 @@ int cdrom_stream_request(void *buffer, size_t size, bool block) {
...
-            else if(syscall_gdrom_dma_check(cmd_hnd, &check_size) == 0) {
-                break;
-            }
-            thd_pass();
-
-        } while(1);
-    }
-    else if(stream_mode == CDROM_READ_PIO) {
-
+    } else {
```
DC-SWAT:
```suggestion
    }
    else {
```

kernel/arch/dreamcast/hardware/cdrom.c
```diff
@@ -591,55 +588,21 @@ int cdrom_stream_request(void *buffer, size_t size, bool block) {
...
-                }
-                break;
-            }
-            thd_pass();
-        } while(1);
+    data = (struct cmd_transfer_data){ cmd_hnd, 0 };
+
+    if(cdrom_poll(&data, 0, cdrom_check_transfer) == ERR_NO_ACTIVE) {
+        cmd_hnd = 0;
+    } else if(stream_mode == CDROM_READ_PIO) {
```
DC-SWAT:
```suggestion
    }
    else if(stream_mode == CDROM_READ_PIO) {
```

DC-SWAT (approved):

QuzarDC (approved):
The changes all track and the refactorization is great for the code.

PR #1091, opened 2025-07-04
examples: Add fb_tex example

pcercuei:
This example demonstrates a trick that allows a RGB565 framebuffer to be used as a texture.

This is useful when additive rendering is needed (where the previous frame is reused as the background of the new frame), or to emulate blur.

### DIFF
--- a/examples/dreamcast/pvr/fb_tex/fb_tex.c
+++ b/examples/dreamcast/pvr/fb_tex/fb_tex.c
@@ -1,0 +1,313 @@
+/* Simple program that shows how the back buffer (the one being currently
+ * presented to the screen) can be used as a texture when rendering to the
+ * front buffer. */
+/* Copyright (C) 2025 Paul Cercueil */
+
+#include <kos.h>
+#include <stdio.h>
+#include <stdbool.h>
+
+/* XXX: KallistiOS has the wrong value for the stride bit. */
+#undef PVR_TXRFMT_STRIDE
+#define PVR_TXRFMT_STRIDE BIT(25)
+
+#define SQUARE_SIZE 64
+
+#define FRONT_RENDER_COLOR 0xfff8f8f8
+
+struct square_fcoords {
+    float x[4];
+    float y[4];
+    float u[4];
+    float v[4];
+};
+
+static bool done;
+
+static uint32_t dr_state;
+
+static void do_exit(uint8_t, uint32_t)
+{
+    done = 1;
+}
+
+alignas(4) static const uint16_t fake_tex_data[] = {
+    /* Alternating 0x8000 / 0x0000 but pre-twiddled */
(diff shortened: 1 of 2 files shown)


pcercuei:
Updated the PR with some more stuff:
- An API function to enable/disable dithering;
- An update to `pvr_get_front_buffer()` which makes it usable with the framebuffer texture trick,
- The fb_tex example now displays better colors and can switch with A/B between a "blur" look, and a regular overpaint.

examples/dreamcast/pvr/fb_tex/fb_tex.c
```diff
@@ -0,0 +1,304 @@
+/* Simple program that shows how the back buffer (the one being currently
+ * presented to the screen) can be used as a texture when rendering to the
+ * front buffer. */
+/* Copyright (C) 2025 Paul Cercueil */
+
+#include <kos.h>
+#include <stdio.h>
+#include <stdbool.h>
+
+/* XXX: KallistiOS has the wrong value for the stride bit. */
```
gyrovorbis, line 10: I meant to ask, there must be some reason why we're not fixing this within the driver, right? Would it break backwards compat? I mean... do we really care, since it's wrong?

examples/dreamcast/pvr/fb_tex/fb_tex.c
```diff
@@ -0,0 +1,304 @@
...
+};
+
+static void render_coords(const struct square_fcoords *coords, float z,
+                          uint32_t argb, float uoffset) {
+    pvr_vertex_t *vert;
+    unsigned int i;
+
+    for(i = 0; i < 4; i++) {
+        vert = pvr_dr_target(dr_state);
+        *vert = (pvr_vertex_t){
```
gyrovorbis, line 79: Damn, I've never done a single assignment like that for the pvr dr pointer... cool.

gyrovorbis (approved):
Absolutely gorgeous and beautiful example both in terms of output and code... 

I could see literally nothing to complain about... Very detailed explanations to help simplify what is a relatively complex novel thing you guys figured out how to do, but also with a real-world use-case that I feel like will benefit any game wanting to do FB effects.

Plus I saw a bunch of "good practices" with the PVR DR API in terms of how to use it well and doing things like compiling the polygon header directly into the SQs, which we have never done before in KOS examples...

Grade A+.

examples/dreamcast/pvr/fb_tex/fb_tex.c
pcercuei: So the story is, BBHoodsta made a PR to fix it in KOS, and added a wrapper to set the stride length register - then rage quit to make his fork. We can either salvage his commits back into KOS (it's BSD license after all) or reimplement something similar.

PR #1094, opened 2025-07-09
timer: Fine-tune timer rate

pcercuei:
The seconds counter has been reported to drift quite significantly vs. a wall clock. This is most likely because until now we considered that the peripheral clock was a perfect 50.0 MHz, while in practice it is closer to 49.9 MHz.

Before merging, somebody should test it against https://github.com/CSword123/Dreamclock-DC-/ and see if the new value works better there. I could do it myself but for the time being I can't find a way to compile Raylib, I get ICEs with many different toolchains.

### DIFF
--- a/kernel/arch/dreamcast/kernel/timer.c
+++ b/kernel/arch/dreamcast/kernel/timer.c
@@ -65,8 +65,11 @@ typedef enum PCK_DIV {
 #define TIMER_TPSC      PCK_DIV_4
 /* Timer IRQ priority levels (0-15) */
 #define TIMER_PRIO      15
-/* Peripheral clock rate (50Mhz) */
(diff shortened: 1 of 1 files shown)


pcercuei:
Related: https://dcemulation.org/phpBB/viewtopic.php?p=1061895

gyrovorbis:
Ugh. Just so you guys know, I actually sat down to start looking at this already, and BAM. Freaking raylib suddenly never wants to compile for me ever again without ICE-ing the compiler.

I think we have a situation here... working on a "patch" we can apply when installing it via kos-ports... Pretty sure if I just break the function for `DrawCylinder()` into multiple smaller (non-inlinable) routines, it'll go away... hopefully...

QuzarDC (approved):

Issue #1044, opened 2025-05-22
RFC: KOS project organization and multi-architecture considerations for 2.3.0/3.0.0 release


darcagn:
I have a number of opinions and considerations for the overall project organization, many (but not all) of which involve necessary changes for restoring multi-arch to KOS. 
I'm not married or firm on any of these opinions, but I have a lot of ideas swirling around so I figured it'd be better to write them all out.
If these changes get large and wide enough, perhaps it might even be enough to justify a 3.x release.

## KOS root
- Do we expect users to maintain separate KallistiOS installations per-platform? e.g. `/opt/toolchains/dc/kos`, `/opt/toolchains/gc/kos`, etc.?
Or do we expect that one KOS installation will support all platforms? (I would prefer this)

  - I propose that we remove the arch from the base dir and make it something like `/opt/toolchains/kos` to contain the KallistiOS environment but `/opt/toolchains/kos/kos` would be awkward for the KOS dir itself. Well, speaking of that...

- We should come up with a new standardized dir for toolchains. `/opt/toolchains/dc/sh-elf` is no longer a great convention since many of us keep multiple toolchains. Right now, I put all of my toolchains in `/opt/toolchains/dc/toolchains/<profile_name>` and use custom helper scripts to change between toolchains and symlink them over to `/opt/toolchains/dc/sh-elf`. 

  - Perhaps we could do something like `/opt/kallistios` for the overall KallistiOS environment, then use `/opt/kallistios/kos` for the `$KOS_BASE`, then `/opt/kallistios/toolchains/${KOS_ARCH}/${PROFILE_NAME}` for toolchains themselves?

    - Then, with true support for multiple toolchain profiles, we could potentially use Makefile variables to enforce the use or non-use of particular profiles (so for example the libmp3 kos-port wouldn't error our on 15.1.0 but would instead use 13.2.0 or something). 

[...]

PR #1097, opened 2025-07-12
DC-chain rework

pcercuei:
This PR reworks DC-chain to make it more cross-platform.

- Drop GDB support, in favor of using gdb-multiarch, which is generally available on Linux distributions, is available on OSX via homebrew, and packaged by mingw-w64.
- Drop old code to build the ARM toolchain, so that DC-chain only targets a single architecture (at a time).
- Misc changes to support multiple architectures
- Bump `__KOS_GCC_PATCHLEVEL__` and drop `__KOS_GCC_32MB__` (KOS will consider that toolchains whose patchlevel is >= the bumped version are assumed compatible with 32 MiB)
- Rework startup code handling; it is no longer copied into the toolchain.
- Shallow clone GIT repositories to boost clone speed.
- Auto-detect number of CPU threads and use that by default.
- Add preliminary PPC/GameCube toolchain support
- Re-introduce support for building an ARM toolchain through the regular code path (that used to be the SH toolchain path)
- Add LRA toolchain based on Oleg's tree.


pcercuei:
> Looking great! Few outstanding things I can think of:
> 
>     1. When running `make distclean` there are still empty `rm -rf` calls.

Done.

>     2. The `kos-gdb` wrapper needs to be more flexible for the binary name. `gdb-multiarch` is on a lot of distros but `gdb-multiarch` is just a custom binary name, and won't exist on a distro like Gentoo where the standard `gdb` package can be compiled with custom arch support. Perhaps a `$KOS_GDB` variable could hold the `gdb` binary name, similar to `$KOS_MAKE`, etc.?

And what would set `$KOS_GDB`, if the name is potentially different on some distribs? Should we have it in the configure.sh, defaulting to `gdb-multiarch`?
 
>     3. In the `Makefile.gamecube.cfg` we should probably use `/opt/toolchains/gc` instead of `/opt/toolchains/dc`, that's what I have been using locally as IMO arch installation dirs will likely be separated for a while until we discuss and implement changes discussed in [RFC: KOS project organization and multi-architecture considerations for 2.3.0/3.0.0 release #1044](https://github.com/KallistiOS/KallistiOS/issues/1044)

Done.
 
>     4. Since this isn't Dreamcast-exclusive anymore, I think we should rename `dc-chain` to `kos-chain` (so change path from `utils/dc-chain` to `utils/kos-chain`, change the names in the headers, change the documentation, etc.).

Yes, although this could be done later, in a PR with no functional change.
 
>     5. The documentation needs changing along with this.
> 
> 
> If you'd like help (especially with dealing with last point, the documentation), let me know; you've already done a ton on this PR and it's great work, thanks!

I did some minor edits of the doc. It might be a bit too soon to talk about how to build GC toolchains if we don't have official KOS support for it, IMHO.

darcagn:
> And what would set `$KOS_GDB`, if the name is potentially different on some distribs? Should we have it in the configure.sh, defaulting to `gdb-multiarch`?

In `environ.sh` like `$KOS_MAKE` is right now. Suitable defaults would be:
- `sh-elf-gdb` for one that is installed manually into the KOS toolchain
- `gdb` for the normal system GDB which may or may not have SuperH support
- `gdb-multiarch` for users running a distro that has the `gdb-multiarch` package installed

Personally _I'd_ think `gdb` is the most logical default since that is just the standard bin name for GDB, but I also get why `gdb-multiarch` could be a better default for end users, so I won't object to any of those three being used as the default.

> Yes, although this could be done later, in a PR with no functional change.

> I did some minor edits of the doc. It might be a bit too soon to talk about how to build GC toolchains if we don't have official KOS support for it, IMHO.

Yeah, we definitely don't need any documentation on GC stuff yet, I just wanted to make sure the instructions for building weren't incorrect, so mostly just stuff like your `Makefile.default.cfg` -> `Makefile.dreamcast.cfg` change. But as it stands the main `README.md` hasn't been updated and is completely wrong, so if we merged this now, it won't be long before users will be asking how to build the toolchain.

Likewise with the `kos-chain` change, I don't like big huge PRs but for end users it'd be better if the big changes were made all at once so the documentation in here and on dreamcast.wiki doesn't have to be changed a bunch of times. 

If you want, you can take care of the GDB change above (unless I find something else in further testing, I think that's the last change I'll be requesting), and I can knock out the documentation changes in a separate PR, and we can have @QuzarDC merge both PRs at once?

pcercuei:
Sorry for the force-push noise.

I updated `doc/environ.sh.sample` to have a (unset by default) `$KOS_GDB` configuration variable that can be used to point the `kos-gdb` script to use: https://github.com/KallistiOS/KallistiOS/pull/1097/files#diff-a745f6ca5e1a8c3333de42c20b7e28a142cb087de107c1d47b7c5de042a2f0a4

The `kos-gdb` script will now try to locate the right GDB program for the target (unless overridden):
https://github.com/KallistiOS/KallistiOS/pull/1097/files#diff-7ee88743c823c470f50f3fcb1e248f5e54b278cca1ff0639c96558c363754c7b

I'll be out most of next week so feel free to work on the doc.

utils/dc-chain/Makefile.dreamcast.cfg
```diff
@@ -5,6 +5,12 @@
 ### TOOLCHAIN PROFILE ###
 #########################
 
+# One of the supported platforms:
+# - dreamcast:    Build a toolchain for the SEGA Dreamcast
+# - aica:         Build a toolchain for the Dreamcast's AICA co-processor
```
sizious, line 10: Although I really like the idea, is the "aica" choice explicit enough? Or should we use something like "dreamcast-aica"?

utils/dc-chain/Makefile
```diff
@@ -69,12 +69,6 @@ include scripts/gcc-pass2.mk
 
 # ---- }}}}
 
-# ---- optional components {{{
-
-include scripts/gdb.mk
```
sizious: Is that possible to first mark it "deprecated" instead of just removing it? I need to evaluate if `gdb-multiarch` embedded in MinGW-w64 can be usable, i.e., my main concern is about if this package provides a static or dynamic linked binary. If the binary is dynamic, then it won't be usable on IDEs on Windows as far as I know.

utils/dc-chain/Makefile.dreamcast.cfg
pcercuei: Can you think of anything else that AICA could stand for? ;)

Besides, I'd rather have names that look different to avoid confusion.

utils/dc-chain/Makefile
pcercuei: Ideally we'd test all that before merging.

QuzarDC:
> > And what would set `$KOS_GDB`, if the name is potentially different on some distribs? Should we have it in the configure.sh, defaulting to `gdb-multiarch`?
> 
> In `environ.sh` like `$KOS_MAKE` is right now. Suitable defaults would be:
> 
>     * `sh-elf-gdb` for one that is installed manually into the KOS toolchain
> 
>     * `gdb` for the normal system GDB which may or may not have SuperH support
> 
>     * `gdb-multiarch` for users running a distro that has the `gdb-multiarch` package installed
> 
> 
> Personally _I'd_ think `gdb` is the most logical default since that is just the standard bin name for GDB, but I also get why `gdb-multiarch` could be a better default for end users, so I won't object to any of those three being used as the default.
> 
> > Yes, although this could be done later, in a PR with no functional change.
> 
> > I did some minor edits of the doc. It might be a bit too soon to talk about how to build GC toolchains if we don't have official KOS support for it, IMHO.
> 
> Yeah, we definitely don't need any documentation on GC stuff yet, I just wanted to make sure the instructions for building weren't incorrect, so mostly just stuff like your `Makefile.default.cfg` -> `Makefile.dreamcast.cfg` change. But as it stands the main `README.md` hasn't been updated and is completely wrong, so if we merged this now, it won't be long before users will be asking how to build the toolchain.
> 
> Likewise with the `kos-chain` change, I don't like big huge PRs but for end users it'd be better if the big changes were made all at once so the documentation in here and on dreamcast.wiki doesn't have to be changed a bunch of times.
> 
> If you want, you can take care of the GDB change above (unless I find something else in further testing, I think that's the last change I'll be requesting), and I can knock out the documentation changes in a separate PR, and we can have @QuzarDC merge both PRs at once?

Sounds good by me. I did start on a pass of reviewing here, largely with an eye towards the documentation and where it might need updating. I figure I'll go ahead and continue that and at worst it can end up being just a secondary checklist for the doc PR.

darcagn:
Perhaps you could add the `__builtin_is_aligned()` macro fix to Newlib 4.3.0 and 4.5.0 in this PR as well?

pcercuei:
@darcagn done.

utils/dc-chain/Makefile
pcercuei: @sizious Could you maybe have a go at mingw-w64's GDB, see if it is usable in our case with the new `kos-gdb` wrapper?

pcercuei:
Force-pushed, with GDB build re-introduced and fixed to build for multiple archs (in theory).

I kept the `kos-gdb` script which will happily work with a sh-elf-gdb as well as gdb-multiarch.

sizious:
Thank you @pcercuei, unfortunately I didn't had time to check this, but I'll be in holidays in 2 weeks, probably I'll be able to test this. I suggest that for now you can keep gdb and as soon as I validate the fact gdb-multiarch is usable, we will drop regular gdb.

pcercuei:
@darcagn force-pushed with `clean` / `distclean` now working properly regarding GDB stuff, and the "drop GDB" entry in the changelog reworked.

This PR did not touch the `README.md` so there's nothing to restore.

darcagn (approved):

QuzarDC (approved):

PR #1099, opened 2025-07-16
Check for call to `snd_stream_poll` before `snd_stream_start`

QuzarDC:
Add a check matching what was done for `snd_stream_fill` that will assert if called before starting the stream. Without this, it can trigger an otherwise mysterious divide by zero error.


gyrovorbis (approved):
Ah, okay, I remember jnmartin bringing this one up. Looks like this is in-line with what `snd_stream_fill()` is doing, plus I like it being a debug-only check that isn't paid for in release builds. Cool.

PR #1085, opened 2025-06-23
Macro cleanups

QuzarDC:
Some minor cleanups to the use of standard macros.

PR #1103, opened 2025-07-22
dmac: Work around bug in `__is_aligned` macro.

QuzarDC:
It seems that the macro can provide incorrect feedback when not used on a single value. Changing this allows the dmac speed test example to perform dmac writes, which otherwise were being flagged as unaligned.

This seems to have slipped by in my v2.2.0 testing as the dmac speedtest was also broken by a dcload-ip bug. Even with this change, the pvr_dma function in the speedtest still seems to hang. I did test reverting the pvr buffers change from [the same commit](https://github.com/KallistiOS/KallistiOS/pull/1085/commits/4111c0c4a42ca96897798012fccbcb1d40406ffe), and that had no impact.

### DIFF
--- a/kernel/arch/dreamcast/hardware/dmac.c
+++ b/kernel/arch/dreamcast/hardware/dmac.c
@@ -143,8 +143,9 @@ int dma_transfer(const dma_config_t *cfg, dma_addr_t dst, dma_addr_t src,
     unsigned int transfer_size = dma_unit_size[cfg->unit_size];
     uint32_t chcr;
 
-    if(!__is_aligned(len | dst | src, transfer_size)) {
-        dbglog(DBG_ERROR, "dmac: src/dst/len not aligned to the bus width\n");
+    if(!__is_aligned((src|dst|len), transfer_size)) {
+        dbglog(DBG_ERROR, "dmac: src=0x%08x dst=0x%08x len=%u not aligned to %u bytes\n",
+               (uintptr_t)src, (uintptr_t)dst, (uintptr_t)len, transfer_size);
         errno = EFAULT;
         return -1;
     }


QuzarDC:
> `otehrwise` -> `otherwise`.
> 
> The issue is actually in Newlib, which defines `__builtin_is_aligned()` without protecting its first argument. I sent a patch upstream to address it.
> 
> You can make the code a little bit simpler (since you should fix the typo anyway ;)) either by using parentheses around "len | src | dst", or by caching this to a u32 temporary variable.

Fixed in the force-push. I went ahead and rolled it back to just adding the parens, but left the nicer debug message (and fixed the typo in the commit message).

In case others wanna see: https://sourceware.org/pipermail/newlib/2025/022055.html

GPF:
```
index [hash]..[hash] 100644
--- a/kernel/arch/dreamcast/hardware/g2dma.c
+++ b/kernel/arch/dreamcast/hardware/g2dma.c
@@ -173,12 +173,7 @@ int g2_dma_transfer(void *sh4, void *g2bus, size_t length, uint32_t block,
     }
 
     /* Check alignments */
-    if(!__is_aligned(sh4, 32)) {
-        dbglog(DBG_ERROR, "g2_dma: Unaligned sh4 DMA %p\n", sh4);
-        errno = EFAULT;
-        return -1;
-    }
-
-    if(!__is_aligned(g2bus, 32)) {
+    if(!__is_aligned(sh4, 32) || !__is_aligned(g2bus, 32)) {
         dbglog(DBG_ERROR, "g2_dma: Unaligned g2bus DMA %p\n", g2bus);
         errno = EFAULT;
         return -1;
```

this  fixes the rest of the /opt/toolchains/dc/kos/examples/dreamcast/basic/dma/speedtest 
DMAC, RAM to RAM:   121.222656 MiB/s
DMAC, RAM to VRAM:  74.109550 MiB/s
DMAC, VRAM to RAM:  38.474205 MiB/s
DMAC, VRAM to VRAM: 26.030235 MiB/s
PVR SQs:            224.486404 MiB/s
PVR DMA, 64-bit:    548.131714 MiB/s
PVR DMA, 32-bit:    333.410492 MiB/s

pcercuei:
> Fixed in the force-push. I went ahead and rolled it back to just adding the parens, but left the nicer debug message (and fixed the typo in the commit message).

Did you force-push? The commit still has the three `__is_aligned()`.
I don't care that much TBH, but it sounded like you wanted a version that simply added parentheses.

QuzarDC:
> > Fixed in the force-push. I went ahead and rolled it back to just adding the parens, but left the nicer debug message (and fixed the typo in the commit message).
> 
> Did you force-push? The commit still has the three `__is_aligned()`. I don't care that much TBH, but it sounded like you wanted a version that simply added parentheses.

Just entirely missed the force-push or it went into the ether. Updated the description text as well just to be more clear, and made sure to see that github got a new force-push.

PR #1104, opened 2025-07-22
Add new arch-agnostic timer API and implement old timer functions on top

pcercuei:
Add new arch-agnostic timer functions, that should be provided by all supported architectures.

The previous API functions are now static inlines that are built on top.

The point of this PR is to reduce the amount of code duplicated between the Dreamcast arch and the other upcoming archs.

### DIFF
--- a/kernel/arch/dreamcast/hardware/g1ata.c
+++ b/kernel/arch/dreamcast/hardware/g1ata.c
@@ -965,7 +965,7 @@ int g1_ata_flush(void) {
 
     /* Select the slave device. */
     g1_ata_select_device(G1_ATA_SLAVE | G1_ATA_LBA_MODE);
-    timer_spin_sleep(1);
+    thd_sleep(1);
 
     /* Flush the disk's write cache to make sure everything gets written out. */
     if(CAN_USE_LBA48())
@@ -973,7 +973,7 @@ int g1_ata_flush(void) {
     else
         OUT8(G1_ATA_COMMAND_REG, ATA_CMD_FLUSH_CACHE);
 
-    timer_spin_sleep(1);
+    thd_sleep(1);
     g1_ata_wait_bsydrq();
     g1_ata_mutex_unlock();
 
@@ -1006,7 +1006,7 @@ static int g1_ata_set_transfer_mode(uint8_t mode) {
 
     /* Send the SET FEATURES command. */
     OUT8(G1_ATA_COMMAND_REG, ATA_CMD_SET_FEATURES);
-    timer_spin_sleep(1);
+    thd_sleep(1);
 
     /* Wait for command completion. */
     g1_ata_wait_nbsy();
@@ -1034,7 +1034,7 @@ static int g1_ata_scan(void) {
     /* For now, just check if there's a slave device. We don't care about the
        primary device, since it should always be the GD-ROM drive. */
     OUT8(G1_ATA_DEVICE_SELECT, 0xF0);
-    timer_spin_sleep(1);
+    thd_sleep(1);
 
     OUT8(G1_ATA_SECTOR_COUNT, 0);
     OUT8(G1_ATA_LBA_LOW, 0);
@@ -1043,7 +1043,7 @@ static int g1_ata_scan(void) {
 
     /* Send the IDENTIFY command. */
     OUT8(G1_ATA_COMMAND_REG, ATA_CMD_IDENTIFY);
-    timer_spin_sleep(1);
+    thd_sleep(1);
     st = IN8(G1_ATA_STATUS_REG);
 
     /* Check if there's anything on the bus. */
--- a/kernel/arch/dreamcast/hardware/g1ata.c
+++ b/kernel/arch/dreamcast/hardware/g1ata.c
@@ -17,7 +17,7 @@
 #include <kos/mutex.h>
 #include <kos/thread.h>
 
-#include <arch/timer.h>
+#include <kos/timer.h>
 #include <arch/cache.h>
 #include <arch/irq.h>
 #include <arch/memory.h>
--- a/kernel/arch/dreamcast/include/arch/timer.h
+++ b/kernel/arch/dreamcast/include/arch/timer.h
@@ -82,8 +82,7 @@ __BEGIN_DECLS
 /** \brief  SH4 Timer Channel 1.
 
     \warning
-    This timer channel is used for the timer_spin_sleep() function, which also
-    backs the kthread, C, C++, and POSIX sleep functions.
+    This timer channel is free to use.
 */
 #define TMU1    1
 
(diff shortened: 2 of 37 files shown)


pcercuei:
The whitespace check failed, but the good kind of "failed" - two lines had extra spaces at the end and were cleaned up.

kernel/arch/dreamcast/include/arch/timer.h
```diff
@@ -437,5 +325,6 @@ void timer_shutdown(void);
 
 __END_DECLS
 
-#endif  /* __ARCH_TIMER_H */
+#include <kos/timer.h>
```
QuzarDC, line 328: If this is needed for any files internally that were getting the API from <arch/timer.h> then they should be updated there as well. We can then just have this as a temporary compatibility measure (unless I missed something and it's required here).

kernel/arch/dreamcast/hardware/g1ata.c
```diff
@@ -965,15 +965,15 @@ int g1_ata_flush(void) {
 
     /* Select the slave device. */
     g1_ata_select_device(G1_ATA_SLAVE | G1_ATA_LBA_MODE);
-    timer_spin_sleep(1);
```
QuzarDC, line 968: I think the reasoning here, generally, was to spin for the minimum time that our API allowed, but still be irq-safe. If going to a thread-based function, then `thd_pass` might be more appropriate as it would be the minimum amount of waiting the thread system can provide. But neither would actually have any sort of delay inside an irq. 

I saw that in the other uses of `timer_spin_sleep` it was being done for inits and resets and such that are more squarely not likely to be things for inside an IRQ, so the rest of the above ends up being fine. I don't have a g1ata and am not too familiar with its API, so just wanted to verify that none of these would be likely to be used in an IRQ.

Perhaps it's a thing as well that @DC-SWAT could weigh in on in his usage of these.

kernel/arch/dreamcast/include/arch/timer.h
```diff
@@ -333,21 +332,7 @@ uint64_t timer_ns_gettime64(void);
...
-
-/** \brief  Spin-loop sleep function.
-    \ingroup tmu_sleep
-
-    This function is meant as a very accurate delay function, even if threading
-    and interrupts are disabled. It uses \ref TMU1 to sleep.
-
-    \param  ms              The number of milliseconds to sleep.
 */
-void timer_spin_sleep(int ms);
```
QuzarDC, line 350: This should just be deprecated, rather than removed. From a quick search on github, it is actually being used in people's codebases unfortunately. And unlike general compatibility, as you rightly point out, this should pretty much never be done and is a valuable thing to warn users about.

It could be easily swapped to a wrapper around `timer_spin_delay_us` and still get rid of the whole unnecessary use of TMU1.

kernel/arch/dreamcast/kernel/timer.c
```diff
@@ -263,7 +263,7 @@ static timer_val_t timer_getticks(const uint32_t *tns, uint32_t shift) {
     integer, otherwise when using the fastest clock speed for timers,
     this value will very quickly overflow mid-expression, before the
     final division. */
-    ticks = ((uint64_t)delta * tns[tmu2 & TPSC]) >> shift;
```
QuzarDC, line 289: Just adding as a note, and no change, but this previously allowed for the timer resolution to be changed at run-time. Considering though that there's no API that allows that, and not really any reason I could imagine why that'd be desirable, it makes sense to work off a compile-time value.

kernel/arch/dreamcast/include/arch/timer.h
```diff
@@ -230,6 +230,22 @@ int timer_ints_enabled(int channel);
...
+
+/** \brief   Get the current uptime of the system (in seconds and ticks).
+    \ingroup tmu_uptime
+
+    This function retrieves the number of seconds and ticks since KOS was
+    started.
+
+    \return                 The time since KOS started as a timer_val_t.
+*/
+timer_val_t timer_get_ticks(void);
```
QuzarDC: Is there a usefulness for this to be part of the arch API aside from how you are using it as static inline to `arch_timer_gettime` ? It seems like it's effectively a private function, and isn't needed from the KOS timer layer right?

Is the naming meant to be what segregates it from part of the arch API, not being prefixed as `arch_timer_get_ticks`? So for the arch headers anything *not* named `arch_` is just part of the platform implementation?

Maybe @darcagn would also have thoughts to add on it.

QuzarDC (changes requested):
Mostly on-board, just some minor things to help compatibility, questions on how we might go about the arch layer, and a note that I'll mark resolved immediately.

kernel/arch/dreamcast/hardware/g1ata.c
pcercuei: I used `thd_sleep(1)` here because it is guaranteed to not run in an IRQ context. There's one call from `g1_ata_shutdown()`, and the other call is from the `kos_blockdev_t.sync()` implementation, which should never be called from an IRQ context either.

`thd_pass()` would work too.

kernel/arch/dreamcast/include/arch/timer.h
pcercuei: This is for compatibility, yes - as `timer_ms_gettime()` and friends are moved to the main `<kos/timer.h>` header, the apps that just include `<arch/timer.h>` will break.

Unless the apps should always just include `<kos.h>`, and in this case I can drop this.

kernel/arch/dreamcast/include/arch/timer.h
pcercuei: I was thinking of making it a deprecated wrapper around `thd_sleep()`.

kernel/arch/dreamcast/include/arch/timer.h
pcercuei: It is not useful in the arch API, but it still needs to be defined somewhere.

So either it's defined in the `<arch/timer.h>` header, so that `arch_timer_gettime()` can be "static inline", or it's defined in a private header e.g. `<dc/tmu.h>`, which means the arch header would have to include it, and pull all the platform-specific TMU stuff into the arch header. Last option is to not make `arch_timer_gettime()` static inline, with a performance penalty (because of the two mults not being folded into one).

QuzarDC: I think what would be sufficient, as we'd discussed via discord, would be then to leave it in this `arch` header, but denote it as platform specific. I don't think there was any objection to simply prefixing it as `__timer_get_ticks()`, but perhaps `__dc_get_ticks()` would more clearly demonstrate that it's dc-specific and not `arch` layer or `kos` layer.

kernel/arch/dreamcast/include/arch/timer.h
QuzarDC: The problem with that is that it isn't the same behavior. It made sense for being able to remove the usages for existing code, as we could know that `thd_sleep()` would work the same way. Since the whole point is to provide compatibility for a version while it's being removed, we shouldn't *also* needlessly change the behavior.

kernel/arch/dreamcast/include/arch/timer.h
QuzarDC: Ok, then other files internal to the codebase should also be updated to include `<kos/timer.h>` in place of `<arch/timer.h>` . 

I checked through kos-ports and fortunately all our hosted ones that use timer are just including <kos.h> and don't need any changes.

PR #1105, opened 2025-07-22
CMake: Use absolute path for KOS variables

pcercuei:
The CMAKE_INSTALL_BINDIR and CMAKE_INSTALL_LIBDIR require absolute paths. Address this issue by using absolute paths for all KOS variables that contain paths.


gyrovorbis (approved):
Good call. Not sure why TF I never encountered that, but thanks for fixing!

PR #1108, opened 2025-07-26
dmac: Restrict dma_init back to NAOMI.

QuzarDC:
Reverts the behavior change from #983 that had
KOS performing the same initialization steps on
DC as NAOMI. This was breaking PVR dma as seen
via the dma speedtest example.


QuzarDC:
> Does the PVR DMA example actually work on Naomi?

I don't have a way to test it, I have yet to be able to run anything KOS on my NAOMI, but #983 should have been transparent on the NAOMI as it's just using the API to set the same regs to the same values that they had been using previously. The functional change was that the same prep was now being done on the DC too.

Was your speedtest example working on a DC for you after #983?

pcercuei (approved):

Issue #1092, opened 2025-07-09
KOS hangs if dctool-ip-siz is used with network based code without forcing `-l` for legacy packet size


QuzarDC:
Moop's updates to dctool-ip expanded the packet size from the legacy size of 1024 supported by mainline dctool-ip. It provides a flag `-l` that can be used to force this legacy mode, but without it KOS will hang in most network functionality. This is because `fs_dclsocket` hard-codes the legacy size for its packets.

This is still supported by our mainline dctool-ip, but most use the newer Siz repo. So some possible work to be done:
1) Update `fs_dclsocket` to be compatible with the larger packet sizes.
2) Update dctool-ip-siz' detection mechanism to work properly in this mode.

More investigation would be needed to address this, but generally the common knowledge for years has been that if you use dctool-ip then you cannot also be using KOS net, while that isn't true. It will work with mainline dctool-ip without disconnecting the console, and will work with the siz version as well as long as the `-l` flag is passed to force legacy compatibility.

PR #1110, opened 2025-07-27
dcload driver refactor

QuzarDC:
Mostly an internal refactoring of the functionality around dcload. Each commit has details of its purpose, but there were a few general goals:
- Disentangle the dcload vfs from a general driver for the dcload syscall.
- Refactor the dcload syscall functionality to match other syscalls.
- Provide access to previously unused dcload commands (chdir, chmod, fstat, time, utime, and exit), and a simpler path for adding new ones.
- Redo concurrence protection which was not being split between protecting the syscall and protecting the FS elements.
- General code health (stdint types, whitespace, removing junk).

The next step I had in mind was to see about replacing the functionality of `fs_dclsocket` entirely and refactor its behavior to happen at the syscall level rather than entangled again with a vfs. Per #1092 would also be looking to ensure it can use the new packet sizes so thought the amount of changes would merit splitting it out rather than having it together with these.


kernel/arch/dreamcast/hardware/dcload.c
```diff
@@ -0,0 +1,134 @@
...
+static int dcload_syscall(dcload_cmd_t cmd, void *param1, void *param2, void *param3) {
+    uintptr_t *syscall_ptr = (uintptr_t *)VEC_DCLOAD;
+    int (*syscall)() = (int (*)())(*syscall_ptr);
+
+    /* Disable IRQs until the syscall returns */
+    irq_disable_scoped();
+
+    /* Ensure that the FIFO buffer is clear */
+    /* XXX - Is this needed? It seems like something only for serial. */
+    while(FIFO_STATUS & FIFO_SH4)            ;
```
DC-SWAT, line 33: I don't see any point in waiting for SH4 FIFO here, including for serial.

QuzarDC: It occurs to me this *might* be intended for `dcload-ip` as a primitive version of `g2_lock`, to help prevent the possibility of breaking in-progress writes. If so though, that should be handled on the dcload side.

Will try to do some testing without it, but may leave removing it to a follow-up PR just to reduce the chance of breakage, and not holding up the rest of the refactoring and cleanup.

DC-SWAT: I just looked at the BBA driver in dcload and realized that they are really cruel to the G2 bus. There are no locks, FIFOs or DMA pauses at all. There is only hope that when the dcload system call is called, interrupts are turned off (which, by the way, leads to lags) and here there is a wait for FIFO, however, once, and not every 8x32-bit words.
Now it is clear why audio streaming does not work with dcload. We need to add to g2_lock(), g2_read() and g2_write() inside dcload, like in KOS, and not turn off interrupts when calling dcload syscall, just sync g2_lock() with KOS by some way.

QuzarDC: After doing some more looking about this, I agree entirely it's something needed but certainly wouldn't be something to hold this cleanup/refactor back.

I've created a new issue #1176 so that we can have some tracking and visibility of the future work needed.

QuzarDC:
Force-pushed to remove conflict created by #1168 and created new issue to track synchronization problems discussed in the comments above.

QuzarDC:
Force-pushed to correct typo of `chmod` rather than `dcload_chmod`.

QuzarDC:
Force pushed to incorporate the offered update in #1206 . Additionally, had to make a further adjustment due to a merge conflict for an include and updated all the intermediate commits to ensure they build. I had originally tested this only with `stable` gcc 13 but on 14 some warnings had become errors.

kernel/arch/dreamcast/fs/fs_dcload.c
```diff
@@ -101,35 +86,19 @@ static void *fs_dcload_open(vfs_handler_t *vfs, const char *fn, int mode) {
             return (void *)NULL;
```
pcercuei, line 87: Leaking memory here ("entry" is not freed).

QuzarDC: Fixed in latest foce-push.

pcercuei (approved):

PR #1111, opened 2025-07-27
Fix strided texture settings, add API and example

QuzarDC:
This PR recreates the changes from #836 . Key differences are that there is no renaming here of the PVR register used for the global setting, the functions are being added to different files due to refactoring that has happened between then and now, some of the descriptions have been changed to emphasize these only can apply to non-twiddled textures, and compatibility consts have been provided for the old define flag names.

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
@@ -43,12 +43,9 @@ size_t pvr_txr_get_stride(void) {
 }
 
 /* Load raw texture data from an SH-4 buffer into PVR RAM */
-void pvr_txr_load(const void *src, pvr_ptr_t dst, uint32 count) {
-    if(count & 3) {
-        count = (count + 4) & ~3;
-    }
-
-    pvr_sq_load((uint32 *)dst, (const uint32 *)src, count, PVR_DMA_VRAM64);
+void pvr_txr_load(const void *src, pvr_ptr_t dst, size_t count) {
+    count = __align_up(count, 4);
+    pvr_sq_load((uint32_t *)dst, (const uint32_t *)src, count, PVR_DMA_VRAM64);
 }
 
 /* Linear/iterative twiddling algorithm from Marcus' tatest */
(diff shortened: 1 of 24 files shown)


kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
```diff
@@ -19,6 +22,28 @@
...
+bool pvr_txr_set_stride(size_t texture_width) {
+    uint32_t temp = texture_width / 32;
+
+    /* The width must be a non-zero multiple of 32 that fits in the 5 bits
+    of the register field. */
+    if(!temp || ((texture_width % 32) != 0) || (temp & ~PVR_TXR_STRIDE_MULT)) {
+        dbglog(DBG_ERROR,
+            "Texture width must be divisible by 32 and be 992 or less.");
+        return false;
+    }
```
pcercuei: This should probably be an assert(), so that it can be disabled in production code.

kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
```diff
@@ -19,6 +22,28 @@
...
+
+    /* Pull the reg, mask out current stride, and rewrite with new */
+    temp |= FIELD_GET(PVR_GET(PVR_TEXTURE_MODULO), ~PVR_TXR_STRIDE_MULT);
+    PVR_SET(PVR_TEXTURE_MODULO, temp);
```
pcercuei, line 35: This only works because the field is in the LSB bits.
It should be:
```c
temp = FIELD_PREP(PVR_TXR_STRIDE_MULT, temp);
temp |= PVR_GET(PVR_TEXTURE_MODULO) & ~PVR_TXR_STRIDE_MULT;
PVR_SET(PVR_TEXTURE_MODULO, temp);
```

kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
```diff
@@ -19,6 +22,28 @@
...
+
+    return true;
+}
+
+size_t pvr_txr_get_stride(void) {
+    return FIELD_GET(PVR_GET(PVR_TEXTURE_MODULO), PVR_TXR_STRIDE_MULT) * 32;
```
pcercuei: I'm not a fan of the nested calls like that, personally.

pcercuei (changes requested):
One major problem to fix, and two less-important comments.

kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
QuzarDC: The first line was being enforced already by the error check, so I ditched it as unnecessary. I presumed the original missed that since it was treating 992 as an arbitrary value (1024-32) rather than as representing the limits of what can be represented in the bitfield.  For it to matter you'd have to have the debug disabled (as with the switch to assert) and pass in invalid data.

In any case I added it back in for you.

QuzarDC:
Addressed feedback in the force-push.

PR #1112, opened 2025-07-29
`fs_romdisk` cleanup and dynamic file table

QuzarDC:
First off, dynamic file table allocation now following the effort in #973 . A TAILQ list holds fd objects and we pass out pointers to them rather than indexes into a table.

For the cleanups:
- Replaced all arch/types with stdint.
- Removed an unneeded param from the fd object.
- A number of refactorings.

Each change is well split out. Tried to have the refactors streamline some of the pieces that would need to change.

### DIFF
--- a/kernel/fs/fs_romdisk.c
+++ b/kernel/fs/fs_romdisk.c
@@ -16,7 +16,6 @@ on sunsite.unc.edu in /pub/Linux/system/recovery/, or as a package under Debian 
 
 */
 
-#include <arch/types.h>
 #include <kos/thread.h>
 #include <kos/mutex.h>
 #include <kos/fs_romdisk.h>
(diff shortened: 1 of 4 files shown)


kernel/fs/fs_romdisk.c
```diff
@@ -109,6 +109,11 @@ static struct {
 #define FH_INDEX_FREE 0
 #define FH_INDEX_RESERVED -1
 
+/* Test if an fd is invalid */
+static inline bool romdisk_fd_invalid(file_t fd) {
+    return (fd >= FS_ROMDISK_MAX_FILES || fh[fd].index == FH_INDEX_FREE)
```
pcercuei: Missing semicolon.

pcercuei (changes requested):
Just one major issue. I know this code gets deleted literally one commit after, but each commit should still result in a working build.

kernel/fs/fs_romdisk.c
QuzarDC: Oh good catch! Fixed and repushed.

pcercuei (approved):

PR #1115, opened 2025-07-30
Thread code cleanups

pcercuei:
My other PR #1113 is probably not going to be merged anytime soon so I decided to fork off these two cleanup commits (and I'll rebase that PR later).


QuzarDC (approved):
Good eye on the genwait refactor.

I wish we had some better insight into why that `thd_schedule` even had the `now` param. It really doesn't seem like it'd ever be useful. Through your further changes in this code it may be worth considering if `genwait_check_timeouts` needs to take the time as a param either. Though in that case it would probably have greater potential use. It seems like the fetching of the time though should be as close to the start of the testing of timeouts as possible.

PR #1118, opened 2025-08-01
dbglog: reduce default level, allow build-time disable, provide source-based defines

QuzarDC:
Does a number of things to try to make dbglog more useful.
-Reduce default level to `DBG_INFO`. It's strange that we've had it at basically the maximum level forever when this is the level noted as 'good'. Directly addresses #543 and the [feedback on the elf refactor](https://github.com/KallistiOS/KallistiOS/commit/4deb1a644bf7d686b64affb0e91f76bcb4cc8170#commitcomment-163175226). 

-Provides a define to entirely disable the debugging. We do have `INIT_QUIET` but this has the effect of simply disabling the ability for data to be sent back out through dcload rather than disabling the debug messages. So the debug will still attempt to be written out but then the end result would be no output. This methodology though should allow for the debug messages to more easily be garbage collected for a release build.

-Add macro to use any define as a `dbglog` level. This allows us to separate the default/general debug level from various subsystem specific debugging that you would only want to enable in certain circumstances and don't want all the extra noise that other debugging might provide. With `DBG_SOURCE(x)` if `x` is defined then the message will be outputted even if level is set to `DBG_DEAD` to disable all *other* output.

-Converted `fs_vmu` to use this system. Based on feedback here would walk through all the other instances in the codebase to implement this, but figured to get feedback before going through all that.


pcercuei:
One thing I had in mind for some time, about logging:

It would be great to also be able to set a configurable maximum log level at compile time. E.g. have a `#define DBG_LOG_LEVEL_SUPPORT DBG_INFO` in `<kos/opts.h>`.

Then `dbglog()` could be a macro:
```c
#define dbglog(lvl, ...) \
do { \
  if ((lvl) <= DBG_LOG_LEVEL_SUPPORT) \
    real_dbglog(lvl, __VA_ARGS__); \
} while(0)
```

The whole point of this, and where I am getting at, is that by being able to set a less-verbose value for `DBG_LOG_LEVEL_SUPPORT`, all the debug strings for the lower-level prints would be automatically garbage-collected. For a release, you could set it to `DBGLOG_DEAD` and have all the debug strings dropped from the release binary, which would effectively replace your `DBGLOG_DISABLED` with the benefits that you wouldn't retain useless data in RAM.

QuzarDC:
> One thing I had in mind for some time, about logging:
> 
> It would be great to also be able to set a configurable maximum log level at compile time. E.g. have a `#define DBG_LOG_LEVEL_SUPPORT DBG_INFO` in `<kos/opts.h>`.
> 
> Then `dbglog()` could be a macro:
> 
> ```c
> #define dbglog(lvl, ...) \
> do { \
>   if ((lvl) <= DBG_LOG_LEVEL_SUPPORT) \
>     real_dbglog(lvl, __VA_ARGS__); \
> } while(0)
> ```
> 
> The whole point of this, and where I am getting at, is that by being able to set a less-verbose value for `DBG_LOG_LEVEL_SUPPORT`, all the debug strings for the lower-level prints would be automatically garbage-collected. For a release, you could set it to `DBGLOG_DEAD` and have all the debug strings dropped from the release binary, which would effectively replace your `DBGLOG_DISABLED` with the benefits that you wouldn't retain useless data in RAM.

Ah, and here I thought my methodology would allow that as well by having the short-circuting in dbglog be set by the define rather than being strictly runtime based. Will rework it from that.

pcercuei:
> Ah, and here I thought my methodology would allow that as well by having the short-circuting in dbglog be set by the define rather than being strictly runtime based. Will rework it from that.

It might work if you have LTO enabled, but not without it - hence the need for a macro here.

QuzarDC:
> > Ah, and here I thought my methodology would allow that as well by having the short-circuting in dbglog be set by the define rather than being strictly runtime based. Will rework it from that.
> 
> It might work if you have LTO enabled, but not without it - hence the need for a macro here.

So I had to make some minor adjustments in order to allow this to still work with the exports system (which may have a cleaner way to do) but otherwise it seems to flow just right in with what I had written. I still left the `DBGLOG_DISABLED` in order to have a simple way to set the most common setting.

Will also need to see about adjusting some of the codebase around dbglog being optional as there's a pile of 'unused variable' references now for things that were only ever used to relay debug info.

pcercuei:
> Will also need to see about adjusting some of the codebase around dbglog being optional as there's a pile of 'unused variable' references now for things that were only ever used to relay debug info.

I don't understand this, `dbglog()` is not optional, it just resolves to `if (0) real_dbglog(lvl, ...)` in case the log level check failed. So the variables etc. will still be referenced and the compiler shouldn't complain about them.

QuzarDC:
> > Will also need to see about adjusting some of the codebase around dbglog being optional as there's a pile of 'unused variable' references now for things that were only ever used to relay debug info.
> 
> I don't understand this, `dbglog()` is not optional, it just resolves to `if (0) real_dbglog(lvl, ...)` in case the log level check failed. So the variables etc. will still be referenced and the compiler shouldn't complain about them.

Absolutely right. Seems it was only happening when I was still making some adjustments for the exports. For that I'd done a round where it was basically just `#define dbglog() (void)0`. Going to take the info from that though to review some of the uses.

QuzarDC:
> About the exports: I suppose you could just export `real_dbglog` instead of `dbglog`, since apps using exports like Dreamshell would still include `<kos/dbglog.h>` and therefore get the macro and try to link vs. `real_dbglog`.

Ok, done, and that worked out just fine.

PR #1119, opened 2025-08-02
Added kos_run_target() to CMake toolchain.

gyrovorbis:
It's currently a total PITA to run a CMake project you're actively building... you gotta go find the binary output wherever CMake puts it by default and call dc-tool with it..

We already have a nice, convenient, standardized rule for this kind of crap in all of our Makefiles, which you can call with "make run." It simply uses the KOS_LOADER environment variable configured from environ.sh.

This PR is an attempt to map this functionality to CMake as directly as possible.

### DIFF
--- a/utils/cmake/dreamcast.cmake
+++ b/utils/cmake/dreamcast.cmake
@@ -58,7 +58,7 @@ function(kos_bin2o inFile symbol)
     if(NOT ${ARGC} EQUAL 3)
         set(outFile ${CMAKE_CURRENT_BINARY_DIR}/${symbol}.o)
     else()
-        set(outFileOH CRA ${ARGV2})
+        set(outFile ${ARGV2})
     endif()
 
     # Custom Command to generate romdisk object file from image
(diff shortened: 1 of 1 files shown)


utils/cmake/dreamcast.cmake
```diff
@@ -34,8 +34,19 @@ endif()
 
 ### Adds a custom "run" target which uses $KOS_LOADER to run the given target. ###
 function(kos_run_target target)
-    add_custom_target(run 
-        COMMAND $ENV{KOS_LOADER} $<TARGET_FILE:${target}> 
+    # targetName is optional and defaults to "run" when not provided.
+    if(NOT ${ARGC} EQUAL 2)
+        set(runTarget run)
+    else()
+        set(runTarget ${ARGN})
```
pcercuei: Gonna be pedantic here, but this should probably be `${ARGV1}`.

gyrovorbis: How come? Also I only did it that way because the other optional arg-take functions in this file do the same thing.

pcercuei: `${ARGV1}` points to the second argument to the function, `${ARGN}` is a list of all the arguments passed to the function.

PR #1121, opened 2025-08-04
Move arch-specific ELF definitions from elf.c to arch.h

darcagn:
Arch-specific definitions and code should be kept to each arch's respective arch dir. There are two files violating this at the moment, `include/kos.h` and `kernel/fs/elf.c`. This PR addresses the latter, moving the `ARCH_ELFCLASS`, `ARCH_ELFDATA`, and `ARCH_CODE` definitions to `arch/arch.h` next to other ELF definitions.

Fixing `include/kos.h` will be for later, as some of the headers need to be reorganized across arch lines first.


QuzarDC (approved):
Good call.

PR #1124, opened 2025-08-05
thread: Correct missing commas from init/reaper attrs.

QuzarDC:
Additionally provide disable_tls for default to avoid warning.


gyrovorbis (approved):
WHOOPS. THANK YOu!!!

PR #1131, opened 2025-08-05
Further updates for `__pure` usage

QuzarDC:
Addressing some feedback on #992 .

- `mutex_is_locked` was missing the `__pure` attribute in mutex.h
- The `kos_get_banner` functions qualify to be the stricter `__pure2` which maps to the gcc function attribute [const](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-const-function-attribute)

### DIFF
--- a/include/kos/mutex.h
+++ b/include/kos/mutex.h
@@ -211,7 +211,7 @@ int mutex_lock_timed(mutex_t *m, int timeout) __nonnull_all;
     \retval 0               If the mutex is not currently locked
     \retval 1               If the mutex is currently locked
 */
-int mutex_is_locked(const mutex_t *m) __nonnull_all;
+int __pure mutex_is_locked(const mutex_t *m) __nonnull_all;
 
 /** \brief  Attempt to lock a mutex.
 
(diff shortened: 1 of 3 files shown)


include/kos/mutex.h
```diff
@@ -211,7 +211,7 @@ int mutex_lock_timed(mutex_t *m, int timeout) __nonnull_all;
     \retval 0               If the mutex is not currently locked
     \retval 1               If the mutex is currently locked
 */
-int mutex_is_locked(const mutex_t *m) __nonnull_all;
+int __pure mutex_is_locked(const mutex_t *m) __nonnull_all;
```
pcercuei, line 214: It's now sandwitched between two attributes, which looks a bit weird, but not enough for me to call a respin.

pcercuei (approved):

include/kos/mutex.h
QuzarDC: Yea, that was a bit weird to me as well, but it felt weirder to me if it was both piled on the end or having `__pure` on the end. Thankfully it's just a bit of style and not something that should need to be looked at frequently.

PR #1132, opened 2025-08-06
CI: Add whitespace check

pcercuei:
Fail if some commits in a given pull request have whitespace issues.

Only runs on pull requests.

Companion to #1128.

### DIFF
--- a/.github/workflows/whitespace.yml
+++ b/.github/workflows/whitespace.yml
@@ -1,0 +1,18 @@
+name: Whitespace check
+
+on: pull_request
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+
+    - name: Whitespace check
+      run: |
+        git fetch origin ${{ github.base_ref }}:${{ github.base_ref }}
+        git diff ${{ github.base_ref }} >/tmp/diff1.txt
+        git diff -w ${{ github.base_ref }} >/tmp/diff2.txt
+        diff /tmp/diff1.txt /tmp/diff2.txt >/dev/null


pcercuei:
@QuzarDC the CI checks (this one and #1026) are more about giving us information. We *can* block merging until all checks are cleared, but we can just as well choose to ignore the CI failing and merge a PR anyway.

pcercuei:
> Wouldn't this just fail for any whitespace change at all? Add a new function that needs to have a newline after it to space it from the next one, and it will have a difference between `diff` and `diff -w`.

This is not the case. I verified it with this change:
```diff
diff --git a/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c b/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
index 871052e2c..0971585b4 100644
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
@@ -214,6 +214,9 @@ void *pvr_sq_set16(void *dest, uint32_t c, size_t n, pvr_dma_type_t type) {
     return dest;
 }
 
+void test_stuff(void) {
+}
+
 /* Fills n bytes at PVR dest with 32-bit c, dest must be 32-byte aligned */
 void *pvr_sq_set32(void *dest, uint32_t c, size_t n, pvr_dma_type_t type) {
     void *dma_area_ptr;
```

QuzarDC (approved):
This should be a good manageable first step into the realm of CI.

QuzarDC:
> > Wouldn't this just fail for any whitespace change at all? Add a new function that needs to have a newline after it to space it from the next one, and it will have a difference between `diff` and `diff -w`.
> 
> This is not the case. I verified it with this change:
> 
> ```diff
> diff --git a/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c b/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
> index 871052e2c..0971585b4 100644
> --- a/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
> +++ b/kernel/arch/dreamcast/hardware/pvr/pvr_dma.c
> @@ -214,6 +214,9 @@ void *pvr_sq_set16(void *dest, uint32_t c, size_t n, pvr_dma_type_t type) {
>      return dest;
>  }
>  
> +void test_stuff(void) {
> +}
> +
>  /* Fills n bytes at PVR dest with 32-bit c, dest must be 32-byte aligned */
>  void *pvr_sq_set32(void *dest, uint32_t c, size_t n, pvr_dma_type_t type) {
>      void *dma_area_ptr;
> ```

Got it. Right, tested some and it's clever enough to ignore whitespace in added rather than modified lines.

PR #1139, opened 2025-08-08
Revert to using patches instead of weak symbols to build libobjc

darcagn:
During #1097, we removed one of the GCC patches to libobjc that was necessary to prevent failing tests due to unresolved symbols, and instead replaced the unresolved symbols with weak symbols to allow compilation.

Unfortunately, toolchains built in this manner now seem to output broken binaries, so we revert this change.


darcagn:
With this PR, the toolchain still has issues unfortunately, so I'm closing until a better solution is found.

PR #1143, opened 2025-08-12
`fs_romdisk` and `exec` example fixup

QuzarDC:
Bundling these together as the exec example breaking revealed the issue with `fs_romdisk_shutdown`. After the romdisk commit, the exec example should work again with its older fragile `fs_mmap` to help validate the issue before then adjusting the example to `fs_close` the subbin.

### DIFF
--- a/examples/dreamcast/basic/exec/exec.c
+++ b/examples/dreamcast/basic/exec/exec.c
@@ -9,20 +9,38 @@
 
 int main(int argc, char **argv) {
     file_t f;
-    void *subelf;
+    size_t s;
+    ssize_t rv;
+    void *subbin;
 
     /* Print a hello */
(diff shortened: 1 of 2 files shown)


examples/dreamcast/basic/exec/exec.c
```diff
@@ -9,20 +9,39 @@
 
 int main(int argc, char **argv) {
     file_t f;
-    void *subelf;
+    size_t s;
+    ssize_t rv;
+    void *subbin;
 
+    dbglog_set_level(999);
```
pcercuei: Is that leftover debug stuff?

QuzarDC: Absolutely was. Removed it now.

pcercuei (approved):

PR #1052, opened 2025-05-26
Extended but backwards compatible purupuru_effect_t definition

dfchil:
- Changed purupuru_effect_t to a union of a uint32_t .raw, the previous 4 uint8_t fields in an unnamed struct and a struct with more specific bitfields.
- Moved documentation on old macros to the corresponding new bitfields
- Changed purupuru_rumble to take advantage of this simpler approach
- Updated rumble example to use the expanded purupuru_effect_t type instead of defining a similar bitfield type in the example code

Issue #941, opened 2025-02-16
KallistiOS's purupuru.c/h driver causes maple bus & controller to stall with official rumble paks.


StrikerTheHedgefox:
Seems that when using KallistiOS's purupuru driver, the controller will randomly stall (causing no input to make it through for a frame or two) or disconnect entirely, and the VMU will also randomly drop connection (causing it to beep as if it was just re-connected), even with the simplest rumble codes, when using the official jump/purupuru pack.

So far, I've tested the following:

- StrikerDC Wireless - Works, no stall.
- Naki Rocker Pack - Works, no stall.
- PuruPuru/Jump Pack - BROKEN.

I only have one controller connected, too.

Apparently, someone was able to work around this by using their own standalone maple code that worked, but using KOS is broken. https://dcemulation.org/phpBB/viewtopic.php?p=528968#p528968
```
I'm going to assume it's something to do with the method KOS uses to
queue up and xmit data to the maple bus. I went back to my stand alone maple
code, and I'm not having the lockup problem, also almost all the testing I've
done was with a performance rumble... you could always try what I did in the end.
run your own maple code to access the rumbles. set up your data and wait until
the DMA isn't busy.. and if I can recall you're not in a Vsync, then start the transfer. If the device doesn't respondind at all reset the bus, and try again.
You could use the code I originally posted as a framework.. there's a mistake
in the section to calc the DST var.. I don't remeber exactly what it was but that
i reversed something, and it won't calc the dst properly beyond the 2nd port.
```

The code in Doom64-DC that I've been working with is pretty simple:
```c
void *I_RumbleThread(void *param) {
	uint32_t packet = (uint32_t)param;
	maple_device_t *purudev = NULL;
	purudev = maple_enum_type(0, MAPLE_FUNC_PURUPURU);
	if (purudev) {
			purupuru_rumble_raw(purudev, packet);
	}
	return (void*)0;
}
```
This only happens once a tick at 30hz, so it shouldn't be overloading things, and going by the behaviour of the two working packs, I highly doubt that's the issue... and again, only one controller connected, which a rumble pak and a vmu. Removing the VMU doesn't fix it either.

So, right now, I'm at a loss.

QuzarDC:
Based on the description by SinisterTengu this may be another expression or variation of this same issue: https://github.com/KallistiOS/KallistiOS/issues/806 . Unfortunately his original code is no longer available.

[...]

PR #1146, opened 2025-08-14
`purupuru`: Refactor effects, fix first-party device support

### LINKED ISSUE #1064: rumble calls disrupts maple controller readings
how to reproduce: 

edit the file: examples/dreamcast/rumble/rumble.c and uncomment the two calls to `plx_fcxt_setsize` (calling this causes the text rendering to not show). 

Then compile and run the rumble.elf program. 

If  you then rapidly press `B` button, which sends a 0x00000000 signal to the purupuru on each press, the screen starts flickering with "Please attach a controller!" or "Please attach a rumbler!"
It does the same thing when using the `A` button to send the currently configured rumble signal.

My assumption is that sending purupuru commands make KOS think that the devices have been removed for a number of frames. 

Note: I've only tried this with original purupuru-packs, I don't have other makes to test.

QuzarDC:
Big fixup to rumble support:
- Added error checking in the maple command irq handler. These were essential to diagnosing the illegal commands that we were sending to official rumble packs. If they trip it likely indicates something that we should be doing differently.
- Cleaned up rumble driver to use stdint types.
- Added new struct members based on info provided by @megavolt85 [here](https://github.com/KallistiOS/KallistiOS/pull/1052#issuecomment-2913983412) I've updated the documentation to include what I've found of error conditions.
- Deprecate old struct members.
- Update `purupuru_rumble` to check for values that would trigger a hardware error on official rumble packs. This is what corrects use with official packs. Previously we were using just a flat `0x00000000` to stop rumbling. On third party rumbles, they just didn't care, but on first-party the motor field triggers an error if set to 0 (and we don't recover from those errors well in the root maple driver).
- Update rumble example to reflect new findings.

This should close #1064 and #941.

From my testing in this PR there are a few big sets of future improvements that should be undertaken as followups (but shouldn't hold back this bugfixing):
- Better handle a maple device throwing an error. Currently we just don't do anything for it. But it seems that it may be a good use case for the same reset command I've implemented in #772 . Something like 'if a device repeatedly errors, send a reset command to it'. (I've opened #1147 to try to block similar errors with VMS devices)
- I found that the rumble pack appears to allow various other commands like `MAPLE_RESPONSE_DATATRF` and `MAPLE_COMMAND_BWRITE` which likely have to do with the 'Auto-Stop Timeout' or other patterns. That it can take larger sets of data would indicate it's possible to upload some sort of ongoing pattern or sequence to the device rather than repeatedly sending individual effects to build a sequence.
- Update the rumble example to display and allow setting of the individual struct fields rather than the raw bitfield.
- The rumble example has a bug where it will periodically drop all the text under the hex values and stop updating the screen. Usually happens on device disconnect but also seemingly randomly on sending a stop. After this point the example continues to function and is intractable a bit via the printf output but appears to hang visually.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/maple/purupuru.h
+++ b/kernel/arch/dreamcast/include/dc/maple/purupuru.h
@@ -36,7 +36,7 @@
 #include <kos/cdefs.h>
 __BEGIN_DECLS
 
-#include <arch/types.h>
+#include <stdint.h>
 #include <dc/maple.h>
 
 /** \defgroup peripherals_rumble    Rumble Pack
@@ -54,16 +54,16 @@ __BEGIN_DECLS
 */
 typedef struct purupuru_effect  {
     /** \brief  The duration of the effect. No idea on units... */
-    uint8 duration;
+    uint8_t duration;
 
     /** \brief  2nd effect field. */
-    uint8 effect2;
+    uint8_t effect2;
 
     /** \brief  1st effect field. */
-    uint8 effect1;
+    uint8_t effect1;
 
     /** \brief  Special effects field. */
-    uint8 special;
+    uint8_t special;
 } purupuru_effect_t;
 
 /* Set one of each of the following in the effect2 field of the
@@ -181,7 +181,7 @@ int purupuru_rumble(maple_device_t *dev, purupuru_effect_t *effect);
     \retval MAPLE_EOK       On success.
     \retval MAPLE_EAGAIN    If the command couldn't be sent. Try again later.
 */
-int purupuru_rumble_raw(maple_device_t *dev, uint32 effect);
+int purupuru_rumble_raw(maple_device_t *dev, uint32_t effect);
 
 /* \cond */
 /* Init / Shutdown */
--- a/kernel/arch/dreamcast/include/dc/maple/purupuru.h
+++ b/kernel/arch/dreamcast/include/dc/maple/purupuru.h
@@ -52,18 +52,23 @@ __BEGIN_DECLS
     This, along with the various macros in this file can give a slightly better
     idea of the effect being generated than using the raw values.
 */
-typedef struct purupuru_effect  {
-    /** \brief  The duration of the effect. No idea on units... */
-    uint8_t duration;
+typedef union purupuru_effect  {
+    /** \brief Access the raw 32-bit value to be sent to the puru */
+    uint32_t raw;
+    /** \brief Deprecated old structure which has been inverted now to union with raw. */
+    struct {
+        /** \brief  Special effects field. */
+        uint8_t special;
 
-    /** \brief  2nd effect field. */
-    uint8_t effect2;
+        /** \brief  1st effect field. */
+        uint8_t effect1;
 
-    /** \brief  1st effect field. */
-    uint8_t effect1;
+        /** \brief  2nd effect field. */
+        uint8_t effect2;
 
-    /** \brief  Special effects field. */
-    uint8_t special;
+        /** \brief  The duration of the effect. No idea on units... */
+        uint8_t duration;
+    };
 } purupuru_effect_t;
 
 /* Set one of each of the following in the effect2 field of the
--- a/kernel/arch/dreamcast/include/dc/maple/purupuru.h
+++ b/kernel/arch/dreamcast/include/dc/maple/purupuru.h
@@ -52,25 +52,19 @@ __BEGIN_DECLS
 /** \brief  Effect generation structure.
 
     This structure is used for convenience to send an effect to the jump pack.
-    This, along with the various macros in this file can give a slightly better
-    idea of the effect being generated than using the raw values.
+    The members in the structure note general explanations of their use as well
+    as some limitations and suggestions. There shouldn't be a need to use the
+    raw accessor with the new fully specified members.
 */
 typedef union purupuru_effect  {
     /** \brief Access the raw 32-bit value to be sent to the puru */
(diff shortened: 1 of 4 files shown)


QuzarDC:
This fails whitespace CI due to the (intentional) changes to the rumble example printouts of the effect data:

```
quzar@Centurion-IV:~/dcdev/KallistiOS$ git diff HEAD~1 > diff.txt
quzar@Centurion-IV:~/dcdev/KallistiOS$ git diff -w HEAD~1 >diffw.txt
quzar@Centurion-IV:~/dcdev/KallistiOS$ diff diff.txt diffw.txt
122,125d121
< -
< -  printf("  .fx1_pulse       =  %u,\n", fields.fx1_pulse);
< -  printf("  .fx1_powersave   =  %u,\n", fields.fx1_powersave);
< -  printf("  .fx1_intensity   =  %u,\n", fields.fx1_intensity);
129,132c125,127
< -  printf("  .fx2_lintensity  =  %u,\n", fields.fx2_lintensity);
< -  printf("  .fx2_pulse       =  %u,\n", fields.fx2_pulse);
< -  printf("  .fx2_uintensity  =  %u,\n", fields.fx2_uintensity);
< -  printf("  .fx2_decay       =  %u,\n", fields.fx2_decay);
---
> -  printf("  .fx1_pulse       =  %u,\n", fields.fx1_pulse);
> -  printf("  .fx1_powersave   =  %u,\n", fields.fx1_powersave);
> -  printf("  .fx1_intensity   =  %u,\n", fields.fx1_intensity);
137a133,137
> -  printf("  .fx2_lintensity  =  %u,\n", fields.fx2_lintensity);
> -  printf("  .fx2_pulse       =  %u,\n", fields.fx2_pulse);
> -  printf("  .fx2_uintensity  =  %u,\n", fields.fx2_uintensity);
> -  printf("  .fx2_decay       =  %u,\n", fields.fx2_decay);
> -
quzar@Centurion-IV:~/dcdev/KallistiOS$

**Review by dfchil — CHANGES_REQUESTED** [review]
Disregard this. I was too hasty.

File: kernel/arch/dreamcast/include/dc/maple/purupuru.h [hunk 1]
```diff
@@ -49,114 +52,56 @@ __BEGIN_DECLS
...
+        uint8_t fpow    : 3;
+        /** \brief Convergent vibration. The rumble will get weaker until it stops. */
+        bool    conv    : 1;
+
+        /** \brief Vibration frequency. for most purupuru 4-59. */
+        uint8_t freq;
+        /** \brief Vibration inclination period. */
+        uint8_t inc;
+    };
 } purupuru_effect_t;
```
dfchil: Could you change this to this: 

```c
typedef union purupuru_effect  {
    /** \brief Access the raw 32-bit value to be sent to the puru */
    uint32_t raw;
    /** \brief Deprecated old structure which has been inverted now to union with raw. */
    struct {
        uint8_t special     __depr("Please see purupuru_effect_t which has new members.");
        uint8_t effect1     __depr("Please see purupuru_effect_t which has new members.");
        uint8_t effect2     __depr("Please see purupuru_effect_t which has new members.");
        uint8_t duration    __depr("Please see purupuru_effect_t which has new members.");
    };
    struct {
        /** \brief Continuous Vibration. When set vibration will continue until stopped */
        bool    cont    : 1;
        /** \brief Reserved. Always 0s */
        uint32_t res    : 3;
        /** \brief Motor number. 0 will cause an error. 1 is the typical setting. 4-bits. */
        uint32_t motor  : 4;
        /** \brief Backward direction (- direction) intensity setting bits. 0 stops vibration. 3-bits. */
        uint32_t bpow   : 3;
        /** \brief Divergent vibration. The rumble will get stronger until it stops. */
        bool    div     : 1;
        /** \brief Forward direction (+ direction) intensity setting bits. 0 stops vibration. 3-bits. */
        uint32_t fpow   : 3;
        /** \brief Convergent vibration. The rumble will get weaker until it stops. */
        bool    conv    : 1;
        /** \brief Vibration frequency. For most purupuru the range is 4-59. */
        uint8_t freq;
        /** \brief Vibration inclination period. */
        uint8_t inc;
    };
} purupuru_effect_t;
```

dfchil: The primary change is to use uint32_t where the bit-size doesn't fit any well known type. Originally I thought that they should all be changed to uint32_t, but I realized that having types that fit the bit-size is a nice hint to the user.

dfchil (changes requested):
I've done a complete rewrite of the rumble example based on this PR, as you can see here: [git@github.com:dfchil/KallistiOS.git](https://github.com/dfchil/KallistiOS/tree/First_rumble_party), I want to make a PR for that once this PR goes through. 

My one request for change (see my comment) is more of a suggestion, if you decline it, I'll happily accept this PR, it is up to you.

kernel/arch/dreamcast/include/dc/maple/purupuru.h
QuzarDC: Yup, makes sense. I was doing parts of `uint8_t`s because I was treating it as four `uint8_t`s rather than the 'raw' value.

I took the opportunity to also add a static assert to validate the size of the struct in order to protect from the chance that this would trigger padding somewhere.

I did leave the spacing separating the two 8-bit chunks in order to help readability of comparing them to the older versions.

dfchil (approved):
Good work. I look forwards to building on this =)

PR #1150, opened 2025-08-18
Make byteorder.h include file cross-platform

pcercuei:
We don't need one `<arch/byteorder.h>` per supported platform.
Factorize code by moving `<arch/byteorder.h>` to `<kos/byteorder.h>`.


darcagn:
> What do you mean by they "actually kinda still are"?

I just mean that they have different behaviors depending on the arch (because different archs have different endiannesses).
But I do think I would qualify them as "not arch-specific abstractions" so I think they should be renamed, but not sure to what. Thoughts @QuzarDC ?

QuzarDC:
> > What do you mean by they "actually kinda still are"?
> 
> I just mean that they have different behaviors depending on the arch (because different archs have different endiannesses). But I do think I would qualify them as "not arch-specific abstractions" so I think they should be renamed, but not sure to what. Thoughts @QuzarDC ?

So I think that *generally* it would be appropriate to have such things be `kos_` even if it's different backing behavior for different archs when it's not specific to the arch. In this case if we added a dozen new archs there'd still be only the two different options here, and gcc provides the mechanism to determine which to use. In this specific case though, the arch functions only really exist to provide backing to the prefixless functions in `koslib/byteorder.c` so to me, it would make most sense to just percolate the changes up into that file (leaving behind a temporary stub for the arch/btyeorder in case anyone was trying to use those directly).

That seems, overall, like a good approach to help reduce the size of the arch API where possible. I don't really see any value added by maintaining the extra layer of abstraction.

pcercuei:
Note that having them static inline in a header is probably a better idea.
Then the compiler can optimize it better. For instance, `kos_htonl(*(u32 *)ptr);` would be one single LWBRX instruction on PowerPC, while `htonl(*(u32 *)ptr)` would be a function call.

QuzarDC:
Yea, as far as I know that shouldn't cause any sort of problem with POSIX compliance so they can just get merged all the way up into `arpa/inet.h`

> Note that having them static inline in a header is probably a better idea. Then the compiler can optimize it better. For instance, `kos_htonl(*(u32 *)ptr);` would be one single LWBRX instruction on PowerPC, while `htonl(*(u32 *)ptr)` would be a function call.

pcercuei:
@QuzarDC so what should I do?
Rename `arch_swap16` / `arch_swap32` to `kos_swap16` / `kos_swap32`, remove the `arch_ntohl` / `arch_htonl` and instead inline them inside <arpa/inet.h>?

PR #1151, opened 2025-08-18
treewide: Update all files to include <kos/cdefs.h>

pcercuei:
Plenty of files would use macros defined within `<kos/cdefs.h>`, but would
include `<sys/cdefs.h>`. Those would only compile because they would also
include `<arch/types.h>`, includes `<arch/_types.h>` that itself includes
`<kos/cdefs.h>`.

This gets nastier, when taking into consideration that the
`<arch/_types.h>` include is only done if the BYTE_ORDER macro is unset.

This is an automated search-and-replace with:
`find include/ kernel/ -name '*.h' -exec sed -i 's/<sys\/cdefs\.h>/<kos\/cdefs.h>/' {} \;`

### DIFF
--- a/include/arpa/inet.h
+++ b/include/arpa/inet.h
@@ -18,7 +18,7 @@
 #ifndef __ARPA_INET_H
 #define __ARPA_INET_H
 
-#include <sys/cdefs.h>
+#include <kos/cdefs.h>
 
 __BEGIN_DECLS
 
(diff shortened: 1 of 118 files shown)


QuzarDC (approved):
See, I had been working to address this but entirely in the other direction, trying to shrink down the kos/cdefs. This makes more sense though, we should presume that the macros there are available and can better prevent stuff from leaking out (like #840) by percolating the includes up into kos-specific files.

PR #1153, opened 2025-08-19
Move banner, license, authors, uname functions from arch to arch-agnostic

darcagn:
More general cleaning up of the "dumping ground" `arch.h` header.

- Generate KOS_GIT_REVISION env var in environ_base.sh instead of in make_banner.sh
- Create kos/banner.h header for banner, license, author functions
- Move banner/authors generation scripts from arch/dreamcast/kernel/ to kernel/
- Add arch/subarch description to the banner
- Add ARCH_NAME define, representing an arch's full name, to arch.h
- Move arch/dreamcast/kernel/uname.c to kernel/libc/koslib/


environ_base.sh
```diff
@@ -6,6 +6,11 @@ export KOS_VERSION_MINOR=`awk '$2 == "KOS_VERSION_MINOR"{print $3; exit}' ${KOS_
 export KOS_VERSION_PATCH=`awk '$2 == "KOS_VERSION_PATCH"{print $3; exit}' ${KOS_BASE}/include/kos/version.h`
 export KOS_VERSION="${KOS_VERSION_MAJOR}.${KOS_VERSION_MINOR}.${KOS_VERSION_PATCH}"
 
+# Grab the git revision, if we're using git.
+if [ -d "$KOS_BASE/.git" ]; then
+  export KOS_GIT_REVISION=`git describe --dirty --always`
```
pcercuei: I'm not sure that works, the current directory may be different than $KOS_BASE.

darcagn: You're right. It would error if cwd is anywhere outside of the KOS tree.
Fixed by specifying `-C ${KOS_BASE}`.

kernel/arch/dreamcast/include/arch/arch.h
```diff
@@ -300,47 +303,6 @@ void hardware_shutdown(void);
 */
 int hardware_sys_mode(int *region);
 
-/* These three aught to be in their own header file at some point, but for now,
```
QuzarDC, line 303: Think deprecated versions of these prototypes can be left in so if they are included from here there's a message? Can drop the docs of course, but this is something we should make sure to not accidentally make it harder to use. It's the exact thing that someone would just comment out if they got an error.

QuzarDC (changes requested):
Just one small change in order to try to help with compatibility.

darcagn:
I re-added the function declarations to the bottom of arch.h in a section explaining that they're deprecated. I don't see how I can make "deprecated versions of these prototypes" though because if both arch.h and banner.h are included in a source file then it would throw a warning.

I do plan on gutting arch.h heavily so this deprecated section is going to grow quite large.

QuzarDC:
> I re-added the function declarations to the bottom of arch.h in a section explaining that they're deprecated. I don't see how I can make "deprecated versions of these prototypes" though because if both arch.h and banner.h are included in a source file then it would throw a warning.

So, I was figuring like this in order to give the warning if only arch/arch was included.
```
#ifndef __KOS_BANNER_H
const char * __pure2 kos_get_banner(void) __depr("Moved to <kos/banner.h>");
const char * __pure2 kos_get_license(void) __depr("Moved to <kos/banner.h>");
const char * __pure2 kos_get_authors(void) __depr("Moved to <kos/banner.h>");
#endif
```

But that's cumbersome too, and only works well to prevent our internal warning. I'll just add as-is and see about putting in something more comprehensive to address the shift to just `<kos.h>` and the internalizing of the `<arch/>` headers en-masse as discussed in the discord.

QuzarDC (approved):

PR #1157, opened 2025-08-20
sys/lock.h: Roll back change to use kos cdefs.

QuzarDC:
The kos cdefs file isn't available (and shouldn't be exposed) to the toolchain building process that uses this header.


pcercuei:
I think it does not go far enough; it looks like all includes from `include/sys/` are copied to the toolchain, so none of them should include `<kos/cdefs.h>`.

I have a RFC fix for this issue in #1158.

PR #1158, opened 2025-08-20
Drop KOS header includes from toolchain files

pcercuei:
Hijacking #1015.

This PR drops all KOS-specific includes from the toolchain files.

This includes all files in `include/sys/` which are copied to the toolchain.

It also updates the `gthr-kos.h`, which now redefines locally all the types and symbols it needs.
To make sure that it stays in sync with the KOS headers, a verification mechanism is in place, automatically enabled for GCC >= 15 (because of the use of C23).

### DIFF
--- a/include/kos/thread.h
+++ b/include/kos/thread.h
@@ -41,6 +41,7 @@ __BEGIN_DECLS
 #include <kos/cdefs.h>
 #include <kos/tls.h>
 #include <arch/irq.h>
+#include <arch/types.h>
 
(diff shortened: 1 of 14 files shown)


pcercuei:
Force pushed, the rule to compile-check the `gthr-kos.h` was moved after the inclusion of Makefile.prefab as it was previously detected as the default rule.

Added C++ guards to `gthr-kos.h` since we now define symbols instead of including C++-guarded headers.

PR #1162, opened 2025-08-24
dc-chain: Export PATH variable

pcercuei:
If we don't export it, the programs called inside the $(shell ) function won't have the updated $PATH.

This could cause the fake-kos.o to not be removed from the libc.a library which in turns caused all kinds of weird issues.


pcercuei:
Closing this PR as it was reported as non-working by @Tchan0 (but it worked for me!)

PR #1165, opened 2025-08-25
Backport Rollup for v2.2.x

QuzarDC:
As with #1156, getting these commits from PRs flagged as v2.2.1 (#1098 #1099 #1100 and #1101) into the v2.2.x stable branch.


QuzarDC:
Whoops. Accidentally set this to merge into master rather than v2.2.x. Couldn't see a way to change that part after the fact so opened #1166

PR #1166, opened 2025-08-25
 Backport Rollup for v2.2.x

QuzarDC:
As with #1156, getting these commits from PRs flagged as v2.2.1 (#1098 #1099 #1100 and #1101) into the v2.2.x stable branch.


darcagn:
#1090 would be a good idea too. When I use 2.2.x I get long stalls/errors building the toolchain.

QuzarDC:
> #1090 would be a good idea too. When I use 2.2.x I get long stalls/errors building the toolchain.

Done. I missed it since I forgot how bad that was.

darcagn (approved):

Issue #1155, opened 2025-08-19
Task: `arch/dreamcast` headers cleanup


darcagn:
We need to put in the work to clear up the organizational lines between `kos`, `arch`, and `dc` headers.

As a reminder:
- `#include <kos/*>` headers are for arch-agnostic, general KallistiOS functionality.
- `#include <arch/*>` headers are for arch-specific abstractions to general KallistiOS functionality.
- `#include <dc/*>`, `#include <gc/*>`, etc. headers are for arch-specific unique functionality.

The `<kos/*>` headers and kernel-level source code are pretty much clean now, due to efforts by several of us in the past few months. 

`<arch/*>` is a mess though. Some files need to be outright moved to `<dc/*>`, some files need to be made into more generic APIs, and some need to be split up (some parts generic and some parts moved to `<dc/*>`.

 I'll list here what I feel needs to be done... this is opinionated and not an authoritative list though:
- `arch.h` currently contains
  1. architecture definitions and functions (should stay)
  2. memory management definitions (should probably move to e.g. a new `arch/mm.h`. Functions moved to `kos/mm.h` in #1286 )
  3. banner definitions (moved to `<kos/*>` in #1153)
  4. video/serial definitions (moved to `<dc/*>` in #1154)
  5. hardware definitions (DC region/revision defines, DC-specific memory defines, DC-specific hardware init functions) should probably move to `<dc/*>`

- [ ] `args.h` is DC-specific (move to `<dc/*>`
- [X] `byteorder.h` (moved to `<kos/*>` in #1150)
- [ ] `cache.h` needs moving to `dc/` but some of the defines can probably stay in `arch/`.
- [ ] `dmac.h` is Dreamcast-specific, should be moved to `<dc/*>`.
- [X] `exec.h` looks OK.
- [ ] `gdb.h` has nothing specific to the console and should probably be moved to `<kos/*>`.
- [X] `init_flags.h` looks OK.
- [X] `irq.h` could use some work, perhaps made more generic with some stuff moved into `<dc/*>`? (At least partially resolved with #1231 )
- [X] `kos.h` looks OK.
- [x] `memory.h` (to be moved to `<dc/*>` in #1265).
- [ ] `mmu.h` is difficult for me to call; I don't know if this can work generically.
- [X] `rtc.h` looks OK.
- [ ] `spinlock.h` looks OK, maybe some of it could be moved to `<kos/*>`.
- [X] `stack.h` looks OK.
[...]

PR #1168, opened 2025-08-26
Move Dreamcast wdt driver from arch/ to dc/

darcagn:
- Move the code from dreamcast/kernel/wdt.c to dreamcast/hardware/wdt.c with other DC-specific hardware drivers
- Move arch/wdt.h to dc/wdt.h, add replacement arch/wdt.h header indicating deprecation
- Update all references to the header
- Update examples to use the new header


kernel/arch/dreamcast/include/arch/wdt.h
```diff
@@ -1,197 +1,9 @@
...
-    Checks to see whether the WDT has been enabled.
-
-    \return     1 if enabled, 0 if disabled
-*/
-int wdt_is_enabled(void);
-
-__END_DECLS
-
-#endif  /* __ARCH_WDT_H */
+#pragma GCC warning "The `<arch/wdt.h>` header has been moved to `<dc/wdt.h>`."
```
QuzarDC: `#warning` is what we've used elsewhere before. Only one currently in the codebase https://github.com/KallistiOS/KallistiOS/blob/7bd48056e76e750d1c90e0cc4aeb156f46da31f8/include/kos/cdefs.h#L31 so if there's a reason to switch to this way, that should be fine. I don't really have an opinion other than all the same.

QuzarDC (commented):
Just one note, otherwise looks fine, and as was discussed in the discord and aligned with #1155 .

kernel/arch/dreamcast/include/arch/wdt.h
darcagn: Nah, it should all be the same, good catch.

QuzarDC (approved):

PR #1169, opened 2025-08-27
Move spinlocks public API from arch/spinlock.h to kos/spinlock.h

darcagn:
Another in the discussion to revamp and de-public the "arch" headers.

Users who `#include <arch/spinlock.h>` will still have it work (for now), but it will warn that the header has moved to `<kos/spinlock.h>`.

Note that because of circular dependencies `kos/spinlock.h` can't include `arch/spinlock.h` at the very top. This will have to be the case with almost all of these unfortunately.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/spinlock.h
+++ b/kernel/arch/dreamcast/include/arch/spinlock.h
@@ -1,40 +1,10 @@
 /* KallistiOS ##version##
 
-   arch/dreamcast/include/arch/spinlock.h
-   Copyright (C) 2001 Megan Potter
+   kernel/arch/dreamcast/include/arch/spinlock.h
 
 */
 
-#ifndef __KOS_SPINLOCK_H
-#pragma GCC warning "The `<arch/spinlock.h>` header has been moved to `<kos/spinlock.h>`."
 #include <kos/spinlock.h>
-#endif
 
-#ifndef __ARCH_SPINLOCK_H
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/include/arch/spinlock.h
```diff
@@ -1,72 +1,31 @@
...
-
-/** \brief  Spinlock data type. */
-typedef volatile int spinlock_t;
-
-/** \brief  Spinlock initializer.
-
-    All created spinlocks should be initialized with this initializer so that
-    they are in a sane state, ready to be used.
+/* Use a test-and-set to attempt to acquire a lock atomically.
+   In one instruction, tas.b writes 0xFF to the spinlock and
```
pcercuei: According to the doc, it does not write 0xff but 0x80.

darcagn: You're correct. Fixed.

darcagn:
Rebased this against latest KOS changes. Not sure what the ultimate fate of spinlocks will be, but if even if it is going to deprecated and eventually removed, it should also be moved out of `arch` for now anyway.

darcagn:
I've adjusted to a new arch-agnostic atomic implementation for `spinlock_trylock` and stripped the `arch/spinlock.h` header down to barebones now.

This implementation passes `examples/dreamcast/basic/threading/spinlock_test` with thread scheduler ramped up to 1000Hz.

PR #1170, opened 2025-08-27
Add -mtas to environ_dreamcast KOS_CFLAGS.

gyrovorbis:
At the moment, the atomic_flag type in C11 is emitting suboptimal code for atomic_flag_test_and_set(), the fundamental operation used to create spinlocks from C's atomics.

Rather than emitting a single tas.b instruction to do the operation atomically in HW, it's emitting byte-wise read-modify-write code with the IRQ disable/reenable preamble/postamble from the imask-soft atomics model.

Turns out that Oleg Endo implemented support for backing atomic_flag by the tas.b instruction, as we'd desire, which can be used in combination with any atomics model type, by simply using the -mtas flag.

This PR simply enables that flag by default, as it's both correct and far more optimal to do for our atomics.

This will also now allow us to implement spinlocks with atomic_flag allowing us to get spinlocks for both SH and PPC that are optimal AND backed by purely standard C code.

### DIFF
--- a/environ_dreamcast.sh
+++ b/environ_dreamcast.sh
@@ -24,7 +24,7 @@ if [ -z "${KOS_SH4_PRECISION}" ] || [ "${KOS_SH4_PRECISION}" = "-m4-single" ]; t
     fi
 fi
 
-export KOS_CFLAGS="${KOS_CFLAGS} ${KOS_SH4_PRECISION} -ml -mfsrra -mfsca -ffunction-sections -fdata-sections -matomic-model=soft-imask -ftls-model=local-exec"
+export KOS_CFLAGS="${KOS_CFLAGS} ${KOS_SH4_PRECISION} -ml -mfsrra -mfsca -ffunction-sections -fdata-sections -mtas -matomic-model=soft-imask -ftls-model=local-exec"
 export KOS_AFLAGS="${KOS_AFLAGS} -little"
 export KOS_LDFLAGS="${KOS_LDFLAGS} ${KOS_SH4_PRECISION} -ml -Wl,--gc-sections"
 export KOS_LD_SCRIPT="-T${KOS_BASE}/utils/ldscripts/shlelf.xc"


gyrovorbis:
Here's a demonstration in Compiler Explorer: https://godbolt.org/z/aE5MYs6Mo.

gyrovorbis:
> So based on the convo in discord it seems this needs to be perf tested to ensure it's a good idea, but just to add another note, it is available all the way back to our oldest gcc version, so at least compatibility shouldn't be a problem: https://gcc.gnu.org/onlinedocs/gcc-9.5.0/gcc/SH-Options.html#index-mtas

Yes, this is the case. Due to the uncertainty of performance impact of using `TAS.B` knowing it has the side-effect of causing its cache-line to be purged, combined with how small and easy these routines are to profile.... Yeah, better wait to benchmark with the perf counters. Will convert to draft.

pcercuei:
Using the following C code:
```c
#include <kos.h>

#include <dc/perf_monitor.h>
#include <stdatomic.h>

#define NB_ENTRIES 16384

static atomic_flag atomic_array_flag[NB_ENTRIES];

static void test_flag(void) {
	unsigned int i;

	perf_monitor();

	for (i = 0; i < NB_ENTRIES; i++)
		atomic_flag_test_and_set(&atomic_array_flag[i]);
}

int main(int argc, char **argv) {

	test_flag();

	perf_monitor_print(stdout);
	perf_monitor_exit();

	return 0;
}
```

With KOS master, compiling with `kos-cc -O2 test-tas.c -o test-tas.o`:
```
Performance monitors:
	test_flag L13: 1 calls
		1643520 ns (1643520.000000 ns/call)
		event 0: 4 (4.000000 event/call)
		event 1: 519 (519.000000 event/call)
```

Compiling with `kos-cc -O2 -mtas test-tas.c -o test-tas.o`:
```
Performance monitors:
	test_flag L13: 1 calls
		2480800 ns (2480800.000000 ns/call)
		event 0: 2 (2.000000 event/call)
		event 1: 2 (2.000000 event/call)
```

Therefore using `-mtas` generates code that's about 50% slower.

gyrovorbis:
Yeah, this is all true. Closing this PR. Good work.

PR #1171, opened 2025-08-28
Enhance mutex code

pcercuei:
- Use `assert()` to check that the mutex type is one of the supported types. This means the check can be disabled for release builds.
- Try to make critical sections as small as possible, by moving the recursive / errcheck tests outside the critical section.
- Trylock / unlock without disabling IRQs. The trylock() can use an atomic compare-and-swap instead, and the unlock() just doesn't need to disable them.
- Wake up all threads waiting on a lock instead of just one, to respect thread priorities.

Tested with Bloom's threaded compiler, which is a very heavy mutex user.

### DIFF
--- a/kernel/thread/mutex.c
+++ b/kernel/thread/mutex.c
@@ -203,8 +203,6 @@ int mutex_trylock(mutex_t *m) {
 static int __nonnull_all mutex_unlock_common(mutex_t *m, kthread_t *thd) {
     int wakeup = 0;
 
-    assert(m->type <= MUTEX_TYPE_RECURSIVE);
-
     irq_disable_scoped();
 
     switch(m->type) {
@@ -255,6 +253,8 @@ static int __nonnull_all mutex_unlock_common(mutex_t *m, kthread_t *thd) {
 int mutex_unlock(mutex_t *m) {
     kthread_t *thd = thd_current;
 
+    assert(m->type <= MUTEX_TYPE_RECURSIVE);
+
     /* If we're inside of an interrupt, use the special value for the thread
        from mutex_trylock(). */
     if(irq_inside_int())
@@ -264,6 +264,8 @@ int mutex_unlock(mutex_t *m) {
 }
 
 int mutex_unlock_as_thread(mutex_t *m, kthread_t *thd) {
+    assert(m->type < MUTEX_TYPE_RECURSIVE); /* Unsafe with recursive mutexes */
+
     /* Make sure we're in an IRQ handler */
     if(!irq_inside_int()) {
         errno = EACCES;
--- a/kernel/thread/mutex.c
+++ b/kernel/thread/mutex.c
@@ -7,6 +7,7 @@
 */
 
 #include <assert.h>
+#include <stdatomic.h>
 #include <stdlib.h>
 #include <errno.h>
 #include <limits.h>
@@ -91,7 +92,7 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
 
     irq_disable_scoped();
 
-    if(!m->count) {
+    if(!m->holder) {
         m->count = 1;
         m->holder = thd_current;
         rv = 0;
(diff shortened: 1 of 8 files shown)


pcercuei:
I initially switched it back to Draft as it was having issues when the atomic model was set to `soft-gusa` (but worked fine with `soft-imask` which we use right now). I debugged it to an issue in my gUSA commits, and can confirm that this PR work fine with the fixed `soft-gusa` as well, so I'm switching it back to ready for review.

kernel/thread/mutex.c
```diff
@@ -160,44 +152,39 @@ int mutex_trylock(mutex_t *m) {
...
-        case MUTEX_TYPE_OLDNORMAL:
-        case MUTEX_TYPE_ERRORCHECK:
-            if(m->count) {
-                errno = EDEADLK;
-                return -1;
-            }
-
-            m->count = 1;
-            break;
+    if(m->holder == thd) {
```
QuzarDC, line 159: I think this could race: Thread A is already holding a recursive mutex and is trying to lock it a second time. After this check that it holds the mutex, it gets scheduled away and Thread B starts running. Thread B is interrupted at some point by an irq handler that uses `mutex_unlock_as_thread` to unlock the mutex as Thread A. The mutex is now unlocked and Thread B can take it. When we switch back to Thread A, it continues along and increments the mutex that Thread B owns.

It's quite contrived, but there may be a simpler way to trip it or maybe I'm missing how this is avoided.

pcercuei: How about we forbid recursive mutexes with `mutex_unlock_as_thread()`?

QuzarDC: That should work. It's not meant to be for common public usage and the few usages we have are all for standard mutexes, so we could update the docs to reflect this limitation and have it assert if a recursive mutex is passed to it. 

Alongside that, the dox for the rest of the functions should also be updated to reflect the reduced possible error returns.

pcercuei: Ok, I'll update the doc and add an assert().

pcercuei:
@QuzarDC Updated with an assert() in `mutex_unlock_as_thread()` to forbid recursive mutexes, and with the doc updated.

kernel/thread/mutex.c
```diff
@@ -215,8 +215,8 @@ static int __nonnull_all mutex_unlock_common(mutex_t *m, kthread_t *thd) {
     if (thd != IRQ_THREAD)
         thd->prio = thd->real_prio;
 
-    /* If we need to wake up a thread, do so. */
-    genwait_wake_one(m);
+    /* If we need to wake up threads, do so. */
+    genwait_wake_all(m);
```
QuzarDC: Not to drag this out too long (I had just not gotten to the last commit of the PR while I was trying to find races), but wouldn't it be better to adjust `genwait` to have awareness of priority? Either during insertion of the objects into the tailq during `genwait_wait` or (probably worse) searching for the highest priority thread when waking? I'm not sure there's a case where we *wouldn't* want `genwait_wake` to be aware of the priority. And as far as I can tell nothing is relying on that specific behavior.

It should be as simple as this I think: https://github.com/KallistiOS/KallistiOS/commit/194da33a373d4c71eb069a82c9e7bc13ab4d1d39 

This wake_all methodology might also have unintentional priority inversion, especially now that `mutex_unlock` is interruptible. 
-Thread A is holding a mutex with B, C, and D waiting in genwait. Their priority is in alpha order.
-Thread A unlocks the mutex, removing B, C, and D from the wait queue. 
-Next schedule Thread B is switched to, acquires the mutex, and starts working.
-Next schedule Thread C being switched to, it retries locking, can't and genwaits.
-Next schedule Thread B is switched back to, finishes with the mutex and unlocks.
-Between setting `m->holder = NULL` and completing `genwait_wake_all`, we get a scheduler interrupt.
-Next schedule Thread D is switched to, because it hadn't yet had a chance to re-lock and Thread C is still in the wait queue.

Certainly, you're already getting more risky and frequent priority inversion with the current `wait_one`, so it's still 'better' at avoiding it, but this just adds to the consideration for me that it's possibly not the right way to go about solving the problem.

pcercuei: Yes, I think your solution is much better. I wasn't 100% happy with mine as it meant waking up threads just to have them go back to sleep right after.

I'll drop my "wake all threads" commit.

PR #1174, opened 2025-08-30
Switch to the gUSA atomic model

pcercuei:
The gUSA atomic model is what Linux uses on SH3 and SH4.

Instead of disabling IRQs around the critical section, when using gUSA GCC will emit code segments that are rewindable: if interrupted in the middle of a critical section, the IRQ handler will just rewind the PC to the beginning of the critical section.

See https://godbolt.org/z/8zbYro41c for a A/B difference between handling atomics by masking IRQs vs. handling atomics the gUSA way.

This requires explicit support for it in the IRQ handler, which this PR adds. The change itself is tiny, and should have a neglectible overhead (a few memory loads from hot cache lines, and two comparisons, we're talking maybe a dozen cycles total).

The effect, however, is not neglectible.

Using the following benchmark C code:
```c
#include <kos.h>

#include <dc/perf_monitor.h>
#include <stdatomic.h>

#define NB_ENTRIES 16384

static atomic_uint atomic_array_u32[NB_ENTRIES];
static atomic_flag atomic_array_flag[NB_ENTRIES];

static void test_u32(void) {
	unsigned int i, zero = 0;

	perf_monitor();

	for (i = 0; i < NB_ENTRIES; i++)
		atomic_compare_exchange_strong(&atomic_array_u32[i], &zero, 1);
}

static void test_flag(void) {
	unsigned int i;

	perf_monitor();

	for (i = 0; i < NB_ENTRIES; i++)
		atomic_flag_test_and_set(&atomic_array_flag[i]);
}

int main(void) {

	test_u32();
	test_flag();

	perf_monitor_print(stdout);
	perf_monitor_exit();

	return 0;
}
```

Compiling with KOS master and `kos-cc -O2 test.c -o test.elf`:
```
Performance monitors:
	test_flag L23: 1 calls
		1672880 ns (1672880.000000 ns/call)
		event 0: 26 (26.000000 event/call)
		event 1: 559 (559.000000 event/call)
	test_u32 L14: 1 calls
		2427520 ns (2427520.000000 ns/call)
		event 0: 92 (92.000000 event/call)
		event 1: 2193 (2193.000000 event/call)
```

Compiling this PR with the same build command (after reloading the environ.sh script):
```
Performance monitors:
	test_flag L23: 1 calls
		678240 ns (678240.000000 ns/call)
		event 0: 1 (1.000000 event/call)
		event 1: 518 (518.000000 event/call)
	test_u32 L14: 1 calls
		1193120 ns (1193120.000000 ns/call)
		event 0: 4 (4.000000 event/call)
		event 1: 2056 (2056.000000 event/call)
```

In this example, with gUSA the "test_flag" benchmark is about 40.5% of the original speed, so a ~59.5% performance improvement. The "test_u32" benchmark is about 49.1% of the original speed, so a ~50.9% performance improvement.

### DIFF
--- a/environ_dreamcast.sh
+++ b/environ_dreamcast.sh
@@ -24,7 +24,7 @@ if [ -z "${KOS_SH4_PRECISION}" ] || [ "${KOS_SH4_PRECISION}" = "-m4-single" ]; t
     fi
 fi
 
-export KOS_CFLAGS="${KOS_CFLAGS} ${KOS_SH4_PRECISION} -ml -mfsrra -mfsca -ffunction-sections -fdata-sections -matomic-model=soft-imask -ftls-model=local-exec"
+export KOS_CFLAGS="${KOS_CFLAGS} ${KOS_SH4_PRECISION} -ml -mfsrra -mfsca -ffunction-sections -fdata-sections -matomic-model=soft-gusa -ftls-model=local-exec"
 export KOS_AFLAGS="${KOS_AFLAGS} -little"
 export KOS_LDFLAGS="${KOS_LDFLAGS} ${KOS_SH4_PRECISION} -ml -Wl,--gc-sections"
 export KOS_LD_SCRIPT="-T${KOS_BASE}/utils/ldscripts/shlelf.xc"


gyrovorbis (approved):
This is a pretty big freaking change to the atomics model that I had originally chosen and had partially implemented SW support for in KOS... but @pcercuei has convinced me it is absolutely warranted.

This is the model that SH Linux uses, and for good reason. It's going to make overall system responsivity better, since we are not constantly masking IRQs out during atomic operations, preventing more important work from interrupting them... PLUS the actual code emitted to perform the atomic operations itself is faster too? So we gain while we gain? 

In addition to his perf profiling, I also saw him rerun my atomics validator example/test, which is pretty damn thorough, and it all passed. Also saw him testing with at least Bloom as a complex real-world scenario, so I'm fully convinced he has also verified correctness here.

I'm approving from my end. Had I known the perf was better and that the `gUSA` model could be implemented with a single conditional in the IRQ handler, I would've gone with that model over the original `imask-soft` model that I went with... but I was a newb and didn't understand that model and thought it took way more OS support to implement.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -216,6 +216,16 @@ void irq_handle_exception(int code) {
     uint32_t evt = 0;
     int handled = 0;
 
+    if(__is_defined(__SH_ATOMIC_MODEL_SOFT_GUSA__)
+       && __unlikely((int32_t)irq_srt_addr->r[15] >= -128
```
QuzarDC, line 220: Could the `-128` be de-magic'd some? I understand its role is to compensate for the DC's memory space, while `gUSA` is defined as setting r15 to just be negative, so it should be possible to have it reference that address to make that more apparent.

QuzarDC:
Sounds good, just had one note about seeing if it could be a bit clearer why it's deviating from the ref implementation.

Just so that it's closer to hand for reference, adding below the current gcc documentation of the model from `\gcc\config\sh\sync.md` in the gcc sources:
```
;; gUSA Software Atomics (-matomic-model=soft-gusa; SH3*, SH4* only)
;;
;; On single-core systems there can only be one execution context running
;; at a given point in time.  This allows the usage of rewindable atomic
;; sequences, which effectively emulate locked-load / conditional-store
;; operations.  This requires complementary support in the interrupt / 
;; exception handling code (e.g. kernel) and does not work safely on multi-
;; core configurations.
;;
;; When an execution context is interrupted while it is an atomic
;; sequence, the interrupted context's PC is rewound to the beginning of
;; the atomic sequence by the interrupt / exception handling code, before
;; transferring control to another execution context.  This is done by
;; something like...
;;
;;	if (interrupted_context_in_atomic_sequence
;;	    && interrupted_pc < atomic_exitpoint)
;;	  interrupted_pc = atomic_entrypoint;
;;
;; This method is also known as gUSA ("g" User Space Atomicity) and the
;; Linux kernel for SH3/SH4 implements support for such software atomic
;; sequences.  It can also be implemented in freestanding environments.
;;
;; For this the following atomic sequence ABI is used.
;;
;; r15 >= 0:	Execution context is not in an atomic sequence.
;;
;; r15  < 0:	Execution context is in an atomic sequence and r15
;;		holds the negative byte length of the atomic sequence.
;;		In this case the following applies:
;;
;;		r0:	PC of the first instruction after the atomic
;;			write-back instruction (exit point).
;;			The entry point PC of the atomic sequence can be 
;;			determined by doing r0 + r15.
;;
;;		r1:	Saved r15 stack pointer before entering the
;;			atomic sequence.
;;
;; An example atomic add sequence would look like:
;;
;;	mova	.Lend,r0		! .Lend must be 4-byte aligned.
;;	mov	r15,r1
;;	.align 2			        ! Insert aligning nop if needed.
;;	mov	#(.Lstart - .Lend),r15	! Enter atomic sequence
;;.Lstart:
;;	mov.l	@r4,r2			! read value
;;	add	   r2,r5			! modify value
;;	mov.l	r5,@r4			! write-back
;;.Lend:
;;	mov	r1,r15		! Exit atomic sequence
;;					! r2 holds the previous value.
;;					! r5 holds the new value.
;;
;; Notice that due to the restrictions of the mova instruction, the .Lend
;; label must always be 4-byte aligned.  Aligning the .Lend label would
;; potentially insert a nop after the write-back instruction which could
;; make the sequence to be rewound, although it has already passed the
;; write-back instruction.  This would make it execute twice.
;; For correct operation the atomic sequences must not be rewound after
;; they have passed the write-back instruction.

kernel/arch/dreamcast/kernel/irq.c
pcercuei: @QuzarDC it really is a arbitrary value. I picked -128 because it can be MOVI'd in one instruction without loading the value from a constant pool. What should the macro be called?

QuzarDC: Oh, ok, that makes more sense. I couldn't tell what the math that was going into reaching that from the negative values that would show up through the casting. I get it now (and hopefully you'll correct me if I'm still not), that an atomic section should never be so big for the negative value to be legitimately that large, and no legitimate address for a stack would be a negative larger than -128.

If my understanding is now correct then yea, no named macro needed. I thought you were setting it to an exact value based on the valid memory ranges for a stack.

pcercuei:
It doesn't really deviate from the reference implementation, does it? The only difference I see is that I do `interrupted_pc != atomic_exitpoint` while the reference does `interrupted_pc < atomic_exitpoint`. If it bothers you I can change it, but did you mean something else?

QuzarDC (approved):

PR #1179, opened 2025-08-31
Drop KOS symbols export list

pcercuei:
Instead of having a list of KOS symbols to export when the `INIT_EXPORT` init flag is set, we defer the responsability of providing the exports list to the applications themselves.

This is the best solution as the list of symbols was completely arbitrary, and would give the false impression of a stable ABI.


pcercuei:
All you need to do is copy the contents of KOS master's `exports.txt` (and/or `kernel/arch/dreamcast/export-*.txt`) into your own exports file.

DC-SWAT:
> All you need to do is copy the contents of KOS master's `exports.txt` (and/or `kernel/arch/dreamcast/export-*.txt`) into your own exports file.

You removed exports files from KOS, that the main issue for future usage. Making it difficult for other users to enter. It's not just me.

DC-SWAT:
> I don't think having an auto-generator is a good idea. Because then you would export every single KOS function, which means that every single KOS function (along with the data it uses etc.) ends up in the executable even if you don't actually need to call any of them, wasting lots of RAM.
> 
> Having to "manually go an fill these functions" is what we've been doing all along and it's been a pain to maintain. I'm just switching the responsibility for doing so from KOS to applications.



> I don't think having an auto-generator is a good idea. Because then you would export every single KOS function, which means that every single KOS function (along with the data it uses etc.) ends up in the executable even if you don't actually need to call any of them, wasting lots of RAM.
> 
> Having to "manually go an fill these functions" is what we've been doing all along and it's been a pain to maintain. I'm just switching the responsibility for doing so from KOS to applications.


The user will choose whether he needs an auto-generator or not. Not everyone has a shortage of RAM in their projects. And someone, like me for example, needs to export all functions.
Well, maybe you're right, it's not up to me to decide. I don’t really understand why this needs to be touched at all now, since there is no alternative solution that won’t further aggravate this problem with exports. Moreover, for those who do not use the `INIT_EXPORT` flag, this will not add memory in any way.

DC-SWAT:
> > The main problem is that if a user starts compiling something, they'll get a ton of errors about missing this and that, and they'll have to spend a lot of time writing all of them into exports. Not an easy start. And then they'll have to keep them up to date with every update. Does he really need that? And again, not everyone has memory issues. Some people prefer an automatic parser to avoid the hassle, and it has enough memory for the table and the entire kernel. Generally, memory issues only occur in heavy games; in all other cases, there's usually enough memory. I have about 2200 more exports in DreamShell, in addition to those in KOS and it works great, and I have plenty of memory. So I think there should be a choice: automatically parse everything or nothing and leave the choice to the developer.
> 
> I think both @DC-SWAT and @pcercuei have valid concerns here. Given that we already have a toggle in `environ.sh` for position-independent code for this scenario, perhaps we create a flag there to enable/disable the exports? Since position-independent code is disabled by default, the exports would be too?

It seems to me that autogeneration should be enabled by default, so that it doesn't cause problems for beginners, and advanced users can consciously turn it off themselves.
Linking of exports is only enabled if you add the flag INIT_EXPORT, that's by default is not used.

PR #1184, opened 2025-09-05
genwait: Keep queues in order by thread priority.

QuzarDC:
Currently, new waiting threads are always added to the end of their genwait tailq and the first thread in the tailq is the first woken. This means that unless a thread is being woken and readded (like can happen with a `genwait_wake_all`) they will always be awoken based on how long they have been waiting.

The downside to this method is that it ignores thread priorities, which should be the controlling factor in which thread gets woken first. By inserting into the queue by priority order, wakes will now happen based on priority rather than time waiting.

The mild overhead of seeking through the waiting queues should be outweighed greatly by allowing priority management via thread priorities.

This is a followup to #1171 and the conversation [here](https://github.com/KallistiOS/KallistiOS/pull/1171#discussion_r2324180337).

~This needs some testing to it.~ Ran against all the threading examples with no issues. The process itself is quite simple, and the only concern I'd have is that it *might* be able to cause some sort of accidental resource starvation.

### DIFF
--- a/kernel/thread/genwait.c
+++ b/kernel/thread/genwait.c
@@ -69,7 +69,7 @@ static kthread_t *tq_next(void) {
 }
 
 int genwait_wait(void *obj, const char *mesg, int timeout, void (*callback)(void *)) {
-    kthread_t   *me;
+    kthread_t   *me, *t;
 
     /* Twiddle interrupt state */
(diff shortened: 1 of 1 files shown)


pcercuei (approved):
This seems to be working well - at least I didn't see any regressions.

PR #1188, opened 2025-09-06
Cleaning doxygen warnings

QuzarDC:
Three sets of warnings that had cropped up in some recent PRs:
- irq.c - Correct param name for `irq_restore`
- purupuru.c - The deprecated struct members had an incorrect `brief`. Set as cond to have them ignored entirely.
- byteorder.h - Remove doxygen entirely for deprecated functions, as well as replace the custom-defined `BYTE_ORDER` with the standard header that defines it.

### DIFF
--- a/kernel/arch/dreamcast/include/arch/byteorder.h
+++ b/kernel/arch/dreamcast/include/arch/byteorder.h
@@ -5,134 +5,44 @@
 
 */
 
-/** \file    arch/byteorder.h
-    \brief   Byte-order related macros.
-    \ingroup system_arch
-
-    This file contains architecture-specific byte-order related macros and/or
-    functions. Each platform should define six macros/functions in this file:
-    arch_swap16, arch_swap32, arch_ntohs, arch_ntohl, arch_htons, and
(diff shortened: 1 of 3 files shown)


darcagn:
> > I'm still getting an error with this latest push:
> 
> > 
> 
> > ```
> 
> > /opt/toolchains/dc/kos-doxycheck/README.md:1: error: explicit link request to 'version' could not be resolved
> 
> > ```
> 
> 
> 
> Doesn't happen here.

What version of Doxygen are you using?

pcercuei:
@darcagn 1.9.8

darcagn (approved):

QuzarDC:
> Approving, but with the latest Doxygen there are still a good number of warnings. No reason it should hold this one up though I guess.

Thanks, that's what I was going for. The amount of changes needed to fix the others were tricky, and with some of us having wildly different versions, the progress on it was a bit stifled. But at least a lot of them were fixed with this :)

And future work won't be needlessly locked out of two of us reviewing like this one was with the multiple authors.

PR #1190, opened 2025-09-09
dc: Add functions to toggle OCINDEX, ICINDEX, OCRAM modes

pcercuei:
Add function cache_write_ccr(), which allows to safely toggle some bits of the CCR register.

Add function dcache_toggle_ocindex(), which can enable or disable OCINDEX mode.

Add function dcache_toggle_ocram(), which can enable or disable OCRAM mode.

Add function icache_toggle_icindex(), which can enable or disable ICINDEX mode.

### DIFF
--- a/kernel/arch/dreamcast/kernel/init.c
+++ b/kernel/arch/dreamcast/kernel/init.c
@@ -22,6 +22,7 @@
 #include <arch/arch.h>
 #include <arch/gdb.h>
 #include <arch/rtc.h>
+#include <dc/cache.h>
 #include <dc/memory.h>
 #include <dc/perfctr.h>
 #include <dc/ubc.h>
@@ -279,6 +280,9 @@ void  __weak_symbol arch_auto_shutdown(void) {
 /* This is the entry point inside the C program */
 void arch_main(void) {
     int rv;
+
+    /* Enable caches */
+    cache_write_ccr((uint32_t)~0, CCR_DEFAULT);
 
     dma_init();
 
--- a/kernel/arch/dreamcast/kernel/startup.S
+++ b/kernel/arch/dreamcast/kernel/startup.S
@@ -34,34 +34,6 @@ start:
 	mov.l	init_sr,r0
 	ldc	r0,sr
 
-	! Run in the P2 area
-	mov.l	setup_cache_addr,r0
-	mov.l	p2_mask,r1
-	or	r1,r0
-	jmp	@r0
-	nop
-
-setup_cache:
-	! Now that we are in P2, it's safe to enable the cache
-	mov.w	ccr_data,r1
-	mov.l	ccr_addr,r0
-	mov.l	r1,@r0
-
-	! After changing CCR, eight instructions must be executed before
-	! it's safe to enter a cached area such as P1
-	nop			! 1
-	nop			! 2
-	nop			! 3
-	nop			! 4
-	nop			! 5 (d-cache now safe)
-	nop			! 6
-	mov.l	init_addr,r0	! 7
-	mov	#0,r1		! 8
-	jmp	@r0		! go
-	mov	r1,r0
-	nop
-
-init:
 	! Save old PR on old stack so we can get to it later
 	sts.l	pr,@-r15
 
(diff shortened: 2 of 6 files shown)


QuzarDC (changes requested):
One change and one question:
- Please add the new authors into the `AUTHORS` file.
- Would it make sense to now to remove the OCRAM setup being done in startup.S, and have the c function called based on the init flag in `arch_main`? Or is that still problematic?

QuzarDC:
> > Would it make sense to now to remove the OCRAM setup being done in startup.S, and have the c function called based on the init flag in `arch_main`? Or is that still problematic?
> 
> To be honest, I'd just drop the INIT_OCRAM completely. I don't think anybody actually uses it, and those who do can switch to using dcache_toggle_ocram() in their main() when they update to KOS 2.3.0.

Alright, I'm on board with that. The only thing I'd add is to stub out the `INIT_OCRAM` macro to a deprecated const int as we've done elsewhere to ensure that there's a message noting that it no longer does anything rather than just a hard error. Before it was fixed last year I think there would have been almost no users, but that sparked a lot of discussion and promotion of the functionality.

kernel/arch/dreamcast/kernel/startup.S
```diff
@@ -44,16 +44,7 @@ start:
 setup_cache:
 	! Now that we are in P2, it's safe to enable the cache
 	! Check to see if we should enable OCRAM.
```
QuzarDC: The comments should be dropped or updated to reflect that this is now just setting the default cache settings. It might also make sense to leverage the new function `cache_write_ccr(~0, ccr_data)` here and not have its internals duplicated, but not sure if that's just overkill.

pcercuei: Ok, force-pushed with just that comment line removed, + added a commit that inits the CCR register in C which cleans up the startup.S ASM file quite a bit.

pcercuei:
> If we're doing our jobs, the user should never have to pull the HW docs out.

I did the job, Falco just didn't bother to read the functions' documentation.

pcercuei:
I don't think we can do that in a arch-agnostic fashion, this looks intrinsically arch-specific to me.

darcagn:
While GameCube does have similar functionality, I think it'd probably be a stretch to try and cover it all with a single API.

gyrovorbis:
So where does that leave this PR?

I actually slightly disagree with the notion that the SH4 and Gekko have too different scratchpad behavior to not be unified by a single API, but... 

1) There should be a low-level direct API for both platforms without the common layer, ie this PR and these routines would be DC-specific.
2) Within either `arch/cache.h` or maybe an `arch/scratchpad.h`, there could definitely be a bare minimum, platform-independent API which uses `OCRAM` on SH4 and emulates `OCRAM` by pinning a static buffer manually to the pinned cache half on Gekko, which would give you almost the exact same behavior with a common API that would simply have something like:

```c
void scratch_set_enabled(bool value);
bool scratch_is_enabled(void);
size_t scratch_size(void);
void* scratch_area(void)
```

This would basically be all of the commonality for a KOS scratchpad, but... are we interested in that being a part of KOS--a common scratchpad API--or do we want the low-level processor-specific, disjoint cache APIs in KOS, and then something like SH4ZAM could offer common abstraction?

I don't care either way, but I am very interested in seeing this functionality either in KOS or hearing from you guys that we aren't interested in having it there, and I'll add to SH4ZAM.

pcercuei:
I do think that an arch-independent scratchpad API is a bad idea (for reasons I explained on discord). Besides that, I do think if such a thing ever comes to light it would be built on top of the API introduced here in a way or another, so it can be designed in an upcoming PR, and we can keep this PR Dreamcast-specific.

pcercuei:
I rebased the PR on top of master with some slight rework of the first commit due to conflicts.
I also added a new commit that introduces the `__ocram` tag, which can be used to put variables in OCRAM space.

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -279,6 +280,9 @@ void  __weak_symbol arch_auto_shutdown(void) {
 void arch_main(void) {
     int rv;
 
+    /* Enable caches */
+    cache_write_ccr((uint32_t)~0, 0x90d);
```
QuzarDC: Now that we have the defines for them in `<dc/cache.h>`,  `(CCR_OCE | CCR_CB | CCR_OCI | CCR_ICE | CCR_ICI)` would prevent it from looking like a magic number. If that seems ugly to add here then it might make sense instead to have something like `CCR_DEFAULT` defined to that to store and describe what the default settings are.

pcercuei: Done and force-pushed.

pcercuei:
Updated the PR to add and use a `CCR_DEFAULT`, and fix the merge conflict with the AUTHORS file.

QuzarDC (approved):

PR #1191, opened 2025-09-10
examples: Fix stackprotector with builtins.

QuzarDC:
With builtins and lto enabled, `strcpy` would optimize better and it seems as though `badfunc` would just not get called at all. Forced optimization off for the example and disabled the new warning it can produce. Due to it being related to a builtins/lto having the warnings disabled in the context of the one function wasn't sufficient anymore.

### DIFF
--- a/examples/dreamcast/basic/stackprotector/Makefile
+++ b/examples/dreamcast/basic/stackprotector/Makefile
@@ -6,6 +6,9 @@
 
 TARGET = stackprotector.elf
 OBJS = stackprotector.o
+#Disable optimization that might avoid the overflow and the warning the intentional error would create
+KOS_CFLAGS += -O0 -Wno-stringop-overflow -Wno-analyzer-out-of-bounds
+#Foce stack protector on
 KOS_CFLAGS += -fstack-protector-all
 KOS_GCCVER_MIN = 4.0.0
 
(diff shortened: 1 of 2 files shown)


examples/dreamcast/basic/stackprotector/Makefile
```diff
@@ -6,6 +6,9 @@
 
 TARGET = stackprotector.elf
 OBJS = stackprotector.o
+#Disable optimization that might avoid the overflow and the warning the intentional error would create
+KOS_CFLAGS += -O0 -Wno-stringop-overflow -Wno-analyzer-out-of-bounds
+#Foce stack protector on
```
darcagn:
```suggestion
# Disable optimization that might avoid the overflow and the warning the intentional error would create
KOS_CFLAGS += -O0 -Wno-stringop-overflow -Wno-analyzer-out-of-bounds
# Force stack protector on
```

QuzarDC: Good catch, fixed in force-push.

darcagn (approved):

QuzarDC:
Oh good. With this one in, all the v2.2.2 labeled PRs have been merged. There haven't been too many and none severe, but it would be good to have another little rollup into v2.2.x soon.

PR #1193, opened 2025-09-10
Reduce example reliance on arch API

QuzarDC:
This is a set of changes to reduce the number of examples that rely on reaching into the arch API for functionality that is now available via standard library functionality. Notably this is:
- `arch_exit` -> `exit`
- `timer_xx_gettime` -> `clock`

Additionally cleaning up the inclusion of various arch headers that were either not needed, or were only being included to access other functionality.

This is towards the various goals around adding a new arch and generally with the intention that users should favor using standard library functionality where available rather than KOS-specific functions.

The timer ones might need extra review as they are not as straight forward as the rest. Additionally, I don't have a g1ata device or lightgun setup to validate those changes. Perhaps clockmaster @gyrovorbis could take a look?

### DIFF
--- a/examples/dreamcast/filesystem/sd/speedtest/sd-speedtest.c
+++ b/examples/dreamcast/filesystem/sd/speedtest/sd-speedtest.c
@@ -14,13 +14,11 @@
 #include <stdlib.h>
 #include <stdint.h>
 #include <stdbool.h>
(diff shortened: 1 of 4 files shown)


QuzarDC:
Just rebasing onto master to be sure it's all still fine. No changes.

PR #1195, opened 2025-09-15
maple: Fix race condition in the auto-detection.

DC-SWAT:
A race condition in the auto-detection logic caused multiple sub-devices to be probed concurrently using a single shared `detect_frame`. This corrupted the first sub-device's initialization data.


DC-SWAT:
Better fix for that: https://github.com/KallistiOS/KallistiOS/pull/1196

PR #1196, opened 2025-09-15
maple: Fix race between driver attach and vblank irq

pcercuei:
The device at unit 0 is probed using a shared frame dedicated to device detection; devices at unit > 0 are probed using the frame of the unit 0.

This means that as soon as a unit 0 device is registered, its frame can be used to auto-detect itself (if previously unplugged) or its children.

Therefore, prior to registering a unit 0 device into the device table, we must ensure that all fields have been properly initialized, or we will race with the vblank interrupt handler.

This PR is an alternative to #1195 that (apparently) fixes the race condition, without changing how the detection mechanism works.


DC-SWAT:
This fixes my issue with initial VMU detection.

DC-SWAT:
This PR fixed my issue with initial detection.

PR #1198, opened 2025-09-19
cmake: Add compatibility cmake file.

QuzarDC:
We've recently changed the old `Dreamcast` to
`kallistios` toolchains. Providing a passthrough
wrapper + warning to aid compatibility.

Since the change was patched in have seen a number of people seeking help with the error due to not people not knowing they need to update their `environ.sh`. This should help some.

### DIFF
--- a/utils/cmake/dreamcast.cmake
+++ b/utils/cmake/dreamcast.cmake
@@ -1,0 +1,2 @@
+message(AUTHOR_WARNING "dreamcast.cmake is deprecated, please use kallistios.cmake")
+include(kallistios)
--- a/utils/cmake/dreamcast.toolchain.cmake
+++ b/utils/cmake/dreamcast.toolchain.cmake
@@ -1,0 +1,2 @@
+message(AUTHOR_WARNING "dreamcast.toolchain.cmake is deprecated, please use kallistios.toolchain.cmake")
+include($ENV{KOS_BASE}/utils/cmake/kallistios.toolchain.cmake)


utils/cmake/Dreamcast.toolchain.cmake
```diff
@@ -0,0 +1,2 @@
+message(AUTHOR_WARNING "Dreamcast.toolchain.cmake is deprecated, please use kallistios.toolchain.cmake")
+include($ENV{KOS_BASE}/utils/cmake/kallistios.toolchain.cmake)
```
pcercuei: `include(kallistios.toolchain)` should work just as well.

QuzarDC: > `include(kallistios.toolchain)` should work just as well.

When I tried that with kos-ports it did not work. My understanding is that since this is before the directory gets added to `CMAKE_MODULE_PATH`, it only looks in the folder that cmake is being invoked at.

Though, yes, the trailing `.cmake` should be able to be removed.

```
Copying KOS files...
CMake Warning (dev) at /home/quzar/dcdev/KallistiOS/utils/cmake/Dreamcast.toolchain.cmake:1 (message):
  Dreamcast.toolchain.cmake is deprecated, please use
  kallistios.toolchain.cmake
Call Stack (most recent call first):
  /usr/share/cmake-3.31/Modules/CMakeDetermineSystem.cmake:146 (include)
  CMakeLists.txt:2 (project)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at /home/quzar/dcdev/KallistiOS/utils/cmake/Dreamcast.toolchain.cmake:2 (include):
  include could not find requested file:

    kallistios.toolchain.cmake
Call Stack (most recent call first):
  /usr/share/cmake-3.31/Modules/CMakeDetermineSystem.cmake:146 (include)
  CMakeLists.txt:2 (project)

**Comment by QuzarDC** [comment]
> Also note that the old name was `dreamcast.toolchain.cmake`, with no capitalization.
 
Oh, whoops. I had worked off someone's error report where it was getting capitalized, might have just been some visual bit because that certainly was the name. I'll correct that.

> While you're at it, you could create a similar wrapper for `dreamcast.cmake`, which was renamed to `kallistios.cmake`. It wouldn't need the warning there; right now it'd only `include(kallistios)` but in the future it would be the place to add macros that are only relevant for the Dreamcast and not the other platforms.

Wouldn't such things in the future go into the `./platform/dreamcast.cmake` or is that for something else? But yes a stub would be good there for the change in location.

File: utils/cmake/Dreamcast.toolchain.cmake (hunk 1 above)
Comment (pcercuei) [reply utils/cmake/Dreamcast.toolchain.cmake]: Ok, I think this is because KallistiOS' cmake modules path is added from the toolchain script itself, so it makes sense that it doesn't work here.

File: utils/cmake/dreamcast.toolchain.cmake [hunk 2]
```diff
@@ -0,0 +1,2 @@
+message(AUTHOR_WARNING "dreamcast.toolchain.cmake is deprecated, please use kallistios.toolchain.cmake")
+include($ENV{KOS_BASE}/utils/cmake/kallistios.toolchain.cmake)
```
pcercuei, line 2: What about `include($ENV{KOS_CMAKE_TOOLCHAIN})`?

utils/cmake/dreamcast.cmake
```diff
@@ -0,0 +1,2 @@
+message(AUTHOR_WARNING "dreamcast.cmake is deprecated, please use kallistios.cmake")
+include($ENV{KOS_BASE}/utils/cmake/kallistios.cmake)
```
pcercuei: This script is sourced manually, so after the toolchain script is loaded - as a result `include(kallistios)` should actually work.

pcercuei (approved):
Approving, with two comments, that you may choose to ignore.

utils/cmake/dreamcast.toolchain.cmake
QuzarDC: While that would still be helpful for those who are manually calling it, it would leave out those with `dreamcast.toolchain.cmake` set in their `environ.sh` as `KOS_CMAKE_TOOLCHAIN` . And part of the gap is not having a good way to signal changes are needed to `environ.sh` after an update.

I'm not sure how cmake handles a recursive call like that, but I presume it would error out in an ugly way. At that point it would probably make more sense to *just* have the message and have it be more verbose. But traditionally our compatibility approach has been to allow the old to function and just warn and eventually shift to the error condition and remove the warning message, rather than providing a message and then failing.

utils/cmake/dreamcast.cmake
QuzarDC: Ah, yes, that makes sense. I updated it. Unfortunately that means stale review, but is a better result.

PR #1205, opened 2025-09-21
tid_ying up arch

QuzarDC:
A set of small changes around cleaning up the `arch` layer:
- Moving the `tid_t` and `prio_t` types to `<kos/thread.h>` from `<arch/types.h>` . That's where they really belong. Additionally dropping the backing `handle_t` type which wasn't used anywhere (though might be a good idea to have something similar in the future that is actually used) and removing the strange usage of `tid_t` as the backing type for library handles.
- Removing stray includes of `<arch/spinlock.h>` . spinlocks simply aren't used in these files.
- Privatizing the prototype for `arch_real_exit`. This was already marked as being 'internal' and 'why would anyone use it?' 

Furthers #840 and #1155 .

### DIFF
--- a/kernel/arch/dreamcast/include/arch/arch.h
+++ b/kernel/arch/dreamcast/include/arch/arch.h
@@ -213,16 +213,6 @@ void * mm_sbrk(unsigned long increment);
 #include <kos/init.h>
 
 /* Dreamcast-specific arch init things */
-/** \brief   Jump back to the bootloader.
-    \ingroup arch
-
-    You generally shouldn't use this function, but rather use arch_exit() or
-    exit() instead.
-
-    \note                   This function will never return!
-*/
-void arch_real_exit(int ret_code) __noreturn;
-
 /** \brief   Initialize bare-bones hardware systems.
     \ingroup arch
 
--- a/kernel/arch/dreamcast/kernel/init.c
+++ b/kernel/arch/dreamcast/kernel/init.c
@@ -42,6 +42,9 @@ extern void _fini(void);
 extern void __verify_newlib_patch();
 extern void dma_init(void);
 
+/* Jump back to the bootloader. From startup.S */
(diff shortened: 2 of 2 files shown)


QuzarDC:
Force pushed to remove an accidentally added white space and correct a typo in one of the commit messages.

kernel/arch/dreamcast/include/arch/types.h
```diff
@@ -48,15 +48,6 @@ typedef volatile int32 vint32;      /**< \brief 32-bit volatile signed type */
 typedef volatile int16 vint16;      /**< \brief 16-bit volatile signed type */
 typedef volatile int8 vint8;        /**< \brief 8-bit volatile signed type */
 
-/* This type may be used for any generic handle type that is allowed
-   to be negative (for errors) and has no specific bit count
-   restraints. */
-typedef int handle_t;               /**< \brief Generic "handle" type */
```
pcercuei, line 54: You dropped `handle_t`, that's just because it's not used anywhere?

QuzarDC: Ah, I thought I'd mentioned that more explicitly in the commit log. Yes, it was only used directly below it for `prio` and `tid`. 

I do think it could be a valuable concept to have in the codebase, to better demonstrate through typing when something is a generic handle and keep them uniform, but there didn't seem to be much value in investing that into this instance in `arch/types.h` that had never been used.

QuzarDC: I guess I also missed out on pointing to #840 generally and the attempt to clear out these unnecessary typedefs from `arch/types.h` in addition to ensuring that they don't easily get forced onto users. But that was the driving motivation to moving these pieces out of the file and reducing what it is providing generally. 

Once we no longer have any code using the old types without `_t` then the file should be able to be dropped entirely.

PR #1207, opened 2025-09-21
Create example for handling raylib Image and Texture2D structures.

SilverNexus2:
Code comments note a couple caveats that may not be obvious to a newbie.

### DIFF
--- a/examples/dreamcast/raylib/texture2d/main.c
+++ b/examples/dreamcast/raylib/texture2d/main.c
@@ -183,6 +183,27 @@ int main() {
 			}
 		}
 		EndDrawing();
+
+		// Check the inputs for an exit combination of ABXYStart on every controller port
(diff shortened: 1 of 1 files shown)


QuzarDC (changes requested):
Sorry this took so long, I'd missed that there were updates. Would it be possible to provide a way to exit the example as well? I see that the loop is over such a flag from raylib, so can a button press be mapped to trigger that?

That's a thing we have for all our other examples (either that or they time out after running) to prevent having to power cycle every time.

SilverNexus2:
Definitely something I can do. I just got sidetracked in life before I got to that one.

SilverNexus2:
The five-button salute (A, B, X, Y, Start at the same time) should now terminate the example.

QuzarDC (approved):

PR #1208, opened 2025-09-24
Updates to our newlib overrides

QuzarDC:
The individual commit messages provide more details but this PR is aimed at modifying places where we have provided our own headers to override/provide newlib functionality that are unneeded or added in the wrong way.
 
 - Remove custom defined `AT_` fcntl defines (provided by newlib `fcntl.h`). 
 - Remove custom defined endianness defines (provided by newlib `machine/endianness.h`)
 - Move `kos/time.h` to `machine/time.h` so that newlib can correctly include it.
 - Remove `<arch/types.h>` from `<sys/types.h>` (moves forward #840 significantly)
 - Remove custom `<sys/sched.h>`


QuzarDC:
Force-pushed as I'd accidentally missed committing the change in `time.h` to change `CLOCKS_PER_SEC` to `_CLOCKS_PER_SEC_` as newlib expects to find.

PR #1209, opened 2025-09-24
g2dma: Fixed race condition for DMA progress checking.

DC-SWAT:
Just added `irq_disable_scoped()` to prevent this.


DC-SWAT:
> @DC-SWAT Looks good to me, but can you add a little bit more information, what was racing exactly?

I have a loop in which I call this function (SPU DMA) without blocking and check the result of errno == EINPROGRESS.
So you can see what happening to me in this case :)
This case is also used in audio streams, but there they are protected by a mutex.
And I'm sure this is the same bug that broke the sound in DCA3 when it still had DMA.

PR #1214, opened 2025-09-28
`fs_ramdisk`: cleanup and dynamic file table

QuzarDC:
General cleanup along with a transition to dynamic file handles like #1112 and #973.

- Use `bools` rather than ints where appropriate.
- Cleanups to spacing.
- Refactoring of validity checks.
- De-magic some constants.
- Fixup usage of LIST macros.
- Provide more complete coverage of errno settings to match posix.
- Correct error return for `close`.

### DIFF
--- a/kernel/fs/fs_ramdisk.c
+++ b/kernel/fs/fs_ramdisk.c
@@ -865,14 +865,11 @@ void fs_ramdisk_shutdown(void) {
 
     /* For now assume there's only the root dir, since mkdir and
        rmdir aren't even implemented... */
-    f1 = LIST_FIRST(rootdir);
-
-    while(f1) {
-        f2 = LIST_NEXT(f1, dirlist);
+    LIST_FOREACH_SAFE(f1, rootdir, dirlist, f2) {
+        LIST_REMOVE(f1, dirlist);
         free(f1->name);
         free(f1->data);
         free(f1);
-        f1 = f2;
     }
 
     free(rootdir);
--- a/kernel/fs/fs_ramdisk.c
+++ b/kernel/fs/fs_ramdisk.c
@@ -38,6 +38,7 @@ cache data from disk rather than as a general purpose file system.
 
 #include <string.h>
 #include <strings.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -122,7 +123,7 @@ static rd_file_t *ramdisk_find(rd_dir_t *parent, const char *name, size_t namele
 
 /* Find a path-named file in the ramdisk. There should not be a
    slash at the beginning, nor at the end. Assumes we hold rd_mutex. */
-static rd_file_t *ramdisk_find_path(rd_dir_t *parent, const char *fn, int dir) {
+static rd_file_t *ramdisk_find_path(rd_dir_t *parent, const char *fn, bool dir) {
     rd_file_t *f = NULL;
     char *cur;
 
@@ -187,7 +188,7 @@ static int ramdisk_get_parent(rd_dir_t *parent, const char *fn, rd_dir_t **dout,
         strncpy(pname, fn, p - fn);
         pname[p - fn] = 0;
 
-        f = ramdisk_find_path(parent, pname, 1);
+        f = ramdisk_find_path(parent, pname, true);
         free(pname);
 
         if(!f)
@@ -203,7 +204,7 @@ static int ramdisk_get_parent(rd_dir_t *parent, const char *fn, rd_dir_t **dout,
 
 /* Create a path-named file in the ramdisk. There should not be a
    slash at the beginning, nor at the end. Assumes we hold rd_mutex. */
-static rd_file_t *ramdisk_create_file(rd_dir_t *parent, const char *fn, int dir) {
+static rd_file_t *ramdisk_create_file(rd_dir_t *parent, const char *fn, bool dir) {
     rd_file_t   *f;
     rd_dir_t    *pdir;
     const char  *p;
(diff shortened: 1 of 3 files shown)


QuzarDC:
Finally ready for review. Had a bug in the first commit that I thought was in a latter one and I had shelved it trying to fix it for too long. Also thought a bug in the filesystem browser example was being caused by this, while using it for testing (#1438). I ended up submitting one of the other cleanups in it during a different PR as well, so edited the description to drop that.

kernel/fs/fs_ramdisk.c
```diff
@@ -102,6 +102,11 @@ static struct {
 /* Mutex for file system structs */
 static mutex_t rd_mutex;
 
+/* Test if an file_t is invalid, presumes mutex is already held. */
+static inline bool ramdisk_fd_invalid(file_t fd) {
+    return((fd >= FS_RAMDISK_MAX_FILES) || (!fh[fd].file));
```
pcercuei: This is funny, because `FILEHND_INVALID` won't be detected as invalid here :)

pcercuei: Actually - probably not needed, I believe this is checked in the upper FS layer already.

kernel/fs/fs_ramdisk.c
```diff
@@ -90,15 +90,17 @@ static rd_dir_t  *rootdir = NULL;
...
     rd_file_t   *file;      /* ramdisk file struct */
     uintptr_t   ptr;        /* Current read position in bytes */
     int         omode;      /* Open mode */
+    TAILQ_ENTRY(rd_fd)  next;   /* Next handle in the linked list */
     dirent_t    dirent;     /* A static dirent to pass back to clients */
     bool        dir;        /* true if a directory */
-} fh[FS_RAMDISK_MAX_FILES];
+} rd_fd_t;
+
+static TAILQ_HEAD(rd_fd_queue, rd_fd) rd_fd_queue;
```
pcercuei, line 103: You do absolutely nothing with this `rd_fd_queue` - just drop it.

pcercuei (dismissed):

kernel/fs/fs_ramdisk.c
QuzarDC: Yup, once the vfs receives a handle from the upper layer it's already checked for `FILEHND_INVALID` and that it has an entry in the global file table. These act more as sanity checks. It should be possible to drop these sorts of checks entirely for asserts, with the notion that the only way this can happen is some sort of corruption but that's a bit out of scope (I'd want to go through the FS layer really closely to be sure). It's probably something that could be cleaned up for other vfs as well.

I'll note that because I did have a few other further fixes for ramdisk that I just didn't get to (and wanted to not let this linger tooo long) like fixing up seek and implementing time stamps for stat.

kernel/fs/fs_ramdisk.c
QuzarDC: Whoops! I forgot to update shutdown to free up all the allocated `rd_fd_t` on shutdown.

kernel/fs/fs_ramdisk.c
```diff
@@ -269,182 +294,190 @@ static void * ramdisk_open(vfs_handler_t * vfs, const char *fn, int mode) {
...
     mutex_lock_scoped(&rd_mutex);
 
-    /* Check that the fd is valid */
-    if(fd < FS_RAMDISK_MAX_FILES && fh[fd].file != NULL) {
-        f = fh[fd].file;
-        fh[fd].file = NULL;
+    /* Check that the fd is invalid */
+    if(ramdisk_fd_invalid(fd)) {
+        errno = EBADF;
+        return 0;
```
andressbarajas:
```suggestion
        return -1;
```

QuzarDC: At least that's not one that I introduced, but was preexisting. Added a commit to update it.

kernel/fs/fs_ramdisk.c
```diff
@@ -269,182 +294,190 @@ static void * ramdisk_open(vfs_handler_t * vfs, const char *fn, int mode) {
...
     /* The rest require a bit more thought */
     if(mm == O_RDONLY) {
         f->openfor = OPENFOR_READ;
-        fh[fd].ptr = 0;
+        fd->ptr = 0;
     }
     else if((mm & O_RDWR) || (mm & O_WRONLY)) {
         if(f->openfor == OPENFOR_READ)
-            goto error_out;
+            return NULL;
```
andressbarajas: I ran cppcheck on your changes and it found a memory leak here. I think we can move this check to before the call to calloc if you want or free here

QuzarDC: Ah, I've only ever gotten pretty simple linty things like some of the 'param can be made const' out of cppcheck or false positives. Scooched that up before the calloc since all the info is already available.

gyrovorbis (approved):
Wow. Didn't realize you had done so much cleanup here. The dynamic list of entries is much better and I see you're properly handling errn now in addition to everything else. Awesome. Looks good here.

pcercuei:
I meant to ask... We have a FD table already in the upper FS layer, it's not possible / easy to use it instead of having another list?

QuzarDC:
> I meant to ask... We have a FD table already in the upper FS layer, it's not possible / easy to use it instead of having another list?

This is in order for calling ramdisk_shutdown to not leak out all it's allocated FDs by closing them itself (the piece I forgot to add on your first comment about the list not being used).

We could probably make for a simpler system by having the vfs shutdown merely inform the upper layer to clean up for it on shutdown rather than doing it itself. Something like `fs_shutdown_vfs(vfs_handler_t *)` to search through fds, null out their assigned vfs and free the backing handle. Then the management of it could live in that layer. On the plus side that would also allow for better handling of such cases as the upper layer could know those fds were from a disconnected fs rather than the errors that would result from calling file ops after the vfs is shut down. 

... Or just make it the user's responsibility to ensure they've closed all fds prior to calling shutdown, but we do currently operate this way for most of our vfs

kernel/fs/fs_ramdisk.c
QuzarDC: Did you still have these concerns here? Didn't want to just dismiss the request for changes without checking.

PR #1217, opened 2025-09-29
Mutex, sem, rwsem, pthreads rework

pcercuei:
This PR reworks how the mutexes, semaphores, read/write semaphores, and pthread work.

- The pthread code was reworked with a new `<kos/errno.h>` header to simplify the code.
- Pthread error-checking mutexes are now implemented inside libpthread itself, on top of recursive mutexes.
- KallistiOS now only supports normal mutexes and recursive mutexes; the only implementation difference between the two is that normal mutexes will assert() if called recursively, which means that when doing a release build where asserts are disabled, they are functionally the same.
- KallistiOS mutexes are now all error-checked using assert(). The old error-checking mutexes are now deprecated and will work like regular mutexes - which might or might not be considered an API break.
- In general, move runtime checks for errors that are never supposed to happen outside of debug builds to assert() calls. For instance, checks that ensures that a mutex_lock is not called from an interrupt context.
- Timeout values for genwait_wait, mutex_lock_timed, sem_wait_timed, rwsem_read_lock_timed, rwsem_write_lock_timed and rwsem_read_upgrade_timed() are now unsigned, which means that the calling code is responsible for passing valid values.
- Add irqsafe and scoped variants to the semaphore code.
- Switch some API functions (mutex_lock, sem_wait, rwsem_read_lock, rwsem_write_lock, rwsem_read_upgrade) to static inline wrappers around their "timed" variant.
- Update all the code that used a mutex locked in a thread and unlocked from an IRQ handler to just use a semaphore instead. This means that there are no callers to `mutex_unlock_as_thread()` anymore, which can then be dropped. This matters because this function wouldn't work with recursive mutexes (due to a race), and this allowed our mutexes to become unconditionally recursive.
- The read/write semaphore ("rwsem") code has been completely rewritten, to now use two mutexes and one atomic counter. It fixes the issue that mutexes used to have, where the lock was incorrectly assumed free after being awaken. It also addresses a limitation of the previous code, which only supported one pending writer. Finally, it allows the rwsem code to take advantage of the recent improvements done to mutexes including priority awareness and thread boosting.

All in all, this big PR makes the mutexes simpler, which also makes them *faster*. Using the following C code, which measures the performance of locking mutexes vs. spinlocks *without contention*:
```c
#include <stdio.h>
#include <kos.h>
#include <dc/perf_monitor.h>

#define NB_ITERATIONS 1024 * 1024
#define USE_MUTEX 1

static int value;

static mutex_t mutex = MUTEX_INITIALIZER;
static spinlock_t spinlock = SPINLOCK_INITIALIZER;

int main(int argc, char **argv)
{
	unsigned int i;

	do {
		perf_monitor();

		for (i = 0; i < NB_ITERATIONS; i++) {
			mutex_lock(&mutex);
			value++;
			mutex_unlock(&mutex);
	} while (0);

	do {
		perf_monitor();

		for (i = 0; i < NB_ITERATIONS; i++) {
			spinlock_lock(&spinlock);
			value++;
			spinlock_unlock(&spinlock);
		}
	} while (0);

	perf_monitor_print(stdout);
	perf_monitor_exit();


	return 0;
}
```

The execution with the current KOS master (8fb6ea27e) returns:
```
Performance monitors:
	main L21: 1 calls
		734836960 ns (734836992.000000 ns/call)
		event 0: 4571 (4571.000000 event/call)
		event 1: 970 (970.000000 event/call)
	main L31: 1 calls
		339610240 ns (339610240.000000 ns/call)
		event 0: 1969 (1969.000000 event/call)
		event 1: 1049032 (1049032.000000 event/call)
```

About 735 milliseconds for mutexes, 340 milliseconds for spinlocks.

The execution with this PR's head returns:
```
Performance monitors:
	main L21: 1 calls
		360799840 ns (360799840.000000 ns/call)
		event 0: 1842 (1842.000000 event/call)
		event 1: 1168 (1168.000000 event/call)
	main L31: 1 calls
		339585840 ns (339585856.000000 ns/call)
		event 0: 1690 (1690.000000 event/call)
		event 1: 1049611 (1049611.000000 event/call)
```

About 361 milliseconds for mutexes now, vs. 340 milliseconds still for spinlocks.

While it is not beating the no-contention locking speed of spinlocks, it gets to the point where the difference is really marginal. Note that the mutexes can be pushed below the 340 milliseconds mark if we simplify them further to forbid their use in interrupt contexts completely, but I am not sure yet if that's a good idea.

The mutex code was battle-tested. However the semaphore / rwsem code has not been tested thoroughly and might still have some bugs.

### DIFF
--- a/include/kos/rwsem.h
+++ b/include/kos/rwsem.h
@@ -100,9 +100,7 @@ int rwsem_destroy(rw_semaphore_t *s) __nonnull_all;
     \retval -1      On error, errno will be set as appropriate.
 
     \par    Error Conditions:
-    \em     EPERM - called inside an interrupt \n
     \em     ETIMEDOUT - the timeout expires before the lock can be acquired \n
-    \em     EINVAL - the semaphore is not initialized
 */
 int rwsem_read_lock_timed(rw_semaphore_t *s, unsigned int timeout) __nonnull_all;
 
@@ -114,12 +112,7 @@ int rwsem_read_lock_timed(rw_semaphore_t *s, unsigned int timeout) __nonnull_all
     call inside of an interrupt; use rwsem_read_lock_irqsafe instead.
 
     \param  s       The r/w semaphore to lock.
-    \retval 0       On success
-    \retval -1      On error, errno will be set as appropriate.
-
-    \par    Error Conditions:
-    \em     EPERM - called inside an interrupt \n
-    \em     EINVAL - the semaphore is not initialized
+    \retval 0       The return value is always 0.
 */
 __nonnull_all
 static inline int rwsem_read_lock(rw_semaphore_t *s) {
@@ -139,7 +132,6 @@ static inline int rwsem_read_lock(rw_semaphore_t *s) {
     \retval -1      On error, errno will be set as appropriate.
 
     \par    Error Conditions:
-    \em     EINVAL - the semaphore is not initialized \n
     \em     EWOULDBLOCK - called inside an interrupt and the semaphore was
                           already locked
 */
@@ -158,9 +150,7 @@ int rwsem_read_lock_irqsafe(rw_semaphore_t *s) __nonnull_all;
     \retval -1      On error, errno will be set as appropriate.
 
     \par    Error Conditions:
-    \em     EPERM - called inside an interrupt \n
     \em     ETIMEDOUT - the timeout expires before the lock can be acquired \n
-    \em     EINVAL - the semaphore is not initialized
 */
 int rwsem_write_lock_timed(rw_semaphore_t *s, unsigned int timeout) __nonnull_all;
 
@@ -172,12 +162,7 @@ int rwsem_write_lock_timed(rw_semaphore_t *s, unsigned int timeout) __nonnull_al
     safe to call inside of an interrupt; use rwsem_write_lock_irqsafe instead.
 
     \param  s       The r/w semaphore to lock.
-    \retval 0       On success.
-    \retval -1      On error, errno will be set as appropriate.
-
-    \par    Error conditions:
-    \em     EPERM - called inside an interrupt \n
-    \em     EINVAL - the semaphore is not initialized
+    \retval 0       The return value is always 0.
 */
 __nonnull_all
 static inline int rwsem_write_lock(rw_semaphore_t *s) {
@@ -197,7 +182,6 @@ static inline int rwsem_write_lock(rw_semaphore_t *s) {
     \retval -1      On error, errno will be set as appropriate.
 
     \par    Error conditions:
-    \em     EINVAL - the semaphore is not initialized \n
     \em     EWOULDBLOCK - called inside an interrupt and the semaphore was
                           already locked
 */
@@ -213,7 +197,6 @@ int rwsem_write_lock_irqsafe(rw_semaphore_t *s) __nonnull_all;
 
     \par    Error Conditions:
     \em     EPERM - the read lock is not currently held \n
-    \em     EINVAL - the semaphore is not initialized
 */
 int rwsem_read_unlock(rw_semaphore_t *s) __nonnull_all;
 
@@ -228,7 +211,6 @@ int rwsem_read_unlock(rw_semaphore_t *s) __nonnull_all;
     \par    Error Conditions:
     \em     EPERM - the write lock is not currently held by the calling
                     thread \n
-    \em     EINVAL - the semaphore is not initialized
 */
 int rwsem_write_unlock(rw_semaphore_t *s) __nonnull_all;
 
@@ -247,7 +229,6 @@ int rwsem_write_unlock(rw_semaphore_t *s) __nonnull_all;
 
     \par    Error Conditions:
     \em     EPERM - the lock is not currently held by the calling thread \n
-    \em     EINVAL - the semaphore is not initialized
 */
 int rwsem_unlock(rw_semaphore_t *s) __nonnull_all;
 
@@ -263,7 +244,6 @@ int rwsem_unlock(rw_semaphore_t *s) __nonnull_all;
 
     \par    Error Conditions:
     \em     EWOULDBLOCK - a call to rwsem_read_lock would block \n
-    \em     EINVAL - the semaphore is not initialized
 */
 int rwsem_read_trylock(rw_semaphore_t *s) __nonnull_all;
 
@@ -279,7 +259,6 @@ int rwsem_read_trylock(rw_semaphore_t *s) __nonnull_all;
 
     \par    Error Conditions:
     \em     EWOULDBLOCK - a call to rwsem_write_lock would block \n
-    \em     EINVAL - the semaphore is not initialized
 */
 int rwsem_write_trylock(rw_semaphore_t *s) __nonnull_all;
 
@@ -303,8 +282,6 @@ int rwsem_write_trylock(rw_semaphore_t *s) __nonnull_all;
     \retval -1      On error, errno will be set as appropriate.
 
     \par    Error Conditions:
-    \em     EPERM - called inside an interrupt \n
-    \em     EINVAL - the semaphore is not initialized \n
     \em     EBUSY - another reader has already requested an upgrade \n
     \em     ETIMEDOUT - the timeout expired before the write lock could be
                         acquired
@@ -329,8 +306,6 @@ int rwsem_read_upgrade_timed(rw_semaphore_t *s, unsigned int timeout) __nonnull_
     \retval -1      On error, errno will be set as appropriate.
 
     \par    Error Conditions:
-    \em     EPERM - called inside an interrupt \n
-    \em     EINVAL - the semaphore is not initialized \n
     \em     EBUSY - another reader has already requested an upgrade
 */
 __nonnull_all
@@ -352,7 +327,6 @@ static inline int rwsem_read_upgrade(rw_semaphore_t *s) {
     \par    Error Conditions:
     \em     EWOULDBLOCK - a call to rwsem_read_upgrade would block \n
     \em     EBUSY - another reader has already requested an upgrade \n
-    \em     EINVAL - the semaphore is not initialized
 */
 int rwsem_read_tryupgrade(rw_semaphore_t *s) __nonnull_all;
 


kernel/thread/mutex.c
```diff
@@ -130,7 +125,7 @@ int mutex_lock_timed(mutex_t *m, int timeout) {
 
             if(timeout) {
                 timeout = deadline - timer_ms_gettime64();
-                if(timeout <= 0) {
+                if((int)timeout <= 0) {
```
QuzarDC, line 119: Would it be faster/safer here to just do `if(deadline >= timer_ms_gettime64())` ? I know that you didn't create the problem with the change here, but the casting highlights that it's a bit off.

include/kos/mutex.h
```diff
@@ -81,7 +68,7 @@ typedef struct kos_mutex {
 */
 #define MUTEX_TYPE_NORMAL       0   /**< \brief Normal mutex type */
 #define MUTEX_TYPE_OLDNORMAL    1   /**< \brief Alias for MUTEX_TYPE_NORMAL */
-#define MUTEX_TYPE_ERRORCHECK   2   /**< \brief Error-checking mutex type */
+#define MUTEX_TYPE_ERRORCHECK   2   __depr("Error-checking mutexes are deprecated")
```
QuzarDC: This won't work in any number of uses. The way we've done it elsewhere is replace the define with a `static const` of the same name and appropriate size, which can then be marked deprecated.

include/kos/mutex.h
```diff
@@ -196,6 +177,27 @@ int mutex_lock_irqsafe(mutex_t *m) __nonnull_all;
...
+    \param  m               The mutex to acquire
+    \retval 0               On success
+    \retval -1              On error, sets errno as appropriate
+
+    \par    Error Conditions:
+    \em     EPERM - called inside an interrupt \n
+    \em     EAGAIN - lock has been acquired too many times (recursive) \n
+    \em     EDEADLK - would deadlock (error-checking)
+*/
+static inline int mutex_lock(mutex_t *m) {
```
QuzarDC, line 187: If there's no a specific reason to drop the `nonnull`, it should stay. If only to match the rest. (I think it would still retain some of the error-checking functionality despite any optimization not likely being relevant unless it happens to not be inlineable)

kernel/thread/mutex.c
pcercuei: It wouldn't work, because the timeout value needs to be updated, since it is in a for() loop.

include/kos/mutex.h
pcercuei: Ok, I dropped it because unlike `mutex_lock_scoped()`, here the compiler has full knowledge that the parameter must be non-null (because it's just a wrapper around a function that's tagged non-null). But I can leave it for consistency.

include/kos/mutex.h
pcercuei: Done.

include/kos/mutex.h
pcercuei: Done.

Note that it now reads like this:
```
__nonnull_all
static inline int mutex_lock(mutex_t *m) {
...
}
```

Because GCC will throw an error if the `__nonnull_all` is placed before the opening brace.

kernel/thread/rwsem.c
```diff
@@ -144,151 +124,59 @@ int rwsem_write_lock_irqsafe(rw_semaphore_t *s) {
...
-        /* No writers were waiting, wake up any readers. */
-        genwait_wake_all(s);
-    }
+    mutex_unlock(&s->read_lock);
+    mutex_unlock(&s->write_lock);
 
     return 0;
 }
 
-int rwsem_unlock(rw_semaphore_t *s) {
```
QuzarDC, line 194: Looks like this got accidentally dropped. Without it. libpthreads can't build.

`kos-cc -o general_pthread_test.elf general_pthread_test.o -lpthread
/opt/toolchains/dc-stable/sh-elf/lib/gcc/sh-elf/13.2.0/../../../../sh-elf/bin/ld: /tmp/ccYde0yo.ltrans0.ltrans.o: in function 'pthread_rwlock_unlock':
/home/quzar/dcdev/KallistiOS/examples/dreamcast/pthread/general/../../../../addons/libpthread/pthread_rwlock_unlock.c:20:(.text.pthread_rwlock_unlock+0x70): undefined reference to 'rwsem_unlock'
make: *** [Makefile:21: general_pthread_test.elf] Error 1`

pcercuei: Ooops. Fixed.

QuzarDC:
I hit some further snags here. It seems that the pthreads/general example is currently stack smashing. If turned on with `-fstack-protector-all` it gives the below. I had unfortunately went and spent a bit much time swapping toolchains and testing at individual commits here before thinking to test just with `master` where it's also broken. Will be trying to track that down, but any extra eyes would be appreciated. Hopefully it's just something minor with pthreads specifically.

```
KOS pthread test program:
Main thread is 0x8c0744a0
Creating detached thread to fill in left half of screen
Starting mutex test...
Thread 0 is 0x8c097c20
Thread 1 is 0x8c0a00c0
Thread 2 is 0x8c0a8560
Thread 3 is 0x8c0b0a00
Thread 4 is 0x8c0b8ea0
Waiting for threads to return...
Filler thread started
Thread 0: Started
Thread 0: Acquired the lock
Thread 0: Sleeping for 3 seconds
Thread 1: Started
Thread 2: Started
Thread 3: Started
Thread 4: Started
Filler thread finished
*** stack smashing detected ***: terminated

arch: exit return code 127
arch: shutting down kernel

File: kernel/thread/rwsem.c (hunk 4 above)
Comment (QuzarDC) [reply kernel/thread/rwsem.c:194]: > Ooops. Fixed.

Content looks fine for it but the dox now also needs adjustment since there's no setting `EPERM` anymore. That check (and similar) also need to get percolated up into the `pthread_rwlock_*` layer in the same way that was done for the mutexes, unless I'm missing how that would already happen.

**Comment by pcercuei** [comment]
@QuzarDC No problem with either `-fstack-protector` or `-fstack-protector-strong`, so maybe the -all variant gets in the way somehow.

With that said I did get some hangs while using files on a CDROM, and this PR updates the G1 code to use a semaphore instead of a mutex. These hangs may or may not be related to this PR so I need to do some more tests there.

**Comment by QuzarDC** [comment]
> @QuzarDC No problem with either `-fstack-protector` or `-fstack-protector-strong`, so maybe the -all variant gets in the way somehow.
> 
Unfortunately, that just says that the function smashing the stack [isn't one that gcc thinks is likely to do so](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html#index-fstack-protector) (things using alloca, allocating local buffers, etc). I don't think we've ever had a false positive for it, but I seem to recall that most of the ones we've found before didn't trip with the basic one (though I don't think I've tried with strong as well in the past to see).

> With that said I did get some hangs while using files on a CDROM, and this PR updates the G1 code to use a semaphore instead of a mutex. These hangs may or may not be related to this PR so I need to do some more tests there.

If you'd want to go through the hassle, it might make sense to split out the first set of mutex and pthread changes (first 5 in your list above ~12-13 commits) and have the rest in another PR. That way the toolchain rebuild can be out of the way and there can be some wider testing on the core mutex stuff while the rest gets more investigation.

**Comment by pcercuei** [comment]
Also note that if #1231 is merged before, I can rework the "irqsafe" variants as static inlines, because then `irq_inside_int()` is part of the high-level API.

File: include/kos/rwsem.h [hunk 5]
```diff
@@ -147,10 +160,9 @@ int rwsem_read_lock_irqsafe(rw_semaphore_t *s) __nonnull_all;
     \par    Error Conditions:
     \em     EPERM - called inside an interrupt \n
```
QuzarDC: This `EPERM` can't happen anymore due to the underlying mutex asserting during irq. That's fine but the documentation should be adjusted to match.

include/kos/rwsem.h
```diff
@@ -91,10 +102,9 @@ int rwsem_destroy(rw_semaphore_t *s) __nonnull_all;
...
     \em     ETIMEDOUT - the timeout expires before the lock can be acquired \n
-    \em     EINVAL - the timeout value is invalid \n
     \em     EINVAL - the semaphore is not initialized
```
QuzarDC: I don't see that the `EPERM` or `EINVAL` are possible anymore either. The same is true of each that have `rwsem_update_timed` as their backing function.

include/kos/sem.h
```diff
@@ -142,8 +159,7 @@ int sem_trywait(semaphore_t *sm) __nonnull_all;
     \retval 0               On success
     \retval -1              On error, sets errno as appropriate
 
-    \par    Error Conditions:
-    \em     EINVAL - the semaphore was not initialized
+    \par    Error Conditions: None defined
```
QuzarDC, line 162: There simply isn't a way for it to have a retval of -1 at all anymore that I see, so better to drop that and the mention of a possible error return. Perhaps noting that it always returns 0 or that the 0 return means 'completion' rather than 'success'?

include/kos/sem.h
```diff
@@ -69,49 +69,48 @@ int sem_init(semaphore_t *sm, int count);
...
     sem_trywait() for a safe function to call in an interrupt.
 
     \param  sm              The semaphore to wait on
+    \param  timeout         The maximum number of milliseconds to block (a value
+                            of 0 here will block indefinitely)
     \retval 0               On success
     \retval -1              On error, sets errno as appropriate
 
     \par    Error Conditions:
     \em     EPERM - called inside an interrupt \n
```
QuzarDC, line 89: `EPERM` is no longer returnable as it just asserts.

kernel/thread/sem.c
```diff
@@ -50,108 +50,80 @@ int sem_destroy(semaphore_t *sm) {
...
+
     /* If there's enough count left, then let the thread proceed */
-    else if(sm->count > 0) {
-        sm->count--;
-    }
-    else {
+    if(sm->count < 0) {
         /* Block us until we're signaled */
-        sm->count--;
         rv = genwait_wait(sm, timeout ? "sem_wait_timed" : "sem_wait", timeout,
```
QuzarDC, line 68: Is there a reason to keep `rv` around here and not drop the `genwait_wait` directly into the if below? After the irq scoped disable cleanup, that's the only use of it.

QuzarDC (changes requested):
Just one set of things to go through and otherwise it all looks great. The same removed errno values apply throughout rwsem.h, but github doesn't allow review comments on parts that did not change, so I couldn't tag them each individually.

kernel/thread/rwsem.c
QuzarDC: I moved the rest of the commentary here into a proper code review. I was entirely wrong on the pthreads part, as we aren't passing along as many errnos now, but for rwsems pthreads wasn't expecting any of them in order to meet spec.

kernel/thread/sem.c
pcercuei: That's just preference, I generally avoid nesting function calls, or calls inside control blocks.
I can update it if you feel strongly about it, otherwise I'd rather keep it like that.

QuzarDC: Nope, no strong feelings, just curious. Especially in case there was some specific technical reason that I wasn't following.

PR #1220, opened 2025-10-01
dc-chain: update dockerfile to properly handle the new config filenames

Arno500:
Allowing the user to choose which target they want to setup the toolchain for. Without this, the latest master do not build using the Dockerfile.


pcercuei:
Note that I already have a similar commit in my CI PR: https://github.com/KallistiOS/KallistiOS/pull/1026

PR #1221, opened 2025-10-02
pvr: Add a way to force vertical scale; fix fb_tex example on non-VGA

pcercuei:
This PR adds a new `pvr_set_vertical_scale()` which can be used to specify a different vertical scale factor, used when the PVR scene is rendered to the framebuffer.

It addresses an issue with the fb_tex example (and my Bloom emulator) when used with a non-VGA cable; in which case the default scale factor was configured to 0.999x instead of 1.0x, which caused a rather neat fading effect towards the center (but unwanted in this example).

### DIFF
--- a/examples/dreamcast/pvr/fb_tex/fb_tex.c
+++ b/examples/dreamcast/pvr/fb_tex/fb_tex.c
@@ -203,6 +203,12 @@ int main(int argc, char **argv) {
     /* Set the stride length for strided textures. */
     pvr_txr_set_stride(640);
 
+    /* The default vertical scale is actually 0.999f when a regular A/V cable
+     * is plugged. This won't work properly here, as the error will accumulate
+     * in the framebuffer, causing a rather neat fading effect towards the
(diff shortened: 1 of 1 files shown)


QuzarDC:
> @QuzarDC My understanding is that scaling by 0.999x acts as a smoothing filter, which enhances the output on non-VGA. Sadly I don't have a good setup to test it properly. If somebody with a good setup (e.g. a good CRT TV or something like a retrotink) could test, it would be amazing.

Got it. I guess the other side of the question would be something like 'why not also have it on for VGA?' I'll see about opening an issue for this and maybe modify an example or two to be able to adjust the value of the scaling to open it up for anyone who can run a bin. Could be a useful thing to provide it as part of a 'suggested setting' based off user display type. The choice to have it this way for VGA/non likely resulted from some minor testing of that sort on displays available ~2000. I could imagine the perceived difference could vary greatly from DCHDMI vs AV to capture card vs LCD VGA vs CRT VGA etc etc.

PR #1223, opened 2025-10-03
cmake: Correct path for addon includes.

QuzarDC:
Was incorrectly defined in #1035. The addons
include DIR does not separate by arch. It's
expected the headers would be shared across
archs.


pcercuei:
In general I agree, but we need to be a bit careful - some libs install headers that contain macros reflecting the lib's configuration, which might differ across archs.

darcagn (dismissed):

Issue #1225, opened 2025-10-06
fs_open() fails after calling vmufb_present() function


kapodamy:
Fails with the message `vmufs_root_read: can't read block 255 on device A1`. 
¿missing maple_frame_unlock() call somewhere? PoC here:

```C
#include <arch/arch.h>
#include <stdio.h>
#include <kos/init.h>
#include <kos/fs.h>
#include <dc/biosfont.h>
#include <dc/maple.h>
#include <dc/maple/vmu.h>
#include <dc/vmu_fb.h>
#include <dc/vmu_pkg.h>
#include <dc/vmufs.h>
#include <errno.h>

KOS_INIT_FLAGS(
    INIT_IRQ | INIT_CONTROLLER | INIT_VMU | INIT_NO_DCLOAD | INIT_LIBRARY | INIT_FS_PTY
);

int main(int argc, char* argv[]) {
    const int PORT = 0;
    const int UNIT = 1;
    
    // insert filename here (optional)
    char PATH[] = "/vmu/up/__STUB_FILENAME__";
    PATH[5] = (char){0x61 + PORT};
    PATH[6] = (char){0x30 + UNIT};
    
    maple_device_t* dev = maple_enum_dev(PORT, UNIT);
    
    vmufb_present(&(vmufb_t){.data = {0}}, dev);
    
    // this fails with "vmufs_root_read: can't read block 255 on device A1"
    file_t vms_file = fs_open(PATH, O_RDONLY);    

#if 0
    // not working STUB for maple_frame_lock() function, maybe the frame is different
    maple_frame_t* frame = &dev->frame;
    if (frame->queued == 1 || frame->state == MAPLE_FRAME_UNSENT) {
        printf("maple_frame_lock() about to return -1.\n");
    }
#endif
 
    if (vms_file != FILEHND_INVALID)
        printf("SUCCESS\n");
    else 
        printf("FAIL\n");

    arch_menu();
    return 0;
}
```

QuzarDC:
This was probably a side effect of vmu drawing no longer waiting for an unlock to return (#904) the intent there was to prevent the need to block for the whole vblank or more while the command executed and returned back its status, especially when there's not really anything the user could/would do if the drawing failed. But that seems to create a situation where it might still be locked when attempting the open and we don't currently signal properly that the user should try again.

[...]

PR #1230, opened 2025-10-13
VMU VFS overhaul

kapodamy:
~~When the vmu header is corrupted, `fs_open` reads the entire file (header+payload), meaning that the first call to `fs_read` will return bytes from the header when the payload is expected. Useless O_META was used for raw access.~~

~~This PR adds another IOCTL to fs_vmu to check if whatever header was corrupted or not. Also serveral changes was made to `fs_vmu`, `vmufs` and `vmu_pkg`.~~

This PR introduces various API breaking changes and rewrites fs_vmu.

what changed:
* vmu_pkg:
    * support for GAME file types.
    * update return values from `vmu_pkg_parse` function.
    * implement `vmu_pkg_parse_ex`, `vmu_pkg_build_ex` and `vmu_pkg_crc_set` functions.
* vmu_fs:
        * return proper error codes like read or FAT errors.
        * add extra output param in `vmufs_read` function to retrieve file type.
        * check available blocks before old file deletion in vmufs_write() function.
* fs_vmu:
    * implement `IOCTL_VMU_GET_HDR_STATE` ro retrieve header status from opened files.  
    * implement `IOCTL_VMU_GET_INTLDATGM` to retrieve innitial data from opened  vmu game files.
    * implement `IOCTL_VMU_GET_REALFSIZE` to retrieve real file size (multiple of 512).
    * support for GAME file types.
    * rewrite file header management.
    * now fs_open set a proper errno before return.
    * _FIXME_: enforce api-breaking `O_CREATE`.
    * enforce openning specific file types: GAME and DATA. May be api-breaking.
    * various fixes.
    * add new example for `vmu_pkg_parse_ex` usage.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/fs_vmu.h
+++ b/kernel/arch/dreamcast/include/dc/fs_vmu.h
@@ -172,7 +172,7 @@ static inline int fs_vmu_get_file_size(file_t fd) {
     \return                 Header status
 */
 static inline vmu_hdr_status_t fs_vmu_get_header_status(file_t fd) {
-    return fs_ioctl(fd, IOCTL_VMU_GET_HDR_STATE);
+    return (vmu_hdr_status_t)fs_ioctl(fd, IOCTL_VMU_GET_HDR_STATE);
 }
 
 /** \brief  Set a default header for newly created VMU files


kernel/arch/dreamcast/include/dc/fs_vmu.h
```diff
@@ -48,9 +48,29 @@ __BEGIN_DECLS
...
+    When O_META flag is not used, fs_open() function will parse the VMU file header.
+
+    If successful the header is hidden, so any operation in the VMU file won't see the header.
+    However, if it fails, the entire file is available in the same way as with O_META (raw access).
+
+    The codes returned by IOCTL_VMU_GET_HDR_STATE indicate status of that parse.
+
+    @{
+*/
+#define FSVMU_HEADEROK     0   /**< \brief Header parsed correctly */
```
gyrovorbis: We've been slowly redoing all public APIs to move away from groups of weakly typed disjoint #defines in favor of strongly typed enums... would definitely do that here,

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -66,6 +66,7 @@ typedef struct vmu_fh_str {
     uint8 *data;                        /* copy of the whole file */
     vmu_pkg_t *header;                  /* VMU file header */
     bool raw;                           /* file opened as raw */
+    int hdr_status;                     /* Parse status of VMU file header */
```
QuzarDC: `vmu_hdr_status_t hdr_status;`

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -280,9 +288,13 @@ static vmu_fh_t *vmu_open_file(maple_device_t * dev, const char *path, int mode)
         }
         datasize = 512;
         memset(data, 0, 512);
-    } else if(!fd->raw && !vmu_pkg_parse(data, datasize, &vmu_pkg)) {
-        fd->header = vmu_pkg_dup(&vmu_pkg);
-        fd->start = (unsigned int)vmu_pkg.data - (unsigned int)data;
+    } else if(!fd->raw) {
+        if(vmu_pkg_parse(data, datasize, &vmu_pkg)) {
```
QuzarDC: It might be worthwhile to update `vmu_pkg_parse` in order to help it better distinguish between the two error conditions it has of 'header wrong size' and 'CRC mismatch' and perhaps add some extra ones. Though I'm not sure if that would change the workflow for the user presuming the file simply didn't have any header saved vs there was likely a header saved but it is now corrupt.

If there wouldn't be any then no need to add further things, but the fact that `STATUS_NEWFILE` was separated out as well made it seem like there might be a desire for that specificity.

kernel/arch/dreamcast/include/dc/fs_vmu.h
```diff
@@ -73,6 +92,55 @@ static inline int fs_vmu_set_header(file_t fd, const vmu_pkg_t *pkg) {
...
+    This function can be used to check if the header was parsed correctly (which contains the
+    metadata, icons...) from an opened VMU file.
+
+    Note: always returns 0 for truncated files and directories.
+
+    \see file_hdr_status
+    \param fd               A file descriptor corresponding to the VMU file
+    \return                 status
+*/
+static inline vmu_hdr_status_t fs_vmu_get_header_parse_status(file_t fd) {
```
QuzarDC: This could probably be `fs_vmu_get_header_status` considering there isn't a different status for it and the term `parse status` wasn't used for the rest.

kernel/arch/dreamcast/include/dc/fs_vmu.h
```diff
@@ -73,6 +92,55 @@ static inline int fs_vmu_set_header(file_t fd, const vmu_pkg_t *pkg) {
...
+    \param  fn              The file path to open.
+    \param  mode            The mode to use with opening the file. This may
+                            include the standard open modes (O_RDONLY, O_WRONLY,
+                            etc), as well as values from the \ref vfs_fopen_modes
+                            "File Open Modes" list. Multiple values can be ORed
+                            together.
+
+    \return                 The new file descriptor on success, -1 on error.
+*/
+static inline file_t fs_vmu_open_constrained(const char* fn, int mode) {
```
QuzarDC: Is there a reason to have this as a static inline? It makes some sense for the get header above, since that's just a thin wrapper around `fs_ioctl`, but here I don't see a reason for it.

kernel/arch/dreamcast/include/dc/fs_vmu.h
```diff
@@ -109,6 +177,7 @@ static inline int fs_vmu_set_default_header(const vmu_pkg_t *pkg) {
 
 /** @} */
 
+
```
QuzarDC: Extra whitespace.

kernel/arch/dreamcast/fs/fs_vmu.c
kapodamy: ok

kapodamy:
splitted in 15 commits by feature and usage (from each previous commit)

kernel/arch/dreamcast/include/dc/vmufs.h
```diff
@@ -325,9 +325,15 @@ int vmufs_readdir(maple_device_t * dev, vmu_dir_t ** outbuf, int * outcnt);
...
+    \param  outtype         Storage for the file type, 0x33=DATA or 0xCC=GAME.
     \return                 0 on success, or <0 on failure.
+
+    \retval 0               On success.
+    \retval -1              On invalid device.
+    \retval -2              On file not found.
+    \retval -3              On read error.
 */
-int vmufs_read(maple_device_t * dev, const char * fn, void ** outbuf, int * outsize);
+int vmufs_read(maple_device_t * dev, const char * fn, void ** outbuf, int * outsize, uint8 * outtype);
```
pcercuei, line 336: You change `vmufs_read()` in this commit but didn't update `kernel/arch/dreamcast/fs/fs_vmu.c`, which means that this commit is not buildable. Please update the function call(s) in this commit as well; this matters for bisectability that all commits are buildable.

kernel/arch/dreamcast/util/vmu_pkg.c
```diff
@@ -168,12 +169,13 @@ int vmu_pkg_parse(uint8_t *data, size_t data_size, vmu_pkg_t *pkg) {
...
     icon_size = 512 * hdr->icon_cnt;
     ec_size = vmu_eyecatch_size(hdr->eyecatch_type);
-    hdr_size = sizeof(vmu_hdr_t) + icon_size + ec_size;
+    hdr_size = sizeof(vmu_hdr_t) + icon_size + (size_t)ec_size;
     total_size = hdr->data_len + hdr_size;
 
-    if(total_size < 0 || (size_t)total_size > data_size) {
-        dbglog(DBG_ERROR, "vmu_pkg_parse: file header is corrupted, or no header?\n");
-        return -1;
+    if (ec_size < 0 || icon_size < 1 || total_size > data_size) {
```
pcercuei: This check is invalid. Since `hdr->icon_cnt` is `int`, if it contains garbage you can get an arbitrary high `icon_size`. So it would be better to check that `hdr->icon_cnt` is between 1 and 3.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -251,36 +261,71 @@ static vmu_fh_t *vmu_open_file(maple_device_t * dev, const char *path, int mode)
...
+        rv = vmu_pkg_parse(data, datasize, &vmu_pkg);
+        if (rv == -1) {
+            /* Invalid checksum */
+            errno = EILSEQ;
+            goto error;
+        } else if(rv == -2) {
+            /* vmufs_read() doesn't lie about the file size, file is totally corrupted */
+            errno = EIO;
+            goto error;
+        }
```
pcercuei, line 297: That is wrong, sorry. `vmu_open_file()` should be able to still open files that have no header. I made sure it was possible when I added support for transparent headers, so that it wouldn't unnecessarily break ABI with old code.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -290,12 +335,17 @@ static vmu_fh_t *vmu_open_file(maple_device_t * dev, const char *path, int mode)
...
     if(fd->filesize == 0) {
         dbglog(DBG_WARNING, "VMUFS: can't open zero-length file %s\n", path);
-        free(fd->data);
-        free(fd);
-        return NULL;
+        errno = ENODATA;
+        goto error;
     }
 
+    errno = 0;
```
pcercuei: Note that you don't have to bother setting errno in the success path, as this variable is assumed set only on error.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -547,39 +554,42 @@ static ssize_t vmu_write(void * hnd, const void *buffer, size_t cnt) {
...
-        n = ((fh->loc + fh->start + cnt) - (fh->filesize * 512));
+        n = ((fh->loc + fh->start + cnt) - (fh->blks * 512));
 
         if(n & 511)
             n = (n + 512) & ~511;
 
         n = n / 512;
 
-        dbglog(DBG_SOURCE(VMUFS_DEBUG), "VMUFS: extending file's filesize by %d\n", n);
+        dbglog(DBG_SOURCE(VMUFS_DEBUG), "VMUFS: extending file's blks by %d\n", n);
```
pcercuei: `blocks`? At least in the debug message.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -547,39 +554,42 @@ static ssize_t vmu_write(void * hnd, const void *buffer, size_t cnt) {
...
     /* insert the data in buffer into fh->data at fh->loc */
     dbglog(DBG_SOURCE(VMUFS_DEBUG), "VMUFS: adding %d bytes of data at loc %ld (%ld avail)\n",
-           cnt, fh->loc, fh->filesize * 512);
+           cnt, fh->loc, fh->blks * 512);
 
     memcpy(fh->data + fh->loc + fh->start, buffer, cnt);
     fh->loc += cnt;
 
+    if((uint32)fh->loc > fh->filelength)
+        fh->filelength = (uint32)fh->loc;
```
pcercuei: Please use `uint32_t` (and in general stdint.h types) instead of `uint32`, which we're trying to get rid of in the code base.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -51,6 +51,18 @@ debug output.
 #define VMU_FILE    1
 #define VMU_ANY     -1  /* Used for checking validity */
 
+/* File header type */
+typedef enum vmufs_hdr_type {
+    VMUFS_HDR_NONE = 0x00,
+    VMUFS_HDR_DATA = 0x33
+} vmufs_hdr_type;
```
pcercuei: We use the `_t` suffix for typedefs.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -268,8 +270,8 @@ static vmu_fh_t *vmu_open_file(maple_device_t * dev, const char *path, int mode)
             /* File not found, flag to setup new blocks. */
             datasize = -1;
         }
-        else if(filetype != 0x33) {
-            dbglog(DBG_WARNING, "VMUFS: file %s isn't DATA type\n", path);
+        else if(filetype != 0x33 && filetype != 0xCC) {
+            /* Unknown file type, this never should happen */
```
pcercuei, line 261: Should probably keep the warning here.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -208,19 +211,24 @@ static vmu_fh_t *vmu_open_dir(maple_device_t * dev) {
...
-    if(vmufs_readdir(dev, &dirents, &dircnt) < 0)
+    if(vmufs_readdir(dev, &dirents, &dircnt) < 0) {
+        errno = ENOENT;
         return NULL;
+    }
 
     /* Allocate a handle for the dir blocks */
-    if(!(dh = malloc(sizeof(vmu_dh_t))))
+    if(!(dh = malloc(sizeof(vmu_dh_t)))) {
+        errno = ENOMEM;
```
pcercuei: No need to set `errno` here (and on the other places you set it to ENOMEM), malloc() already does.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -308,31 +284,27 @@ static vmu_fh_t *vmu_open_file(maple_device_t * dev, const char *path, int mode)
 
     /* We were flagged to set up new blocks */
     if(datasize == -1) {
-        fd->blks = 1;
+        fd->blks = 2;
```
pcercuei: This transparently went from 1 to 2, for some reason. Please explain why it's needed.

kernel/arch/dreamcast/include/dc/fs_vmu.h
```diff
@@ -51,6 +51,7 @@ int fs_vmu_shutdown(void);
 #define IOCTL_VMU_SET_HDR          0x564d5530 /* "VMU0" */
 #define IOCTL_VMU_GET_REALFSIZE    0x564d5531 /* "VMU1" */
 #define IOCTL_VMU_GET_HDR_STATE    0x564d5532 /* "VMU2" */
+#define IOCTL_VMU_GET_INTLDATGM    0x564d5533 /* "VMU3" */
```
pcercuei: Nitpick, but I'm not fond of the macro name, `INTLDATGM` doesn't mean much to me. What about `INIT_DATA`?

examples/dreamcast/vmu/vmu_pkg_ex/test_vmudata.c
```diff
@@ -0,0 +1,359 @@
...
+    .type _RAW_VMS_HEADER_START, @object                                        \n\
+    .global _RAW_VMS_HEADER_SIZE_PTR                                            \n\
+    .type _RAW_VMS_HEADER_SIZE_PTR, @object                                     \n\
+    _RAW_VMS_HEADER_START:                                                      \n\
+        .incbin \"SMPLTXTSD.VMS\"                                              \n\
+    _RAW_VMS_END:                                                               \n\
+        .set _RAW_VMS_HEADER_SIZE_PTR, _RAW_VMS_END - _RAW_VMS_HEADER_START     \n\
+");
+extern uint8_t RAW_VMS_HEADER_START;
+extern uint8_t RAW_VMS_HEADER_SIZE_PTR;
```
pcercuei: Could you move the VMS file to the romdisk instead? This would make the code simpler and closer to how people might actually want to use it.

pcercuei (changes requested):
First round of review, a few comments inline.

One thing I don't understand, why do you need a way to force reading a file when the CRC is wrong? If the CRC fails, the app still has the option to open the file with `O_META` if it really needs to read the corrupted stuff.

examples/dreamcast/vmu/vmu_pkg_ex/test_vmudata.c
kapodamy: ok

kernel/arch/dreamcast/include/dc/fs_vmu.h
kapodamy: `IOCTL_VMU_GET_INIT_DATA` will be

kernel/arch/dreamcast/util/vmu_pkg.c
kapodamy: only was checking if was less than 1.

kernel/arch/dreamcast/fs/fs_vmu.c
kapodamy: In this commit the file is rejected as-is.
This changes with the implementation of `IOCTL_VMU_GET_HDR_STATE`  and `vmu_hdr_status_t` in this commit 9016adb1c7433cee78c4106e4929b3c74e0f9026.

`/* Parse header but ignore any bad checksum */`
I forgot move this one (also belongs to 9016adb1c7433cee78c4106e4929b3c74e0f9026 too)

kernel/arch/dreamcast/fs/fs_vmu.c
kapodamy: missed explanation for this one.
Header takes 640 bytes (128 in fields and 512 single icon) thats why is 2 "blocks", the idea is have some preallocated space. I will add the corresponding comment later.

kernel/arch/dreamcast/util/vmu_pkg.c
pcercuei: Yes but your `icon_size` is a `size_t` so the check is invalid (it only checks that it's non-zero, not negative), and you'd better check the value of `hdr->icon_cnt` instead.

kapodamy:
Applied all requested changes,  i also dropped `vmu_pkg_t::initial_data` change in favor of `vmu_pkg_build_ex`() function.

examples/dreamcast/vmu/vmu_pkg/vmu.c
```diff
@@ -134,8 +134,10 @@ void write_entry(void) {
     vmu_pkg_load_icon(&pkg, "/rd/ebook.ico");
     vmu_pkg_build(&pkg, &pkg_out, &pkg_size);
 
-    fs_unlink("/vmu/a1/TESTFILE");
-    f = fs_open("/vmu/a1/TESTFILE", O_WRONLY);
+    /* Delete existing entry */
+    //fs_unlink("/vmu/a1/TESTFILE");
+
+    f = fs_open("/vmu/a1/TESTFILE", O_WRONLY | O_CREAT);
```
DC-SWAT: `O_WRONLY | O_CREAT | O_TRUNC`

examples/dreamcast/vmu/vmu_pkg_ex_data/vmudata.c
```diff
@@ -0,0 +1,362 @@
...
+#include <kos/init.h>
+#include <kos/thread.h>
+
+
+// no stdio here, except for this one
+extern int snprintf(char* s, size_t n, const char* format, ...);
+
+
+KOS_INIT_FLAGS(
+    INIT_IRQ | INIT_CONTROLLER | INIT_VMU | INIT_LIBRARY | INIT_FS_PTY | INIT_FS_ROMDISK
```
DC-SWAT: Better use INIT_DEFAULT in examples if you don't need a special case.
Also doesn't clear for what you need `INIT_LIBRARY` and `INIT_FS_PTY` here.

kernel/arch/dreamcast/fs/fs_vmu.c
```diff
@@ -636,37 +835,63 @@ static dirent_t *vmu_readdir(void * fd) {
...
+                /* Not applicable */
+                return VMUHDR_STATUS_OK;
+            }
+            return fh->header_status;
+        case IOCTL_VMU_GET_INIT_DATA:
+            out_intial_data = va_arg(ap, const void **);
+
+            if(!out_intial_data) {
+                return -1;
+            } else if(fh->strtype != VMU_FILE) {
```
DC-SWAT: `else` must be from new line by code style.
It was the old way here because the code was very old.

examples/dreamcast/vmu/vmu_pkg_ex_data/vmudata.c
kapodamy: functions like `printf()` needs `INIT_FS_PTY` otherwise doesn't print. `INIT_LIBRARY` can be removed

pcercuei: @kapodamy That's still in the default flags. You can probably drop the whole KOS_INIT_FLAGS().

kapodamy: ok

examples/dreamcast/vmu/vmu_pkg_ex_data/vmudata.c
```diff
@@ -20,9 +20,8 @@
 extern int snprintf(char* s, size_t n, const char* format, ...);
 
 
-KOS_INIT_FLAGS(
-    INIT_IRQ | INIT_CONTROLLER | INIT_VMU | INIT_LIBRARY | INIT_FS_PTY | INIT_FS_ROMDISK
-);
+KOS_INIT_FLAGS(INIT_DEFAULT | INIT_CONTROLLER);
```
DC-SWAT: `INIT_CONTROLLER` already in default.

examples/dreamcast/vmu/vmu_pkg/vmu.c
```diff
@@ -134,10 +134,10 @@ void write_entry(void) {
     vmu_pkg_load_icon(&pkg, "/rd/ebook.ico");
     vmu_pkg_build(&pkg, &pkg_out, &pkg_size);
 
-    /* Delete existing entry */
+    /* Delete existing entry (optional, is done in fs_close() func) */
     //fs_unlink("/vmu/a1/TESTFILE");
```
DC-SWAT: For what this? You'll always rewrite file with `O_TRUNC`.

pcercuei: But it's commented out? ...

examples/dreamcast/vmu/vmu_pkg_ex_data/vmudata.c
pcercuei: Yeah, you can drop the `KOS_INIT_FLAGS()` completely.

examples/dreamcast/vmu/vmu_pkg/vmu.c
DC-SWAT: > But it's commented out? ...

What I mean is that this comment is also useless here. Both lines can to be removed.

kapodamy: here is clarified why was commented out: `optional, is done in fs_close() func`

addons/libkosutils/netcfg.c
```diff
@@ -54,7 +54,7 @@ void netcfg_vmuify(const char *filename_in, const char *filename_out) {
     dbgp("Closing source file\n");
 
     dbgp("Opening output file\n");
-    fd = fs_open(filename_out, O_WRONLY);
+    fd = fs_open(filename_out, O_WRONLY | O_CREAT);
```
DC-SWAT:
```suggestion
    fd = fs_open(filename_out, O_WRONLY | O_CREAT | O_TRUNC);
```

kapodamy:
@pcercuei writing is not guaranteed from block 0 because `vmufs_write` function is used.

pcercuei:
First quick test on Bloom and it just doesn't work. It won't find my save states.
```
vmu_block_read failed: DATATRF(8)/82010008
vmufs_root_read: can't read block 255 on device A1
No VMU file found, loading dummy memcard
vmu_block_read failed: DATATRF(8)/82414008
vmufs_root_read: can't read block 255 on device B1
No VMU file found, loading dummy memcard
```

kapodamy:
@pcercuei maybe related to https://github.com/KallistiOS/KallistiOS/issues/1225 do `thd_pass` and then `fs_open` again, example:
```C
size_t open_attempts = 3;
L_try_open:
    errno = 0;
    file_t vms_file = fs_open(FUNKINSAVE_VMU_SAVEFILE_PATH, O_RDONLY);

    if (vms_file == FILEHND_INVALID) {
        if (errno == 5 && open_attempts > 0) {
            open_attempts--;
            thd_pass();
            goto L_try_open;
        }

        logger_info("no savedata present in %s", FUNKINSAVE_VMU_SAVEFILE_PATH);
        return 2;
    }
```

PR #1231, opened 2025-10-18
Add and use `<kos/irq.h>` header

pcercuei:
Move a few functions and macros to a new arch-agnostic `<kos/irq.h>`.
Having these functions and macros as arch-agnostic will allow to write a bigger part of KallistiOS as arch-agnostic as well.

The second commit drops some useless `<arch/irq.h>` includes, and changes some more to `<kos/irq.h>`.

This is meant to be a first step towards an arch-agnostic IRQ API; a lot of it has still to be cleaned up (IRQ codes, priorities, IRQ context creation, registering / unregistering of handlers, etc).

### DIFF
--- a/kernel/arch/dreamcast/include/arch/irq.h
+++ b/kernel/arch/dreamcast/include/arch/irq.h
@@ -268,18 +268,6 @@ static inline void arch_irq_enable(void) {
     @{
 */
 
-/** Resume normal execution from IRQ context.
-
-    Pretend like we just came in from an interrupt and force a context switch
-    back to the "current" context.
-
-    \warning
-    Make sure you've called irq_set_context() before doing this!
-
-    \sa irq_set_context()
-*/
-void irq_force_return(void);
-
 void arch_irq_create_context(irq_context_t *context,
                              uintptr_t stack_pointer,
                              uintptr_t routine,
--- a/kernel/exports.txt
+++ b/kernel/exports.txt
@@ -199,7 +199,6 @@ dbgio_read_buffer
 dbgio_printf
 
 # Interrupt / Exception handling
-irq_force_return
 irq_disable
 irq_enable
 irq_restore


pcercuei:
Rebased since it was conflicting with a doxygen fix.
I also updated the commit that introduces `<kos/arch_types.h>` since C++ does not like C-style forward declaration of enums.

include/kos/irq.h
```diff
@@ -0,0 +1,232 @@
...
+
+    \return                 The current IRQ context.
+
+    \sa irq_set_context()
+*/
+static inline irq_context_t *irq_get_context(void) {
+    return arch_irq_get_context();
+}
+
+/** @} */
```
QuzarDC, line 198: The end group marker here doesn't have a group associated with it. It seems like this was matched to the irq context group in the original.

kernel/thread/thread.c
```diff
@@ -425,13 +425,13 @@ kthread_t *thd_create_ex(const kthread_attr_t *restrict attr,
...
-            params[1] = (uint32_t)param;
+            params[0] = (uintptr_t)routine;
+            params[1] = (uintptr_t)param;
             params[2] = 0;
             params[3] = 0;
             irq_create_context(&nt->context,
-                               ((uint32_t)nt->stack) + nt->stack_size,
-                               (uint32_t)thd_birth, params, 0);
+                               ((uintptr_t)nt->stack) + nt->stack_size,
+                               (uintptr_t)thd_birth, params);
```
QuzarDC, line 434: btw, great catch here. This had shown up in my 'null arch' branch (building the kernel with the host compiler) as being mismatched sizes, and hadn't yet gotten to fixing it up.

```
kos-cc   -c thread.c -o thread.o
thread.c: In function ‘thd_create_ex’:
thread.c:428:25: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
  428 |             params[0] = (uint32_t)routine;
      |                         ^
thread.c:429:25: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
  429 |             params[1] = (uint32_t)param;
      |                         ^
thread.c:433:33: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
  433 |                                ((uint32_t)nt->stack) + nt->stack_size,
      |                                 ^
thread.c:434:32: warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]
  434 |                                (uint32_t)thd_birth, params, 0);
      |                                ^

QuzarDC:
> @QuzarDC I could solve the problem without `<kos/arch_types.h>` by being smarter about how the arch-specific and arch-agnostic files include each other. So the types are now properly declared in `<kos/irq.h>`, and the arch-specific types are defined in `<arch/irq.h>`.

I had started trying to work through the problem you mentioned this morning, and I basically had the same notion. `arch/irq.h` doesn't need to be free-standing anymore so the types can be defined ahead of it. Unfortunately that still leaves some of the need for the c/c++ hackiness it seems, but that seems a better organization already.

Regarding that, it could make sense to have the underlying type specification also happen if we're in C23 since it's also supported there. Not sure if it would cause problems if in C gcc randomly decided the underlying type was `int` and in cpp it was `unsigned int` but it might merit some static asserts or something to make sure the size ends up as expected.

Still have that stray dox bit too though that should at least be removed, but if possible the right opening brace should be added to get the grouping to look good. I have a feeling though that there's gonna need to be an independent effort after all the arch shuffling to get the docs into a better organization generally.

pcercuei:
Updated with new doc sections in the kos/irq.h header.

pcercuei:
Rebased once again because of conflicts.

QuzarDC (approved):

PR #1232, opened 2025-10-24
PVR and SPU memory expansion for Dreamcast/Pristine target.

DC-SWAT:
Also added a similar check to detect the video cable.
This will allow developers to use more memory on NAOMI while maintaining full compatibility with the Dreamcast through the use of DreamShell.
Nothing else needs to be changed because I was able to get Maple port A to work on the NAOMI using [simple hardware mod]([url](http://www.dc-swat.ru/blog/hardware/1146.html)). System calls are emulated by DreamShell. All devices supported by Dreamcast also should work on NAOMI. So we just get improved Dreamcast. This could popularize support for NAOMI in homebrew after I release BIOS and ISO Loader for NAOMI.

### DIFF
--- a/kernel/arch/dreamcast/hardware/hardware.c
+++ b/kernel/arch/dreamcast/hardware/hardware.c
@@ -19,16 +19,15 @@
 #include <dc/vblank.h>
 
 static int initted = 0;
+static uint32_t sysmode;
 
 #define SYSMODE_REG 0xA05F74B0
 
 int hardware_sys_mode(int *region) {
-    uint32 sm = *((vuint32 *)SYSMODE_REG);
-
     if(region)
-        *region = sm & 0x0F;
+        *region = sysmode & 0x0F;
 
-    return (sm >> 4) & 0x0F;
+    return (sysmode >> 4) & 0x0F;
 }
 
 int hardware_sys_init(void) {
@@ -38,6 +37,7 @@ int hardware_sys_init(void) {
     /* VBlank multiplexer */
     vblank_init();
 
+    sysmode = *((volatile uint32_t *)SYSMODE_REG);
     initted = 1;
 
     return 0;
--- a/kernel/arch/dreamcast/include/arch/arch.h
+++ b/kernel/arch/dreamcast/include/arch/arch.h
@@ -291,6 +291,8 @@ void hardware_shutdown(void);
                             -- otherwise, you must retrieve the region from the
                             flashrom.
     \return                 The console type (one of the \ref hw_consoles).
+
+    \note    Do not use before hardware_sys_init() has been called.
 */
 int hardware_sys_mode(int *region);
 


kernel/arch/dreamcast/include/dc/pvr/pvr_regs.h
```diff
@@ -175,7 +175,7 @@ __BEGIN_DECLS
 #define PVR_RAM_BASE        0xa5000000  /**< \brief VRAM 32-bit, P2 area, PVR->VRAM */
 #define PVR_RAM_INT_BASE    0xa4000000  /**< \brief VRAM 64-bit, P2 area, PVR->VRAM */
 
-#define PVR_RAM_SIZE_MB     (KOS_PLATFORM_IS_NAOMI ? 16 : 8)    /**< \brief RAM size in MiB */
+#define PVR_RAM_SIZE_MB     (hardware_sys_mode(NULL) ? 16 : 8)  /**< \brief RAM size in MiB */
```
pcercuei: This slightly bothers me, because then every time you reference `PVR_RAM_SIZE` it will make a function call and read a hardware register.

It would be better to initialize a "pvr_ram_size" variable at init and use that instead.

pcercuei: (Or maybe even better, cache the hardware sys mode into a variable)

DC-SWAT: Yeah, that bothered me too, but then I looked at the code and couldn't find anything that could have caused the negative effect. But you're right, it's better to be on the safe side in the function itself.

kernel/arch/dreamcast/hardware/hardware.c
```diff
@@ -19,16 +19,18 @@
...
 
 static int initted = 0;
+static uint32_t sysmode = (uint32_t)-1;
 
 #define SYSMODE_REG 0xA05F74B0
 
 int hardware_sys_mode(int *region) {
-    uint32 sm = *((vuint32 *)SYSMODE_REG);
+    if(sysmode == (uint32_t)-1)
+        sysmode = *((vuint32 *)SYSMODE_REG);
```
pcercuei: Move the initialization to `hardware_sys_init()` maybe?

kernel/arch/dreamcast/sound/snd_mem.c
```diff
@@ -94,7 +95,17 @@ int snd_mem_init(uint32 reserve) {
...
+    if(hardware_sys_mode(NULL) == HW_TYPE_RETAIL) {
+        blk->size = 2;
+    }
+    else {
+        blk->size = 8;
+    }
+
+    blk->size <<= 20;
+    blk->size -= reserve;
+
```
pcercuei, line 103: Why make it so complicated?
```c
if (hardware_sys_mode(NULL) == HW_TYPE_RETAIL)
    blk->size = 2 * 1024 * 1024 - reserve;
else
    blk->size = 8 * 1024 * 1024 - reserve;
```

kernel/arch/dreamcast/sound/snd_mem.c
```diff
@@ -137,7 +148,7 @@ void snd_mem_shutdown(void) {
 /* Allocate a chunk of SPU RAM; we will return an offset into SPU RAM. */
 uint32 snd_mem_malloc(size_t size) {
     snd_block_t *e, *best = NULL;
-    size_t best_size = 4 * 1024 * 1024;
+    size_t best_size = 8 * 1024 * 1024;
```
pcercuei: So I went to read the code to understand why you were doing this. This change is OK, but the new value makes it look like it actually means something - while in practice it's just an impossibly high value for a block size, so I'd advise to use `SIZE_MAX`, just so that it's clear.

kernel/arch/dreamcast/include/dc/pvr/pvr_regs.h
pcercuei: Slight preference for `hardware_sys_mode(NULL) == HW_TYPE_RETAIL ? 8 : 16`
Otherwise you expect `HW_TYPE_RETAIL` to always be 0, which might change (even accidentally) in the future.

kernel/arch/dreamcast/hardware/hardware.c
```diff
@@ -19,16 +19,15 @@
 #include <dc/vblank.h>
 
 static int initted = 0;
+static uint32_t sysmode = (uint32_t)-1;
```
pcercuei: No need to set it to -1.

pcercuei (changes requested):
Overall, looks good, with just a few minor comments, all related to the code style and not the functionality.

kernel/arch/dreamcast/hardware/hardware.c
DC-SWAT: If you use hardware_sys_mode() in the EarlyInit function, you will get undefined behavior.

DC-SWAT: And this isn't much better, because we still can't get the value. The first option was more flexible...

PR #1243, opened 2025-11-20
sd: Improved robustness and data writing.

DC-SWAT:
Updated the SPI read and write data functions to return status codes. This change improves error handling in the SD card read and write operations. Additionally, introduced a reinitialization mechanism for the SD card in case of read/write failures, enhancing robustness. Also impoved data writing for SCI interface by changing per-byte operations to per-sector.

### DIFF
--- a/kernel/arch/dreamcast/hardware/sd.c
+++ b/kernel/arch/dreamcast/hardware/sd.c
@@ -40,8 +40,8 @@ static uint8_t (*spi_rw_byte)(uint8_t data) = NULL;
 static void (*spi_set_cs)(bool enabled) = NULL;
 static int (*spi_init)(bool fast) = NULL;
 static void (*spi_shutdown)(void) = NULL;
-static void (*spi_read_data)(uint8_t *data, size_t len) = NULL;
-static void (*spi_write_data)(const uint8_t *data, size_t len) = NULL;
+static int (*spi_read_data)(uint8_t *data, size_t len) = NULL;
+static int (*spi_write_data)(const uint8_t *data, size_t len) = NULL;
 static uint8_t (*spi_read_byte)(void) = NULL;
 static void (*spi_write_byte)(uint8_t data) = NULL;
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/hardware/sd.c
```diff
@@ -454,8 +477,11 @@ static int read_data(size_t bytes, uint8 *buf) {
     }
 }
 
-int sd_read_blocks(uint32 block, size_t count, uint8 *buf) {
-    int rv = 0;
+int sd_read_blocks(uint32_t block, size_t count, uint8_t *buf) {
```
QuzarDC: sd.h should be updated to have the prototype still match with these stdint types, or just not change them from the original types. Either way would be fine.

QuzarDC (changes requested):
Looks good to me, aside from the one noted type update. Also, would there be any merit in having a flexible number of retries? I'm not sure what scenarios would cause the need to retry, something like the reader/card becoming unseated and de-initted between operations?

DC-SWAT:
> Looks good to me, aside from the one noted type update. Also, would there be any merit in having a flexible number of retries? I'm not sure what scenarios would cause the need to retry, something like the reader/card becoming unseated and de-initted between operations?

If a single reinitialization wasn't enough, that means only a power reset will help, which we can't do. I've already discovered this experimentally, so there's no point in trying more in a row.

QuzarDC:
> > Looks good to me, aside from the one noted type update. Also, would there be any merit in having a flexible number of retries? I'm not sure what scenarios would cause the need to retry, something like the reader/card becoming unseated and de-initted between operations?
> 
> If a single reinitialization wasn't enough, that means only a power reset will help, which we can't do. I've already discovered this experimentally, so there's no point in trying more in a row.

Ok, great. Any idea though what makes them need the reinit, or is that still not clear and it just happens sometimes?

PR #1244, opened 2025-11-23
Add thread-level polling functionality

pcercuei:
This PR adds the function `thd_poll()`, which can be used to poll ie. a hardware state change. The difference vs. the usual loop + thd_pass() is that the polling is done by the scheduler, which saves the back-and-forth between the scheduler and the thread.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_iso9660.c
+++ b/kernel/arch/dreamcast/fs/fs_iso9660.c
@@ -693,6 +693,10 @@ static int iso_close(void * h) {
     return 0;
 }
 
+static int iso_stream_done(size_t *remain_size) {
+    return cdrom_stream_progress(remain_size) != 1;
+}
+
 /* Read from a file */
 static ssize_t iso_read(void * h, void *buf, size_t bytes) {
     int rv, c;
@@ -802,9 +806,7 @@ static ssize_t iso_read(void * h, void *buf, size_t bytes) {
                 goto read_error;
             }
 
-            while(cdrom_stream_progress(&remain_size) == 1) {
-                thd_pass();
-            }
+            thd_poll((thd_cb_t)iso_stream_done, &remain_size, 0);
 
             memcpy(outbuf, fd->stream_data, toread);
             fd->stream_part = toread & 31;
--- a/kernel/arch/dreamcast/hardware/biosfont.c
+++ b/kernel/arch/dreamcast/hardware/biosfont.c
@@ -82,10 +82,15 @@ uint32_t bfont_set_background_color(uint32_t c) {
     return rv;
 }
 
+static int bfont_lock(void *d) {
+    (void)d;
+
+    return syscall_font_lock() == 0;
+}
+
 int lock_bfont(void) {
     /* Just make sure no outside system took the lock */
-    while(syscall_font_lock() != 0)
-        thd_pass();
+    thd_poll(bfont_lock, NULL, 0);
 
     return 0;
 }
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -85,22 +85,11 @@ int cdrom_set_sector_size(int size) {
 }
 
 static int cdrom_poll(void *d, uint32_t timeout, int (*cb)(void *)) {
-    uint64_t start_time;
     int ret;
 
-    if(timeout)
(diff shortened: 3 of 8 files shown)


QuzarDC:
> > Though that's all kinds of wild conjecture since we've had a variety of different discussions of how a dynticks would work.
> 
> I've changed my mind a few times about how it should work, so the term "dynticks" have meant a few different things over time (at least for me).
> 
> My current (and hopefully final) idea is that the current thread will always run for a time slot multiple of the 1/HZ value. The multiplier would be 1 when multiple threads are running, and otherwise, would be computed based on the soonest expiration time in the genwait list. If no threads are waiting, the sole running thread would get infinite time.
> 
> The other side of the mechanism would treat the idle thead differently, which would be the only one to get a variable time slot, equal to the soonest expiration time in the genwait list; so that we don't give it unnecessarily more time than we need.

For what it's worth, it seems as though the changes for the idle thread taking a more minimal amount of time could be done independently of the more generalized dynamic scheduler. I'm not sure that the different methodologies that would be applied to the general case would really change what the best case for idle is.

kernel/thread/thread.c
```diff
@@ -226,7 +241,10 @@ static void *thd_idle_task(void *param) {
     (void)param;
 
     for(;;) {
-        arch_sleep();   /* We can safely enter sleep mode here */
+        if(thd_has_polls())
```
QuzarDC, line 244: Why drop out into the idle thread at all? From the scheduler we should know before scheduling the idle thread if we have pollers and could just continue polling them. Is it just to dip out and back into an interrupt context?

kernel/thread/thread.c
```diff
@@ -871,6 +910,35 @@ int thd_detach(kthread_t *thd) {
...
+    kthread_t *thd = thd_current;
+    int ret;
+
+    assert(cb != NULL);
+    assert(!irq_inside_int());
+
+    irq_disable_scoped();
+
+    /* Try to poll once, just in case */
+    ret = cb(data);
```
QuzarDC, line 925: Is there any difference to be had that this would be the only time the callback is called outside of an irq context? I can't think of where it would make a difference, but just in case. I presume that disabling the irqs ahead of it was intended to help make it function more similarly to how it would when called in the scheduler and that certainly narrows the differences.

QuzarDC (commented):
Some feedback I'd just entirely forgotten that I'd not added here. Otherwise though was still a bit concerned about the discussion here:

https://discord.com/channels/488332893324836864/614912396460556298/1454078200253644952

Where using this poll mechanism counterintuitively degraded performance for the W5500 driver. Since that driver got added in, this PR could be updated to incorporate to there and then tested with it further to see if it's a peculiarity of the driver (especially if one that we can document and possibly need to adjust some other cases to not use poll to avoid the same) or if there's something that needs to be changed in the core here.

kernel/thread/thread.c
pcercuei: Yes, that's it - if we just continue polling them in the scheduler, we'd stay in interrupt context for way too long which is never a good idea, and also risk a deadlock, as some polls may be waiting on an hardware IRQ to complete.

kernel/thread/thread.c
pcercuei: I don't think it would matter much that it's called outside an IRQ context.

The IRQ disable is needed after the callback anyway, but I disable before the callback just to make it consistent, because then the callback function can assume that it will never run with IRQs enabled.

pcercuei:
I've been testing with the following (rather extreme) test program: [test_thd_poll.c](https://github.com/user-attachments/files/24525792/test_thd_poll.c)

It creates a configurable number of threads, each one polling for one particular value of a counter, then updating it.
When `TEST_THD_POLL` is set to 0, each thread will poll once then call `thd_pass()` if the poll function failed. When set to 1, it will use `thd_poll()` instead.

Without TEST_THD_POLL:
`Test completed in 10894955 us.`

With TEST_THD_POLL:
`Test completed in 209441 us.`

For this particular example, it is a 52x speedup.

kernel/thread/thread.c
QuzarDC: Ok, I guess it could be a future kind of thing (with the support for nested IRQs as in #618) to have such a condition just reenable IRQs rather than have the rest of the overhead.

QuzarDC (approved):

PR #1256, opened 2025-12-15
Minor macro and build cleanups

QuzarDC:
Mostly around using and standardizing cdefs macros.

- Removed duplicated DECLS in init.h
- Use `alignas` where possible over gcc attributes.
- Use `_Static_Assert` rather than faked one.
- Use `__is_aligned` and `__align_up` macros.
- Use our regfield `BIT` macro.

and two small cleanups for build:
- Remove duplicated `rm` commands from `dreamcast/kernel/Makefile`
- Replace reference to AR env var with our `kos-ar` wrapper

### DIFF
--- a/kernel/arch/dreamcast/include/dc/flashrom.h
+++ b/kernel/arch/dreamcast/include/dc/flashrom.h
@@ -27,6 +27,7 @@
 __BEGIN_DECLS
 
 #include <arch/types.h>
+#include <kos/regfield.h>
 
 /** \defgroup flashrom  Flashrom
     \brief              Driver for the Dreamcast's Internal Flash Storage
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
```diff
@@ -272,23 +268,23 @@ void pvr_allocate_buffers(const pvr_init_params_t *params) {
...
+        outaddr = __align_up(outaddr, 128);
 
         /* Output buffer */
         fbuf->frame = outaddr;
         fbuf->frame_size = pvr_state.w * pvr_state.h * vid_pmode_bpp[vid_mode->pm];
         outaddr += fbuf->frame_size;
 
         /* N-byte align */
-        outaddr = APPLY_ALIGNMENT(outaddr);
+        outaddr = __align_up(outaddr, 128);
```
pcercuei, line 287: Nitpick, but the alignment size is now distributed across calls instead of being defined at a single place. I don't think much of it, thought I'd mention it nonetheless.

pcercuei (approved):

PR #1257, opened 2025-12-15
Misc fixes + arch-specific exports

pcercuei:
A few misc cleanups, and a factorization on the exports lists by moving the symbols shared between the "pristine" and "naomi" platforms into a new arch-specific list.

### DIFF
--- a/utils/ldscripts/shlelf.xc
+++ b/utils/ldscripts/shlelf.xc
@@ -8,6 +8,7 @@ INPUT(crtbegin.o)
 INPUT(crtend.o)
 INPUT(crtn.o)
 LOAD_OFFSET = DEFINED(LOAD_OFFSET) ? LOAD_OFFSET : 0x8c010000 ;
+ICACHE_SIZE = 0x2000 ;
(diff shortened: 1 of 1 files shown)


kernel/exports/exports.c
```diff
@@ -44,10 +44,23 @@ static symtab_handler_t st_arch = {
...
+        0x00010000,
+        0,
+        NMMGR_TYPE_SYMTAB,
+        NMMGR_LIST_INIT
+    },
+    subarch_symtab
+};
+
 void export_init(void) {
     /* Add our two export tables */
```
QuzarDC, line 60: *three (not a required change, but I had to mention once I noticed)

DC-SWAT:
> All looks fine, but would like to see if @DC-SWAT has any feedback on the exports change. That seems like it would require changes on the user side, but trivial ones, and makes sense for factorization.

I don't see that I'll have to change anything on user side. And these changes are truly necessary.

QuzarDC (approved):

PR #1258, opened 2025-12-16
Maple: fix rumble

pcercuei:
One cleanup commit + one fix for the rumble.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_irq.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_irq.c
@@ -131,14 +131,8 @@ static void vbl_autodet_callback(maple_state_t *state, maple_frame_t *frm) {
     if(resp->response == MAPLE_RESPONSE_NONE) {
         /* No device, or not functioning properly; check for removal */
         if(u == 0) {
-            if(dev) {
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/hardware/maple/maple_queue.c
```diff
@@ -173,7 +172,7 @@ void maple_frame_init(maple_frame_t *frame) {
     frame->length = 0;
     frame->queued = 0;
     frame->dev = NULL;
-    frame->send_buf = NULL;
+    frame->send_buf = (uint32_t *)frame->recv_buf;
```
dfchil, line 175: This is probably fine, but goes against all the other lines in this PR  where lines like ```c
send_buf = (uint32_t *)dev->frame.recv_buf;``` are removed. I probably just don't get it because I'm insufficiently versed in the inner workings of the maple driver. 

So I'm just adding this comment to make you double check that this is what you want.

pcercuei: The frame's `send_buf` has to be set somewhere. So it is now set in the `maple_frame_init()`.

PR #1262, opened 2025-12-21
Tls cleanup (no more spinlocks)

QuzarDC:
- Add a check for possible malloc use in IRQ that was missing in key setting.
- Update check for possible malloc use in IRQ for key creation to only apply if malloc will be needed.
- Use `LIST_FOREACH_SAFE` rather than replicating the macro by pieces.
- Move `kthread_key_delete` from thread.c to tls.c.
- Replace spinlock with an atomic to protect the key counter and a mutex to protect the destructor list.
- Clean up and update documentation, reduce unnecessary exposure of internal API.

### DIFF
--- a/include/kos/tls.h
+++ b/include/kos/tls.h
@@ -117,11 +117,6 @@ int kthread_setspecific(kthread_key_t key, const void *value);
 */
 int kthread_key_delete(kthread_key_t key);
 
-/** \cond */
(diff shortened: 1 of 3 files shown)


pcercuei (dismissed):
LGTM, thanks.

pcercuei (dismissed):

pcercuei (approved):

PR #1266, opened 2025-12-23
CI: Add no-tabulation check

pcercuei:
Add a check that will make sure that tabulations characters are not added in the source code.


QuzarDC:
Good call on the new label and the idea of this generally. Any chance it could be easily made to exclude Makefiles since we expect those to have tabs?

QuzarDC (approved):

PR #1270, opened 2025-12-24
More newlib fixes

QuzarDC:
Followup to #1208 and trying to get our interface with newlib cleaned up (old patches removed, place implementation-specific stuff in the right files, etc):

- Remove `<kos/stdlib.h>` - This was added in #268 in order to provide `posix_memalign` by forcing it into the environment via `<sys/_types.h>`. This is no longer needed as newlib provides the function prototype now in its main `<stdlib.h>` without our patch. 
- Remove custom `<libgen.h>` - This was [added in 2020 (pre-PR)](https://github.com/KallistiOS/KallistiOS/commit/e8e63c1c37018b83336704dc9e406dc81d1a6b3d) in order to provide `dirname()` and `basename()`. The newlib version of the header now provides those and with better handling of overlapping with the string.h version of basename and for language standards levels.
- Remove custom `<malloc.h>` - This predates our use of newlib. Instead of providing it and overriding the newlib version, add a `<machine/malloc.h>` which newlib includes and provides access to our custom extensions in the same way that they were with the custom one.
- Remove custom `<threads.h>` - Like the malloc change, move implementation dependent components to `<machine/_threads.h>` to let newlib handle the standard parts.
- Reduce duplicated definitions from `<sys/dirent.h>` - This file is meant to only provide the implementation-specific pieces, and shouldn't be referenced directly. Changed all references to it in the codebase to point to the correct `<dirent.h>` and removed the definitions from the sys version that already existed in the newlib top-level include. Made sure to also add some comments noting what we're leaving out.

### DIFF
--- a/include/sys/dirent.h
+++ b/include/sys/dirent.h
@@ -103,147 +103,21 @@ typedef struct {
     };
 } DIR;
 
-// Standard UNIX dir functions. Not all of these are fully functional
-// right now due to lack of support in KOS.
+/* The following functions that will be defined in <dirent.h> are not currently
+    implemented in KOS:
 
-/** \brief  Open a directory based on the specified name.
-
-    The directory specified is opened if it exists. A directory stream object is
-    returned for accessing the entries of the directory.
-
-    \note               As with other functions for opening files on the VFS,
-                        relative paths are permitted for the name parameter of
-                        this function.
-
-    \param  name        The name of the directory to open.
-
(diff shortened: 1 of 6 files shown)


QuzarDC:
Had accidentally reintroduced some tab characters in `sys/dirent.h` so updated the whitespace in that commit to get them back out. Thanks Tabulation CI check!

pcercuei (approved):
LGTM.

PR #1274, opened 2025-12-27
atomics: Ensure builtins unavailable before declaring.

QuzarDC:
On non-sh4 based platforms, these may be available, and defining them unconditionally would lead to a conflict. This was found by trying to build for x86-64 on the [`null_arch`](https://github.com/KallistiOS/KallistiOS/commits/null_arch/) branch

@darcagn not sure how these have worked out for PPC so far. It seems though that they can normally be overridden without gcc complaining, and I only noticed it was happening because the prototypes were different for x86-64 and gave a warning. It may be the case that the whole file should be gated or moved over into the dc arch.

@gyrovorbis do you happen to know on the dc how these interact with a non soft-imask model? It *might* be that these are overriding or doubling up with the gusa system when that's being used, both adding the gusa guardrails *and* disabling irqs by calling these. I think we missed raising the question when we switched. Hopefully it's not a problem but would be good to be sure and at least maybe update the comments in this atomics file too.

Edit: Addressed in #1314

### DIFF
--- a/kernel/libc/c11/atomics.c
+++ b/kernel/libc/c11/atomics.c
@@ -92,17 +92,39 @@
         return ret;  \
     }
 
-/* GCC provides us with all primitive atomics except for 64-bit types. */
-ATOMIC_LOAD_N_(unsigned long long, 8)
-ATOMIC_STORE_N_(unsigned long long, 8)
-ATOMIC_EXCHANGE_N_(unsigned long long, 8)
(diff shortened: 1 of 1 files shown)


darcagn:
I hadn't noticed anything adverse from these on PPC, but I agree it might make sense to move to DC. When I rebase the GC port against the latest KOS soon I'll investigate further. Should be in a week or so.

PR #1277, opened 2025-12-28
irq: Fix C++ compatibility

pcercuei:
Add two fixes for C++ compatibility with the new IRQ code.

PR #1278, opened 2025-12-29
irq: C++ compatibility and fixes

QuzarDC:
My take on #1277 .

If there's a better middle ground, I'm open to it. What I was thinking to work on from here was to reduce the surface area that either irq.h has on the API. It's currently being included in a lot of places where it doesn't need to be for relatively dubious or arcane reasons. It isn't directly used by the toolchain, for example but we keep on having to jump through hoops for it because it gets into the C++ toolchain through the various other primitives headers (thread, mutex, cond).

Edit: Via #1279 I did already reduce the surface area of irq.h and at the very least this should no longer cause problems for the toolchain. However, the check on the enum is still false in that the functionally simply isn't supported pre-C++11. While we have been moving our own code forward our public headers should at least strive for as much compatibility as possible to facilitate interoperability with other libraries.

### DIFF
--- a/include/kos/irq.h
+++ b/include/kos/irq.h
@@ -26,14 +26,7 @@ __BEGIN_DECLS
 #include <stdbool.h>
 #include <stdint.h>
 
-/** \cond INTERNAL */
-struct irq_context;
-#ifdef __cplusplus
-enum irq_exception: unsigned int;
-#else
-enum irq_exception;
-#endif
-/** \endcond */
+#include <arch/irq.h>
 
 /** Architecture-specific structure for holding the processor state.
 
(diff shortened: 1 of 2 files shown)


QuzarDC:
The CI failures here don't seem related to the changes. I was able to build the stable toolchain in an alpine VM as well as 13.4.1-dev and 15.2.0 on debian WSL2 without issue. Might have to do with the update process? Or that I added an extra commit after they kicked off? Not sure.

addons/libpthread/pthread_getsetconcurrency.c
```diff
@@ -7,23 +7,20 @@
 #include "pthread-internal.h"
 #include <pthread.h>
 #include <errno.h>
-#include <arch/irq.h>
+#include <stdatomic.h>
 
-static int pth_concurrency = 0;
+static atomic_int pth_concurrency = 0;
```
pcercuei: You can keep it `int`, the atomic types are only atomic when used with the corresponding atomic API functions.

QuzarDC: Ah, of course

MastaG:
Would it be possible for this PR to get some traction. We're waiting for this in order to bump the KOS rev for DreamShell :)

PR #1279, opened 2025-12-30
Reduce toolchain interface

QuzarDC:
This is independent of #1278 but may make it less necessary and help reduce further issues in the future that are related to the same.

The goal here is to ensure that the minimum amount is included/exposed from the kos side in the toolchain patch, with special focus on avoiding `irq.h` from leaking out to there. The previous changes for this same thing ended up being limited solely to C due to the need to have headers for mutex, cond and others available in full for C++. This had not been a problem for thread.h but it was being included in the other headers (and thus including irq.h with it). So this update does the following:

- Update includes across the codebase that were implicitly getting thread or irq functions from mutex.h or cond.h
- Adjust mutex.h, cond.h, and rwsem.h to not include thread.h 
- Adjust the toolchain patch to always include the (now) irq-less mutex.h and cond.h as well as the other headers that had not been specifically contributing to the issue.

In doing so, now only the cond, mutex, once, and tls headers need to be compliant with all c++ versions.

### DIFF
--- a/utils/dc-chain/patches/gcc/gthr-kos.h
+++ b/utils/dc-chain/patches/gcc/gthr-kos.h
@@ -40,78 +40,40 @@ __BEGIN_DECLS
 #define __GTHREADS_CXX0X 1
 #define __GTHREAD_HAS_COND 1
 
-#if defined(TEST_GTHR_KOS_API) || defined(__cplusplus)
+/* Include thread to test for discrepancies with our prototypes */
+#if defined(TEST_GTHR_KOS_API)
 #include <kos/thread.h>
-#include <kos/tls.h>
-#include <kos/mutex.h>
-#include <kos/once.h>
-#include <kos/cond.h>
-#include <arch/irq.h>
 #endif
 
-/* 9.5.0 somehow requires this. Remove when no longer supported */
-#include <pthread.h>
-#include <stdbool.h>
-#include <time.h>
-
-#if !defined(__cplusplus)
-/* KOS types */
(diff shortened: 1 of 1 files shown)


pcercuei:
The CI fails, I think independently from your PR:
```
 ../../gcc-13.2.0/intl/dcigettext.c:549:17: error: too many arguments to function 'getcwd'; expected 0, have 2
   549 |           ret = getcwd (dirname, path_max);
       |                 ^~~~~~  ~~~~~~~
 ../../gcc-13.2.0/intl/dcigettext.c:151:7: note: declared here
   151 | char *getcwd ();
       |       ^~~~~~
```

This is probably a consequence of this: https://gcc.gnu.org/gcc-15/porting_to.html#c23-fn-decls-without-parameters

utils/dc-chain/patches/gcc/gthr-kos.h
```diff
@@ -40,78 +40,40 @@ __BEGIN_DECLS
...
 #endif
 
-/* 9.5.0 somehow requires this. Remove when no longer supported */
-#include <pthread.h>
-#include <stdbool.h>
-#include <time.h>
-
-#if !defined(__cplusplus)
-/* KOS types */
+/* Forwrd-declare needed types and functions from thread.h
```
pcercuei: Forward*

pcercuei (changes requested):
Minor typo.

pcercuei (approved):

Issue #886, opened 2025-01-07
CMake toolchain simplification breaks CLion static analysis


michael-fadely:
It would appear that in #501, things such as [the removal of `include_directories`](https://github.com/KallistiOS/KallistiOS/pull/501/files#diff-cc0ced613b1f1f7266a5352c9fd54ee0af7b68629d71ad30086494904e370572L107) from `dreamcast.toolchain.cmake` breaks static analysis in CLion. It also requires the entire IDE to be run with all of the KOS environment variables set (i.e. `source environ.sh`), which was not previously necessary. This new requirement completely breaks my WSL workflow.

If you're running on Linux directly and set all the variables before running CLion, CMake configuration and build will work, but static analysis will fail, particularly for includes like `kos.h`.

![image](https://github.com/user-attachments/assets/42eb8a4b-5e4e-48fa-bf00-838abd9678c4)

It's super easy to reproduce. Here's a little test case:
`CMakeLists.txt`:
```cmake
cmake_minimum_required(VERSION 3.28)

project(cmake-testbed)

add_executable(cmake-testbed "source/main.cpp")

target_include_directories(cmake-testbed INTERFACE "${KOS_PORTS}")
```

`source/main.cpp`:
```cpp
#include <kos.h>

int main(int argc, char** argv)
{
	return 0;
}
```

As a workaround I'm sure I could export `compile_commands.json`, but this is much less desirable than CMake integration. Before these changes, I was able to get by using CLion's CMake environment variable configuration to set just a few of the critical ones--`KOS_BASE`, `KOS_CC_BASE`, `KOS_PORTS`, and `KOS_SUBARCH`--and the CMake toolchain would handle the rest.

michael-fadely:
> As a workaround I'm sure I could export `compile_commands.json`
Scratch that. Because of the build wrapper, the flags within the script are obfuscated and don't get included in the output.

pcercuei:
> It also requires the entire IDE to be run with all of the KOS environment variables set (i.e. `source environ.sh`), which was not previously necessary. This new requirement completely breaks my WSL workflow, as it's not practical to set environment variables in WSL and expect them to stick when running a Windows program, particularly for something like an IDE.

Sorry but that's a bogus argument. You want to develop using a Windows IDE, use DreamSDK.

michael-fadely:
[...]

PR #1283, opened 2026-01-01
Expand @-style parameters to fix CLion compatibility (#886)

kouta-kun:
This commit fixes issue #886.

When CLion processes a CMake file, it executes CMAKE_C_COMPILER and CMAKE_CXX_COMPILER with `-xc++ --sysroot=/opt/toolchains/dc/kos-ports -g -std=gnu++20 -DFRAME_POINTERS -fno-omit-frame-pointer -fpch-preprocess -v -dD -E -D___CIDR_DEFINITIONS_END`, but it passes these arguments through a file in /tmp using @-parameter (https://gcc.gnu.org/onlinedocs/gcc/Overall-Options.html#Overall-Options). Since the kos-cc wrapper was not expanding this file, it did not correctly detect it and thus did not include the import directories in the parameter list.

Expanding these kinds of parameters allows CLion to correctly detect the include directories configured in the CMakeLists both of the project itself and the toolchain.

### DIFF
--- a/utils/build_wrappers/kos-cc
+++ b/utils/build_wrappers/kos-cc
@@ -5,6 +5,13 @@ ARGS=$*
 USEMODE=0
 for i in $ARGS; do
 	case "${i}" in
+		@*)
+			ARGS="${ARGS} $(cat ${i:1})"
+		;;
+	esac
+done
+for i in $ARGS; do
+	case "${i}" in
 		-o)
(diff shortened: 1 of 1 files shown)


QuzarDC (approved):
Thanks for the contribution! Glad we could finally get this resolved and hopefully it'll improve compatibility for other usages as well.

PR #1284, opened 2026-01-01
User configurable dbgio interfaces

darcagn:
- Change dbgio interface list from static array to singly linked list
- Create dbgio_add_handler() and dbgio_remove_handler() functions to add and remove from the dbgio handler list
- Create dbgio_set_first() function to set dbgio handler to the first available in list
- Make dbgio_init() a weak symbol that calls dbgio_set_first(). Users can override this weak symbol to inject their own dbgio handling code early in kernel initialization.
- Add conio_dbgio example to demonstrate user configurable dbgio interfaces

This PR basically allows users to integrate KOS debug output with their application's logger. Let's say the user is building an FPS game with a console that can be opened with `~`. With this PR, the KOS user can set up a dbgio handler so that KOS debug output can be redirected to this console. And if `dbglog()` is used with their own application's debug output, they can use the dbglog levels to fine tune their own output as well.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -42,7 +42,7 @@ Josh Pearson: 2013, 2014, 2015, 2016
 Joe Fenton: 2016
 Stefan Galowicz: 2016, 2017
 Luke Benstead: 2020, 2021, 2022, 2023
-Eric Fradella: 2023, 2024, 2025
+Eric Fradella: 2023, 2024, 2025, 2026
 Falco Girgis: 2023, 2024, 2025
 Ruslan Rostovtsev: 2014, 2016, 2023, 2024, 2025
 Colton Pawielski: 2023


include/kos/dbgio.h
```diff
@@ -102,17 +105,49 @@ typedef struct dbgio_handler {
...
-/* These two should be initialized in arch. */
-extern dbgio_handler_t *dbgio_handlers[];
-extern const size_t dbgio_handler_cnt;
+/** \brief   List entry initializer for static structs.
+    \ingroup logging
 
+    If you are creating dbgio handlers, this is what you should initialize the
+    entry member with.
+*/
+#define DBGIO_LIST_INIT { NULL }
```
pcercuei: I think this is not needed, since you are using `SLIST_INSERT_HEAD()`, the entry member will be initialized there.

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -174,6 +164,13 @@ int  __weak_symbol arch_auto_init(void) {
     /* Init SCIF for debug stuff (maybe) */
     scif_init();
 
+    /* Add dbgio handlers for our arch, from last to first in precedence */
+    dbgio_add_handler(&dbgio_fb);
+    dbgio_add_handler(&dbgio_null);
+    dbgio_add_handler(&dbgio_scif);
+    dbgio_add_handler(&dbgio_dcls);
+    dbgio_add_handler(&dbgio_dcload);
```
pcercuei, line 172: I want to say keep the `dbgio_handlers` array, make it static, and iterate over it to call `dbgio_add_handler` in a loop in reverse order.

kernel/arch/dreamcast/fs/fs_dcload.c
```diff
@@ -424,7 +424,8 @@ dbgio_handler_t dbgio_dcload = {
     NULL,
     NULL,
     NULL,
-    NULL
+    NULL,
+    DBGIO_LIST_INIT
```
pcercuei: Maybe it'd be a good idea to switch to designated initializers, so that fields that should be NULL can simply be omitted...

include/kos/dbgio.h
```diff
@@ -129,6 +164,20 @@ extern dbgio_handler_t dbgio_null;
...
+    This function selects the first valid dbgio interface in the list.
+
+    \retval 0               On success
+
+    \retval -1              On error
+
+    \par    Error Conditions:
+    \em     ENODEV - No devices could be detected/initialized
+*/
+int dbgio_dev_set_first(void);
```
QuzarDC: Mind if this is something more like `dbgio_dev_select_auto`? My thinking is that handlers like `fb` could be set to `auto = false` so that they can't be automatically found by this selection system. And it would leave it more flexible if we want to have a different selection/prioritization process since the current selection process isn't transparent to the user (or guaranteed).

Either that or maybe just not expose it? Is the idea that someone would set their own init, install some, then call this? Or call it on context switches to find which have been enabled/disabled? Because if they install some they should be able to also just know the specific one to enable. The same would be the case for something like redirecting it.

kernel/arch/dreamcast/kernel/init.c
QuzarDC: Why have the array at all though if you've got the list? In the followup I was describing in discord these would each be moved into the init functions for the io subsystem so then at least it wouldn't *look* like it was still just a fixed array.

include/kos/dbgio.h
darcagn: If I omit the entry member, GCC complains that the initializers are incomplete.
If I put just `NULL` there, then it complains about the missing braces.
So here I just copied what we're doing with the nmmgr (`#define NMMGR_LIST_INIT { NULL }`).

But, I will switch to designated initializers for all the dbgio handlers which allows me to remove this.

QuzarDC: Oh right, underlying it is a struct, so it would need to have the extra parens.

kernel/arch/dreamcast/fs/fs_dcload.c
darcagn: Done.

kernel/arch/dreamcast/kernel/init.c
darcagn: > Why have the array at all though if you've got the list? In the followup I was describing in discord these would each be moved into the init functions for the io subsystem so then at least it wouldn't _look_ like it was still just a fixed array.

It might look better since it would allow us to write them in forward order then loop in reverse order. But I think since we're going to change this anyway, I might as well leave as it is here.

include/kos/dbgio.h
darcagn: The idea was initially to decouple the "set first" behavior from "init" (since this was all that init did anyway). But I can see usefulness in exposing the behavior to the end user. For example, if multiple dbgio handlers are installed by the end user, some of which may not be "detected" and some which may, depending on particular conditions, this would allow KOS to find the right one to select.

I have no problem naming it `dbgio_dev_select_auto` though, so that's what I've done. 👍🏼

kernel/arch/dreamcast/fs/fs_dcload.c
```diff
@@ -415,16 +415,16 @@ static int never_detected(void) {
...
+    .name = "fs_dcload_uninit",
+    .detected = never_detected,
+    .init = NULL,
+    .shutdown = NULL,
+    .set_irq_usage = NULL,
+    .read = NULL,
+    .write = NULL,
+    .flush = NULL,
+    .write_buffer = NULL,
+    .read_buffer = NULL
```
pcercuei: No need to initialize fields to NULL, that's the default.

darcagn: Good point, I wasn't even thinking about it there.

include/kos/dbgio.h
QuzarDC: Hrm, ok yea that tracks too. I kind of expected that even in those conditional situations they'd rather manage via testing against whether dev_select worked or not, since it would be fiddly to rely on the insertion order and that would only give a singular immutable (I guess you could remove/readd to shuffle priority, but that's even more fiddly) priority list.

QuzarDC:
> As of now, `dbgio_dev_select()` (by name) does not call `dbgio_enable()` when successful, while `dbgio_dev_select_auto()` does. I have just pushed a new commit to change this behavior so it does call `dbgio_enable()` which I believe to be the real intention.

Hrm. Dunno if that's a good idea. The enable/disable was otherwise decoupled from selecting the handler to use when it is enabled. Disable, for instance, doesn't de-select the selected handler. The enabling in the auto select was due to it being the initialization as well, so perhaps rather the weak init should be `dbgio_dev_select_auto(); dbgio_enable();` ?

This way would force user code to check if dbgio is enabled/disabled (which we actually don't have a way to do ...) before attempting to set a different handler for the context they're in, or maintain tracking on their side. Which, I guess isn't necessarily a bad thing, but it's just a tangible difference compared to the current behavior.

PR #1286, opened 2026-01-02
Improve mm code and make it generic

pcercuei:
The mm_sbrk() function will be implemented more or less the same across all architectures, and therefore can be moved to the arch-agnostic part of the kernel.

The algorithm was also rewritten to use atomics.

### DIFF
--- a/kernel/mm/mm.c
+++ b/kernel/mm/mm.c
@@ -16,11 +16,11 @@
 #include <arch/arch.h>
 #include <arch/stack.h>
 #include <kos/dbglog.h>
-#include <kos/irq.h>
 #include <kos/mm.h>
 #include <kos/linker.h>
 #include <errno.h>
 #include <inttypes.h>
+#include <stdatomic.h>
 #include <stdint.h>
 #include <stdio.h>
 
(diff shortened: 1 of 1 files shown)


include/kos/mm.h
```diff
@@ -0,0 +1,39 @@
...
+int mm_init(void);
+
+/** \brief   Request more core memory from the system.
+    \ingroup mm
+
+    \param  increment       The number of bytes requested.
+    \return                 A pointer to the memory.
+    \note                   This function will panic if no memory is available.
+*/
+void * mm_sbrk(unsigned long increment);
```
QuzarDC: Extra space after star could be removed. I think it'd make sense too while at it to swap this over to a `size_t`. That could allow dropping the cast in the newlib shim and be more idiomatic.

Otherwise the rest looks fine.

QuzarDC: It might be worth it to take a look at #592 in case you'd be interested in some of the extra API

pcercuei: Anything worth saving there?

QuzarDC: The thing I was imagining was that it might be worthwhile to have the API to access the otherwise internal values for `_arch_mem` and/or the moving of those out of the 'misc' file of `arch/arch.h` into something like `arch/mm.h` to centralize the arch side of this.

One other thing that was much more on the big picture and certainly not for this immediately would be to possibly have it made flexible to manage multiple memory pools. So that there's no need to reinvent the same core functionality for video or audio ram. `mm_sbrk_ex(increment, pool_id)` and such.

If none of it seems worthwhile, then no worries and just the whitespace and type change.

pcercuei: Updated with the whitespace change + prototype change.

Even if we could have multi-pool `sbrk`, our malloc implementation does not support it. We would benefit from switching to TLSF, which supports multi-pool, but does not use sbrk(), so in any case a mm_sbrk_ex() wouldn't serve any purpose.

Another interesting malloc implementation to look at is jemalloc, which supports online de-fragmentation.

QuzarDC: Oh, I didn't mean to indicate the multiple pools would be within the single malloc (which I see now is exactly how that would sound) merely that we'd be reusing the same code to do this sbrk operation for each of the pools. So instead of having `pvr_int_sbrk` duplicated as an independent function in `pvr_mem.c` we'd have `mm_sbrk_ex(amount, pvr_pool);`

None of that changes what needs to be in this PR though so I'll still just go ahead, but hopefully that makes more sense than how I'd described it before.

kernel/libc/newlib/newlib_sbrk.c
QuzarDC: The cast can now be dropped as `mm_sbrk` is going to take `size_t`

pcercuei: Thanks, fixed.

PR #1287, opened 2026-01-03
Update dcplib examples to use `pvr_filter_mode_t`.

QuzarDC:
In order to update our API to use this type, it must first be updated in C++ examples to prevent illegal int->enum conversions. This will be followed by a PR to update libdcplib then to pvr.h


QuzarDC:
Once this is ready, then https://github.com/KallistiOS/libdcplib/pull/4 can go.

pcercuei (approved):

PR #1288, opened 2026-01-03
`pvr.h`: Further retyping and cleanup

QuzarDC:
Yet another pvr cleanup split off from #539.

- Create `pvr_legacy.h` to be a repository for deprecated defines.
- Use stdint types throughout.
- Replace custom debug macro with new dbgio source function.
- Retype members of `pvr_*_ctx_t` types to use bool for 0/1 and move old defines to `pvr_legacy.h`
- Retype members of `pvr_*_ctx_t` types to use enums rather than define groups where possible.
- Consolidate the multiple usages of pvr list types and defines.

This should only go after #1287 https://github.com/KallistiOS/libparallax/pull/4 and https://github.com/KallistiOS/libdcplib/pull/4

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
@@ -142,19 +142,19 @@ void pvr_int_handler(uint32_t code, void *data) {
     // What kind of event did we get?
     switch(code) {
         case ASIC_EVT_PVR_OPAQUEDONE:
-            pvr_state.lists_transferred |= BIT(PVR_OPB_OP);
+            pvr_state.lists_transferred |= BIT(PVR_LIST_OP_POLY);
             break;
         case ASIC_EVT_PVR_TRANSDONE:
-            pvr_state.lists_transferred |= BIT(PVR_OPB_TP);
+            pvr_state.lists_transferred |= BIT(PVR_LIST_TR_POLY);
             break;
         case ASIC_EVT_PVR_OPAQUEMODDONE:
-            pvr_state.lists_transferred |= BIT(PVR_OPB_OM);
+            pvr_state.lists_transferred |= BIT(PVR_LIST_OP_MOD);
             break;
         case ASIC_EVT_PVR_TRANSMODDONE:
-            pvr_state.lists_transferred |= BIT(PVR_OPB_TM);
+            pvr_state.lists_transferred |= BIT(PVR_LIST_TR_MOD);
             break;
         case ASIC_EVT_PVR_PTDONE:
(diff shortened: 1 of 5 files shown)


kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
```diff
@@ -161,7 +159,7 @@ void pvr_int_handler(uint32_t code, void *data) {
             pvr_state.lists_transferred |= BIT(PVR_OPB_PT);
             break;
         case ASIC_EVT_PVR_RENDERDONE_TSP:
-            //DBG(("irq_renderdone\n"));
+            dbglog(DBG_SOURCE(PVR_RENDER_DBG), "pvr_irq: ASIC_EVT_PVR_RENDERDONE_TSP\n");
```
pcercuei: Actually scrap my ACK, you un-commented these and I think you should just drop them. If I want to debug the PVR driver by enabling PVR_RENDER_DBG I want to log the issues below, not the "list done" / "render done".

kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
```diff
@@ -171,33 +169,33 @@ void pvr_int_handler(uint32_t code, void *data) {
...
+            break;
+
+        case ASIC_EVT_PVR_OPB_OUTOFMEM:
+            dbglog(DBG_SOURCE(PVR_RENDER_DBG), "pvr_irq: ASIC_EVT_PVR_OPB_OUTOFMEM\n");
+            dbglog(DBG_SOURCE(PVR_RENDER_DBG), "pvr_irq: PVR_TA_OPB_START: %08lx\n",
+                PVR_GET(PVR_TA_OPB_START));
+            dbglog(DBG_SOURCE(PVR_RENDER_DBG), "pvr_irq: PVR_TA_OPB_END: %08lx\n",
+                PVR_GET(PVR_TA_OPB_END));
+            dbglog(DBG_SOURCE(PVR_RENDER_DBG), "pvr_irq: PVR_TA_OPB_POS: %08lx\n",
+                PVR_GET(PVR_TA_OPB_POS) << 2);
```
pcercuei: You could also factorize that to a single dbglog() call.

kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
QuzarDC: Ok, that makes sense. Updated to just drop them.

kernel/arch/dreamcast/hardware/pvr/pvr_irq.c
QuzarDC: Good call, I'd just gathered the existing lines and moved them. Done.

PR #1290, opened 2026-01-04
Maple cleanups

QuzarDC:
Some really trivial bits:
- Use correct initialization macro for maple_state driver list rather than just letting it be 0 initted.
- Don't look into maple driver list internals to check for double adding, better checks already exist.
- Drop dreameye attach callback which just cleared individual members of `status` which is cleared by the global attach.
- Fix typo, some whitespace, and stdints.

Then one more involved changes:
- Drop `status_valid` private matter from `maple_state_t`. It does not serve a purpose as:
     - `status` is now calloc'd so there isn't a concern about getting random garbage data from it.
     - It was often being set on attach (not on first poll as it was advertising) which is the same as it always being set, as that is all done during the initial scan/attach irq so there's no race against `maple_dev_status` where it was being tested.
     - Where it's expected to contain specific data from the start (like SIP and Dreameye) that already gets done during attach, so it will always be set by the time `maple_dev_status` can check it.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/dreameye.c
+++ b/kernel/arch/dreamcast/hardware/maple/dreameye.c
@@ -383,29 +383,11 @@ int dreameye_erase_image(maple_device_t *dev, uint8_t image, int block) {
     return MAPLE_EOK;
 }
 
-static int dreameye_attach(maple_driver_t *drv, maple_device_t *dev) {
-    dreameye_state_t *de;
-
-    (void)drv;
-
-    de = (dreameye_state_t *)dev->status;
-    de->image_count = 0;
-    de->image_count_valid = 0;
-    de->transfer_count = 0;
-    de->img_transferring = 0;
-    de->img_buf = NULL;
-    de->img_size = 0;
-    de->img_number = 0;
-
-    return 0;
-}
-
 /* Device Driver Struct */
 static maple_driver_t dreameye_drv = {
     .functions = MAPLE_FUNC_CAMERA,
     .name = "Dreameye (Camera)",
-    .status_size = sizeof(dreameye_state_t),
-    .attach = dreameye_attach
+    .status_size = sizeof(dreameye_state_t)
 };
 
 /* Add the Dreameye to the driver chain */


kernel/arch/dreamcast/hardware/maple/maple_irq.c
```diff
@@ -56,6 +56,8 @@ static void vbl_chk_disconnect(maple_state_t *state, int p, int u) {
 
         if(maple_driver_detach(p, u) >= 0) {
             assert(!maple_dev_valid(p, u));
+            /* If we're still scanning, reset this port's scan_ready state */
+            state->scan_ready_mask &= ~(1 << p);
```
pcercuei: Not sure about this one... this looks like potential race material. What's wrong with waiting for the next rescan, which is just one frame away?

QuzarDC: That sort of questions the purpose of waiting for all devices to have been scanned once before proceeding then generally. If by the end of that we have the ability to know it's not true and there's some device already that hasn't been scanned, it seemed right to ensure we rescan. But the notion of it being accurate isn't right I guess, it's not that we want to assure there's an accurate scan of the devices, just that we haven't handed off control to the user before we've at least tried on each device.

Will go ahead and just drop it.

kernel/arch/dreamcast/hardware/maple/keyboard.c
```diff
@@ -537,7 +537,7 @@ kbd_state_t *kbd_get_state(maple_device_t *device) {
     if(!device)
         return NULL;
 
-    if(!device->status_valid)
+    if(!device->status)
```
pcercuei, line 540: You could check for `dev->valid` there instead (or not, up to you).

One thing I see here, unrelated to your PR, is that nothing prevents the maple core from pulling the device under your feet after `kbd_get_state()` has been called, freeing the status buffer, causing the calling function to deference a dangling pointer.

Maybe the simplest fix would be to not free the status buffer on driver detach...

pcercuei (approved):

kernel/arch/dreamcast/hardware/maple/keyboard.c
QuzarDC: Yea, unfortunately that's a problem generally with the maple device objects. It would probably make sense though to extend the change from #804 to `status` so that it's at least uniform in nature as you're describing. I'll take some time to see about updating the detach code to keep status allocated and trying to clean up/standardize the validity checks.

The validity checks are unfortunately lax throughout all the maple code, if you look at the function directly below (`kbd_queue_pop`) it also presumes that the dev and state are valid despite the fact that even if the user tests it first, it could become invalid before disabling irqs. Since it all used to be statically allocated, the worst case was just having stale data rather than random garbage, so it wasn't as problematic.

PR #1291, opened 2026-01-06
`dbgio` cleanup, denullify, and multi-output

QuzarDC:
The cleanup:
- Whitespace cleanup.
- Switch spinlocks to mutexes.
- Refactor `dbgio_dev_select_auto` to avoid repeated change of output.
- Update `dbgio_handler::detected()` documentation and drop unnecessary stubs.
- Merge DC and NAOMI fb dbgio handlers, and correct its `detected` behavior to avoid auto-detect.
- Add error handling to `dbgio_remove_handler()`
- Actually use `dbgio_handler::shutdown()`. This was documented and available but not actually used. Now, call the shutdown when a handler is deselected or uninstalled.
- Refactor dcload syscall detection and initialization out of `fs_dcload` and into the syscall driver.
- Prevent adding an already registered handler name.

Denullify / dynamic handlers :
- Allow silent failure if no handler is selected. Previously this was a disallowed state.
- Additionally allow failure if handler members are missing.
- Remove default `null` handler. This only existed to prevent failure if no other handler was available. Now if no handler is available the behavior just defaults to the same `null handler` behavior.
- Remove different handler's stub functions. No member function is required anymore, so stubs are not necessary.
- Have the dcload and scif subsystems install their own dbgio handlers if they successfully initialize.

Multi-output:
- Add a flag `dbgio_handler::output` that sets whether a handler will be forced to be outputting even when not selected.
- Add `dbgio_dev_output` which allows setting whether or not forced output is enabled. If changing the setting, the handler will be automatically initialized/shutdown as appropriate.
- Add a small example of the behavior into the `conio_dgbio` example.


kernel/debug/dbgio.c
```diff
@@ -84,23 +85,19 @@ int dbgio_dev_select_auto(void) {
...
-        if(cur->detected()) {
-            /* Select this device. */
-            dbgio = cur;
-
+        if(cur->detected && cur->detected()) {
             /* Try to init it. If it fails,
                then move on to the next one anyway. */
-            if(!dbgio->init()) {
-                /* Worked. */
+            if(!cur->init()) {
```
pcercuei: Could make the `.init` callback optional too, if missing assume it worked.

kernel/debug/dbgio.c
```diff
@@ -116,79 +113,79 @@ int __weak_symbol dbgio_init(void) {
 
 int dbgio_set_irq_usage(int mode) {
     if(dbgio_enabled) {
```
pcercuei, line 205: I think, while you are at it, instead of checking for `dbgio_enabled` all the time, you could set dbgio to NULL when disabled.

QuzarDC: So we already had the inverse of that in the last PR. The problem then is that the disable/enable of the functionality would be tied to the installed handler. So if you disabled and reenabled you'd not have a handler selected, would have to go through autoselect, or would have to store the 'last enabled' which all make it less than just a straight simplicity win.

And the goal is additionally to end up replacing the singular dbgio with a mechanism to allow for a list of them, which would make it even more important for the enable/disable to be independent of the selected handlers.

kernel/debug/dbgio.c
QuzarDC: Ah, good catch! I'd missed doing it for init.

kernel/debug/dbgio.c
```diff
@@ -170,7 +171,7 @@ int dbgio_write_str(const char *str) {
 
 /* Not re-entrant */
 static char printf_buf[1024];
-static spinlock_t lock = SPINLOCK_INITIALIZER;
+static mutex_t lock = MUTEX_INITIALIZER;
```
pcercuei, line 279: I wonder if that's really needed. The `printf_buf` could be allocated on the stack, which would remove the need for locking.

QuzarDC: Hrm, sure. Would probably still want a 'one at a time' mechanism though, maybe around the core `dbgio_write_str`, so two don't overlap. To prevent `First\nSecond\n` from possibly coming out `FirSecond\nst\n`

pcercuei: I don't think that can happen, at least not with dcload-serial (where syscalls get a lock AFAIK) or dcload-ip (where the network output is protected with a mutex).

QuzarDC: It can with scif, fb, or custom installable handlers which are now a thing with the recent expansion of dbgio. It could be that the answer is just to clean those up to ensure they protect themselves, or simply document that dbgio expects its handlers to figure it out for themselves.

QuzarDC:
@gyrovorbis - tagged you to check this as it basically obsoletes the work you were trying for with #587 . Rather than having a dedicated 'aux' slot though, allows output to be multiplexed through any number of installed handlers.

kernel/arch/dreamcast/fs/fs_dcload.c
```diff
@@ -410,64 +410,29 @@ static vfs_handler_t vh = {
...
-    return 0;
-}
-
 dbgio_handler_t dbgio_dcload = {
-    .name = "fs_dcload_uninit",
-    .detected = never_detected
+    .name = "fs_dcload",
+    .detected = syscall_dcload_detected,
+    .write_buffer = dcload_write_buffer
+    // .read = dcload_read_cons;
```
andressbarajas: .read_buffer is missing. It was in void fs_dcload_init_console() [dbgio_dcload.read_buffer = dcload_read_buffer;] in fs_dcload.c

kernel/debug/dbgio.c
```diff
@@ -29,17 +30,21 @@ struct dbgio_handlers_list dbgio_handlers;
...
 
-int dbgio_dev_select(const char * name) {
+int dbgio_dev_select(const char *name) {
     dbgio_handler_t *cur;
 
     SLIST_FOREACH(cur, &dbgio_handlers, entry) {
         if(!strcmp(cur->name, name)) {
-            /* Try to initialize the device, and if we can't then bail. */
-            if(cur->init()) {
+            /* If it has init, try to and kick on failure. */
```
andressbarajas: Should add a check in this foreach to make sure the user doesn't select the same dbgio twice otherwise it would be shutdown on line 45 because curr is equal to dbgio.

/* Already selected, nothing to do. */
if(cur == dbgio)
    return 0;

Should use !cur->output instead of !dbgio->output because dbgio should not decide if cur should be inited. If cur->output is true then we don't need to reinit it.

/* If it has init, try to and kick on failure. */
if(!cur->output && cur->init && cur->init()) {

kernel/debug/dbgio.c
```diff
@@ -65,42 +112,66 @@ void dbgio_disable(void) {
...
-
-            /* Try to init it. If it fails,
+        if(cur->detected && cur->detected()) {
+            /* If inittable, try to init it. If it fails,
                then move on to the next one anyway. */
-            if(!dbgio->init()) {
-                /* Worked. */
+            if(cur->output || !cur->init || !cur->init()) {
+                /* If it has a shutdown, do so */
+                if(!dbgio->output && dbgio->shutdown)
```
andressbarajas:
```suggestion
                if(dbgio && dbgio != cur && !dbgio->output && dbgio->shutdown)
```
I get a crash here on PS2. dbgio is NULL when dbgio_dev_select_auto() is first called. I guess on the Dreamcast it gets lucky dereferencing what ever is at memory 0 for output and/or shutdown.

kernel/debug/dbgio.c
```diff
@@ -65,42 +112,66 @@ void dbgio_disable(void) {
 }
 
 int dbgio_add_handler(dbgio_handler_t *handler) {
+    int rv = 0;
```
andressbarajas, line 126:
```suggestion
    dbgio_handler_t *cur;
    int rv = 0;

    /* Make sure we don't add a handler that is already in the list. */
    SLIST_FOREACH(cur, &dbgio_handlers, entry) {
        if(cur == handler)
            return 0;
    }

```

kernel/arch/dreamcast/fs/fs_dcload.c
QuzarDC: Good catch. Corrected.

kernel/debug/dbgio.c
QuzarDC: Makes sense, though the issue predated this cleanup. I implemented it slightly differently in that I made a helper function for that search and made it so that it test for the name, since that (rather than the pointer to the handler) acts as the key for most of the API and we shouldn't have two with the same name.

kernel/debug/dbgio.c
QuzarDC: Fixed.

kernel/debug/dbgio.c
QuzarDC: Fixed.

QuzarDC:
I think I got all the feedback. Additionally to implement one of the gaps, did a small refactor, and I noticed that we weren't null-checking any of the input parameters so marked those functions with the `__nonnull` attribute to codify that.

PR #1294, opened 2026-01-07
Filesystem code cleanup & API update

pcercuei:
This set of commits update the top-level filesystem code, fixing some thread-unsafe code that probably triggered real-word consequences, and update a few prototypes to enhance the API and fix error handling.


kernel/fs/fs.c
```diff
@@ -476,15 +476,15 @@ size_t fs_total(file_t fd) {
 
     /* Prefer the 32-bit version, but fall back if needed to the 64-bit one. */
     if(h->handler->total)
-        return h->handler->total(h->hnd);
+        return (ssize_t)h->handler->total(h->hnd);
```
QuzarDC: Instead of casting here, it really should be updating the prototypes of `vfs_handler::total{64}`

PR #1296, opened 2026-01-09
thd: Enable thread ageing support

### LINKED ISSUE #1295: Low-priority threads get zero CPU time
When more than 3 threads with different priorities are all running, the scheduler will never preempt the thread with the lowest priority thread.

Reproducer:  [test_thd_issue.c](https://github.com/user-attachments/files/24525833/test_thd_issue.c)

pcercuei:
Until now, a low-priority thread could be completely starved when the scheduler always had higher-priority running threads to preempt.

Address this issue by implementing thread ageing. This mechanism increases a thread priority at every given time interval (harcoded to 32 milliseconds here) since the last preemption, giving it a chance to eventually run.

Note that this means we're now parsing the entire run queue to find the thread that will eventually be preempted. This should be fine, KallistiOS is not Linux and the number of running threads is generally minimal.

Fixes #1295.

### DIFF
--- a/include/kos/thread.h
+++ b/include/kos/thread.h
@@ -216,7 +216,7 @@ typedef struct __attribute__((aligned(32))) kthread {
     */
     uint64_t wait_timeout;
 
-    /** \brief Per-Thread CPU Time. */
+    /** \brief Per-Thread CPU Time, in milliseconds. */
     struct {
         uint64_t scheduled; /**< \brief time when the thread became active */
         uint64_t total;     /**< \brief total running CPU time for thread */
--- a/include/kos/thread.h
+++ b/include/kos/thread.h
@@ -89,6 +89,9 @@ __BEGIN_DECLS
 
     This macro defines the maximum value for a thread's priority. Note that the
     larger this number, the lower the priority of the thread.
+
+    Priority values above this threshold are still supported, with the caveat
+    that the scheduler might not give any CPU time to the thread.
 */
 #define PRIO_MAX 4096
 
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -121,7 +121,7 @@ int thd_each(int (*cb)(kthread_t *thd, void *user_data), void *data) {
 }
 
 int thd_pslist(int (*pf)(const char *fmt, ...)) {
-    uint64_t cpu_time, ns_time, cpu_total = 0;
+    uint64_t cpu_time, ms_time, cpu_total = 0;
     kthread_t *cur;
 
     pf("All threads (may not be deterministic):\n");
@@ -129,7 +129,7 @@ int thd_pslist(int (*pf)(const char *fmt, ...)) {
 
     irq_disable_scoped();
     thd_get_cpu_time(thd_get_current());
-    ns_time = timer_ns_gettime64();
+    ms_time = timer_ms_gettime64();
 
     LIST_FOREACH(cur, &thd_list, t_list) {
         pf("%08lx  ", CONTEXT_PC(cur->context));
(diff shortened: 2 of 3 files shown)


include/kos/thread.h
```diff
@@ -218,6 +218,8 @@ typedef struct __attribute__((aligned(32))) kthread {
     */
     uint64_t wait_timeout;
 
+    uint64_t last_sched;
```
QuzarDC: Pretty sure `cpu_time.scheduled` could be used instead here. It's set to the current time on rescheduling, and only otherwise in the current thread if we're requesting the thread's time usage.

kernel/thread/thread.c
```diff
@@ -31,6 +32,10 @@
 #include <arch/stack.h>
 #include <arch/tls_static.h>
 
+/* Time interval in milliseconds since a thread's last preemption, after which
+ * the thread's priority is doubled */
+#define THD_AGEING_THRESHOLD_MS 32
```
QuzarDC: Are there pros/cons to having this be independent of the scheduler frequency? It seems like they could or should be related. In the sense of having it be something like 'if it hasn't run after X scheduler ticks'.

QuzarDC (commented):
My apologies for how long this is, but there's just a lot of extra considerations I'm seeing. tl;dr: Why not reuse boosting? Should we keep the run queue prio sorted? What about the idle thread?

Two specific things (noted in the code) and three more generalized comments:

Maybe it just wouldn't be workable here but I had imagined that the way this would have been done would be via the priority boosting mechanism. `if(thd->last_sched > threshold) do_boost()`. Was there a specific reason to not integrate this into priority boosting? The biggest downside I see to this as-is would be that you'd have to be shuffling around the order in the run queue since it's currently sorted by priority. Which leads to:

If we're already going to now be stepping through the whole run queue on each schedule maybe we should reconsider changing the sorting. If it were sorted by `last_sched` then the going through the whole run queue could be done in two phases:
- First start from the least recently run thread and step backwards through the queue doing the full set of testing.
- Once you reach a thread that doesn't age out, start a new loop through that thread where you *just* test the prios to not have to keep doing the `thd_calc_prio` testing

Even without that two-phase loop through the queue, sorting that way would allow for
putting the current thread back into the queue on reschedule to be less expensive as it would always go back to the head as the most recently run. The reason to sort by priority was, after all, to optimize these two sets of operations (find next thread, and insert old thread).

Tied into that sorting is also a question on the idle thread. I don't see any protection here from the idle thread aging and getting run even when there are other ready threads. In a simple way we could have an exception where the lowest priority value allowed is exempt from aging. But going along with the issue of the priority sorting, having the idle thread in the run queue at all is tied to that sorting by priority and allowing it to naturally get picked if we can't otherwise find a thread to run. If we don't need that anymore then it might make sense to have selecting it be treated as a special case.

kernel/thread/thread.c
pcercuei: I thought about that and ideally it should depend on the scheduler frequency, yes. However I don't see how to do it cleanly without adding an integer division in there and I'd rather avoid that.

gyrovorbis (dismissed):
Okay, I'm back.

I like his latest implementation. He's gone out of his way to ensure the minimal amount of work is done to calculate the aged thread priority level, and I'm okay with the CPU time counter being reduced to `ms` to allow for it. It does indeed solve the aforementioned starvation issue, as verified with the test app he attached to the related issue (plus dicked with it more to make it more insidious, and it still worked).

Tested:

1. His app
2. A bunch of threading examples/tests
3. Mario Kart 64
4. Sonic Mania
5. Bruce's Balls

No regression spotted, and it does fix what I see as a pretty bad issue in the kernel scheduler. Ship it.

kernel/thread/thread.c
```diff
@@ -585,20 +585,18 @@ tid_t thd_get_id(const kthread_t *thd) {
 /*****************************************************************************/
 /* Scheduling routines */
 
-static void thd_update_cpu_time(kthread_t *thd) {
-    const uint64_t ns = timer_ns_gettime64();
```
QuzarDC, line 607: What's the reason to shift this out to being provided externally? It seems that otherwise the change could be just to `timer_ms_gettime64` and have the proto for `thd_schedule_inner` and behavior of `thd_get_cpu_time` stay the same. With the change to `thd_get_cpu_time` the time for the current thread will always be missing the time elapsed since it's last run.

Maybe the idea is to get ahead of the change being needed to support the age checking that is added in the next commit? Or to narrow the amount of the scheduler time that's being counted for the thread?

pcercuei: At that point we already did a call to `timer_ms_gettime64` in the main path (from thd_schedule), so there's no point in calling it a second time. It'd just add unnecessary overhead.

QuzarDC (approved):

PR #1299, opened 2026-01-16
Optimized and fixed buggy/slow IPv4 checksum alg.

gyrovorbis:
I've taken the liberty of gifting KOS back my hand-optimized ipv4 checksum that was previously part of SH4ZAM. It was written in C, but optimized by hand within Compiler Explorer.

1) net_ipv4.c
    - net_ipv4_checksum() was calculating the checksum extremely inefficiently, doing byte-wise masking + shifting within a loop for every 16-bit word. The algorithm is now directly summing each 16-bit word and is only doing the shifty loop for carry bits once at the end.
    - net_ipv4_checksum() was also TOTALLY INCORRECT for the unaligned path, due to a bug on line 38 (*ptr + 1): which is adding one to the dereferenced value, not using the value of the next byte address... So basically this whole path is wrong and we've gotten lucky that the "data" has always been at least 2 byte aligned. 
2) AUTHORS
    - added 2026 copyright date, since this is my first contribution this year.

Tested by running all network examples. Everything passed and seems to be in working order.

### DIFF
--- a/kernel/net/net_ipv4.c
+++ b/kernel/net/net_ipv4.c
@@ -5,6 +5,8 @@
    Copyright (C) 2005, 2006, 2007, 2008, 2009, 2010, 2012, 2013,
                  2016 Lawrence Sebald
 
+   Copyright (C) 2026 Falco Girgis
+
    Portions adapted from KOS' old net_icmp.c file:
    Copyright (C) 2002 Megan Potter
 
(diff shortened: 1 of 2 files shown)


kernel/net/net_ipv4.c
```diff
@@ -29,32 +31,27 @@ static net_ipv4_stats_t ipv4_stats = { 0 };
...
-
-            while(sum >> 16)
-                sum = (sum >> 16) + (sum & 0xFFFF);
         }
     }
 
+    i &= 0x1;
+
     /* Handle the last byte, if we have an odd byte count */
     if(i)
```
pcercuei: It looks like `i` can be dropped entirely. You could check for `if(bytes & 1)` here.

gyrovorbis: Oh shit, nice. Going to do that.

QuzarDC (approved):

PR #1305, opened 2026-01-21
libkosfat: fix fat_search_long longname_buf for LFN

japanese-cake:
This patch fixes the longname_buf null character offset which was set at the end of the non null-terminated string + 2. As the buffer contains ucs2 data from previous iterations, if a non-null char ends up there, the memcmp test fails.

Notes to reviewers:
* 1st contrib so please let me know if there is something missing or wrong :)
* It may not be 100% reproductible but here is how I tested it: with a fat32 formated sdcard on windows, create a file named "ABCDEFGHI.123" at the root of the sdcard and try to load it --> file not found.


pcercuei (approved):
I don't know much about FAT but the short filenames are in [8.3 format](https://en.wikipedia.org/wiki/8.3_filename), and 8+1+3 == 12, so the 13th character should be the NULL character, it makes sense to me.

PR #1311, opened 2026-01-26
kernel: Set default init flags as weak symbol

pcercuei:
If not set as a weak symbol, the linker will pick the first one it finds, which may just be KOS' default one if libkallisti.a is placed before the application's objects on the command line.


gyrovorbis (approved):
Already talked about this. Definitely needed to happen.

darcagn (approved):

PR #1313, opened 2026-01-28
Newlib/Binutils and other toolchain updates

darcagn:
- Binutils updated to `2.45.1` for all toolchains that aren't pinned to particular versions.
- Newlib updated to `4.6.0` for all toolchains that aren't pinned to particular versions.
- GCC master is now listed as `16.0.1` instead of `16.0.0` to reflect its updated development stage.
- `-dev` patches updated to patch cleanly by fixing offsets and patch fuzz.
- Updated documentation and changelog

All profiles build fine on Gentoo Linux with host GCC `15.2.1_p20251122`, no further testing done beyond that yet.

### DIFF
--- a/utils/dc-chain/Makefile.dreamcast.cfg
+++ b/utils/dc-chain/Makefile.dreamcast.cfg
@@ -23,7 +23,7 @@ platform=dreamcast
 # - 14.3.1-dev    Bleeding edge GCC 14 series from git.
 # - 15.0.0-lra    GCC 15 branch containing experimental SuperH fixes for LRA.
 # - 15.2.1-dev    Bleeding edge GCC 15 series from git.
-# - 16.0.0-dev    Bleeding edge GCC 16 series from git.
+# - 16.0.1-dev    Bleeding edge GCC 16 series from git.
 # If unsure, select stable. See README.md for more detailed descriptions.
 toolchain_profile=stable
 
--- a/utils/dc-chain/doc/CHANGELOG.md
+++ b/utils/dc-chain/doc/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 | Date<br/>_____________ | Author(s)<br/>_____________ | Changes<br/>_____________ |
 |:-----------------------|:----------------------------|---------------------------|
+| 2026-01-28 | Eric Fradella | Update binutils to 2.45.1 and Newlib to 4.6.0 for most profiles |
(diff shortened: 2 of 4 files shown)


utils/dc-chain/patches/targets/gcc-13.4.1-kos.diff
```diff
@@ -1,7 +1,7 @@
 diff -ruN gcc-13.4.1/gcc/config/elfos.h gcc-13.4.1-kos/gcc/config/elfos.h
 --- gcc-13.4.1/gcc/config/elfos.h	2025-04-17 16:01:33.790051712 -0600
 +++ gcc-13.4.1-kos/gcc/config/elfos.h	2025-04-17 16:01:42.910094466 -0600
-@@ -486,3 +486,6 @@
+@@ -482,3 +482,6 @@
```
QuzarDC, line 4: Why are there patch updates needed here? The config files change within the same version? Or were they changes missed when the original patches were added for these versions because the patching could still just work good enough (since there's no changes to what's being patched just stuff around that and line numbers).

darcagn: Normally I always keep the diff files updated regularly with the correct offsets when I do one of these batches of toolchain updates, but #1026 re-used patches across the toolchain profiles without ensuring the offsets were perfect for each profile. It technically didn't matter and not something worth holding the PR up over, so I approved #1026 as it was, figuring I'd just clean up the offsets myself in a later PR instead of asking @pcercuei to do that work.

That being said it is possible that any of these files could change within the same version since the `-dev` profiles are active branches.

Note that I do manual adjustments on the diff files. The fastest way for me is to just apply the diff, and if `patch` says that it used an offset adjustment, I manually update the diff with the new offset and then re-verify. So, if the change is just an adjusted offset (as this one you've highlighted is), you won't see the dates being updated.

If any of the _text_ within the diff changes -- if the patch has a fuzz instead of just an offset adjustment -- then I do regenerate that block of the diff file with an updated date.

QuzarDC: Alrighty, basically what I had figured it was. Since they were all in the 'config' related stuff though I thought I might have missed a bit about how the config.guess component works or something.

QuzarDC (approved):

PR #1320, opened 2026-02-08
Stabilize 15.2.0, toolchain builder updates and bug fixes

darcagn:
- Complete transition from `dc-chain` to `kos-chain` by renaming dir, paths, and comments
- Fix incorrect `auto_fixup_sh4_newlib` variable
- Fix `config.guess` not always being run
- Fix `--with-native-system-header-dir` argument
- Fix ability to apply multiple patches to GDB
- Fix `use_custom_dependencies` option to work once more
- ...and other various cleanups

See individual commits for more detail on each.

Tested building GCC and GDB on Gentoo Linux. macOS was not fully tested as it fails to build with the existing `dc-chain`, too. This needs fixing in a future PR. A further `doc` dir refresh is needed as well, since many of the instructions in there are outdated or redundant.

EDIT:
I've also added an update to this PR to stabilize 15.2.0 and mark 13.2.0 as legacy. I think it's time to move forward with this.

- 15.x builds KOS/kos-ports/examples without issue now in my testing, with both safe CFLAGS and aggressive CFLAGS
- 15.x has better language support (C23 and C++23)
- 15.x is more likely to build for newer systems
- People have been using 15.x for 10 months now without any issues
- DreamShell is now using 15.x as well

I used the new Newlib and Binutils for this, but they've been working fine and I believe putting them in the stable config would be low risk.

### DIFF
--- a/utils/kos-chain/scripts/phony.mk
+++ b/utils/kos-chain/scripts/phony.mk
@@ -1,11 +1,10 @@
 # KallistiOS Toolchain Builder (kos-chain)
 
 .PHONY: all
-.PHONY: fetch fetch-gdb
+.PHONY: fetch
 .PHONY: patch $(patch_targets)
-.PHONY: $(patch_binutils) $(patch_gcc) $(patch_newlib)
 .PHONY: build $(build_targets)
 .PHONY: build-binutils build-newlib build-gcc-pass1 build-gcc-pass2 fixup-newlib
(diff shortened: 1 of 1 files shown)


darcagn:
> You'll also need to `find .github -type f -exec sed -i 's/dc-chain/kos-chain/' {} \;`

Yep, forgot to do a search outside of the `dc-chain` dir. 🤦🏼 

Fixed and pushed.

PR #1323, opened 2026-02-14
pvr: Simplify direct rendering

pcercuei:
Instead of requiring the user to supply a 'pvr_dr_state_t' that is used as a selector between the two SQs, we can have it inside the PVR driver itself. It is then defined as an extern value inside pvr.h, and pvr_dr_target() has been modified to use it.

The pvr_dr_target() also returns a (void *) pointer, which is better to avoid having to cast anything that's not a vertex struct. For good measure, the macro will now tell the compiler that the pointer is 32-byte aligned.

pvr_dr_target() ideally should be defined as taking no argument, but I set it to accept any number of arguments to retain compatibility with old code.

We also don't need pvr_dr_init() / pvr_dr_finish(), which only initialized a "dr_used" variable inside the driver, that was never actually used anywhere afterwards. So those were moved as dummies to the legacy API header.

### DIFF
--- a/examples/dreamcast/pvr/fb_tex/fb_tex.c
+++ b/examples/dreamcast/pvr/fb_tex/fb_tex.c
@@ -18,7 +18,6 @@ struct square_fcoords {
 
 static bool done;
 
-static uint32_t dr_state;
 static uint32_t fbuf_color = 0xffffffff;
 
 static void change_color(uint8_t, uint32_t btns) {
@@ -71,7 +70,7 @@ static void render_coords(const struct square_fcoords *coords, float z,
     unsigned int i;
 
     for(i = 0; i < 4; i++) {
-        vert = pvr_dr_target(dr_state);
+        vert = pvr_dr_target();
         *vert = (pvr_vertex_t){
             .flags = i == 3 ? PVR_CMD_VERTEX_EOL : PVR_CMD_VERTEX,
             .x = coords->x[i],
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -957,31 +957,16 @@ int pvr_prim(const void *data, size_t size);
...
-typedef uint32_t pvr_dr_state_t;
-
-/** \brief   Initialize a state variable for Direct Rendering.
-
-    Store Queues are used.
-
-    \param  vtx_buf_ptr     A variable of type pvr_dr_state_t to init.
-*/
-void pvr_dr_init(pvr_dr_state_t *vtx_buf_ptr);
+/** \brief   Current SQ address for direct rendering. */
```
gyrovorbis: Personally... I'd rather hide this mofo between `\cond` and `\encond` and not even reveal it as part of the documented API, since it's more if an unavoidably leaking implementation detail, but... I guess I'm fine either way.

dfchil (approved):
This is a tremendous improvement of the previous pvr_dr_* API.

kernel/arch/dreamcast/include/dc/pvr.h
pcercuei: Done. And added one comment on `pvr_dr_target()`'s doc that says that it actually expects no argument.

This dismissed your review so you'll need to ACK again.

QuzarDC (changes requested):
Could you please update the examples that are using the deprecated methods/functionality?

Otherwise looks fine. Seems like it should have happened at the time of the introduction of the hybrid rendering generally.

PR #1324, opened 2026-02-19
examples: Fix dangling reference in concurrency.cpp

darcagn:
Using `decltype(ff)` here creates a dangling reference, which can cause a crash (seems to always happen on GameCube). Switching this to `std::decay_t<FF>` passes the value which fixes the crash. Verified by @gyrovorbis this is necessary to fix.


gyrovorbis (approved):
Yeah, this is actually correct and is fixing a very subtle bug that caused a dangling reference.

PR #1325, opened 2026-02-21
Fix DMA transfer lock for non-blocking mode

nextgeniuspro:
As we discussed in Simulant discord chat this fix eliminates freezing issue when manipulating files in different threads the same or almost the same time

### DIFF
--- a/kernel/arch/dreamcast/hardware/cdrom.c
+++ b/kernel/arch/dreamcast/hardware/cdrom.c
@@ -557,7 +557,7 @@ int cdrom_stream_request(void *buffer, size_t size, bool block) {
     }
 
     params.size = size;
-    sem_wait_scoped(&_g1_ata_sem);
+    sem_wait(&_g1_ata_sem);
 
     if(stream_dma) {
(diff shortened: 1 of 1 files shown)


darcagn:
To clarify for those not looking at the Discord -- with the scoped semaphore macro, it signals on this path at line 575:

```c
if(!block) {
    return ERR_OK;
}
```

but it's not supposed to, which this fixes.

nextgeniuspro:
> To clarify for those not looking at the Discord -- with the scoped semaphore macro, it signals on this path at line 575:
> 
> ```c
> if(!block) {
>     return ERR_OK;
> }
> ```
> 
> but it's not supposed to, which this fixes.

Thanks @darcagn, I should have included better explanation

darcagn:
> Thanks @darcagn, I should have included better explanation

No worries! Thanks very much for tracing down the issue and opening a PR. Very appreciated.

QuzarDC (approved):
Thanks for the contribution! Makes perfect sense.

PR #1331, opened 2026-02-27
net: Fix checksum algorithm

pcercuei:
The recent commit to improve the checksum algorithm actually broke it, as shown by the NTP example not working anymore.

Rework the algorithm to make it work again, and improve its performance by always processing two bytes at a time.

### DIFF
--- a/kernel/net/net_ipv4.c
+++ b/kernel/net/net_ipv4.c
@@ -27,37 +27,30 @@
 
 static net_ipv4_stats_t ipv4_stats = { 0 };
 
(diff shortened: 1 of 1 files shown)


gyrovorbis (approved):
Damn, thanks so much.

I had Googled multiple times and had read that the carry didn't have to be taken into account during the main sum loop and that it could be handled at the end... So either I read some BS, or my original implementation was flawed still.

Either way, this is a more robust and more elegant algorithm. I really like what you did here. Thanks. ACK.

PR #1333, opened 2026-02-27
New cache ARCH API

pcercuei:
Add a new cache ARCH API, compatible with existing code but with slightly different names.

Note that some functions have been introduced, e.g. functions to invalidate / flush / purge individual cache lines.

This PR also rewrites most of the cache code from ASM to C, because it really does not need this level of complexity.

### DIFF
--- a/include/kos/cache.h
+++ b/include/kos/cache.h
@@ -1,0 +1,303 @@
+/* KallistiOS ##version##
+
+   include/kos/cache.h
+   Copyright (C) 2001 Megan Potter
+   Copyright (C) 2014, 2016, 2023 Ruslan Rostovtsev
+   Copyright (C) 2023 Andy Barajas
+   Copyright (C) 2025 Eric Fradella
+   Copyright (C) 2026 Falco Girgis
+*/
+
+/** \file    include/kos/cache.h
+    \brief   Cache management functionality.
+    \ingroup system_cache
+
+    This file contains definitions for functions that manage the caches,
+    including functions to flush, invalidate, purge, prefetch and allocate
+    individual cache lines or address ranges.
+
+    \author Megan Potter
+    \author Ruslan Rostovtsev
+    \author Andy Barajas
+    \author Falco Girgis
+*/
+
+#ifndef __KOS_CACHE_H
+#define __KOS_CACHE_H
+
+#include <kos/cdefs.h>
+__BEGIN_DECLS
+
+#include <arch/cache.h>
+
+#include <stdint.h>
+
+/** \defgroup system_cache Cache
+    \brief                 Driver and API for managing the SH4's cache
+    \ingroup               system
+
+    @{
+*/
+
+/** \brief  Level 1 instruction cache size.
+
+    The capacity of the L1 instruction cache in bytes.
+*/
+#define CACHE_L1_ICACHE_SIZE \
+    ARCH_CACHE_L1_ICACHE_SIZE
+
+/** \brief  Level 1 instruction cache associativity.
+
+    Number of ways in the L1 instruction cache.
+*/
+#define CACHE_L1_ICACHE_ASSOC \
+    ARCH_CACHE_L1_ICACHE_ASSOC
+
+/** \brief  L1 instruction cache line size.
+
+    The size of each cache line in the L1 instruction cache.
+*/
+#define CACHE_L1_ICACHE_LINESIZE \
+    ARCH_CACHE_L1_ICACHE_LINESIZE
+
+/** \brief  Level 1 data cache size.
+
+    The capacity of the L1 data cache in bytes.
+*/
+#define CACHE_L1_DCACHE_SIZE \
+    ARCH_CACHE_L1_DCACHE_SIZE
+
+/** \brief  Level 1 data cache associativity.
+
+    Number of ways in the L1 data cache.
+*/
+#define CACHE_L1_DCACHE_ASSOC \
+    ARCH_CACHE_L1_DCACHE_ASSOC
+
+/** \brief  L1 data cache line size.
+
+    The size of each cache line in the L1 data cache.
+*/
+#define CACHE_L1_DCACHE_LINESIZE \
+    ARCH_CACHE_L1_DCACHE_LINESIZE
+
+/** \brief  Level 2 cache size.
+
+    The capacity of the L2 cache in bytes.
+*/
+#define CACHE_L2_CACHE_SIZE \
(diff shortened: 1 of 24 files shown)


pcercuei:
Note that the whitespace check fails because of the whitespace cleanups, so CI can  be ignored here.

kernel/arch/dreamcast/include/arch/cache.h
```diff
@@ -209,20 +209,6 @@ static __always_inline void dcache_pref_block(const void *src) {
     );
 }
 
-/** \brief  Write-back Store Queue buffer to external memory
-
-    This function initiates write-back for one Store Queue.
-
-    \param  src             The SQ mapped address to write-back.
-*/
-static __always_inline void dcache_wback_sq(void *src) {
```
QuzarDC: I agree this should be over with SQ stuff, but please add some sort of compatibility define or warning so it's not just a hard break.

pcercuei:
Updated.
- `dcache_flush_range` is now called `dcache_wback_range`
- `dcache_flush_all` is now called `dcache_wback_all`
- Added `dcache_alloc_line_with_value` as well as `dcache_zero_alloc_line`.

include/kos/cache.h
```diff
@@ -0,0 +1,311 @@
...
+/** \brief  Purge all the data/operand cache.
+
+    This function flushes the entire data/operand cache, ensuring that all
+    cache blocks marked as dirty are written back to memory and all cache
+    entries are invalidated. It does not require an additional buffer and is
+    preferred when memory resources are constrained.
+
+    \note
+    dcache_purge_range() is faster than dcache_purge_all() if the count
+    param is 39936 or less.
```
QuzarDC: The note would certainly be arch-specific to the DC. Perhaps it should clarify that.

QuzarDC (commented):
One very specific doc note, but then also there should be some compatibility messages for the dropped functions, especially `dcache_wback_sq` which is probably the most externally visible of the removed.

include/kos/cache.h
pcercuei: Good point. I can drop that - the DC-specific code will already automatically call `dcache_purge_all()` from `dcache_purge_range()` above a certain threshold.

pcercuei:
Updated.
- The notes about "range" functions being faster than "all" functions above a certain threshold have been removed.
- dcache_wback_sq() is back, but marked as deprecated. (Note that it does not call sq_flush() to avoid leaking <dc/sq.h> into the arch API).

QuzarDC:
> > The Dreamcast has "alloc a cache line while writing a 32-bit value", and the GameCube has "alloc a cache line which is zeroed out by default". What I see here, is that both could be used to implement a "alloc a cache line" operation. I think the cache ARCH API shouldn't try to cover every single little bit of functionality of the supported platforms, but to be a superset of what's supported on all of them.
> 
> If so, then `dcache_alloc_line()` would drop its `src` argument, and a separate version would exist on Dreamcast with the `src` argument?
> 
> > For that reason I agree that it's too much (for this API) to try to support the "locked" feature of PPC, or OCINDEX of the SH4... those would be better in their respective platform-specific API headers.
> 
> That's fine, but what would that look like in practice? Would a user have to include both `<kos/cache.h>` plus `<dc/cache.h>` to get everything? Does one automatically include the other? I ask because I can see that getting ugly fast.
> 
> > You mention "dcache invalidate all", and "icache invalidate all". I don't think we need those. I can't think of any particular usecase where those would be useful.
> 
> I use them at startup, but that's in assembly and doesn't need to be part of the API so that's fine.

Were your various concerns well addressed by the updates? This one otherwise seems ready but wasn't sure since there was a lot of commentary without clear resolution.

include/kos/cache.h
```diff
@@ -0,0 +1,303 @@
...
+    The size of each cache line in the L1 data cache.
+*/
+#define CACHE_L1_DCACHE_LINESIZE \
+    ARCH_CACHE_L1_DCACHE_LINESIZE
+
+/** \brief  Level 2 cache size.
+
+    The capacity of the L2 cache in bytes.
+*/
+#define CACHE_L2_CACHE_SIZE \
```
andressbarajas, line 88: For L2 #defines I noticed that we leave out the 'D' in DCACHE. Maybe add it so it follows the same naming scheme as CACHE_L1_DCACHE_*? 

EDIT: I just saw that you just ported exactly what was there before. Could be in a follow up.

darcagn: That's intentional. Those defines are named to correlate with the same ones used by glibc's extensions to POSIX sysconf: [in KOS](https://github.com/KallistiOS/KallistiOS/blob/45b72287db497d57ee22c969905325e9422d61d4/kernel/libc/posix/sysconf.c#L54), [in glibc](https://github.com/zerovm/glibc/blob/3f07350498160f552350dc39f6fe6d237c7c3b03/sysdeps/x86_64/cacheinfo.c#L35)

It's very rare for an architecture to have a split L2 cache. None of the architectures KOS supported in the past have one, nor do any of the ones on our radar for the future. So it makes most sense to just stick with what we (and glibc) are already doing.

darcagn:
> Were your various concerns well addressed by the updates? This one otherwise seems ready but wasn't sure since there was a lot of commentary without clear resolution.

I think so, I've just been extremely busy and haven't had the time to sit down and properly review this. I'll do that by Monday night at the latest.

pcercuei:
Rebased to master, so that the whitespace CI check passes.

kernel/arch/dreamcast/include/arch/cache.h
```diff
@@ -23,231 +22,199 @@
...
+}
 
-    This function flushes a range of the data/operand cache, forcing a write-
-    back and then invalidates all of the data in the specified range.
+static inline void arch_dcache_alloc_line(void *src) {
+#if __GNUC__ <= 9
+    /* Avoid ICE on GCC 9 */
+    uint32_t r0 = 0;
+#else
+    register uint32_t r0 asm("r0");
```
QuzarDC: This should be `__asm__`. Surprisingly it manages to break libmp3 otherwise.

QuzarDC (changes requested):
Seems all the structural feedback is done, and if more's needed it can be adjusted afterwards. Just found one issue that would need to get corrected when rebuilding all the kos-ports with it.

pcercuei:
Fixed `asm` -> `__asm__` and force-pushed.

QuzarDC (approved):

PR #1336, opened 2026-03-08
Preliminary cleanups for AICA port

pcercuei:
Here is a set of (mostly) bug fixes, indirectly related to my upcoming KallistiOS port for the AICA.


kernel/libc/newlib/newlib_exit.c
```diff
@@ -30,5 +30,5 @@ __used void _exit(int code) {
            "arch: _exit(%d) called without SHUTDOWN flag enabled!\n",
            code);
 
-    arch_menu();
+    arch_abort();
```
QuzarDC: Might make sense to combine this and the warning and just have this call `panic`.

kernel/exports.txt
```diff
@@ -210,21 +210,13 @@ irq_get_global_handler
...
-arch_exec
-
 # Misc other kernel control
 arch_exit
 arch_abort
 
 # Panic!
 arch_panic
 
-# Real-time clock
```
QuzarDC, line 224: These currently provide the backing for `gettimeofday` and `clock_{g/s}ettime`'s `CLOCK_REALTIME`. I agree that they aren't appropriately factored/named for `arch` but they are being used in arch-agnostic code.

utils/kos-chain/patches/targets/newlib-4.6.0.20260123-kos.diff
```diff
@@ -236,7 +246,7 @@ diff --color -ruN newlib-4.6.0.20260123/newlib/libc/include/sys/_types.h newlib-
  #ifndef __machine_ino_t_defined
  #if (defined(__i386__) && (defined(GO32) || defined(__MSDOS__))) || \
 -    defined(__sparc__) || defined(__SPU__)
-+    defined(__sparc__) || defined(__SPU__) || defined(__sh__) || defined(__PPC__)
++    defined(__sparc__) || defined(__SPU__) || defined(__sh__) || defined(__PPC__) || defined(__ARMEL__)
```
QuzarDC, line 249: Certainly not a blocker for this, but perhaps we should finally consider just a `__KOS__` define or similar for this and other patches where applicable? @darcagn

kernel/fs/fs_pty.c
QuzarDC, line 1: Was there a change elsewhere that should have made files reachable at `/dev/pty`? To my understanding it's always been `/pty` and I've not seen anything else that tries to access it differently from that.

include/kos/init.h
```diff
@@ -139,7 +139,7 @@ extern const void * __kos_romdisk;
 
 /** Default init flags (IRQs on, preemption enabled, romdisks). */
 #define INIT_DEFAULT    (INIT_IRQ     | INIT_THD_PREEMPT | INIT_FS_ALL | \
-                         INIT_LIBRARY | INIT_DEFAULT_ARCH)
```
QuzarDC, line 142: Just giving a tag to @DC-SWAT here for visibility. I'd presume in all your usage you're already manually adding `INIT_EXPORT` anyways, but just in case it causes an issue.

QuzarDC (changes requested):
Most optional or just notes, but not sure about moving rtc with some of the clock functionality requiring it, and am unclear on the reasoning for the `/pty`->`/dev/pty` move.

kernel/fs/fs_pty.c
pcercuei: The reason is that `/pty` does not exist on the filesystem. It has always been `/dev/pty`. So it does not make sense that you can open the "ghost file" `/pty/sl01`, and get a ENOENT error when trying to open the actually existing `/dev/pty/sl01` file.

pcercuei: "it has always been `/dev/pty`" -> or maybe it used to be different, which is possible given that I have the memory of a fish - my point is that this is the current status now and has been for ages, at least since the /dev filesystem was introduced and maybe earlier.

QuzarDC: I added `/dev` as a discrete filesystem and it was to allow proper navigation and interaction with random/urandom which had been directly mapped under it (to which I added null). What you are changing here is exactly what defines the filesystem location to be `/pty`, and to my knowledge the only thing that does.

I wouldn't mind the change as it makes sense and possibly should have been done at the time of adding /dev, but it seems like there's something else at play here if you couldn't open the files where they were.

pcercuei: I think I got really confused somewhere...

The "filesystem/browse" example on KOS master does indeed show that it's `/pty` and not `/dev/pty`. Apologies for that.

So I'll just drop this commit.

kernel/exports.txt
pcercuei: I decided to solve this with a new RTC ARCH API: https://github.com/KallistiOS/KallistiOS/pull/1354

QuzarDC (dismissed):
I did realize one extra thing, and if it's easy enough to add in quickly, would be best to do so here but if not it can go in a followup PR: updating the prebuilt driver to use the updated toolchain. That way it'll get more testing in case there are any unexpected side-effects.

QuzarDC (approved):

PR #1348, opened 2026-03-27
BBA / network stack cleanups

pcercuei:
This PR fixes a panel of issues in the BBA driver as well as in the network stack, which caused all kind of issues like excessive packet drops, terrible performance, and lockups.

After this PR the RX speed is boosted from ~30 KiB/s to more than 2.2 MiB/s, and more importantly, it seems to be more reliable.

### DIFF
--- a/kernel/net/net_tcp.c
+++ b/kernel/net/net_tcp.c
@@ -1270,12 +1270,17 @@ static ssize_t net_tcp_recvfrom(net_socket_t *hnd, void *buffer, size_t length,
         sock->data.rcv.wnd += size;
         sock->data.rcvbuf_cur_sz -= size;
 
-        /* When the receive window was zero, the sender has entered TCP
-           persist mode. Send a window-update ACK so it knows the window
-           has reopened — without this the connection deadlocks. */
+        /* BUG FIX: When the receive window was zero, the sender has
+           entered TCP persist mode (zero-window probing). We must send
+           a window-update ACK so it knows the window has reopened.
+           Without this, the sender never learns and the connection
+           deadlocks permanently. */
         if(old_wnd == 0 && sock->data.rcv.wnd > 0) {
             tcp_send_ack(sock);
         }
+        /* Flush any pending delayed ACK now that the app has read.
+           This gets the ACK out immediately instead of waiting up to
+           10ms for the timer, reducing effective RTT significantly. */
         else if(sock->data.ack_pending > 0) {
             tcp_send_ack(sock);
             sock->data.ack_pending = 0;
@@ -3017,7 +3022,8 @@ static void tcp_thd_cb(void *arg) {
             case TCP_STATE_ESTABLISHED:
             case TCP_STATE_CLOSE_WAIT:
 
-                /* Flush pending delayed ACK */
+                /* Flush pending delayed ACK. Without this, a single
+                   unACKed segment stalls the sender indefinitely. */
                 if(i->data.ack_pending > 0) {
                     tcp_send_ack(i);
                     i->data.ack_pending = 0;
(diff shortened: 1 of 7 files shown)


kernel/net/net_tcp.c
```diff
@@ -858,8 +871,10 @@ static int net_tcp_bind(net_socket_t *hnd, const struct sockaddr *addr,
                     return -1;
                 }
 
-                if(iter->local_addr.sin6_port == port) {
+                if(iter->local_addr.sin6_port == htons(port)) {
+                    mutex_unlock(&iter->mutex);
                     ++port;
+                    if(port > 65000) port = 1024;
```
ljsebald, line 877: Out of curiosity, why stop here at 65000? Why not go all the way to the last port?

Bruceleeto:
```
bruce@fedora:~/Downloads/gpsp/pc/assets/dc$ cat /proc/sys/net/ipv4/ip_local_port_range
32768   60999
```
https://github.com/torvalds/linux/blob/9147566d801602c9e7fc7f85e989735735bf38ba/net/ipv4/af_inet.c#L1799

I dont have a solid reason it just seems like norm. Lol.

ljsebald (dismissed):
I don't see anything glaringly wrong, but someone else should probably review/test the BBA stuff, especially.

kernel/net/net_tcp.c
```diff
@@ -3096,7 +3102,7 @@ static fs_socket_proto_t proto = {
 };
 
 int net_tcp_init(void) {
-    if((thd_cb_id = net_thd_add_callback(tcp_thd_cb, NULL, 50)) < 0)
+    if((thd_cb_id = net_thd_add_callback(tcp_thd_cb, NULL, 10)) < 0)
```
QuzarDC, line 3127: Should the thread callback time also be adjusted for `dhcp_cbid` ? Or figure it's fine to have it stick at a slower rate? Seems like it should be fine to be slower than the tcp thread, but wanted to make sure it wasn't just an oversight.

pcercuei:
Rebased on master (no functional change) as it had conflicts.

QuzarDC (commented):
Ok, all looks good, just had one note on something that might have been overlooked.

UnknownShadow200:
> I tested this on all the examples on master / 14.2.0. It also works on Meese's stripped example.
> 
> But I dont know of any other homebrew games / ports that are built with KOS that use the net stack to properly battle test it on which is quite unfortunate and it's an issue that needs to at some point be addressed by adding a proper example so contributors have a better way to track regressions / improvements as this PR really came about because the net stack regressed without anyone really knowing til months later.

Biased, but [ClassiCube](https://github.com/ClassiCube/ClassiCube) has a dreamcast port that uses the networking stack. It's fairly common to run into random network data corruption at any point, even with the initial https request at startup to fetch latest version.

pcercuei:
> Biased, but [ClassiCube](https://github.com/ClassiCube/ClassiCube) has a dreamcast port that uses the networking stack. It's fairly common to run into random network data corruption at any point, even with the initial https request at startup to fetch latest version.

That's before this PR, after this PR, or both?

kernel/net/net_tcp.c
pcercuei: Probably a question for @Bruceleeto but I would think it's fine.

Bruceleeto: dropping to 10 from what I remember did have a significant up tick in performance. the dhcp_cbid doesnt need to be as fast. I dont think anyway. so should be fine as is.

Like I said earlier though there is not a lot of apps to test on to verify. 
I havent tested classic cube yet. So i will try and do that next if it builds without giving me too much grief.

QuzarDC (approved):

QuzarDC:
> > Biased, but [ClassiCube](https://github.com/ClassiCube/ClassiCube) has a dreamcast port that uses the networking stack. It's fairly common to run into random network data corruption at any point, even with the initial https request at startup to fetch latest version.
> 
> That's before this PR, after this PR, or both?

I'm just presuming it was before this PR, due to the description, and intended to point out another possible thing to use to test that leverages the networking stack.

PR #1351, opened 2026-03-29
malloc: Fix priority inversion

meese4:
This PR fixes the malloc priority inversion bug described in #1350.


kernel/libc/koslib/malloc.c
```diff
@@ -41,17 +41,16 @@
...
-/* Enable thread locking. Because mutexes actually require malloc themselves,
-   we use spinlocks here. They are also MUCH faster. As a general rule this
-   shouldn't be a big problem. */
-static spinlock_t mALLOC_MUTEx = SPINLOCK_INITIALIZER;
-#define MALLOC_PREACTION   ({ spinlock_lock(&mALLOC_MUTEx); 0; })
-#define MALLOC_POSTACTION  ({ spinlock_unlock(&mALLOC_MUTEx); 0; })
+/* Enable thread locking. Because spin locks can cause priority inversion,
+   we use mutexes here. */
+static mutex_t mALLOC_MUTEx = MUTEX_INITIALIZER;
+#define MALLOC_PREACTION   ({ mutex_lock_irqsafe(&mALLOC_MUTEx); 0; })
```
pcercuei: I'd suggest to wrap this into an `assert()`, as it is never supposed to fail.

pcercuei:
CI complains about `result` being unused when compiling without asserts.

Two options:
- Add a `(void)result;` after the assert to silence the issue;
- The best solution IMHO, return `result` instead of `0`. This introduces a behaviour change but the mutex lock/unlock already return 0 on success, and if those fail, it's probably a better idea to just return the error.

QuzarDC:
Merged with review feedback in #1379 left you as the author @meese4 , and thank you for the contribution :)

PR #1353, opened 2026-03-31
Fix: Format-specific AICA loop boundary in snd_stream

KatanaDreamsDev:
Issue:

    PCM (AICA_SM_16BIT, AICA_SM_8BIT): Requires an inclusive boundary (length). The length - 1 offset drops a sample on every buffer wrap-around, causing a periodic clicking artifact.

    ADPCM (AICA_SM_ADPCM_LS): Requires the length - 1 offset to prevent stream corruption during wrap-around.

Solution:
Apply the offset exclusively for ADPCM streams.

### DIFF
--- a/kernel/arch/dreamcast/sound/snd_stream.c
+++ b/kernel/arch/dreamcast/sound/snd_stream.c
@@ -561,7 +561,13 @@ static void snd_stream_start_type(snd_stream_hnd_t hnd, uint32_t type, uint32_t 
     chan->length = bytes_to_samples(hnd, streams[hnd].buffer_size);
     chan->loop = 1;
     chan->loopstart = 0;
-    chan->loopend = chan->length;
+
+    if (type == AICA_SM_ADPCM_LS) {
+        chan->loopend = chan->length - 1;
+    }
+    else {
+        chan->loopend = chan->length;
+
     chan->freq = freq;
     chan->vol = 255;
     chan->pan = streams[hnd].channels == 2 ? 0 : 128;
--- a/kernel/arch/dreamcast/sound/snd_stream.c
+++ b/kernel/arch/dreamcast/sound/snd_stream.c
@@ -561,7 +561,7 @@ static void snd_stream_start_type(snd_stream_hnd_t hnd, uint32_t type, uint32_t 
     chan->length = bytes_to_samples(hnd, streams[hnd].buffer_size);
     chan->loop = 1;
     chan->loopstart = 0;
-    chan->loopend = chan->length - 1;
+    chan->loopend = chan->length;
     chan->freq = freq;
     chan->vol = 255;
     chan->pan = streams[hnd].channels == 2 ? 0 : 128;


pcercuei:
I think the fix makes sense given the doc we have, the `aica_v08.txt` file says:
```
pos:	Set Loop End Address in samples (0x0001 to 0xffff, Loop End must be > Loop
		Start Address) 
		if no looping is desired this register must be set to the sample length
		in samples before a channel is played
```

EDIT: But @DC-SWAT's explanation makes sense too.

DC-SWAT:
> I think the fix makes sense given the doc we have, the `aica_v08.txt` file says:
> 
> ```
> pos:	Set Loop End Address in samples (0x0001 to 0xffff, Loop End must be > Loop
> 		Start Address) 
> 		if no looping is desired this register must be set to the sample length
> 		in samples before a channel is played
> ```
> 
> EDIT: But @DC-SWAT's explanation makes sense too.

It turns out that for looping, we need to specify -1, but without looping, we don't need -1.
Looping are used in streams, so...

KatanaDreamsDev:
Wait a minute. @pcercuei, your quote from the manual just made something click in my head. I was staring at the 16-bit register limit (0xFFFF), and then I looked back at the ADPCM buffer size clamp.

Guys, I think I just realized why ADPCM was breaking without the -1 in the first place. It’s a literal integer overflow.

ADPCM is 4-bit (2 samples per byte). The code clamps the buffer to exactly 32KB (32 << 10 = 32768 bytes).
32768 bytes * 2 = 65536 samples (0x10000).

Since the AICA loop register is strictly 16-bit, 0x10000 overflows perfectly to 0x0000, silently breaking the playback. Subtracting 1 gave 65535 (0xFFFF), which miraculously fit the register and "fixed" ADPCM... but applying it globally meant it started dropping the last valid sample for standard PCM (since a 64KB PCM16 buffer is only 32768 samples and never overflowed to begin with).

If we want to avoid modifying loop boundaries and dropping samples, we just need to keep the max ADPCM buffer slightly under the 16-bit limit, keeping AICA DMA 32-byte alignment in mind.

For example, clamping it to 30 << 10 (30KB) gives exactly 61440 samples (0xF000).
0xF000 fits beautifully in 0xFFFF.

If we just change the ADPCM limit from 32 << 10 to 30 << 10, the hardware overflow completely disappears. We can safely remove the -1 hack entirely, ADPCM plays flawlessly, and the PCM click is gone.

Unless I'm missing something obvious, I believe this solves the root cause for both formats.

pcercuei:
Why not limit to `0x10000 - 32` then?

KatanaDreamsDev:
If we limit samples to 0x10000 - 32 (65504 samples), the buffer size becomes 32752 bytes.
32752 is not divisible by 32, which breaks AICA DMA alignment.
​We need to subtract 32 from the bytes instead:
32768 - 32 = 32736 bytes (perfectly 32-byte aligned).
That gives exactly 65472 samples (0xFFC0), safely avoiding the 0xFFFF hardware overflow limit.

DC-SWAT:
I tried 30KB without -1 and ADPCM plays well.

KatanaDreamsDev:
Closing this since the fix was implemented by QuzarDC in #1380. The issue is finally resolved.

PR #1354, opened 2026-04-01
New RTC ARCH API

pcercuei:
Add a new ARCH API for the RTC + some cleanups.


pcercuei:
Rebased to master as it had conflicts. No other change.

kernel/arch/dreamcast/include/arch/rtc.h
```diff
@@ -29,93 +33,33 @@ __BEGIN_DECLS
...
+    return dc_boot_time;
+}
 
-/* \cond INTERNAL */
-/* Internally called Init / Shutdown */
-int rtc_init(void);
-void rtc_shutdown(void);
-/* \endcond */
+static inline int arch_rtc_init(void) {
+    dc_boot_time = arch_rtc_unix_secs();
```
gyrovorbis, line 57: Is there a reason this is even specific to Dreamcast?

Like shouldn't the generic, arch-independent layer be able to have this variable get set automatically from `rtc_init()` itself, before delegating to `arch_rtc_init()`? Would deduplicate that logic across platforms.

pcercuei: Something like this? https://github.com/pcercuei/KallistiOS/commit/9a1238b317e3a216fec0e99f4fe40385fea40d10

QuzarDC: @darcagn - in case you already have the GC part ready.

pcercuei:
Rebased to master so that the whitespace CI check passes.

QuzarDC (approved):
Since there wasn't any further feedback forthcoming, it all seemed fine to me. It should be simple to modify if needed for any of the other concerns in the future.

PR #1355, opened 2026-04-01
Bump the LRA toolchain to gcc 16

MastaG:
Also rebased the patch.
Oleg Endo's repo is actually at gcc version 16.0.1. However due to the way we apply patches, they apply to both gcc-16.0.1-dev and gcc-16.0.1-lra.
So I leave the version at 16.0.0 (its a WIP branch anyways).


MastaG:
Putting into draft until Oleg merges the patch into his repo.

darcagn:
Superseded by #1470

PR #1356, opened 2026-04-04
dbglog: Simplify logging facility

pcercuei:
Instead of calling vsnprintf() to process the format string + arguments into a non-reentrant static buffer, that then needs to be protected against concurrent accesses, just rely on the fact that the C library already provides such a mechanism.

Update dbglog() to a simple macro that will basically call printf() for standard messages, below the warning threshold, and fprintf(stderr) for the important messages.

The macro will also check that the requested log level is below the one set to KallistiOS at compilation time, and the one dynamically set with dbglog_set_level().

Note the use of __builtin_choose_expr(), this is to allow using printf() / fprintf(stderr) instead of fprintf(stdout) / fprintf(stderr). The distinction is important, as the compiler will change the calls to printf() with static strings with no format arguments to calls to puts(). On the ARM port of KallistiOS, these strings can then be transmitted to the SH4 side much more efficiently and asynchronously.

### DIFF
--- a/include/kos/dbglog.h
+++ b/include/kos/dbglog.h
@@ -1,7 +1,8 @@
 /* KallistiOS ##version##
 
    kos/dbglog.h
-   Copyright (C)2004 Megan Potter
+   Copyright (C) 2004 Megan Potter
+   Copyright (C) 2026 Paul Cercueil
 
 */
 
@@ -23,9 +24,10 @@
 __BEGIN_DECLS
 
 #include <kos/opts.h>
+#include <stdio.h>
 
 /** \defgroup logging   Logging
-    \brief              KOS's Logging API 
+    \brief              KOS's Logging API
     \ingroup            debugging
 */
 
@@ -37,17 +39,14 @@ __BEGIN_DECLS
     info you want to see (or want your users to see).
 
     \param  level           The level of importance of this message.
-    \param  fmt             Message format string.
-    \param  ...             Format arguments
+    \param  ...             Format string + Format arguments
     \see    dbglog_levels
 */
-void __real_dbglog(int level, const char *fmt, ...) __printflike(2, 3);
-
-/* This wrapper allows for the garbage collection of unneeded debug data */
 #define dbglog(lvl, ...) \
 do { \
-  if ((lvl) <= DBGLOG_LEVEL_SUPPORT) \
-    __real_dbglog(lvl, __VA_ARGS__); \
+    int __dbglog_lvl = (lvl); \
+    if(__dbglog_lvl <= DBGLOG_LEVEL_SUPPORT && __dbglog_lvl <= dbglog_level) \
+        printf(__VA_ARGS__); \
 } while(0)
 
 /** \defgroup   dbglog_levels   Log Levels
@@ -88,6 +87,10 @@ do { \
     \see    dbglog_levels
 */
 void dbglog_set_level(int level);
+
+/** \cond */
+extern int dbglog_level;
+/** \endcond */
 
 __END_DECLS
 
--- a/kernel/exports.txt
+++ b/kernel/exports.txt
@@ -176,7 +176,8 @@ thd_block_now
 #library_close
 
 # Low-level debug I/O
-__real_dbglog
+dbglog_level
+dbglog_set_level
 dbgio_set_irq_usage
 dbgio_enable
 dbgio_disable
--- a/kernel/libc/koslib/dbglog.c
+++ b/kernel/libc/koslib/dbglog.c
@@ -1,23 +1,14 @@
 /* KallistiOS ##version##
 
    dbglog.c
-   Copyright (C)2000,2001,2004 Megan Potter
+   Copyright (C) 2000,2001,2004 Megan Potter
+   Copyright (C) 2026 Paul Cercueil
 */
 
 #include <stdio.h>
 #include <stdarg.h>
-#include <unistd.h>
-#include <string.h>
 
 #include <kos/dbglog.h>
-#include <kos/thread.h>
-#include <kos/dbgio.h>
-#include <kos/fs.h>
-#include <kos/spinlock.h>
-
-/* Not re-entrant */
-static char printf_buf[1024];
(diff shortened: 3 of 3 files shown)


QuzarDC:
I don't see a reason to have the `printf`/`fprintf(stderr)` split in here and it might point to a misunderstanding of the original dbgio fallback. The intent there was to make sure that high priority dbglog messages got through by bypassing the fs layer, not to distinguish them between two different outputs (`stdout`/`stderr`). So that you don't have to worry about high priority messages getting lost if there's something wrong in the fs layer (or with other low level stuff like the mallocing that the printf family require). Not only that, but the `dbgio_write_str` acted as the fallback in case the `fs_write` failed.

I agree generally with the notion of collapsing away the `vsnprintf`. As well as having `stderr` as the target output for debug rather than `stdout`. But it seems that there should still be some mechanism to pipe directly to dbgio, or at the very least an error check to write out to dbgio directly if the `fprintf(stderr)` fails (even if it's just the unformatted string that couldn't be printed).

pcercuei:
I think the dbgio fallback is flawed and doesn't serve much of a purpose.
- Fallback or not, `vsnprintf()` was called, so if a malloc failed in there, the message was silently omitted.
- The return value of `dbgio_write_str` in the fallback is not checked. `dcls_writebuf` can fail, for instance when trying to print a string in interrupt context when its mutex is already locked. ~~I *assume* that when using stdio, if a write() fails, the internal buffer is not flushed, and the string will be printed the next time the stdout/stderr is written to, which is better than nothing.~~
- Neither `fs_write()` nor `pty_write()` need locking, and the latter will directly call `dbgio_write_buffer_xlat`. Given that `dbgio_write_str` is just a tiny wrapper around `dbgio_write_buffer_xlat`, and that we use line buffering by default, they are both functionally the same, and the fallback is pointless.

QuzarDC:
> I think the dbgio fallback is flawed and doesn't serve much of a purpose.
> 
>     * Fallback or not, `vsnprintf()` was called, so if a malloc failed in there, the message was silently omitted.
> 
>     * The return value of `dbgio_write_str` in the fallback is not checked. `dcls_writebuf` can fail, for instance when trying to print a string in interrupt context when its mutex is already locked. ~I _assume_ that when using stdio, if a write() fails, the internal buffer is not flushed, and the string will be printed the next time the stdout/stderr is written to, which is better than nothing.~
> 
>     * Neither `fs_write()` nor `pty_write()` need locking, and the latter will directly call `dbgio_write_buffer_xlat`. Given that `dbgio_write_str` is just a tiny wrapper around `dbgio_write_buffer_xlat`, and that we use line buffering by default, they are both functionally the same, and the fallback is pointless.

Alright, that's fine I guess. At the very least the previous implementation's ignoring the spinlocks would allow errors generated while already calling a dbglog to take precedence, but the printf internals should be more robust and we don't really have dbglog calls that could happen here. Additionally, I was thinking that there was something to tie the interaction between dbglog and dbgio, but I guess not, was just the prefix having me think they should be related.

But why introduce the split between stdout/stderr for the outputs? I'm guessing this could allow more control, but at in an arbitrary way. Would it make more sense, for instance, for the user to be able to set the output fd so that dbglog could be independent of either? I'm not sure how someone would want to leverage that stdout/stderr split here.

pcercuei:
I can leave the stdout/stderr split idea for later and just use stdout for this PR.

The idea with the split was that we may want different output paths for debug/info messages than for warning/error messages, e.g. errors could be displayed on the framebuffer but the debug messages should only go through the debug cable. It was also done to mimic unix behaviour.

include/kos/dbglog.h
```diff
@@ -37,17 +39,16 @@ __BEGIN_DECLS
...
 */
-void __real_dbglog(int level, const char *fmt, ...) __printflike(2, 3);
-
-/* This wrapper allows for the garbage collection of unneeded debug data */
 #define dbglog(lvl, ...) \
 do { \
-  if ((lvl) <= DBGLOG_LEVEL_SUPPORT) \
-    __real_dbglog(lvl, __VA_ARGS__); \
+  if((lvl) <= DBGLOG_LEVEL_SUPPORT && (lvl) <= dbglog_level) { \
+    __builtin_choose_expr((lvl) <= DBG_WARNING, \
```
QuzarDC: [`__builtin_choose_expr`](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html#index-_005f_005fbuiltin_005fchoose_005fexpr) is only available in C, so this breaks dbglog in C++ with errors like:

```
In file included from /home/quzar/dcdev/KallistiOS/include/kos.h:57,
                 from /home/quzar/dcdev/kos-ports/libGL/build/libGL-1.1.0/samples/cubes/main.cpp:7:
/home/quzar/dcdev/kos-ports/libGL/build/libGL-1.1.0/samples/cubes/main.cpp: In function ‘void debugLog(const char*)’:
/home/quzar/dcdev/KallistiOS/include/kos/dbglog.h:48:5: error: ‘__builtin_choose_expr’ was not declared in this scope; did you mean ‘__builtin_cosf32x’?
   48 |     __builtin_choose_expr((lvl) <= DBG_WARNING, \
      |     ^~~~~~~~~~~~~~~~~~~~~
/home/quzar/dcdev/kos-ports/libGL/build/libGL-1.1.0/samples/cubes/main.cpp:127:9: note: in expansion of macro ‘dbglog’
  127 |         dbglog(DBG_KDEBUG, "%s\n", msg);
      |         ^~~~~~

QuzarDC:
> I can leave the stdout/stderr split idea for later and just use stdout for this PR.
> 
> The idea with the split was that we may want different output paths for debug/info messages than for warning/error messages, e.g. errors could be displayed on the framebuffer but the debug messages should only go through the debug cable. It was also done to mimic unix behaviour.

Ok, that makes sense. Found an issue will trying to rebuild all of kos-ports with it though.

pcercuei:
Updated, to get rid of the stdout / stderr distinction. This also dropped `__builtin_choose_expr` so your issue should be fixed.

QuzarDC (approved):

PR #1357, opened 2026-04-04
PTY related filesystem fixes

pcercuei:
This PR addresses a few issues relative to the file descriptors. Until now, the filesystem core would allocate file descriptors starting from 0. This does not work well alongside the C library, which reserves FDs 0/1/2 for stdin/stdout/stderr.

What happened then, is that the PTY driver would open `/pty/sl00` as the FD 0, and dup2 that FD to 1 and 2 (aka. stdout / stderr). As a result, this file descriptor was also used implictly as the stdin, without it being actually dup2'd.

This was worse on the ARM side, as another file is opened as FD 0, causing the PTY driver's `/pty/sl00` to be opened as FD 1. Then the dup2() call that maps it to stdout would actually be a `dup2(1, 1)`, causing the file descriptor to be closed because of a bug in the dup2 implementation.

Address these issues by handling the case where oldfd == newfd in dup2(), and start the FD values at 3 in the core filesystem code.

### DIFF
--- a/kernel/fs/fs.c
+++ b/kernel/fs/fs.c
@@ -194,7 +194,7 @@ static int fs_hnd_assign(fs_hnd_t *hnd) {
 
     fs_hnd_ref(hnd);
 
-    for(i = 0; i < FD_SETSIZE; i++) {
+    for(i = 3; i < FD_SETSIZE; i++) {
         fs_hnd_t *old = NULL;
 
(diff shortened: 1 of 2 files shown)


QuzarDC (approved):
That all tracks, and is a better way of specifying it, as well as avoiding accidental STDIN/OUT/ERR redirection if pty is disabled.

PR #1358, opened 2026-04-06
Add Remote Procedure Call (RPC) API

pcercuei:
Introduce a new API that can be used for sending RPC messages to a remote processor, and handling RPC messages from a remote processor.

This will be used for the communication between the SH4 and the ARM on the Dreamcast, when the ARM will be running KallistiOS. It will allow the ARM to output debug messages, access files, or anything really.

It will also allow the SH4 and the ARM to work better together; for instance, by having the ARM processor responsible for managing the sound RAM, and have commands to request memory.

### DIFF
--- a/include/kos/rpc.h
+++ b/include/kos/rpc.h
@@ -1,0 +1,159 @@
+/* KallistiOS ##version##
+
+   include/kos/rpc.h
+   Copyright (C) 2026 Paul Cercueil
+*/
+
+/** \file    kos/rpc.h
+    \brief   Remote procedure call support
+
+    This file contains an API that can be used for sending RPC messages to a
+    remote processor and handling RPC messages from a remote processor.
+
+    \author Paul Cercueil
+*/
+#ifndef __KOS_RPC_H
+#define __KOS_RPC_H
+
+#include <sys/cdefs.h>
+__BEGIN_DECLS
+
+#include <kos/regfield.h>
+#include <stdint.h>
+
+/** \brief  RPC command code.
+
+    A RPC command code identifies a command that can be executed on the remote
+    processor.
+
+    \headerfile kos/rpc.h
+*/
+typedef unsigned char rpc_code_t;
+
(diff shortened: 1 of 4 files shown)


include/kos/rpc.h
```diff
@@ -0,0 +1,159 @@
...
+    A RPC command code identifies a command that can be executed on the remote
+    processor.
+
+    \headerfile kos/rpc.h
+*/
+typedef unsigned char rpc_code_t;
+
+/** \brief  Special command code that represents a response to a command
+    received from the remote. */
+#define RPC_CMD_RESP    0xff
```
DC-SWAT, line 35: If there's a memory issue on G2, you'll always read 0xff from there. Maybe it makes sense to choose a different value?

pcercuei: Why would there be a memory issue?

DC-SWAT: > Why would there be a memory issue?

This is of course a rare case, but I encountered it once when I was experimenting with initialization and firmware for AICA. It costs nothing to replace the value, why not do it.

pcercuei: Because if there's a rare case of memory issue I want to know when it happens, and why it happens. I'm not going to add workarounds for things I don't know about and haven't seen myself. Besides, if there's a chance of reading 0xff there, that means there's a chance of reading 0xff everywhere else on the sound RAM, so adding a small workaround here won't change anything.

DC-SWAT: This is not a workaround in my opinion, it is a more accurate error check.
I'm not insisting, just making a suggestion.

QuzarDC:
So this all looks fine, but I don't think it's ripe without a use for it. Since it doesn't really interact with anything it shouldn't be any trouble to rebase once there'd be a use for it.

pcercuei:
@QuzarDC It's a dependency for KOS on the AICA. I can submit these when sending the big PR, if you prefer.

PR #1359, opened 2026-04-07
Replace frame-pointer stack tracing with heuristic SP scanning

andressbarajas:
GCC does not reliably maintain r14-based frame pointers, even with `-fno-omit-frame-pointer`, resulting in incomplete or empty stack traces. This has been the case for many years. 

Using the algorithm that SKMP proposed here: https://discord.com/channels/488332893324836864/614912396460556298/1347632664408101026 I replace the FP-chain walker with a stack pointer scanner that walks upward from SP, validating saved PR values against preceding BSR/BSRF/JSR instructions. This removes the `FRAME_POINTERS` build-time dependency so stack traces work unconditionally! 

I also expanded the stracktrace example to demonstrate recursive and indirect (function pointer) call scenarios.

### DIFF
--- a/kernel/arch/dreamcast/kernel/stack.c
+++ b/kernel/arch/dreamcast/kernel/stack.c
@@ -13,12 +13,114 @@
 #include <arch/arch.h>
 #include <arch/stack.h>
 #include <stdint.h>
+#include <stdbool.h>
 
 static uintptr_t arch_stack_16m_dft = 0x8d000000;
 static uintptr_t arch_stack_32m_dft = 0x8e000000;
 
 extern uintptr_t arch_stack_16m __attribute__((weak,alias("arch_stack_16m_dft")));
 extern uintptr_t arch_stack_32m __attribute__((weak,alias("arch_stack_32m_dft")));
+
+/* Resolve the scan end for a stack walk.
+
+   When SP falls within the current thread's stack, use the actual end of that
+   stack so repeated unwind steps remain within one fixed stack window.
+   Otherwise, fall back to a conservative kernel-stack-sized scan. */
+static uintptr_t arch_stk_scan_end(uintptr_t sp) {
+    uintptr_t scan_end = sp + THD_KERNEL_STACK_SIZE;
+
+    if(thd_current && thd_current->stack && thd_current->stack_size) {
+        uintptr_t stack_base = (uintptr_t)thd_current->stack;
+        uintptr_t stack_end = stack_base + thd_current->stack_size;
+
+        if(sp >= stack_base && sp < stack_end)
+            scan_end = stack_end;
+    }
+
+    if(scan_end > _arch_mem_top)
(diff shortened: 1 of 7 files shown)


kernel/arch/dreamcast/kernel/stack.c
```diff
@@ -28,43 +102,45 @@ void arch_stk_setup(kthread_t *nt) {
 /* Do a stack trace from the current function; leave off the first n frames
    (i.e., in assert()). */
 __noinline void arch_stk_trace(int n) {
-    arch_stk_trace_at(arch_get_fptr(), n + 1);
+    uintptr_t sp;
+
+    __asm__ __volatile__("mov r15,%0\n" : "=r"(sp));
```
pcercuei: Slight preference for `register uintptr_t sp asm("sp")`

andressbarajas: Done

QuzarDC:
Just for reference, the post from discord was also codified here: https://gitlab.com/skmp/dca3-kos/-/blob/main/kernel/arch/dreamcast/kernel/stack.c in the dca3 fork.

andressbarajas:
Reviewed the original SKMP source and see that Im doing one thing wrong so Im putting this into draft.

andressbarajas:
The thing I was doing wrong was that each unwind step could start a new scan window from the updated SP, so the stack walk could scan outside the current thread's frames.

The fix was to keep the scan bounded to one fixed end address for the whole walk. When possible, that end comes from the current thread’s real stack (`thd_current->stack + thd_current->stack_size`); otherwise it falls back to `THD_KERNEL_STACK_SIZE`.

That keeps the trace anchored to the actual stack in normal crash cases, with a safe fallback for other callers.

pcercuei (dismissed):
Tested and works fine.

QuzarDC:
There's two relatively small other cleanups that should go with this:
1) The `environ.sh` sample should be updated to not present the frame pointers option, and add `-fomit-frame-pointer`
2) `startup.S` no longer needs to set r14 to -1 to signal the last frame.

andressbarajas:
@QuzarDC I agree with #2 but was unsure about 1 because that will be used across ports to other consoles and it might work for them but just not for Dreamcast. Maybe a dreamcast specific note? or I can make the change you requested and let the next person handle it. Up to you.

darcagn:
> @QuzarDC I agree with #2 but was unsure about 1 because that will be used across ports to other consoles and it might work for them but just not for Dreamcast. Maybe a dreamcast specific note? or I can make the change you requested and let the next person handle it. Up to you.

The current sample `environ.sh` should be considered Dreamcast-specific. For other consoles, a different setup would be necessary. There are too many differences in configuration to have one universal `environ.sh`.

andressbarajas:
@QuzarDC Done

QuzarDC (approved):

Issue #797, opened 2024-10-08
DC-Tool/DC-Load Version 3.0 Wishlist


gyrovorbis:
Okay, guys, it should probably be clear to everyone at this point that we are overdue for a big update to DC-Tool and DC-Load to address some deficiencies and add some new features. There has been so much talk of new features and so many new requests, that I've decided to start this issue as a way to capture it all. 

There is a lot of discussion about whether this next release should be a total overhaul/rewrite or something incremental... That's not really the point of discussion here. We're just gathering a wish-list, and I'm just going to call it "DC-Tool 3.0" regardless of whether it's something new or not. :)

Bugs to Fix:
- ISO Redirection
- EXTREMELY slow `printf()`/`stdout` performance
- Fairly slow FILE performance
- chroot inconsistencies across platforms (goes to root directory on Linux, which requires sudo)

Right now, here's the feature wish-list:
- Command-line argument support
- Return value support
- Environment variables accessible from host
- `stdin`, `scanf()`, etc implemented by taking input from host
- `exec()` family from C stdlib
- command-line argument for KOS to automatically sync the DC's RTC to the host's
- command-line argument for KOS dbglog verbosity level

Please feel free to respond with your own ideas!

Tchan0:
Naomi / System SP support :)
Can easily be done - just requires a hardcoding to VGA, like I do in my JinGasa custom Bios (https://github.com/Tchan0/JinGasa)
Before compiling, hardcode dcload-ser to VGA:

    in target-src/dcload/video.s - _check_cable: change the 3 first commands:
    MOV #0,R0 !by T - for Naomi & System SP - Tmp test, since KOS also does not know how to check the cable
    RTS ! note: I commented out the 3 next commands, so that the size of the bin remains the same
    NOP
    ! set PORT8 and PORT9 to input
    ! mov.l porta,r0
    ! mov.l pctra_clr,r2
    ! mov.l @r0,r1

andressbarajas:
Loving the wishlist so far.  I have a few comments. Do we update the KOS repo version of dcload-ip to match Sizious's in one fell swoop since its the one just about everyone uses anyway? The understanding of moops changes can be in the next phase which is paying the technical debt by cleaning/organizing the code.  While cleaning we can understand the code better. I think a good plan that will make it possible to possible to even have a DC-Tool/DC-Load 3.0 release within the next year is:

1. Update dcload-ip repo to match Sizious.
[...]

PR #1361, opened 2026-04-08
Fix dcload stdin for stdio input

andressbarajas:
This enables taking STDIN input from the dcload console without a dcload update. Takes care of one of the requests from this issue: https://github.com/KallistiOS/KallistiOS/issues/797

Tested against these methods:

read(STDIN_FILENO, buf, n)
getchar()
fgetc(stdin)
fgets(buf, sizeof(buf), stdin)
scanf("%d")
scanf("%s")
scanf("%c")
scanf(" %c")
fread(buf, 1, n, stdin)
ungetc()


kernel/fs/fs_pty.c
```diff
@@ -512,8 +512,12 @@ static ssize_t pty_read(void *h, void *buf, size_t bytes) {
     }
 
     /* Special case the unattached console */
-    if(ph->id == 0 && !ph->master && ph->other->refcnt == 0)
-        return pty_read_serial(fdobj, ph, buf, bytes);
+    if(ph->id == 0 && !ph->master && ph->other->refcnt == 0) {
+        if(!strcmp(dbgio_dev_get(), "fs_dcload"))
```
pcercuei: I think you can drop that check, call `dbgio_read_buffer` unconditionally and trash the `pty_read_serial` function.

Doing so would get rid of the last references to `dbgio_write()` / `dbgio_read()`, which can consequently be trashed as well, as well as the dbgio's .read / .write callbacks. In this PR if you feel like it, or in a subsequent PR by anyone motivated.

`kernel/arch/dreamcast/kernel/ser_console.c` use those as well but that file is never compiled, so we can consider it dead code and remove it as well.

That sounds like a great cleanup!

pcercuei (changes requested):

kernel/fs/fs_pty.c
andressbarajas: ~~Im good with calling dbgio_read_buffer unconditionally and trashing the pty_read_serial but when I tried to remove references to dbgio_write() / dbgio_read()/.read/.write callbacks it also affected scif.c because it uses those too and Im not too sure about removing that. @DC-SWAT What do you think?~~

~~`scif.c:415:6: error: ‘dbgio_handler_t’ {aka ‘struct dbgio_handler’} has no member named ‘read’
  415 |     .read = scif_read,
      |      ^~~~
scif.c:416:6: error: ‘dbgio_handler_t’ {aka ‘struct dbgio_handler’} has no member named ‘write’
  416 |     .write = scif_write,
      |      ^~~~~
scif.c:416:14: error: initialization of ‘int (*)(const uint8_t *, int,  int)’ {aka ‘int (*)(const unsigned char *, int,  int)’} from incompatible pointer type int (*)(int)’ [-Wincompatible-pointer-types]
  416 |     .write = scif_write,
      |              ^~~~~~~~~~
scif.c:416:14: note: (near initialization for ‘dbgio_scif.write_buffer’)
scif.c:308:5: note: ‘scif_write’ declared here
  308 | int scif_write(int c) {
`~~

EDIT: @pcercuei Nvm. Done.

andressbarajas:
@DC-SWAT I touched exports.txt. Do you use dbgio_write() or dbgio_read() for anything?

pcercuei (approved):

pcercuei:
Whitespace CI can be ignored, the PR cleans up some whitespace issues.

PR #1362, opened 2026-04-11
Add Workqueues

pcercuei:
Add support for work queues, use them in network code, and drop jobs support in worker threads.

A work queue is a thread that will execute tasks (aka. jobs) that are enqueued by client code, at a predetermined moment in time. Multiple jobs can be enqueued. Once a job is executed, it is removed from the execution queue.

Use a work queue to replace the network thread.

Since work queues support the "job list" functionality that worker threads already had, without being used anywhere, drop jobs support from the worker threads, to reduce functionality duplication.

I don't think worker thread jobs were used anywhere, so I decided to just drop them without tagging them as deprecated.


pcercuei:
FYI this will conflict with #1348, but each one can be rebased on top of the other quite easily.

include/kos/workqueue.h
```diff
@@ -0,0 +1,129 @@
...
+/** \brief       Create a new work queue.
+    \relatesalso workqueue_t
+
+    This function will create a new work queue.
+
+    \return                 The new work queue on success, NULL on failure.
+
+    \sa workqueue_destroy
+*/
+workqueue_t *workqueue_create(void);
```
gyrovorbis, line 66: If `workqueue_t` is used as an opaque pointer which is internally allocated/freed, the API really needs a way to associate a `void*` userdata with it... Because otherwise you will not be able to "extend" the workqueue to add more fields to it AND you will not be able to write a C++, Rust, whatever OOP wrapper around a workqueue that delegates work to virtual method or rust trait, as it will not be able to store the wrapper object within the queue.

Idc whether `workqueue_create()` takes a `void* userdata` or whether we add a `workqueue_set_userdata()`/`workqueue_get_userdata()`, but I think it's paramount for making this API as extensible and robust as possible.

include/kos/workqueue.h
```diff
@@ -0,0 +1,129 @@
...
+    \brief   Opaque structure describing one work queue.
+*/
+typedef struct workqueue workqueue_t;
+
+/** \struct  workqueue_job_t
+    \brief   Structure describing a job for the work queue.
+*/
+typedef struct workqueue_job {
+    /** \brief  Routine to call. */
+    void (*cb)(struct workqueue_job *job);
```
gyrovorbis, line 47: Are we sure that we shouldn't also be passing the opaque `workqueue_t` handle back to the job's execution routine? That way if it's some periodic shit, it can reschedule itself within its own handler? Or you can store some static/shared shit within the workqueue itself that the handlers access/modify/update?

include/kos/workqueue.h
gyrovorbis: Also... perhaps we want a public, user-facing `workqueue_attrib_t` which has basic thread attributes like label + stack size, so that we have optional fine-grained controls over these queues? Could be a separate constructor, could be an arg of the default one that just uses internal defaults if passed `NULL`?

kernel/thread/workqueue.c
```diff
@@ -0,0 +1,133 @@
...
+        prev = elm;
+    }
+
+    if(!elm)
+        STAILQ_INSERT_TAIL(&wq->jobs, job, entry);
+
+    cond_signal(&wq->cond);
+}
+
+void workqueue_cancel(workqueue_t *wq, workqueue_job_t *job) {
```
gyrovorbis, line 111: This might need more robustness.

Imagine you added something to a queue like a timer that will run at some point in the future, then later on, you want to cancel it, but you don't know if it has already been canceled or not. Is this going to gracefully handle the scenario where it's not even in the queue anymore? Maybe return a `bool` that's `false` if it wasn't even there to cancel?

kernel/thread/workqueue.c
```diff
@@ -0,0 +1,133 @@
...
+    wq->thd = thd_create_ex(&workqueue_attrs, workqueue_thread, wq);
+    if(!wq->thd) {
+        free(wq);
+        return NULL;
+    }
+
+    return wq;
+}
+
+void workqueue_enqueue(workqueue_t *wq, workqueue_job_t *job) {
```
gyrovorbis, line 85: Should we at least assert if the thing you're trying to enqueue is already enqueued? Or.... is that a legit way to like tweak its timeout and reschedule it?

include/kos/workqueue.h
pcercuei: @gyrovorbis We can add `workqueue_set_userdata()` when it's actually needed. Right now virtually nothing in KOS does that. The only times we store user pointers is for passing them to callbacks.

include/kos/workqueue.h
pcercuei: I *could* pass the `workqueue_t` there, I just don't think it's needed. "Periodic shit" already works, see the "net: Switch to using a work queue" commit.

kernel/thread/workqueue.c
pcercuei: I don't see how that's a problem. Whether or not the job was in  the queue, the behaviour of `workqueue_cancel()` is always the same, once the call finishes the requested job won't be in the queue anymore.

For your example you'd stop the timer unconditionally before removing the job, even if it already expired.

kernel/thread/workqueue.c
pcercuei: To reschedule you can cancel + enqueue.

Walking the jobs list to check whether a job has already been enqueued is costly, so I'd prefer to not do that. I could add a "enqueued" field to the job struct, but that's another field with a "private stuff do not touch" in a public API...

include/kos/workqueue.h
pcercuei: Done.

pcercuei:
Rebased as it conflicted with the BBA+net PR.

The callback now gets a pointer to the `workqueue_t` too.

kernel/thread/workqueue.c
gyrovorbis: Oh. My bad. Sounds good.

gyrovorbis (approved):
Yes, okay. Spent a considerable amount of time reviewing the architecture and API here, plus @pcercuei and I had a brainstorming session in Discord where I created a proof-of-concept user-leve C++ layer around it, extending its functionality at the app level, outside of the kernel, which is what I adamantly wanted the API to be robust enough to support.

I can confirm that everything that for the kernel, its way more than good enough, plus it is also sufficient for my needs beyond the kernel as a C++ user.

I kind of would hope that eventually there could be an example of using this workqueue construct and the derived oneshot timers from a user-app perspective eventually, because I insist that it can be a great way to solve problems for users at the app level as well.

Anyway, approved

PR #1363, opened 2026-04-12
CI: Ignore removed whitespaces in whitespace check

pcercuei:
A PR that removes errorneous whitespaces shouldn't trigger the whitespace check.
    
Update the whitespace check to only consider the whitespaces that were added by a PR.


gyrovorbis (approved):
My bad. Should've approved this forever ago. lol.

PR #1364, opened 2026-04-16
pthread: Add pthread_cleanup_{push,pop}

pcercuei:
Add the functions pthread_cleanup_push() and pthread_cleanup_pop().

These functions can be used to register (and respectively unregister) cleanup callbacks, that will be called when pthread_exit() is called (and *not* when the thread routine returns normally).

I don't actually have much of an use for these functions - however, autotools' "ax_pthread.m4" script, used to detect pthread, will try to compile a test program that happens to call these two functions. Adding these functions to our libpthread means that autotools will properly detect pthread now.

### DIFF
--- a/addons/libpthread/pthread_exit.c
+++ b/addons/libpthread/pthread_exit.c
@@ -2,12 +2,38 @@
 
    pthread_exit.c
    Copyright (C) 2023 Lawrence Sebald
+   Copyright (C) 2026 Paul Cercueil
 */
 
 #include "pthread-internal.h"
 #include <pthread.h>
 #include <kos/thread.h>
 
+struct pthread_cleanup_cb {
+    void (*cb)(void *d);
+    void *arg;
+};
+
+static _Thread_local struct pthread_cleanup_cb cleanups[16];
+static _Thread_local unsigned int nb_cleanups;
+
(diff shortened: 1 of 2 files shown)


addons/libpthread/pthread_exit.c
```diff
@@ -2,12 +2,38 @@
...
 #include "pthread-internal.h"
 #include <pthread.h>
 #include <kos/thread.h>
 
+struct pthread_cleanup_cb {
+    void (*cb)(void *d);
+    void *arg;
+};
+
+_Thread_local struct pthread_cleanup_cb cleanups[16];
```
gyrovorbis:
```suggestion
_Thread_local static struct pthread_cleanup_cb cleanups[16];
```
We don't want this to have external linkage here, and `thread_local` doesn't imply internal linkage/static.

gyrovorbis (commented):
Awesome and good call on this one. Had one teensy nitpick. Also lol, good use of TLS.

addons/libpthread/pthread_exit.c
pcercuei: Good point.

pcercuei:
Updated, thanks.

QuzarDC (approved):

PR #297, opened 2023-09-08
Additional debugging capabilities for GDB stub

gyrovorbis:
Our GDB stub is actually really good and has saved my ass and made debugging easier on multiple occasions... Since a bunch of us are using it so much with command-line GDB or our IDEs, I've decided to look into fleshing it out further and adding support for other advanced GDB features such as:

- Multithreading support
- Command-line argument support
- TLS data support

I've also taken the liberty of cleaning up the file which had horrible formatting issues.

PR #1365, opened 2026-04-16
GDB stub rewrite

andressbarajas:
### PR: Rewrite and Modularize GDB Stub for Dreamcast

This PR rewrites the legacy `gdb_stub.c` into a modular implementation. It addresses historical bugs and adds modern debugging features. This replaces this PR: https://github.com/KallistiOS/KallistiOS/pull/297

#### Bug Fixes
- **RLE Encoding**: Fixed infinite loop in run-length encoding logic by adding bounds checking (`runlen == 99`).
- **Breakpoint Handling**: Enabled breakpoints and single-stepping; avoided placing breakpoints in delay slots for SH4 instructions.
- **Cache Management**: Added instruction cache flushing after memory modifications to ensure consistency.

#### New Functionality
- **Modular Architecture**: Split `gdb_stub.c` into 10 focused files (`gdb.c`, `gdb_break.c`, etc.) for better organization.
- **Advanced Packet Support**: Added `v` packet handling (e.g., `vCont` for thread-aware execution control, `vKill` for process termination).
- **Thread-Aware Debugging**: Support for multi-threading with thread selectors (`GDB_THREAD_ANY`, `GDB_THREAD_ALL`).
- **Extended Registers**: Improved CPU register handling, including floating-point registers.
- **Query Commands**: Support for GDB query packets (`qSupported`, `qAttached`) for better integration.
- **Enhanced Control**: Robust software breakpoints, single-stepping, and exception handling.


pcercuei:
The awk script doesn't like your change on the AUTHORS file, try with this:
```diff
diff --git a/kernel/make_authors.awk b/kernel/make_authors.awk
index 4ee8981f1..03e28379d 100644
--- a/kernel/make_authors.awk
+++ b/kernel/make_authors.awk
@@ -23,12 +23,11 @@ BEGIN {
         phase = 2;
         print "static const char authors[] = ";
     }
-    else if($0 == "Files with Specific licenses:") {
-        print ";\n";
-        exit;
-    }
     else if(phase == 2 && $0 != "") {
         gsub(/"/, "\\\"", $1)
         print "\"Copyright (C)" $2, $1 "\\n\"";
     }
 }
+END {
+    print ";\n";
+}
```

andressbarajas:
Thanks I didn't think to compile after modifying the AUTHORS file. Your changes worked.

pcercuei:
@andressbarajas if you can (and know how) please merge the commit modifying make_authors.awk into the one removing the old license from the AUTHORS file. The reason is that between the two commits KOS is not buildable and it'd be better to avoid that.

andressbarajas:
@pcercuei Makes sense. Done.

kernel/arch/dreamcast/kernel/gdb/gdb.c
```diff
@@ -0,0 +1,243 @@
...
+}
+
+/*
+   Triggers a software breakpoint using TRAPA #0xFF.
+
+   Typically called at the start of a program to establish a connection
+   with the debugger, or used to halt execution and enter the debugger manually.
+*/
+void gdb_breakpoint(void) {
+    __asm__("trapa	#0xff"::);
```
gyrovorbis: I'm pretty sure this has to be `volatile`, otherwise there are no actual side-effects here, and it risks being optimized away, especially with LTO.

gyrovorbis (commented):
Got a shitload of testing to do, but goddamn, this looks very nice in terms of feature-set.

kernel/arch/dreamcast/kernel/gdb/gdb.c
andressbarajas: I thought so too but its not what legacy had: `#define BREAKPOINT()   __asm__("trapa	#0xff"::);` so I stuck with it and haven't had an issue using it in VSCode. I haven't played around much with LTO though so I'll make the change.

kernel/arch/dreamcast/gdb/gdb_query.c
```diff
@@ -0,0 +1,161 @@
...
+
+typedef struct {
+    char *out;
+    size_t remaining;
+    bool first;
+} thread_list_state_t;
+
+static size_t align_to(size_t value, size_t alignment) {
+    return (value + (alignment - 1)) & ~(alignment - 1);
+}
```
pcercuei, line 42: Note that we have an `__align_up(val, align)` macro that you could use.

kernel/arch/dreamcast/gdb/gdb_query.c
```diff
@@ -0,0 +1,161 @@
...
+    const size_t tbss_size = (size_t)(&_tbss_size);
+    size_t align = 8;
+
+    if(tdata_size && (size_t)_tdata_align > align)
+        align = (size_t)_tdata_align;
+    if(tbss_size && (size_t)_tbss_align > align)
+        align = (size_t)_tbss_align;
+
+    return align_to(sizeof(gdb_tcbhead_t), align);
+}
```
pcercuei, line 71: I'm not fond of this being in the GDB code. You also redefine structs (gdb_tcbhead_t) that's already in our TLS code. Can you move this function to `kernel/arch/dreamcast/kernel/tls_static.c`?

pcercuei (changes requested):
I skimmed through the PR (due to the amount of changes), overall looks like a fantastic change. Two comments on the code though.

kernel/arch/dreamcast/kernel/gdb/gdb.c
```diff
@@ -0,0 +1,293 @@
...
+    gdb_handle_exception(EXC_TRAPA);
+}
+
+/* This function will generate a breakpoint exception.  It is used at the
+   beginning of a program to sync up with a debugger and can be used
+   otherwise as a quick means to stop program execution and "break" into
+   the debugger. */
+
+void gdb_breakpoint(void) {
+    __asm__ __volatile__("trapa	#0xff"::);
```
pcercuei: One last thing - the tab check doesn't like this line as there's a tab character between "trapa" and "#0xff", you could replace it with a space and the CI check should pass.

pcercuei:
This actually does not work at all, the GDB code is not being compiled at all. Something to do with the subdirs I think.

To get it to compile I actually moved the "gdb" folder one directory up, alongside "kernel" and friends. Which is probably better than having it hidden in "kernel" anyway.

Then the GDB code fails to build on a missing include of `<arch/tls_static.h>` in `gdb_query.c`, I added it and the GDB code is now properly compiled and added to the libkallisti.a.

andressbarajas:
> This actually does not work at all, the GDB code is not being compiled at all. Something to do with the subdirs I think.
> 
> To get it to compile I actually moved the "gdb" folder one directory up, alongside "kernel" and friends. Which is probably better than having it hidden in "kernel" anyway.
> 
> Then the GDB code fails to build on a missing include of `<arch/tls_static.h>` in `gdb_query.c`, I added it and the GDB code is now properly compiled and added to the libkallisti.a.

Sorry about that. I'll make those changes today

kernel/arch/dreamcast/gdb/gdb_mem.c
```diff
@@ -32,6 +39,10 @@ static bool is_valid_memory_address(uintptr_t addr, bool is_write) {
     if(arch_valid_address(normalized) || arch_valid_text_address(normalized))
         return true;
 
+    if(!is_write && normalized >= MEM_AREA_FIRMWARE_BASE &&
+            normalized < MEM_AREA_FIRMWARE_BASE + MEM_AREA_FIRMWARE_SIZE)
+        return true;
+
```
pcercuei, line 45: You hid this in an unrelated commit, can you shave it off onto its own commit?
And rebase on master. Then I think the PR is good to go for me.

pcercuei (changes requested):

kernel/arch/dreamcast/gdb/gdb_mem.c
andressbarajas: Done: https://github.com/KallistiOS/KallistiOS/pull/1365/changes/20eba72eeec0160c3841c9534b3a0a48390d1464

pcercuei (approved):

PR #1366, opened 2026-04-16
Fix and update pvr_set_presort_mode()

pcercuei:
`pvr_set_presort_mode()` was just not working, as it would call `pvr_init_tile_matrix()`, causing the tile matrix offset to be modified to an invalid value.

Update pvr_init_tile_matrix() so that the tile matrix offset is not modified, and update pvr_set_presort_mode() so that it can be used at any time during the scene, and not just at the very beginning.

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
@@ -30,6 +30,8 @@
 
 #define LIST_ENABLED(i) (pvr_state.lists_enabled & BIT(i))
 
+#define PVR_TILE_MATRIX_HEADER_SIZE 0x48
+
 
 /* Fill Tile Matrix buffers. This function takes a base address and sets up
    the rendering structures there. Each tile of the screen (32x32) receives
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/hardware/pvr/pvr_buffers.c
```diff
@@ -56,7 +56,7 @@ static void pvr_init_tile_matrix(int which, bool presort) {
     */
 
     /* Header of zeros */
-    vr += BYTES_TO_WORDS(buf->tile_matrix);
+    vr += BYTES_TO_WORDS(buf->tile_matrix - 0x48);
```
andressbarajas, line 59: Not sure if this is too nitpicky so feel free to ignore but maybe define the magic value 0x48 somewhere. Its kinda clear from the comments what the value is(TILE_MATRIX_HEADER_SIZE).

pcercuei:
Updated to use a `PVR_TILE_MATRIX_HEADER_SIZE` macro.

andressbarajas (approved):

QuzarDC (approved):

PR #1371, opened 2026-04-21
nmmgr cleanups

pcercuei:
Check for the return code of a `mutex_lock_irqsafe()`, and switch to using a single-linked list.

### DIFF
--- a/kernel/exports/nmmgr.c
+++ b/kernel/exports/nmmgr.c
@@ -39,7 +39,7 @@ nmmgr_handler_t * nmmgr_lookup(const char *fn) {
     size_t          cur_len = 0, tmp_len;
 
     /* Scan the handler table and look for the best path match */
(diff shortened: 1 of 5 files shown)


kernel/exports/nmmgr.c
```diff
@@ -82,7 +82,8 @@ int nmmgr_handler_remove(nmmgr_handler_t *hnd) {
     nmmgr_handler_t *c, *tmp;
     int rv = -1;
 
-    mutex_lock_irqsafe(&mutex);
+    if(mutex_lock_irqsafe(&mutex) < 0)
```
QuzarDC, line 85: Maybe it and a few other of our primitives should get `__attribute__ ((warn_unused_result))` added? There are multiple other cases where this isn't being done.

QuzarDC (approved):

PR #1376, opened 2026-04-28
arch_stk_trace_at: Add heuristic-based stack trace fallback

skmp:
Not as accurate as using frame pointers, has helped us find many bugs in dca.


skmp:
Closing in favor of #1359

PR #1377, opened 2026-04-28
spu: Enhance reset and support for extended RAM.

DC-SWAT:
- Preserve AICA RAM mode and stereo settings when resetting channels
- Reset additional channel registers to sane defaults
- Use extended SPU RAM size on non-retail hardware

### DIFF
--- a/kernel/arch/dreamcast/hardware/spu.c
+++ b/kernel/arch/dreamcast/hardware/spu.c
@@ -2,10 +2,12 @@
 
    spu.c
    Copyright (C) 2000, 2001 Megan Potter
-   Copyright (C) 2023, 2024 Ruslan Rostovtsev
+   Copyright (C) 2023, 2024, 2026 Ruslan Rostovtsev
  */
 
 #include <kos/thread.h>
+#include <kos/regfield.h>
+#include <arch/arch.h>
 #include <dc/spu.h>
 #include <dc/g2bus.h>
 #include <dc/sq.h>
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/hardware/spu.c
```diff
@@ -326,18 +340,27 @@ static void spu_cdda_init(void) {
...
+                (val & ~0x800f) | (volume & 0xf) | (stereo ? 0 : 0x8000));
 }
 
 /* Initialize the SPU; by default it will be left in a state of
    reset until you upload a program. */
 int spu_init(void) {
+    bool is_retail = hardware_sys_mode(NULL) == HW_TYPE_RETAIL;
+
+    /* Set the RAM mode (2MB or 8MB) and default to stereo/min volume */
+    g2_write_32(SNDREGADDR(0x2800), is_retail ? 0 : BIT(9));
```
pcercuei, line 357: Both the Naomi and the `Set5.xx` devkit have 8 MiB of sound RAM?

DC-SWAT: > Both the Naomi and the `Set5.xx` devkit have 8 MiB of sound RAM?

Good question. As far as I remember, there are two versions: one has memory like the Naomi, the other like the Dreamcast. But I don't know how to detect them; I don't have either one.
Maybe @Tchan0  can help with it.

QuzarDC: > > Both the Naomi and the `Set5.xx` devkit have 8 MiB of sound RAM?
> 
> Good question. As far as I remember, there are two versions: one has memory like the Naomi, the other like the Dreamcast. But I don't know how to detect them; I don't have either one. Maybe @Tchan0 can help with it.

Would it make sense to simply do a R/W test like is done for checking main RAM size?

DC-SWAT: > > > Both the Naomi and the `Set5.xx` devkit have 8 MiB of sound RAM?
> > 
> > 
> > Good question. As far as I remember, there are two versions: one has memory like the Naomi, the other like the Dreamcast. But I don't know how to detect them; I don't have either one. Maybe @Tchan0 can help with it.
> 
> Would it make sense to simply do a R/W test like is done for checking main RAM size?

It's probably possible. But this needs to be done centrally, since we also only check RETAIL to determine the amount of video memory.
Something like `hardware_mem_mode()`.

DC-SWAT: So, maybe we'll do that later? Anyway, the video memory is detected the same way, and I don't think there are many Set5.xx owners which launch something on it from KOS software...

DC-SWAT:
Just rebased without changes.

kernel/arch/dreamcast/hardware/spu.c
Tchan0: > Both the Naomi and the `Set5.xx` devkit have 8 MiB of sound RAM?

Naomi 1, Naomi 2 and System SP have 8 MiB.
(just checked my System SP, it has sound ram chip Samsung K4S641632H-TC75, which is a 8 MiB chip)

As for the Set5 devkits:
- the big majority of them have standard DC specs
- on very very rare occasions, you can find a devkit with 32 MiB RAM, but that doesn't say anything about the sound ram. I checked my devkit, it has 32 MiB, but only 2 MiB sound RAM (nec d4516161AG5). Doesn't mean all were like that of course.

Standalone Soundboxes:
- default 2 MiB, but seems you could ask to upgrade them to 8 MiB
- it seems that sound boxes with 2 AICA's in them (unobtanium) had  2 x 8 MiB RAM
- there was a dip switch I think to switch between 2 MiB and 8 MiB (of course not recommended to use on a 2 MiB soundbox), I imagine that would set the RAM size bit in 0x00702800 / 0x00802800 (not tested)

QuzarDC: > So, maybe we'll do that later? Anyway, the video memory is detected the same way, and I don't think there are many Set5.xx owners which launch something on it from KOS software...

Absolutely right and I'll make sure to keep in mind to try to check these out next time I get mine set up.

QuzarDC (approved):

PR #1379, opened 2026-05-12
malloc: Fix priority inversion

QuzarDC:
Incorporates the feedback from #1351 in a clean commit.

If the application is multithreaded with high and
low priority threads, and the low priority thread
gets preempted during a malloc/free, then the high priority thread will block indefinitely on malloc.

Resolves #1350

### DIFF
--- a/kernel/libc/koslib/malloc.c
+++ b/kernel/libc/koslib/malloc.c
@@ -24,7 +24,7 @@
 
 #include <kos/dbglog.h>
 #include <kos/opts.h>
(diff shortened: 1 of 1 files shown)


gyrovorbis (approved):
Oh snap, that's a good find.

Between not wasting CPU cycles unnecessarily waking up the spinning thread to attempt to acquire the lock, plus now with mutexes implementing priority boosting, this is way cleaner and maybe even gainzier. Nice.

PR #1380, opened 2026-05-12
sound_stream: Prevent dropping last sample on loop.

QuzarDC:
Discussed in more detail in #1353. Reducing the loopend size by one was required for ADPCM but would cause the loss of one sample at the end of a loop for other formats.

The issue had to do with the AICA loop register overflowing the sample count with the amount of ADPCM samples that can fit in a single buffer as the 4-bits in 32kb allowed a max of 0x10000 samples.

The fix is to remove the dropping of the last sample from loopend and adjust the maximum ADPCM buffer size to prevent it from being able to overflow (while still keeping 32-byte alignment).

@KatanaDreamsDev - This what you were describing by the end of #1353 ? If you'd like to update it there, that's fine too.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/sound/stream.h
+++ b/kernel/arch/dreamcast/include/dc/sound/stream.h
@@ -46,7 +46,7 @@ __BEGIN_DECLS
 #define SND_STREAM_BUFFER_MAX_PCM8  (64 << 10)
 
 /** \brief  The maximum buffer size for each channel of ADPCM stream. */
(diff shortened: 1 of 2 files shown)


KatanaDreamsDev:
Yes, exactly! That's exactly the approach we came to by the end of #1353. Please go ahead and merge it to put an end to this issue!

gyrovorbis (approved):
JESUS. OKAY. I just followed along with the whole conversation that transpired here...

...what a bitchy little obscure bug from an integer overflow, jfc... THAT was a good find. Good job, @KatanaDreamsDev! 

Looks good to me.

PR #1228, opened 2025-10-10
sound playback improvements

kapodamy:
This PR changes how stream and sfx is started:

Changes:
* Procedure to start streams by implementing `snd_stream_setup` which is also API breaking.
* Implementation of `snd_stream_queue*` functions are completly removed, now streams are "queued" by default.
* Several changes was done on AICA side in order to implement new commands like `AICA_CH_STOP_SYNC`.
* Streams and sfx no longer blocks the AICA queue (newest AICA changes allow this).
* bye to `snd_stream_reinit` function.
* Now is posible set volume and panning before start an stream.
* implemented `snd_stream_start_streams` to start multiple streams at once.

Since snd_sh4_to_aica_start/snd_sh4_to_aica_stop are no longer used interally, the user has full control on the AICA queue.

PR #1381, opened 2026-05-13
Remove broken mutex

Kazade:
The queue_proc_mutex was attempted to be unlocked before even locking. It doesn't look like it's useful.

More context can be found in this Discord thread: https://discord.com/channels/488332893324836864/488333450886250496/1503912965328343181

### DIFF
--- a/kernel/arch/dreamcast/sound/snd_iface.c
+++ b/kernel/arch/dreamcast/sound/snd_iface.c
@@ -28,11 +28,6 @@
 /* Are we initted? */
 static int initted = 0;
 
-/* The queue processing mutex for snd_sh4_to_aica_start and snd_sh4_to_aica_stop.
-   There are some cases like stereo stream control + stereo sfx control
-   at the same time in separate threads. */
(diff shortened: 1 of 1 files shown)


QuzarDC:
Think you could try out the change as provided in #1228 that changes the logic a bit and switches to atomics, and see if that also resolves your issue?

https://github.com/kapodamy/KallistiOS/blob/62ccbb6036fa2428372ca207a3350cb159a6eb61/kernel/arch/dreamcast/sound/snd_iface.c

DC-SWAT:
Comment from the code:
> The queue processing mutex for snd_sh4_to_aica_start and snd_sh4_to_aica_stop.
> There are some cases like stereo stream control + stereo sfx control at the same time in separate threads.

Just do not call `snd_stream_queue_go()` before snd_stream_start, this is fundamentally incorrect.
Or better don't use the queue mode at all, because it’s clumsy and blocks all AICA commands handling before you go.
It would be better to add a check to `snd_stream_queue_go()` so that it cannot be called until the stream has started and it in queued mode.

PR #1382, opened 2026-05-13
Fix assert stack tracing

QuzarDC:
Following #1359 , remove further references that had been left in around the frame-pointers based stack tracing.

Additionally remove the places where we were intentionally ignoring layers of the tracing in order to avoid things like `arch_stk_trace` or `assert` showing up in the trace. While this might lead to some extra lines to trace, it prevents the bigger problem of accidentally skipping layers when something gets inlined


andressbarajas:
Whoops, sorry about that. Great cleanup!

gyrovorbis (approved):
Good cleanup and fixes.

PR #1383, opened 2026-05-14
libkosfat: Partial LFN entry filename fix.

### LINKED ISSUE #1326: libkosfat: LFN entries padded with 0x0000 instead of 0xFFFF after null terminator
When `fat_add_dentry()` creates long filename entries, partial LFN entries (where the filename doesn't fill all 13 UCS-2 character slots) are padded with `0x0000` after the null terminator. The FAT specification requires these positions to be `0xFFFF`.

This causes host operating systems with conformant FAT implementations (which likely includes Windows, Linux, macOS, and others) to ignore the LFN entries and fall back to displaying the 8.3 short filename (e.g. `MK5115~1.VMU` instead of `MK5115450-1.VMU`). Filenames that exactly fill one or more LFN entries with no partial trailing entry are unaffected.

**Root cause:**

In `fat_add_dentry()` inside `directory.c`:

`memset(longname_buf2 + len, 0, 13 * sizeof(uint16_t));`

This fills the remaining slots with `0x0000`. Per the FAT spec (Microsoft FAT specification, section 7.3), character positions after the null terminator must be filled with `0xFFFF`.

Hex dump example (a partial LFN entry for a 15-character filename where second entry holds 2 chars + null + padding):

**Before fix (broken):**
```
  42 4D 00 55 00 00 00 00 00 00 00 0F 00 48 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
                  ~~~~~~~~~~~~~~~~~~~              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~
                  name1: 0x0000                    name2: 0x0000                              name3: 0x0000
```

**After fix (correct):**
```
  42 4D 00 55 00 00 00 FF FF FF FF 0F 00 48 FF FF FF FF FF FF FF FF FF FF FF FF 00 00 FF FF FF FF
                  ~~~~~~~~~~~~~~~~~~~              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ~~~~~~~~
                  name1: 0x0000,0xFFFF             name2: 0xFFFF                              name3: 0xFFFF
```

**Fix:**
```
memset(longname_buf2 + len, 0xFF, 13 * sizeof(uint16_t));
longname_buf2[len] = 0x0000;
```

The `0xFF` fill sets each uint16_t to `0xFFFF`, then the null terminator is restored at position len.

QuzarDC:
When fat_add_dentry() creates long filename entries, partial LFN entries (where the filename doesn't fill all 13 UCS-2 character slots) are padded with 0x0000 after the null terminator. The FAT specification requires these positions to be 0xFFFF.

This causes host operating systems with conformant FAT implementations (which likely includes Windows, Linux, macOS, and others) to ignore the LFN entries and fall back to displaying the 8.3 short filename (e.g. MK5115~1.VMU instead of MK5115450-1.VMU). Filenames that exactly fill one or more LFN entries with no partial trailing entry are unaffected.

Closes #1326

### DIFF
--- a/addons/libkosfat/directory.c
+++ b/addons/libkosfat/directory.c
@@ -1323,7 +1323,8 @@ int fat_add_dentry(fat_fs_t *fs, const char *fn, fat_dentry_t *parent,
         /* Make things easier later... */
         if(len % 13) {
             ++dents;
-            memset(longname_buf2 + len, 0, 13 * sizeof(uint16_t));
+            memset(longname_buf2 + len, 0xFF, 13 * sizeof(uint16_t));
+            longname_buf2[len] = 0x0000;
         }
 
         /* Come up with the short name the file will have. */


ljsebald (approved):
Looks good to me.

Issue #1180, opened 2025-09-01
Spinlocks should be dropped


pcercuei:
KallistiOS uses a synchronization mechanism called "spinlocks", where the lock is taken if free, and if it is not, the thread will yield until it is.

The assumption seems to be that it is faster and thus a better idea instead of using a regular mutex.

However, this is just not true: `spinlock_trylock()` is implemented using `tas.b`, which results in slower code than if it was just disabling/enabling IRQs (see https://github.com/KallistiOS/KallistiOS/pull/1170#issuecomment-3239212419) because of how this instruction messes with the cache. With my recent work on mutexes (https://github.com/KallistiOS/KallistiOS/pull/1171), it is very possible that for reserving a non-taken lock, mutexes are faster than spinlocks when it comes to the execution speed of the locking routine.

Even if locking a mutex was slower, the time it takes to *reserve* the locking object is really marginal and not that much important in comparison to the overhead the locking mechanism has when there is contention. Sure, trying to reserve a locked mutex will cause the thread to be put to sleep and added to the genwait queue, which cost CPU cycles. However, until the lock is released, this thread won't be preempted by the scheduler, and therefore we won't waste CPU cycles trying to execute a thread waiting on a lock. This however cannot be said for spinlocks, which will endlessly request the hand from the scheduler, until the lock is released.

If the cost of the pointless context switches isn't enough, having a thread virtually waiting on a lock while staying active means that the scheduler will see this thread as active: if a second thread is trying to lock a spinlock, the scheduler will see two active threads; if it tries to lock a mutex instead, it will see just one active thread. Having just one thread running at a time has a huge benefit: in the future, we can implement "dynamic ticks", where the scheduler's timer is stopped as long as there's only one thread to run, which would prevent a lot of useless context switches.

Finally, spinlocks assume yielding the thread will always end with the second thread holding the lock releasing it. This is wrong, as proven in my thread boost mechanism PR here: https://github.com/KallistiOS/KallistiOS/pull/783. With enough threads of different priorities, a low-priority thread holding a spinlock would have very low chances of being preempted, resulting in a lockup. It also assumes that the scheduler will give a fair chance to each thread; but this is just one possible behaviour of the scheduler (to be fair, the only one implemented right now), which could have other operating modes, e.g. a mode where higher-priority threads always get the hand over lower-priority threads.

If you still need some convincing, the Linux kernel does not have these spinlocks. Their "spinlocks" are basically our "irq_disable() / irq_restore()".

[...]

PR #1386, opened 2026-05-25
Reduce and clean up spinlock usage

QuzarDC:
Furthering #1180 , reduce the usage of spinlocks and clean up mentions of them that had already been migrated to mutexes. This migrates the usage of spinlocks to mutexes for the PVR and AICA memory managers ~~as well as newlib's generic locks~~.

Additionally performed some minor cleanups while there.

### DIFF
--- a/include/machine/malloc.h
+++ b/include/machine/malloc.h
@@ -44,7 +44,7 @@ __BEGIN_DECLS
 
 /** \brief  Determine if it is safe to call malloc() in an IRQ context.
 
-    This function checks the value of the internal spinlock that is used for
+    This function checks the value of the internal mutex that is used for
     malloc() to ensure that a call to it will not freeze the running process.
     This is only really useful in an IRQ context to ensure that a call to
     malloc() (or some other memory allocation function) won't cause a deadlock.
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_mem_core.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_mem_core.c
@@ -95,13 +95,21 @@ static struct mallinfo mALLINFo();
 
 #ifdef USE_MALLOC_LOCK
 
-#include <kos/spinlock.h>
+#include <kos/mutex.h>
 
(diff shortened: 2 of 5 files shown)


QuzarDC:
Per @gyrovorbis breaks toolchain building as various uses of lock.h can't find kos/mutex.h. We were never actually including spinlock.h there before so it may be necessary to fake the mutex types in lock.h or otherwise update to avoid issues. If no solution is forthcoming relatively quickly will just drop that commit as the rest should still be good on their own.

Edit: Addressed by modifying the toolchain newlib patching to happen prior to building. This may have unintended consequences that need to be fully vetted, but no noticeable build failures.

utils/kos-chain/scripts/newlib.mk
```diff
@@ -28,8 +28,11 @@ ifeq (1,$(do_auto_fixup_newlib))
   fixup-newlib: fixup-newlib-apply
 endif
 
+# Build newlib only after the above fixup
+fixup-newlib: $(build_newlib)
```
pcercuei: This one is backwards I think.

QuzarDC: > This one is backwards I think.

Before `fixup-newlib` was calling `fixup-newlib-init` which called `$(build_newlib)` before applying the patch. Now `fixup-newlib-init` no longer depends on `$(build_newlib)` and instead it's done after the patching. Both before and after, the building of newlib is being triggered by `fixup-newlib`, now it's just doing it after the fixup part (as long as `do_auto_fixup_newlib` is selected which is the default and would only be turned off for KOS-less toolchains).

pcercuei:
Sadly this breaks the AICA and GameCube toolchains, as they are the ones without KOS support.

QuzarDC:
> Sadly this breaks the AICA and GameCube toolchains, as they are the ones without KOS support.

Aha, good point. 'Yet' at least :) Dropped that commit and left the rest.

pcercuei (approved):

PR #1387, opened 2026-05-25
C++ std::chrono precision toolchain fixes + clock example upgrade.

gyrovorbis:
This PR is a series of commits which works its  way up to fixing the horrible 1 second maximum resolution of C++'s `std::chrono` API in addition to revamping the C++ clock example to... y'know... actually use real C++ code and not just be C++98 calling into C APIs for date/time.

Basically `std::chrono` went from utterly useless for C++ developers to having the full 80ns granularity of `clock_gettime()` and `timespec_get()`, which is awesome. Plus, the clock example now actually flexes `std::chrono` for a cleaner implementation that also validates modern C++ APIs and features for our toolchains.

Here are the highlights of the commits:

1.  First I had to break our arch/timer API... because SOMEONE named a public API routine, `timer_gettime()` which is literally the name of a POSIX Timer API routine, which I have had to stub out as part of this PR... This could be seen as an API break, sorta, technically, but there's just no way around this. That routine should've never been named that to begin with, and we should absolutely rename it ASAP before more damage is done.

2. I had to stub out the entirety of the POSIX asynchronous timer API routines, such as the *real* `timer_gettime()`, which are just returning EINVAL/ENOSUP atm as stubs. These are needed, because libstdc++ REQUIRES `_POSIX_TIMERS` feature support in order to back `std::chrono` with `clock_gettime()` for its full resolution.

3. I had to remove our declarations for the little bit of the POSIX timer API that we do support from `machine/time.h` and instead get those declarations plus the rest of the POSIX timer API declarations handled by Newlib, by patching its top-level `sys/features.h` header, which is where each platform supported by Newlib conditionally enables the POSIX API features which it supports.

4. I had to finally patch GCC versions 15 and 16 so that `libstdc++`'s configure script which tests for routines to back `std::chrono` with would properly find and use `clock_gettime()` from POSIX, which offers full 80ns resolution, rather than the fallthrough implementation that just uses C89's `time()` routine, with only 1 second resolution.

5. I up-ported the C++ clock example to unapologetically modern AF C++23 code for god, glory, toolchain testing, and educational purposes. :kekw:

EDIT: Oh yeah, I rebuilt both a GCC16 and a GCC15 toolchain with those patches and confirmed everything works great with the new `std::chrono`-based C++ clock example:

https://github.com/user-attachments/assets/c71cbafa-50b6-4e90-8205-ed5419eecdce

### DIFF
--- a/examples/dreamcast/cpp/clock/Makefile
+++ b/examples/dreamcast/cpp/clock/Makefile
@@ -1,18 +1,22 @@
 #
 # DC Clock App
 # Based on Peter Hatch's Font Test
-# (c)2002 Megan Potter
-#   
+#
+# Copyright (C) 2002 Megan Potter
+# Copyright (C) 2026 Falco Girgis
+#
 
 TARGET = clock.elf
 OBJS = clock.o romdisk.o
 KOS_ROMDISK_DIR = romdisk
-
-KOS_CPPFLAGS += -std=gnu++17
-
-all: rm-elf $(TARGET)
+KOS_CPPFLAGS += -std=gnu++23
+KOS_GCCVER_MIN = 13.0.0
 
 include $(KOS_BASE)/Makefile.rules
+
+ifeq ($(call KOS_GCCVER_MIN_CHECK,$(KOS_GCCVER_MIN)),1)
+
+all: rm-elf $(TARGET)
 
 clean: rm-elf
 	-rm -f $(OBJS) 
@@ -30,3 +34,7 @@ dist: $(TARGET)
 	-rm -f $(OBJS) romdisk.img
 	$(KOS_STRIP) $(TARGET)
 
+else
+  all $(TARGET) clean rm-elf run dist:
+	$(KOS_GCCVER_MIN_WARNING)
+endif
--- a/examples/dreamcast/cpp/clock/clock.cc
+++ b/examples/dreamcast/cpp/clock/clock.cc
@@ -1,76 +1,85 @@
-// Originally from plib_examples
-// Ported to Dreamcast/KOS by Peter Hatch
-// read_input () code from KOS examples
-// Converted to a clock by Megan Potter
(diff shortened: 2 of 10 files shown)


gyrovorbis:
DOH. Goddamn, dudes, we have freaking GCC9 in CI? ...yeah, okay, I guess I broke KOS with older GCC toolchains that don't use the same Newlib version as the one I'm patching to enable all the stuff I removed from KOS's `time.h`.

Go ahead and keep reviewing the PR for other issues, but standby, cuz I'll work on fixing it ASAP tomorrow.

pcercuei:
You can probably cheese it by adding the `_POSIX_TIMERS` define to our <sys/_types.h>`.

Not that it'd be the right header to add that, but at least it wouldn't require patching newlib.

gyrovorbis:
I'm about to give this a try when I get a chance. Thanks.

QuzarDC:
>     1. First I had to break our arch/timer API... because SOMEONE named a public API routine, `timer_gettime()` which is literally the name of a POSIX Timer API routine, which I have had to stub out as part of this PR... This could be seen as an API break, sorta, technically, but there's just no way around this. That routine should've never been named that to begin with, and we should absolutely rename it ASAP before more damage is done.
> 
I wouldn't call it an `API Break` because it changes API introduced since the last stable version (was added in #1104, a few weeks after v2.2.0). That's how I've been going about it anyways. 

> You can probably cheese it by adding the `_POSIX_TIMERS` define to our <sys/_types.h>`.
> 
> Not that it'd be the right header to add that, but at least it wouldn't require patching newlib.

Got rid of that in #1408 . But our `include/machine/time.h` is included by `<time.h>` as well and that seems like an appropriate place for them. Would make that commit look nicer too. Rather than defining the standards ourselves in the file, define the flags newlib expects to do its part.

QuzarDC:
Finished via #1518

PR #1392, opened 2026-05-27
examples: Drop KGL examples.

QuzarDC:
We marked KGL as deprecated some time ago, and consistently warn new users against it. No reason to continue to host and test against examples for it.


gyrovorbis (approved):
Yeah, agreed. We had already discussed eventually not including examples for ports within kos itself... and then regardless, this is some reesty, ancient, deprecated stuff that has to be maintained and built, and... yeah, screw it. lol.

Especially if we got good ports like that one accelerated math lib that don't even have examples in kos itself either.... ;)

PR #1395, opened 2026-05-29
mutex: skip priority-restore when there is no current thread

andressbarajas:
Now that malloc uses mutexes, mutex_unlock() restores the holder's pre-boost priority via `thd->prio = thd->real_prio` on the final unlock. But the first malloc runs during early boot inside thd_init -> thd_create_ex (allocating the kernel thread) before thd_current is assigned, so thd_current is NULL there. That made mutex_unlock write through a NULL pointer and crashes on some archs.


gyrovorbis (approved):
VERY good find. Jesus. That is bad.

QuzarDC (approved):

PR #1401, opened 2026-06-01
pvr_mem: Don't mark prototypes __weak_symbol

andressbarajas:
Got the latest master and found that I was having issues with my own PVR example I use to test a lot. Launched into automatic reset of the Dreamcast. I tested against dreamcast/pvr/plasma example and got the same behavior. Turns out the issue was adding __weak_symbol to the pvr_mem_* declarations in <dc/pvr/pvr_mem.h>, not just the definitions in pvr_mem.c. That makes every caller emit a *weak* undefined reference, and an ELF linker does not extract an archive member to satisfy a weak reference. As a result pvr_mem.o is never pulled out of libkallisti.a: pvr_mem_malloc(), pvr_mem_reset() and pvr_mem_initialize() all resolve to address 0, so pvr_init() jumps to 0 and the console resets the moment any textured program runs.

Keep __weak_symbol on the definitions (so a custom allocator can still override them) but use plain prototypes in the header, restoring the strong references that force archive extraction.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/pvr/pvr_mem.h
+++ b/kernel/arch/dreamcast/include/dc/pvr/pvr_mem.h
@@ -64,7 +64,7 @@ typedef void *pvr_ptr_t;
     
     \return                 A pointer to the memory on success, NULL on error
 */
-pvr_ptr_t __weak_symbol pvr_mem_malloc(size_t size);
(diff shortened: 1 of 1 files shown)


GPF:
with #1399 PR , fixes the black screen issue with GLdc examples for me, and SDL2 integration is working again.

QuzarDC (approved):

PR #1404, opened 2026-06-04
Newlib 4.6.0 Patch: Weak ASM memory symbols.

gyrovorbis:
This patch is the result of my evil experimentation with testing my shz_memcpy() routines as the toolchain's default implementation.

So how can I achieve such a thing while still being able to iterate on my implementation? How can I use it for system memcpy() but not have to commit to a particular implementation and keep control over my own implementation, as I iterate on it?

...this PR was my idea. All of the standard C memory routines from Newlib that are implemented in out-of-line ASM: memcpy(), memmove(), memset(), strlen(), etc, are all declared via the same macro. This PR modifies that macro so that all of these routines which get declared with it are now declared WEAK symbols.

So now, simply add a memcpy, memset, memmove, or whatever function into your project and forward it on to shz_memcpy() or libfastmem or whatever you want to back it with!

### DIFF
--- a/utils/kos-chain/patches/targets/newlib-4.6.0.20260123-kos.diff
+++ b/utils/kos-chain/patches/targets/newlib-4.6.0.20260123-kos.diff
@@ -287,3 +287,15 @@ diff --color -ruN newlib-4.6.0.20260123/newlib/libc/stdlib/assert.c newlib-4.6.0
 +// This is put in here to cause link errors if a proper newlib isn't present.
 +int __newlib_kos_patch = 1;
 +
+diff --color -ruN newlib-4.6.0.20260123/newlib/libc/machine/sh/asm.h newlib-4.6.0.20260123-kos/newlib/libc/machine/sh/asm.h
+--- newlib-4.6.0.20260123/newlib/libc/machine/sh/asm.h	2026-01-23 16:12:49.000000000 -0600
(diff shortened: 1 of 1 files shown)


QuzarDC:
From what I can tell for SH the full list of functions is: strlen, strcmp, strcpy, strncpy, setjmp, longjmp, memcpy, and memset. memmove is in asm for some platforms but not SH.

Otherwise looks fine, though think you could add a comment in the patch to explain it? Just as simple as "Allow overriding memcpy, memset, etc" or such would be plenty. I know we can track back to this PR via history, but with the patches each newlib version gets a new file and it becomes trickier to go back through the history to find why things are being patched.

Also, you should really consider offering some improvements upstream. The full on sh4zam versions of these might be too specific (fpu and all) but the updates could probably percolate over to saturn folks at least. With this change things can go as fast as they can, but with some tinkering of the upstream you can raise the floor :)

pcercuei:
Is that really needed? ...

If you use `kos-cc`, then libkallisti.a and libc.a are linked last, which means that any symbols from any object file you link, including static libraries, take precedence over whatever is in KOS or the libc.

QuzarDC:
Additionally see https://github.com/pcercuei/bloom/commit/a2b5de395802b06957a75ee4c652195908fb4e39 as a working example of how to achieve the desired outcome via CMAKE.

Issue #1199, opened 2025-09-19
RFC: Migrate KallistiOS org `kos-ports` into `addons`


QuzarDC:
This would apply to [`libconio`](https://github.com/KallistiOS/libconio), [`libkosh`](https://github.com/KallistiOS/libkosh), [`libparallax`](https://github.com/KallistiOS/libparallax), [`libtsunami`](https://github.com/KallistiOS/libtsunami), and [`libkmg`](https://github.com/KallistiOS/libkmg).

Each of these are internal to and written by KallistiOS org members, have no external dependencies, and have fallen into some disrepair (in part) due to their relegation to `kos-ports` and the code to them not being made available by default when pulling kos.

I think doing this would help to clarify a division between `addons` and `kos-ports`:
`addons` are:
- Written/maintained by kos org members, hosted by kos org.
- Licensed under the KOS license.
- Have no external dependencies.
and either
- Provide and API layer of higher abstraction than we want in the kernel (the libs above, `libkosutils`, `libpthread`).
AND/OR
- Are optional modules that would benefit from being dynamically loadable rather than statically enabled/disabled via INIT flags (`libkosext2fs`, `libkosfat`, `libnavi`, `libppp`).

Comment encouraged on the above division. This seems to be a sensible outline for what should/could be an `addon`.

QuzarDC:
Opened #1405 to start this. I had made the mistake in the original description with libparallax as it relies on libpng and libjpg. I still think it would be beneficial to find a way to bring that in but it won't be as straightforward as the others.

PR #1405, opened 2026-06-04
libconio, libkosh, and libkmg migration to addons

QuzarDC:
Following the reasons from #1199 , migrating these ports from separate repos into addons.

Since libs and includes from addons take precedence over kos-ports, those can stay to ensure examples still build on versions prior while new work on the libs would be done in the mainline repo.

Aside from the raw moving of files from their independent repos the following changes were made:
- Replace deprecated keyboard functions in libconio (to avoid warnings)
- Fix warning in libkosh from constifying of readdir in #1294
- Clean up DECLS in kmg.h
- Add the `lib_` functions required by our library system for correct building of a libkmg klf
- Update addons make to use `kos-ar` rather than the env version (no functional change there)
- Update addons readme with these new libs as well as libpthread


QuzarDC:
Apologies for the noise, hadn't noticed the CI failing and not sure what was going on with my own environ that I simply wasn't getting the unused param warnings

PR #1406, opened 2026-06-05
Add gprof profiling runtime (libgprof)

andressbarajas:
This has been been PRed here before and now Im bringing it back cleaned up and with no ugly toolchain/startup.S changes(thanks to build_wrapper changes)! This adds `libgprof`: a timer-sampled flat profile plus a call graph recorded at each function entry, written out as a standard `gmon.out` for host-side `gprof`.

- Portable core (sampler, BST call graph, `gmon.out` writer) + a per-arch hook; this PR ships the **SH4/Dreamcast** backend (the `trapa #33` that stock `gcc -pg` already emits, so **no toolchain changes**).

- Just build/link with `-pg` — the `kos-cc` wrapper auto-links the runtime (no `-lgprof`), and profiling auto-starts before `main()` and flushes on exit.

- Public API in `<gprof/gmon.h>`: `monstartup`, `moncontrol`, `_mcleanup`.

```make
CFLAGS = -pg -fno-omit-frame-pointer -fno-inline
$(TARGET): $(OBJS)
	kos-cc -pg -o $(TARGET) $(OBJS)
```


addons/include/gprof/gmon.h
```diff
@@ -0,0 +1,120 @@
...
+    1. Build with profiling flags. \c -pg must be on both the compile and the
+       link (as on a normal gprof system). KOS applies \c CFLAGS automatically
+       when compiling, but the link command is hand-written, so add \c -pg to it
+       directly:
+
+        ```make
+        CFLAGS += -pg -fno-omit-frame-pointer -fno-inline
+
+        $(TARGET): $(OBJS)
+        	kos-cc -pg -o $(TARGET) $(OBJS)
```
pcercuei: Stray `\n` here.

pcercuei (changes requested):
One comment inline.
The no-tab CI check will still fail afterwards as it chokes on `kos-cc` but we can ignore it.

addons/include/gprof/gmon.h
andressbarajas: I found a stray \t and replaced with spaces. Is that what you were talking about?

pcercuei: Yes. Thanks

addons/libgprof/gmon.c
```diff
@@ -0,0 +1,576 @@
...
+
+#include <gprof/gmon.h>
+#include <kos/thread.h>
+#include <kos/dbglog.h>
+#include <arch/irq.h>
+
+#include "gprof_internal.h"
+
+#define ROUNDDOWN(x, y)  (((x) / (y)) * (y))
+#define ROUNDUP(x, y)    ((((x) + (y) - 1) / (y)) * (y))
```
pcercuei: We have `__align_down()` and `__align_up()` already.

addons/libgprof/gmon.c
```diff
@@ -0,0 +1,576 @@
...
+    .nodes = NULL,
+    .node_count = 0,
+    .ncounters = 0,
+    .allocate_size = 0,
+    .histogram = NULL,
+    .main_thread = NULL,
+    .histogram_thread = NULL,
+    .running_thread = false,
+    .initialized = false
+};
```
pcercuei, line 133: So basically the only field that's not initialized to zero is `.state`, so you can drop all the others.

addons/libgprof/gmon.c
```diff
@@ -0,0 +1,576 @@
...
+    /* Write out gmon.out when the program exits. */
+    atexit(_mcleanup);
+
+    /* Turn profiling on. */
+    moncontrol(true);
+}
+
+/* Strong override for the weak gprof_init() in arch_main; pulled into the link
+   by the -pg instrumentation (MIPS) or by --whole-archive (SH4). Starts
+   profiling over the whole program .text range. */
```
pcercuei, line 549: Instead of having a weak `gprof_init` in KOS' init itself, I think you only need to have this `gprof_init` tagged with `__attribute__((constructor))` and it will be called by arch_main().

pcercuei (dismissed):
Some more comments.

addons/libgprof/gmon.c
andressbarajas: I do it this way to guarantee that no other constructors end up getting profiled. If I go the constructor attribute route there is a chance that other constructors will run after gprof_init() which means they get included in the profiled data. AFAIK there is no way to order the list of constructors. This may not effect C programs much but I THINK C++ programs would be heavily affected and produce graphs with a bunch of constructor nodes. I'll still make the change just letting you know why I did it specifically this way. Let me know if I should keep it this way or go constructor route.

andressbarajas: Actually thinking about it now. gcov instruments constructors so there is no good reason gprof shouldn't. I just did more research and it looks like Linux runs the gprof_init() before even the constructors so all of them are guaranteed to be profiled. So instead of calling gprof_init() after _init(); I should call it before to make sure all constructors are profiled. If I just add the constructor attribute to gprof_init() there is a chance that some constructors would be left out of being profiled.

addons/libgprof/gmon.c
andressbarajas: Done

addons/libgprof/gmon.c
andressbarajas: Done.

addons/libgprof/gmon.c
andressbarajas: @pcercuei I moved gprof_init() above _init() to make sure all constructors are profiled

kernel/arch/dreamcast/kernel/init.c
```diff
@@ -277,6 +277,10 @@ void  __weak_symbol arch_auto_shutdown(void) {
     rtc_shutdown();
 }
 
+/* Weak default; the libgprof addon provides the strong version that starts
+   gprof profiling. Called after constructors so it never profiles them. */
+void __weak_symbol gprof_init(void) { }
```
pcercuei: You should also be able to use a weak definition, e.g.
```c
__weak_symbol void gprof_init(void);
```

Then in code:
```c
if(gprof_init)
    gprof_init();
```

addons/libgprof/gmon.c
```diff
@@ -0,0 +1,576 @@
...
+                cxt->histogram[index]++;
+            }
+        }
+
+        /* Always sleep before the next sample */
+        thd_sleep(HISTOGRAM_INTERVAL_MS);
+    }
+
+    return NULL;
+}
```
pcercuei, line 289: I feel like your histogram_thread and the code around it could be simplified quite a bit, using a oneshot timer (`<kos/oneshot_timer.h>`) that you just reset at the end of the callback function.

kernel/arch/dreamcast/kernel/init.c
andressbarajas: Done

addons/libgprof/gmon.c
andressbarajas: I looked into it and I dont think the oneshot_timer fits here. It'll shoot off a thread with PRIO_DEFAULT with no way to set priority or label it, so in the end I still use a thread while costing control I want.

What I should probably do is use a timer interrupt instead. My current approach reads only the main thread's saved PC, so anything running in other threads are missed. Plus, if the users program has many threads the histogram thread would be fighting to fire every 10ms. A timer interrupt gets around all that (I'll get to sample other threads and fire every 10 ms).

Every TMU was taken when I first wrote this code over a year ago so that is why I went with the thread approach. I see now that TMU1 is free... or would you rather I not use TMU1? Any other ideas?

pcercuei: 10 ms is exactly one time slice at the default HZ. You could use a thread that runs `thd_poll()` with a callback that returns non-zero only on exit, and do your stuff in the callback. It will run at exactly every 10 ms.
Another alternative would be to use the vsync IRQ that runs once every 16.6 ms.

andressbarajas: Thanks! I went with the first idea. Committed the change. I tested it and its working fine against my example.

examples/dreamcast/profiling/gprof/Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+TARGET = gprof.elf
+OBJS = gprof.o
+
+# -pg instruments every function for profiling. -fno-omit-frame-pointer is
+# REQUIRED only because GCC refuses -pg with an omitted frame pointer (KOS builds
+# at -O2, which omits it by default); the call graph itself reads the caller from
+# a register, so it does NOT depend on the frame pointer -- which is just as well,
+# since recent SH4 GCC frame pointers are unreliable. -fno-inline keeps each function
+# a distinct symbol in the report but is not, and -fno-ipa-icf stops GCC from folding
```
QuzarDC: Maybe I'm misreading but is there something missing after the 'is not'?

examples/dreamcast/profiling/gprof/gprof.c
```diff
@@ -0,0 +1,120 @@
...
+
+   Nothing here calls into the runtime directly: linking with -pg auto-starts
+   profiling before main() (via the weak gprof_init() hook) and flushes gmon.out
+   on exit. Build/run/report steps are in README.md.
+
+   A couple of details keep the optimizer from spoiling the demo:
+
+     - the helpers are __attribute__((noinline, noclone)) so each stays a
+       distinct symbol in the report (the Makefile also passes -fno-inline, and
+       -fno-ipa-icf so hot_path/warm_path -- which are identical -- aren't folded
```
QuzarDC: Since you're already having to add attributes to the functions, you might keep all of this to attributes and avoid the `-fno-ipa-icf` by passing the [`no-icf`](https://gcc.gnu.org/onlinedocs/gcc/Common-Attributes.html#index-no_005ficf) and [`noipa`](https://gcc.gnu.org/onlinedocs/gcc/Common-Attributes.html#index-no_005ficf) attributes that can be used for that same effect on just the specific functions. Not sure if it was intentionally in just the makefile, but since you doubled up on the noinline, it seemed like an oversight. With so many it might also look nicer to come up with a named define for the big attribute set, similar to the ones in cdefs.

Doing something like `__attribute__(optimize("-Og"))` per-function would do similar as well, or passing `Og` or `O0` in the makefile to override the environ default. Unless the intent was to keep as many optimizations as possible and only exclude the specific ones that would interfere. Either kind of thing can make a sense, and I'm not sure exactly what the most common interactions would be with all this in order to tell what would be more useful for the user.

examples/dreamcast/profiling/gprof/Makefile
andressbarajas: Good catch, that's a leftover from an edit.

EDIT: Updated and force pushed

examples/dreamcast/profiling/gprof/gprof.c
andressbarajas: Updated and force pushed. I removed -noinline from the makefile and add no_icf attributes to correct functions. I dont expect people to use all of these in their applications. I introduced them in this example to I could produce a nice example graph with a simple example.

QuzarDC (approved):

QuzarDC:
> Some more comments.

As far as I can tell they were all addressed, anything else you wanted from them?

PR #1407, opened 2026-06-05
Add gcov coverage + PGO runtime (libgcov)

andressbarajas:
This has been been sitting in my fork for over a year now as PR. Since that is over and done with Im bringing it here.  This adds `libgcov`: KOS programs built with `--coverage` / `-fprofile-generate` now emit `.gcda` on hardware (GCC 13+ format), enabling coverage reports (lcov/genhtml) and PGO (`-fprofile-use`). Checkout gcov.h for more details.

- Implements the gcov runtime (`__gcov_init/_dump/_reset/_exit` + merge/TopN), writing `.gcda` to `/pc` (relocatable via `GCOV_PREFIX`/`GCOV_PREFIX_STRIP`) and merging across runs.

- Arch-agnostic addon, selected per console via `kos/<arch>.cnf`.

```make
CFLAGS   = --coverage -fprofile-values
PROFLIBS = -lgcov
```


addons/include/gcov/gcov.h
```diff
@@ -0,0 +1,206 @@
...
+        		--gcov-tool=$(KOS_GCOV) \
+        		--directory . \
+        		--base-directory . \
+        		--capture \
+        		--rc geninfo_unexecuted_blocks=1 \
+        		--ignore-errors inconsistent,unused,source,empty \
+        		--output-file coverage.info
+        	genhtml coverage.info --output-directory html \
+        		--ignore-errors inconsistent,source
+        	open html/index.html
```
pcercuei: Stray tabs here.

addons/libgcov/reset.c
```diff
@@ -0,0 +1,49 @@
...
+
+    reset.c
+    Copyright (C) 2025 Andy Barajas
+
+*/
+
+#include "gcov.h"
+
+/*
+    Zeroes out all active counters associated with each function in
```
pcercuei: There's a space character at EOL here.

pcercuei (changes requested):
Some cosmetic nitpicks, which should make the CI happy.

addons/include/gcov/gcov.h
andressbarajas: Done

addons/libgcov/reset.c
andressbarajas: Done

pcercuei:
@andressbarajas That's all your code? I'm asking in regards to the license, to make sure it's not adapted from e.g. GPL code.

andressbarajas:
> @andressbarajas That's all your code? I'm asking in regards to the license, to make sure it's not adapted from e.g. GPL code.

Yes, all my code WITH AI assistance

pcercuei:
I wonder, why didn't you use GCC's libgcov? My understanding is that we only needed to provide a few callback functions.

andressbarajas:
> I wonder, why didn't you use GCC's libgcov? My understanding is that we only needed to provide a few callback functions.

There was a reason but I can't remember. I think I initially thought because we have more control over the output path so it can be exported to /pc but I don't think that is the case anymore. I'll look into it tonight and get back to you.

andressbarajas:
@pcercuei You are right. Just a little change to KOS_LIBS and a callback or two to write out the data! I'll get on it tonight.

@QuzarDC @pcercuei Should I close this PR and start a new one? Clear the history of this PR and start from scratch?

pcercuei:
> @QuzarDC @pcercuei Should I close this PR and start a new one? Clear the history of this PR and start from scratch?

If you force-push then Github will still mention the git hash of the old HEAD so we still have access to the old ones. So it's up to you really, I don't mind either way.

andressbarajas:
@pcercuei Now instead of using a custom libgcov we use GCC's freestanding libgcov so we only have to supply the missing file-I/O piece for KOS and wires it into the toolchain so coverage "just works." . Requires GCC 12+ (when freestanding libgcov was introduced).

pcercuei:
For GCC < 12:
```patch
diff --git a/addons/libkosgcov/Makefile b/addons/libkosgcov/Makefile
index 1cd479d08..8744eef55 100644
--- a/addons/libkosgcov/Makefile
+++ b/addons/libkosgcov/Makefile
@@ -6,5 +6,19 @@
 
 TARGET = libkosgcov.a
 OBJS = kos_gcov.o
+KOS_GCCVER_MIN = 12.0.0
+
+.PHONY: libkosgcov
+
+all: libkosgcov
 
 include $(KOS_BASE)/addons/Makefile.prefab
+
+ifeq ($(call KOS_GCCVER_MIN_CHECK,$(KOS_GCCVER_MIN)),1)
+
+libkosgcov: defaultall
+
+else
+  libkosgcov:
+       $(KOS_GCCVER_MIN_WARNING)
+endif
```

addons/libkosgcov/kos_gcov.c
```diff
@@ -0,0 +1,155 @@
...
+    }
+}
+
+/* Auto-register the flush at startup, so coverage is written at exit with no
+   source changes. */
+__attribute__((constructor))
+static void kos_gcov_autostart(void) {
+    dbglog(DBG_NOTICE, "[GCOV]: coverage instrumentation active\n");
+
+    atexit(kos_gcov_dump);
```
pcercuei, line 154: You can have `kos_gcov_dump` use `__attribute__((destructor))` I think, so you wouldn't need this "autostart" function.

pcercuei (changes requested):
You can discriminate on GCC's version, see my patch suggestion above.
One small comment to the code too.

addons/libkosgcov/kos_gcov.c
andressbarajas: When a program is compiled with GCOV turned "ON" it makes the program run slower which is why I included "[GCOV]: coverage instrumentation active" to let the user know in the beginning they have a gcov build running. I sort of do the same for GPROF too. Is it okay to keep this approach?

pcercuei: Sure.

pcercuei (dismissed):
ACK.
No-tab CI check can be ignored, the tabs were added to `kos-cc`.

QuzarDC:
The minversion gcc test should get added to the example makefile as well.

QuzarDC (approved):

PR #1408, opened 2026-06-05
Newlib patch reduction

QuzarDC:
Some cleanups to our newlib integration, mostly in deleting unneeded headers:
- Remove `<sys/_pthreads.h>` - This was a patch apparently needed for newlib versions prior to `<sys/_pthreadtypes.h>`. No version of newlib that we support does this anymore.
- Stop copying `<arch>` into newlib's includes. - This was being done so that `<arch/irq.h>` was accessible as it was called by `<kos/thread.h>` via `<kos/mutex.h>`. This hasn't been needed since #1279 where we stopped exposing thread.h in mutex.h
- Remove `<sys/_types.h>` - This required adjusting some headers first that were referencing `IOV_MAX` via it. That define though already existed in `<sys/syslimits.h>`. After that though, nothing else was being used in the header that newlib's isn't providing.


pcercuei:
I checked that the 16.0.1-dev toolchain (which tracks GCC master) does build fine locally, so the CI fail seems to be triggered by one of your commits.

QuzarDC:
Cleared a conflict in sysconf.c with the force-push ahead of testing more locally. And a second force-push to correct a break that caused with mmu.h

QuzarDC:
> I checked that the 16.0.1-dev toolchain (which tracks GCC master) does build fine locally, so the CI fail seems to be triggered by one of your commits.

With the force push that had trivial changes (nothing presented to the toolchain builder would have changed) it now built, and I was able to build it locally as well with no problem. The error that came up was unrelated to any sort of patched stuff so I'd presume it was either a roll of the dice or that upstream changed in the meantime?

kernel/libc/posix/sysconf.c
```diff
@@ -83,18 +83,18 @@ long sysconf(int name) {
...
             return mallinfo().fordblks / PAGESIZE;
 
-        case _SC_NPROCESSORS_CONF: 
-        case _SC_NPROCESSORS_ONLN: 
+        case _SC_NPROCESSORS_CONF:
+        case _SC_NPROCESSORS_ONLN:
             return 1;
 
-        default: 
+        default:
```
pcercuei, line 97: Lots of whitespace cleanups here, which is fine, but in a separate commit would be even better ;)

QuzarDC: Yea, I did those on a whim when I noticed them while resolving the merge conflict. Separated out now from the functional change.

pcercuei:
Great, now sourceware.org is down so it can't download Newlib.

pcercuei (approved):

PR #1410, opened 2026-06-05
thread: Fix POSIX clock reporting

andressbarajas:
While playing around with improving dcprofiler I encountered an issue with POSIX's clock_gettime(). I was getting 0's back. This PR #1296 (commit cdcb8ef) reduced per-thread CPU-time accounting to millisecond resolution to avoid reading the timer twice (once in ms and then in ns) on every reschedule.

That broke POSIX's clock_gettime(): CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID divide the counter by 1e9 (treating it as nanoseconds), so with a millisecond counter both clocks reported times 1,000,000 times too small.

This PR restores nanosecond resolution without reintroducing the cost that #1296 removed(reading the timer twice in two different resolutions). The scheduler keeps a millisecond timestamp for its priority-ageing logic; the public CPU-time counters are kept in nanoseconds.

Verified all basic/threading examples all pass.

### DIFF
--- a/kernel/thread/thread.c
+++ b/kernel/thread/thread.c
@@ -138,7 +138,6 @@ int thd_pslist(int (*pf)(const char *fmt, ...)) {
     pf("addr\t  tid\tprio\tflags\t  wait_timeout\t  cpu_time\t      state\t  name\n");
 
     irq_disable_scoped();
-    thd_get_cpu_time(thd_get_current());
     ms_time = timer_ms_gettime64();
 
     LIST_FOREACH(cur, &thd_list, t_list) {
--- a/kernel/libc/posix/clock_gettime.c
+++ b/kernel/libc/posix/clock_gettime.c
@@ -29,10 +29,9 @@ int clock_getcpuclockid(pid_t pid, clockid_t *clock_id) {
 
 int clock_getres(clockid_t clk_id, struct timespec *ts) {
     switch(clk_id) {
+        /* Backed by the nanosecond resolution */
         case CLOCK_REALTIME:
         case CLOCK_MONOTONIC:
-        case CLOCK_PROCESS_CPUTIME_ID:
-        case CLOCK_THREAD_CPUTIME_ID:
             if(!ts) {
                 errno = EFAULT;
                 return -1;
@@ -41,7 +40,19 @@ int clock_getres(clockid_t clk_id, struct timespec *ts) {
             ts->tv_sec = 0;
             ts->tv_nsec = 1;
             return 0;
-            
+
+        /* Backed by the millisecond resolution */
+        case CLOCK_PROCESS_CPUTIME_ID:
+        case CLOCK_THREAD_CPUTIME_ID:
+            if(!ts) {
+                errno = EFAULT;
+                return -1;
+            }
+
+            ts->tv_sec = 0;
+            ts->tv_nsec = 1000000;
+            return 0;
+
         default:
             errno = EINVAL;
             return -1;
@@ -50,7 +61,6 @@ int clock_getres(clockid_t clk_id, struct timespec *ts) {
 
 int clock_gettime(clockid_t clk_id, struct timespec *ts) {
     lldiv_t  div_result;
-    uint64_t ns64;
     uint32_t secs, nsecs, secs_offset=0;
 
     if(!ts) {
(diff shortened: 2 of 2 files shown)


QuzarDC:
Couldn't this be done with just a few changes with something like this? It leaves the structural changes in place that reduced the calls to time and stopped counting as much scheduler time as part of the thread's time, and just adds that extra struct member to track the ms value alongside the ns value. We could even do `>>20` or whatnot for the division if that's too heavy since the absolute accuracy of the ms is not necessary, just relative accuracy.

Edit: pushed it as a branch to be easier to test: https://github.com/KallistiOS/KallistiOS/tree/thread_ns
 
Edit2: I see how I entirely missed the call for ms in `thd_schedule`. Will have to look more carefully but still seems like more adjustments than needed. I agree though that we should certainly have the internals back on ns if possible.

andressbarajas:
@QuzarDC Pumped this PR full of Ozempic and shes trimmed down.

pcercuei:
> That broke POSIX's clock_gettime(): CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID divide the counter by 1e9 (treating it as nanoseconds), so with a millisecond counter both clocks reported times 1,000,000 times too small.

Given that our CPU time is counted with millisecond granularity now, I would instead update `clock_gettime()` to just divide by 1e3.

andressbarajas:
> > That broke POSIX's clock_gettime(): CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID divide the counter by 1e9 (treating it as nanoseconds), so with a millisecond counter both clocks reported times 1,000,000 times too small.
> 
> Given that our CPU time is counted with millisecond granularity now, I would instead update `clock_gettime()` to just divide by 1e3.

Im willing to do that. @QuzarDC cool with you?

andressbarajas:
> > That broke POSIX's clock_gettime(): CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID divide the counter by 1e9 (treating it as nanoseconds), so with a millisecond counter both clocks reported times 1,000,000 times too small.
> 
> Given that our CPU time is counted with millisecond granularity now, I would instead update `clock_gettime()` to just divide by 1e3.

@pcercuei Done

kernel/thread/thread.c
```diff
@@ -91,6 +91,19 @@ static size_t thd_count = 0;
...
+
+/* If the thread is currently running, add the time it has been running */
+static inline uint64_t thd_cpu_time_now(const kthread_t *thd, uint64_t now_ms) {
+    uint64_t total_ms = thd->cpu_time.total;
+
+    if(thd == thd_current && thd->state == STATE_RUNNING)
+        total_ms += now_ms - thd->cpu_time.scheduled;
+
+    return total_ms;
+}
```
pcercuei: I kinda disagree with this commit. Not that it's *wrong*, but do we really care about the few milliseconds of the running thread since it's been preempted? It will be completely dwarfed by the value in `thd->cpu_time.total`. I just don't see the usecase.

pcercuei: ("this commit" being the second one)

andressbarajas: Done. Reverted that commit.

pcercuei:
Can you just drop 7f60d5f and the commit that reverts it, then force-push?

kernel/thread/thread.c
QuzarDC: > I kinda disagree with this commit. Not that it's _wrong_, but do we really care about the few milliseconds of the running thread since it's been preempted? It will be completely dwarfed by the value in `thd->cpu_time.total`. I just don't see the usecase.

Kind of irrelevant now but it'd come back up if you ever get the scheduler disabling/0Hz tick/lazy scheduler scheme going where a single threaded program by-and-large avoids the scheduler.

QuzarDC:
> > > That broke POSIX's clock_gettime(): CLOCK_PROCESS_CPUTIME_ID and CLOCK_THREAD_CPUTIME_ID divide the counter by 1e9 (treating it as nanoseconds), so with a millisecond counter both clocks reported times 1,000,000 times too small.
> > 
> > 
> > Given that our CPU time is counted with millisecond granularity now, I would instead update `clock_gettime()` to just divide by 1e3.
> 
> Im willing to do that. @QuzarDC cool with you?

Since this is aimed at fixing a bug caused by other changes, sure. But generally I'm of the attitude that the whole of the timing for the scheduler and genwait should get shifted to `ns` . That's a much bigger change, but I don't know why we're continuing to try to go backwards in the granularity of the scheduler when it seems that there have been so many complaints about the low resolution nature of the scheduler forcing clunky behavior. And we've had #537 rotting to try to enhance it all.


Aside from all that it seems that for this change all that's relevant is that first commit to update the clock_gettime. I'd see as well about summoning clockmaster @gyrovorbis over for input. This shouldn't interfere at all with his #1387 but might have cropped up in the testing for it.

pcercuei:
> But generally I'm of the attitude that the whole of the timing for the scheduler and genwait should get shifted to `ns` . That's a much bigger change, but I don't know why we're continuing to try to go backwards in the granularity of the scheduler when it seems that there have been so many complaints about the low resolution nature of the scheduler forcing clunky behavior. And we've had #537 rotting to try to enhance it all.

Our current tick period is 10 milliseconds.
Going to 1 millisecond means one thousand timer IRQs per second. Having the scheduler work with a granularity lower than a millisecond would just result in an extremely high overhead. There is no benefit in trying to go lower.

pcercuei (approved):

QuzarDC:
> > But generally I'm of the attitude that the whole of the timing for the scheduler and genwait should get shifted to `ns` . That's a much bigger change, but I don't know why we're continuing to try to go backwards in the granularity of the scheduler when it seems that there have been so many complaints about the low resolution nature of the scheduler forcing clunky behavior. And we've had #537 rotting to try to enhance it all.
> 
> Our current tick period is 10 milliseconds. Going to 1 millisecond means one thousand timer IRQs per second. Having the scheduler work with a granularity lower than a millisecond would just result in an extremely high overhead. There is no benefit in trying to go lower.

That's where your 'dynamic ticks' comes into play ;) , so that it's not about ticking more often, but being able to with less latency when needed.

Issue #1310, opened 2026-01-24
RFC: Move away from ANSI C variable definition


QuzarDC:
This may have made sense at a time, but I really don't see a good rationale for maintaining it other than as a style preference. To my sensibility it makes the code a bit more difficult to read. Especially in big functions I end up hunting to tell what the scope is for different variables. I presume that GCC's smart enough a majority of the time that this shouldn't really improve codegen, but it could also do that in some cases.

If there's no objections then there's not really anything 'todo'. This isn't mentioned in the style guide (which is woefully in need of rewriting anyways), it's just something we've always done and there's only a few places in the codebase where it's violated. I wouldn't expect that we do a bunch of cleanups immediately for this, just to take it into account if useful in the future.

QuzarDC:
I'm going to close this as done. I've started doing this in various cleanups which were approved and I would approve PRs that do so sensibly.

PR #1412, opened 2026-06-06
`file_t` and general fs cleanups

QuzarDC:
A number of relatively small cleanups related to our vfs system:
- Use `FILEHND_INVALID` where appropriate. Most places in the codebase were checking for a hardcoded `-1` as an error return/invalid `file_t`. There is though a define that is meant just for that. It turns out in searching for this, there were also many places that treated any negative value in `file_t` as an error, and others that seemed to think that a value of `0` was an error (and so would have pretty much never been able to detect an error).
- Adjustments for casts of pointers to appropriate types. In fs, fs_dev, and fs_ramdisk there were usages of plain ints to store cast pointers. In testing out the '[null arch](https://github.com/KallistiOS/KallistiOS/tree/null_arch)' on a 64-bit host, these cropped up as issues.
- In fs_pty use `file_t` types rather than generic ints. Like the use of the error, this presumed the backing value of the type.
- General cleanups in fs.c. Lots of whitespace cleanup and reducing loc via [ignoring ANSI C variable declaration](#1310 ).

The only bits that qualify as 'bug' is some of the checks for invalid `file_t` that were looking for a 0 to indicate invalid. The rest should be  transparent.

### DIFF
--- a/kernel/fs/fs.c
+++ b/kernel/fs/fs.c
@@ -814,7 +814,7 @@ int fs_symlink(const char *path1, const char *path2) {
     }
 }
 
-int fs_readlink(const char *path, char *buf, size_t bufsize) {
+ssize_t fs_readlink(const char *path, char *buf, size_t bufsize) {
     vfs_handler_t *vfs;
     char fullpath[PATH_MAX];
 
--- a/kernel/fs/fs.c
+++ b/kernel/fs/fs.c
@@ -61,12 +61,8 @@ static fs_hnd_t *fs_root_opendir(void) {
 static dirent_t root_readdir_dirent;
 static dirent_t *fs_root_readdir(fs_hnd_t *handle) {
     nmmgr_handler_t *nmhnd;
-    nmmgr_list_t    *nmhead;
-    uintptr_t        cnt;
-
-    cnt = (uintptr_t)handle->hnd;
-
-    nmhead = nmmgr_get_list();
+    nmmgr_list_t    *nmhead = nmmgr_get_list();
+    uintptr_t        cnt = (uintptr_t)handle->hnd;
 
     SLIST_FOREACH(nmhnd, nmhead, list_ent) {
         if((nmhnd->flags & NMMGR_FLAGS_INDEV))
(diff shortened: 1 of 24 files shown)


kernel/fs/fs_ramdisk.c
```diff
@@ -94,7 +94,7 @@ static rd_dir_t  *rootdir = NULL;
 static struct {
     rd_file_t   *file;      /* ramdisk file struct */
     int         dir;        /* >0 if a directory */
-    uint32_t    ptr;        /* Current read position in bytes */
+    uintptr_t   ptr;        /* Current read position in bytes */
     dirent_t    dirent;     /* A static dirent to pass back to clients */
     int         omode;      /* Open mode */
 } fh[FS_RAMDISK_MAX_FILES];
```
pcercuei, line 100: In that case, you might want to move `omode` after `dir` just to have the struct more packed on 64-bit systems. Unless it causes problems elsewhere.

kernel/fs/fs.c
```diff
@@ -331,7 +332,7 @@ file_t fs_dup2(file_t oldfd, file_t newfd) {
 /* Returns a file handle for a given fd, or NULL if the parameters
    are not valid. */
 static fs_hnd_t *fs_map_hnd(file_t fd) {
-    if(fd < 0 || fd >= FD_SETSIZE) {
+    if(fd == FILEHND_INVALID || fd >= FD_SETSIZE) {
```
pcercuei: Maybe keep the `< 0` here (and above), otherwise you're changing the behaviour of the error-check.

pcercuei (commented):
Some thoughts on this PR.

kernel/fs/fs.c
QuzarDC: Ok, yea, that tracks. I wasn't looking at the wider scope that it's taking in the `file_t` basically direct from the user. A next step was going to be to have `file_t` backed by an unsigned value to further simplify the checks, but that was going to require some deeper changes, so it should be able to error still if the user passes in garbage (`(file_t)-7` or such).

I'm still adding in the test against `FILEHND_INVALID` for visibility. gcc should be able to just optimize it away as redundant while it's set to a negative value.

kernel/fs/fs_ramdisk.c
QuzarDC: Aha, good catch. It shouldn't cause any issues as the layout is internal and doesn't map to anything else.

QuzarDC:
Force pushed the first time to address review feedback. The second (and third) was that I noticed one place that I had missed where `fs_hnd_assign` would return `-1` rather than `FILEHND_INVALID`.

pcercuei (approved):

PR #1414, opened 2026-06-10
fs_iso9660: Check for CD init in stat() callback

### LINKED ISSUE #1413: stat() doesn't find files in /cd if it's the first operation... or something
In Simulant, when you want to open a file it searches a set of registered search paths for the relative path. It does this by calling stat() and if the result is > -1 then it returns that path.

When I run Simulant games over dc-load, files are read from /pc and everything is fine (stat works perfectly). If I run from CD, no files are ever found (stat always returns -1)

If however you call opendir("/cd/") before doing anything then stat works fine from /cd. Of course calling fopen() instead and checking the result of that succeeds. 

The problem is reproducible on hardware, and emulators (Nitrocast + Flycast).

pcercuei:
The stat() callback does not take an opened file descriptor as its argument, but a filename. Therefore, if fs_stat() is called before fs_open(), the code would return an error as the CD-Rom hardware was not initialized.

Address this issue by also conditionally initializing the CD-Rom hardware from the stat() callback.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_iso9660.c
+++ b/kernel/arch/dreamcast/fs/fs_iso9660.c
@@ -39,6 +39,7 @@ ISO9660 systems, as these were used as references as well.
 #include <kos/opts.h>
 #include <kos/dbglog.h>
 
+#include <stdbool.h>
 #include <stdlib.h>
 #include <stdio.h>
 #include <stdalign.h>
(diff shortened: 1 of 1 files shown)


QuzarDC (approved):
We really need to refactor this all at some point. Looks good.

Kazade:
Thanks! Sorry I didn't get time to test but I'm sure this was the issue :)

PR #1415, opened 2026-06-10
aica: Enhanced driver initialization.

DC-SWAT:
Updated the AICA initialization routine to include additional setup for interrupt sources and FIQ handling.
This is a sound fix for NAOMI and NAOMI 2 systems running BIOS without startup sound, as well as for Dreamcast on  stock BIOS when returning to KOS from another SDK.


pcercuei (approved):
Tab CI check can be ignored.

PR #1416, opened 2026-06-10
pvr: Avoid memcpy() to VRAM

pcercuei:
Calling memcpy() to copy from/to a VRAM address is dangerous, as the implementation might try to do 64-bit I/O, which is not possible on this bus.


gyrovorbis (approved):
This is exactly right!!! We’re playing with overriding memcpy() with my implementation which is actually smart enough to do 8-byte reads/writes, so it exposed the fact that we’re doing the wrong thing by using a generic memcpy() here!

PR #1417, opened 2026-06-10
Examples cleanups

QuzarDC:
A pile of small cleanups for our examples. Some are just kind of trivial `-Wextra` stuff, but most either clear away actual warnings or prevent issues with different gcc or language revisions.

The only behavior change is in the keyboard example where it will no longer treat all keyboards as JP region, but read the region from the dev state for key translation. This should allow it to work properly on euro keyboards (though I don't have any to test with).


pcercuei (approved):
All good.

PR #1419, opened 2026-06-11
Cleanups around stricter type usage

QuzarDC:
Various places where we've done retyping and refactoring of different internal and public types had small instances where we were still using the older/generic types. These changes help move those to use the proper types.

Additionally included a similar set of fixes around function prototypes. These were all found using `-Wc++-compat` as it is much stricter about implicit type conversion it was able to point them out.

### DIFF
--- a/kernel/arch/dreamcast/hardware/network/lan_adapter.c
+++ b/kernel/arch/dreamcast/hardware/network/lan_adapter.c
@@ -556,8 +556,6 @@ static void la_irq_hnd(uint32_t code, void *data) {
 /****************************************************************************/
 /* Netcore interface */
 
-netif_t la_if;
-
 static void set_ipv6_lladdr(void) {
     /* Set up the IPv6 link-local address. This is done in accordance with
        Section 4/5 of RFC 2464 based on the MAC Address of the adapter. */
--- a/kernel/arch/dreamcast/kernel/irq.c
+++ b/kernel/arch/dreamcast/kernel/irq.c
@@ -25,8 +25,8 @@
 
 /* Macros for accessing related registers. */
 #define TRA    ( *((volatile uint32_t *)(0xff000020)) ) /* TRAPA Exception Register */
-#define EXPEVT ( *((volatile uint32_t *)(0xff000024)) ) /* Exception Event Register */
-#define INTEVT ( *((volatile uint32_t *)(0xff000028)) ) /* Interrupt Event Register */
+#define EXPEVT ((irq_t) *((volatile uint32_t *)(0xff000024)) ) /* Exception Event Register */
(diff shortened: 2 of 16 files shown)


kernel/arch/dreamcast/hardware/network/lan_adapter.c
```diff
@@ -556,8 +556,6 @@ static void la_irq_hnd(uint32_t code, void *data) {
 /****************************************************************************/
 /* Netcore interface */
 
-netif_t la_if;
-
```
pcercuei, line 560: `duplciate` -> `duplicate`

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -24,9 +24,9 @@
 #include <kos/timer.h>
 
 /* Macros for accessing related registers. */
-#define TRA    ( *((volatile uint32_t *)(0xff000020)) ) /* TRAPA Exception Register */
-#define EXPEVT ( *((volatile uint32_t *)(0xff000024)) ) /* Exception Event Register */
-#define INTEVT ( *((volatile uint32_t *)(0xff000028)) ) /* Interrupt Event Register */
+#define TRA    ((irq_t) *((volatile uint32_t *)(0xff000020)) ) /* TRAPA Exception Register */
+#define EXPEVT ((irq_t) *((volatile uint32_t *)(0xff000024)) ) /* Exception Event Register */
+#define INTEVT ((irq_t) *((volatile uint32_t *)(0xff000028)) ) /* Interrupt Event Register */
```
pcercuei, line 29: `TRA` does not contain an `irq_t` value but a trap code, so while I kinda agree about the other two, this one should probably not use this type.

kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -257,13 +257,11 @@ void irq_handle_exception(int code) {
...
 static void irq_handle_trapa(irq_t code, irq_context_t *context, void *data) {
-    const struct irq_cb *hnd, *handlers = data;
-    uint32_t vec;
-
     (void)code;
+    const struct irq_cb *hnd, *handlers = data;
 
     /* Get the trapa vector */
-    vec = TRA >> 2;
+    irq_t vec = TRA >> 2;
```
pcercuei: This is not an `irq_t`, rather an offset to search the `trapa_handlers` array.

The actual `irq_t` would be `IRQ_TRAP_CODE(TRA >> 2)`, which we should actually be passing as "vec" to the handler (it's been working until now because nobody checks the irq code in their handlers).

QuzarDC: Oho! Right. Updated. Glad this little pedantic cleanup led to finding an actual bug :)

pcercuei (approved):

PR #1422, opened 2026-06-13
sd: Use time-based timeouts for SD card busy waits

andressbarajas:
While looking at sci code I see that we dont have a scif_spi_write_data() equivalent of scif_spi_read_data(). I was able to add that and it increased write speeds by about ~10% but writes to my old KINGSTON 8gb SDHC card started to fail using it. Looking at the SD code we don't follow the write timeout spec which says we should have about a 500ms timeout to support SD cards all the way up to SDXC. The current timeout is 100ms which is fine for reads but not enough for writes.

This PR adds an sd_wait_ready() util function that polls against a time-based deadline and replaces every section of code that waits for the card to not be busy.

### DIFF
--- a/kernel/arch/dreamcast/hardware/sd.c
+++ b/kernel/arch/dreamcast/hardware/sd.c
@@ -22,11 +22,14 @@
 #include <kos/net.h>
 
 #include <kos/blockdev.h>
+#include <kos/timer.h>
 #include <kos/dbglog.h>
 
-#define MAX_RETRIES     5000
-#define READ_RETRIES    50000
-#define WRITE_RETRIES   50000
+#define MAX_RETRIES       5000
+#define READ_RETRIES      50000
+
+/* SD spec allows a card to signal busy for up to 500ms */
+#define READY_TIMEOUT_MS  500
 
 #define CMD(n) ((n) | 0x40)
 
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/hardware/sd.c
```diff
@@ -176,20 +179,26 @@ static int sci_init_wrapper(bool fast) {
...
-    uint8_t pkt[6];
+    uint64_t timeout = timer_ms_gettime64() + READY_TIMEOUT_MS;
 
-    /* Wait for the SD card to be ready to accept our command... */
     spi_rw_byte(0xFF);
     do {
         rv = spi_rw_byte(0xFF);
-        ++i;
-    } while(rv != 0xFF && i < MAX_RETRIES);
+    } while(rv != 0xFF && timer_ms_gettime64() < timeout);
```
pcercuei: I would suggest you to use `thd_poll()` here instead, especially since the polling is expected to take a very long time.

andressbarajas: Using thd_poll() made all writes ~4% slower so I timed sd_wait_ready() on both my best and worst SD cards and found that the busy loop almost always completes within ~12–44 us, with a few cases reaching ~8 ms on both cards. I got 178ms for the worst case on my worst SD card. So I went with a hybrid approach instead so we spin for the most common short case, and only fall back to thd_poll() when the card stalls long enough to be worth yielding for.

pcercuei: 8 ms is two complete scheduler ticks, it's already quite a long time. Can you try with the fast spin limited to e.g. 200 µs?

andressbarajas: Done and force pushed.

pcercuei: Does it work just as well?

andressbarajas: Yup, I also tested it with the other PR stacked on this one and it works fine.

pcercuei (approved):

PR #1424, opened 2026-06-13
dc: Add scif_spi_write_data() and speed up SPI transfers

andressbarajas:
While working on the SD driver I noticed scif_spi_read_data() had a fast path but there was no scif_spi_write_data() counterpart. This PR adds the missing function and then takes both bulk paths a step further.

- New scif_spi_write_data() — clocks the buffer out 4 bytes at a time instead of per byte. Rather than a straight mirror of the read function, it handles any alignment and length: unaligned leading bytes and leftover trailing bytes go one at a time, the aligned middle uses the fast loop, and short transfers skip the setup entirely.

- Reworked scif_spi_read_data() the same way, so it no longer drops to a byte read just because the buffer is unaligned or the length isn't a multiple of 4.

- Dropped SD_WAIT() from scif_spi_write_byte() — the inter-bit nop delay didn't appear to do anything on the hardware I tested, and removing it picked up speed(+ 1 %).

- Wired the SD and W5500 drivers to use the new bulk write.

Net result: ~10% faster SD writes over SCIF-SPI on real hardware.  This speedup broke writing many blocks at a time on my KINGSTON 8gb SD card but this PR #1422 fixes that and should be merged first.

This is a draft because I want @DC-SWAT input on if we can remove SD_WAIT entirely (its still in scif_spi_rw_byte()). If so, then I can also include those in this PR.

EDIT: Was given his approval on Discord

### DIFF
--- a/kernel/arch/dreamcast/hardware/scif-spi.c
+++ b/kernel/arch/dreamcast/hardware/scif-spi.c
@@ -324,14 +324,18 @@ void scif_spi_read_data(uint8_t *buffer, size_t len) {
     uint32_t data;
     uint32_t *ptr;
 
-    /* Less optimized version for unaligned buffers or lengths not divisible by
-       four. */
-    if((((uint32_t)buffer) & 0x03) || (len & 0x03)) {
+    /* Not worth the alignment/setup overhead for short reads. */
+    if(len < 16) {
         while(len--) {
             *buffer++ = scif_spi_read_byte();
         }
-
         return;
+    }
+
+    /* Read single bytes until the buffer is 4-byte aligned. */
+    while(((uint32_t)buffer) & 0x03) {
+        *buffer++ = scif_spi_read_byte();
+        --len;
     }
 
     b = 0xff;
@@ -339,7 +343,7 @@ void scif_spi_read_data(uint8_t *buffer, size_t len) {
     ptr = (uint32_t *)buffer;
     SCSPTR2 = tmp;
 
-    for(; len > 0; len -= 4) {
+    for(; len >= 4; len -= 4) {
         SCSPTR2 = tmp | PTR2_CTSDT;
         b = (b << 1) | (SCSPTR2 & PTR2_SPB2DT);   /* 7 */
         SCSPTR2 = tmp;
@@ -443,6 +447,13 @@ void scif_spi_read_data(uint8_t *buffer, size_t len) {
         b = (b << 1) | (SCSPTR2 & PTR2_SPB2DT);   /* 0 */
         SCSPTR2 = tmp;
         data |= b << 24;
+
         *ptr++ = data;
+    }
+
+    /* Read any leftover bytes. */
+    buffer = (uint8_t *)ptr;
+    while(len--) {
+        *buffer++ = scif_spi_read_byte();
     }
 }
--- a/kernel/arch/dreamcast/hardware/scif-spi.c
+++ b/kernel/arch/dreamcast/hardware/scif-spi.c
@@ -169,7 +169,7 @@ void scif_spi_write_byte(uint8_t b) {
 
 void scif_spi_write_data(const uint8_t *buffer, size_t len) {
     uint16_t tmp;
-    uint8_t bit;
+    uint32_t bit;
     uint32_t data;
     const uint32_t *ptr;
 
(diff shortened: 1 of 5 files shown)


kernel/arch/dreamcast/hardware/scif-spi.c
```diff
@@ -187,6 +187,111 @@ void scif_spi_write_byte(uint8_t b) {
...
+        SCSPTR2 = tmp | (bit = (data >> 28) & 0x01);
+        SCSPTR2 = tmp | bit | PTR2_CTSDT;
+        SCSPTR2 = tmp | (bit = (data >> 27) & 0x01);
+        SCSPTR2 = tmp | bit | PTR2_CTSDT;
+        SCSPTR2 = tmp | (bit = (data >> 26) & 0x01);
+        SCSPTR2 = tmp | bit | PTR2_CTSDT;
+        SCSPTR2 = tmp | (bit = (data >> 25) & 0x01);
+        SCSPTR2 = tmp | bit | PTR2_CTSDT;
+        SCSPTR2 = tmp | (bit = (data >> 24) & 0x01);
+        SCSPTR2 = tmp | bit | PTR2_CTSDT;
```
pcercuei: I feel like a few loops would work just as well while keeping the code (both C and ASM) tiny.

Something like:
```c
unsigned int i, j;
int32_t tmp2;

for (j = 4; j > 0; j++) {
    tmp2 = (int32_t)(data << (8 * (j - 1)));

    for (i = 0; i < 8; i++) {
        SCSPTR2 = tmp | (tmp2 < 0);
        SCSPTR2 = tmp | (tmp2 < 0) | PTR2_CTSDT;
        tmp2 <<= 1;
    }
}
```

pcercuei: Probably even better:
```c
unsigned int i;
int32_t tmp2;

/* If source is little-endian, we need to swap the bytes */
tmp2 = (int32_t)htonl(data);

for (i = 0; i < 32; i++) {
    SCSPTR2 = tmp | (tmp2 < 0);
    SCSPTR2 = tmp | (tmp2 < 0) | PTR2_CTSDT;
    tmp2 <<= 1;
}
```

andressbarajas: I made the change and I tested it out on three different SD cards. I lose about a kilobyte per second when writing but I think it's worth the save in text space.

pcercuei: That's 1 KiB/s out of how many?

andressbarajas: For my old ("Bad"/ worst performance) SD Card

Master Branch:
499–502 ms   (~1048 KB/s)

Before (Unrolled):
454–457 ms (~1152 KB/s)

After (Nested for-loop):
456–459 ms (~1149 KB/s)

So instead of a ~10% speed up we get ~9.7% instead.

pcercuei: One last test, can you try to add `__attribute__((optimize("unroll-loops")))` to the function declaration?
On compiler-explorer it seems to generate better code than the hand-unrolled one.

andressbarajas: ~2 percent regression with __attribute__((optimize("unroll-loops"))):
465–467 ms (~1125 KB/s)

Using KOS default flags and no LTO.

kernel/arch/dreamcast/hardware/scif-spi.c
```diff
@@ -192,75 +192,15 @@ void scif_spi_write_data(const uint8_t *buffer, size_t len) {
...
-        SCSPTR2 = tmp | (bit = (data >> 27) & 0x01);
-        SCSPTR2 = tmp | bit | PTR2_CTSDT;
-        SCSPTR2 = tmp | (bit = (data >> 26) & 0x01);
-        SCSPTR2 = tmp | bit | PTR2_CTSDT;
-        SCSPTR2 = tmp | (bit = (data >> 25) & 0x01);
-        SCSPTR2 = tmp | bit | PTR2_CTSDT;
-        SCSPTR2 = tmp | (bit = (data >> 24) & 0x01);
-        SCSPTR2 = tmp | bit | PTR2_CTSDT;
+        /* Data is transmitted in big-endian */
+        data = __builtin_bswap32(*ptr++);
```
pcercuei, line 196: One nit here - it's not that the data is transmitted as big-endian, it's that we're reading it from little-endian memory (my original comment was wrong) in 4-byte chunks, so the first byte ends up in the lowest bits.

pcercuei (dismissed):
ACK from me, with or without the nitpick below (I'll re-ack if you update it).

kernel/arch/dreamcast/hardware/scif-spi.c
andressbarajas: @pcercuei Updated comment and force pushed.

pcercuei (approved):

PR #1425, opened 2026-06-14
ldscripts: Shave off common ldscript bits to common.xc

pcercuei:
A good part of the linker script is composed of arch-independent stuff. Move it to a common.xc script that can then be shared across the supported archs.


pcercuei:
Force-pushed with some whitespace fixes while I'm at it.

pcercuei:
Closing as deemed too early.

PR #1430, opened 2026-06-16
ldscripts: Fix c++ exceptions

andressbarajas:
Implemented fix shown in this discord post:

https://discord.com/channels/488332893324836864/614912396460556298/1515986867051561083

Verified against cpp/out_of_memory example that was failing before

### DIFF
--- a/utils/ldscripts/shlelf.xc
+++ b/utils/ldscripts/shlelf.xc
@@ -143,9 +143,9 @@ SECTIONS
   PROVIDE (__etext = .);
   PROVIDE (_etext = .);
   PROVIDE (etext = .);
-  .rodata         : 
-  { 
-    *(.rodata .rodata.* .gnu.linkonce.r.*) 
+  .rodata         :
+  {
+    *(.rodata .rodata.* .gnu.linkonce.r.*)
     . = ALIGN(4);
     __tdata_align = .;
     LONG (ALIGNOF(.tdata));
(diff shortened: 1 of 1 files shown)


pcercuei (changes requested):
Not ACKing this as-is.
- Zero information in the commit message about what was the problem and why the fix makes sense.
- I can drop lines of the fix and the example still works, which makes me think that not all of this is useful.
- Whitespace fixes scattered around in the same commit.

pcercuei:
Btw the lines I'm talking about that I can drop are the ones with the `?` symbol, I think we never have anything else than `crtbegin.o` / `crtend.o` in our toolchains. However I understand the script already referenced those in ctors/dtors, so I guess they belong here too. I am just making a mental note to see if we can drop them later.

andressbarajas:
I'll circle back to this PR tonight with more info in the commits and slimming down the changes.

andressbarajas:
@pcercuei Done. Split into two commits and added explanation. 

> Btw the lines I'm talking about that I can drop are the ones with the `?` symbol, I think we never have anything else than `crtbegin.o` / `crtend.o` in our toolchains. However I understand the script already referenced those in ctors/dtors, so I guess they belong here too. I am just making a mental note to see if we can drop them later.

I researched and it turns out GCC ships two flavors of those startup files. The plain ones (crtbegin.o, crtend.o) we use and the S flavor (crtbeginS.o, crtendS.o) which are only used for (shared libraries / PIE) that KOS never uses so I did not include them.

QuzarDC (approved):
With the explanations added and the trimming down this now all tracks fine.

@Kazade please be sure to let us know if this doesn't end up solving the original issue you'd been describing on discord.

pcercuei (approved):

Issue #1025, opened 2025-05-20
Some examples will not build with multiple jobs


darcagn:
`2ndmix` and `library` seem to frequently error out if `examples` are built with `-j2` or higher. Maybe others too.

For example, with `2ndmix`, `bin2c` errors out:
```
/opt/toolchains/dc/kos/utils/bin2c/bin2c romdisk.img romdisk_tmp.c romdisk
error: can't open input or output file
```

causing the project to fail to build:
```
/opt/toolchains/dc/toolchains/stable/lib/gcc/sh-elf/13.2.0/../../../../sh-elf/bin/ld: romdisk.o:/opt/toolchains/dc/kos/kernel/romdisk/romdiskbase.c:11:(.data.__kos_romdisk+0x0): undefined reference to `romdisk_data'
```

For `library`, it fails to build its dependencies in order.

pcercuei:
@darcagn I can't seem to reproduce...

pcercuei:
I know, that's why I switched it back to draft (It's probably still a good idea to have it).

The problem is more nasty and I'm working on it.

PR #1432, opened 2026-06-17
Examples makefile rules updates

QuzarDC:
Some changes drawn out from #1038 . They still don't correct #1025 and allow foolproof multithreaded building of examples, but should help some. I left behind the changes that change the rules into a provider of all the make targets as that's a much heavier change.

- Remove gcc version checks for unsupported versions.
- Ensure git ignores example build error files.
- Integrate the gcc version check logic just like we do for romdisk and exports, just define the version number.
- Add dependencies to romdisk objects to prevent failures.

The gcc version checker probably needs a bit further tweaking. As far as I can tell it works fine for our examples but I'm not too fluent with makefiles.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,5 @@ romdisk.img
 html/
 /doc/reference/
 /ports/
+/examples/errors.txt
+/examples/error_count.txt
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -95,10 +95,10 @@ $(patsubst %, _clean_dir_%, $(SUBDIRS)):
 
 # Define KOS_ROMDISK_DIR in your Makefile if you want these two handy rules.
 ifdef KOS_ROMDISK_DIR
-romdisk.img:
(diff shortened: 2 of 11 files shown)


pcercuei:
We have another GCC version check now, in `addons/libkosgcov/Makefile`, that should be updated too.

QuzarDC:
> We have another GCC version check now, in `addons/libkosgcov/Makefile`, that should be updated too.

Updated. Tested with 9.5.0, 16.0.1, and switching to naomi subarch. Seems libkosgconv happily builds with 9.5.0 though and just gives a warning about implicit declaration of the unsupported functions (and of course simply wouldn't work).

Had to tinker with the addons prefab to get this working, as it was operating off 'defaultall' for everything rather than 'all'.

gyrovorbis:
I'm laying on the floor with my phone about to pass out, so I haven't had a chance to test the changes yet, but...

Omg this is SO much better for the GCC min version checking, plus I see you added missing dependencies to a bunch of the rules. Love it.

PR #1433, opened 2026-06-17
Modernize CI

andressbarajas:
This PR updates the Github actions to the current versions so we don't get "Node.js 20 actions are deprecated" anymore. Also extend the no-tab and whitespace checks to display file and line number on failure.


pcercuei (approved):
That's some complex AWK magic.
I'll just trust you on that one.

PR #1434, opened 2026-06-17
maple: Don't reset controllers on shutdown.

### LINKED ISSUE #1384: sometimes L + R triggers won't read
I have found an issue when testing KOS `master`, commit `327a4a4`.

I noticed that every other boot or so, the L and R triggers on the controller would not respond.

I tested multiple controllers to be sure it wasn't bad hardware.

I created the following MRE:

```
#include <kos.h>
#include <dc/maple.h>
#include <dc/maple/controller.h>
#include <dc/biosfont.h>
#include <dc/video.h>

KOS_INIT_FLAGS(INIT_DEFAULT);

static void draw_str(int x, int y, const char *str) {
    bfont_draw_str(vram_s + y * 640 + x, 640, 0, str);
}

int main(void) {
    vid_set_mode(DM_640x480, PM_RGB565);

    maple_device_t *cont;
    cont_state_t *cs;
    char buf[128];
    int frame = 0;

    for (;;) {
        frame++;

        cont = maple_enum_type(0, MAPLE_FUNC_CONTROLLER);
        if (!cont) {
            vid_clear(0, 0, 0);
            draw_str(32, 32, "No controller found");
            vid_flip(1);
            continue;
        }

        cs = (cont_state_t *)maple_dev_status(cont);
        if (!cs) {
            vid_clear(0, 0, 0);
            draw_str(32, 32, "No status");
            vid_flip(1);
            continue;
        }

        vid_clear(0, 0, 0);

        draw_str(32, 32, "Trigger Test - press L/R triggers");
 
        sprintf(buf, "ltrig: %3d   rtrig: %3d", cs->ltrig, cs->rtrig);
        draw_str(32, 80, buf);
    
        sprintf(buf, "buttons: 0x%04x", cs->buttons);
        draw_str(32, 112, buf);
    
        sprintf(buf, "A:%d B:%d X:%d Y:%d St:%d",
                (cs->buttons & CONT_A) ? 1 : 0,
                (cs->buttons & CONT_B) ? 1 : 0,
                (cs->buttons & CONT_X) ? 1 : 0,
                (cs->buttons & CONT_Y) ? 1 : 0,
                (cs->buttons & CONT_START) ? 1 : 0);
        draw_str(32, 144, buf);
        
        sprintf(buf, "joyx: %4d  joyy: %4d", cs->joyx, cs->joyy);
        draw_str(32, 176, buf);
            
        sprintf(buf, "frame: %d", frame);
        draw_str(32, 224, buf);

        if (cs->ltrig > 0 || cs->rtrig > 0)
            draw_str(32, 272, "** TRIGGERS ACTIVE **");
        else
            draw_str(32, 272, "triggers idle");
            
        vid_flip(1); 
         
        if ((cs->buttons & CONT_START) && cs->ltrig > 200 && cs->rtrig > 200)
            break;
    }
        
    return 0;
}
```

andressbarajas:
This addresses #1384.

First, I tested on real hardware with the retail Marvel vs. Capcom 2 to see what the production behavior was. With the triggers held down and the stick offset I plugged the controller into the game and it also shows miscalibrated stick and trigger input. That confirms the controller calibrates its analog inputs whenever power is provided to it (or maple RESET command). If the controls aren't in their default positions at that moment, tough luck, the software can't prevent it and does not try to fix it.

So we should avoid the RESET command on controller init/shutdown because the controller was already calibrated "correctly" when the console powered on or when the controller was inserted into the port while the console is on. If a program uses those buttons to exit back to dcload, the controller is left miscalibrated and stays that way on the next run.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
@@ -80,7 +80,9 @@ static void maple_dev_reset(maple_device_t *dev) {
 }
 
 static void maple_dev_reset_all(void) {
-    MAPLE_FOREACH_BEGIN(MAPLE_FUNC_ANY, void, st)
+    /* Skip controllers: a RESET re-zeroes their analog axes at the current
+       position, so triggers/stick held at exit stay miscalibrated. */
+    MAPLE_FOREACH_BEGIN(MAPLE_FUNC_ANY & ~MAPLE_FUNC_CONTROLLER, void, st)
         maple_dev_reset(__dev);
         (void)st;
     MAPLE_FOREACH_END()


QuzarDC (approved):
Thanks, forgot to put it in after suggesting it. Tested as well to make sure that it stopped the VMU beeping and it does.

I remember when originally testing there were problems with stopping the beeping if everything wasn't being reset, but that was probably to do with some other aspects of the change.

PR #1435, opened 2026-06-18
Add nonnull, pure, and const where appropriate

QuzarDC:
I *think* a few of the `__pure` could be `__pure2` (`__const__`). There are also a few other places where a function could be pure if it didn't use `assert`, as we allow custom assert handlers. It *could* be worthwhile to come up with a `__pure_assert` or such that assigns pure when `NDEBUG` is set.

### DIFF
--- a/addons/libkosext2fs/ext2fs.h
+++ b/addons/libkosext2fs/ext2fs.h
@@ -60,11 +60,11 @@ typedef struct kos_blockdev {
     uint32_t l_block_size;
     int (*init)(struct kos_blockdev *d);
     int (*shutdown)(struct kos_blockdev *d);
-    int (*read_blocks)(struct kos_blockdev *d, uint32_t block, size_t count,
+    int (*read_blocks)(const struct kos_blockdev *d, uint32_t block, size_t count,
                        void *buf);
-    int (*write_blocks)(struct kos_blockdev *d, uint32_t block, size_t count,
+    int (*write_blocks)(const struct kos_blockdev *d, uint32_t block, size_t count,
                         const void *buf);
-    uint32_t (*count_blocks)(struct kos_blockdev *d);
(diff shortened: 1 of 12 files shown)


pcercuei:
Your `__pure` tags do nothing.
- In the ISO code and the PVR code, you use it on static functions. This adds no information that GCC doesn't already know.
- In `hw_to_dma_addr()` you add the tag to the declaration and not the prototype. This tag adds context to the callers of the functions, so it should be moved to the function prototype.

QuzarDC:
> Your `__pure` tags do nothing.
> 
>     * In the ISO code and the PVR code, you use it on static functions. This adds no information that GCC doesn't already know.

Ah, I guess that makes sense. The doc on it doesn't specify it's just for linking but it makes sense that gcc would already have the full scope of info to make those determinations for a static function. Perhaps it would be applicable just for cases where gcc wouldn't be able to determine that like a similar function that read from hardware that we knew to be invariant (though in that kind of case we could probably inform gcc with a `const` cast or something rather than using this attribute). Dropped all the iso and pvr ones.

>     * In `hw_to_dma_addr()` you add the tag to the declaration and not the prototype. This tag adds context to the callers of the functions, so it should be moved to the function prototype.

Fixed.

pcercuei:
I think for `hw_to_dma_addr()` you could actually use a `__pure2`, since the output is derived from the input value directly and exclusively.

QuzarDC:
> I think for `hw_to_dma_addr()` you could actually use a `__pure2`, since the output is derived from the input value directly and exclusively.

Makes sense, updated.

pcercuei (dismissed):

QuzarDC:
Ugh, the default for 'Update branch' button was via merge commit and I didn't notice. Was just trying to test out the new CI tests. Force pushed to get rid of the merge commit.

pcercuei (approved):
np.

PR #1436, opened 2026-06-18
pvrtex: Remove unused variables to silence warnings

andressbarajas:
While working on KOS its nice not to have to double check if the warnings in the terminal are KOS kernel related. Here are fixes for warnings I get when compiling KOS using make in the kos directory.

I dropped the unused opaquepix counter in pteAutoSelectPixelFormat and #if 0 out the unused dither map[]/pal[] tables in dither.cpp.


andressbarajas:
pvr_texture_encoder.c uses tabs instead of spaces throughout the whole file

utils/pvrtex/dither.cpp
```diff
@@ -237,6 +238,7 @@ static const unsigned char map[8*8] = {
 	10,58, 6,54, 9,57, 5,53,
 	42,26,38,22,41,25,37,21
 };
+#endif
```
pcercuei: Why not just remove those entirely?

andressbarajas: There is a main program at the bottom of this file that is also #if 0 out that references this. Thought Id take the same approach incase that program needs to be used in the future.

pcercuei: I'd drop all of that. If we want to retrieve the code in the future it's not lost, it's in the git history.

andressbarajas: Split this PR into two commits (1 per file) and force pushed. I found another #if 0 section of code that could be removed too so I did.

QuzarDC (approved):

PR #1438, opened 2026-06-18
filesystem_browse: Handle browsing via `.` and `..`

QuzarDC:
Without this, if navigating solely using the `A` button and going 'up' using `..` current_directory only removes the `..` and tries to navigate from the previous subfolder into another root folder. So going into `/vmu` then `..` and `/rd` will attempt to browse to `/vmu/rd`.

Open to more efficient or comprehensive ways to solve the issue, but it seems quite a bit that normalizing the path is what's required (though perhaps it could/should happen elsewhere).


pcercuei (approved):
Ack. Not sure the memset is needed though?

QuzarDC:
> Ack. Not sure the memset is needed though?

No, it's not, but I figured to follow the practice being used throughout the rest of the example for those global buffers and just clear it as well.

PR #1439, opened 2026-06-19
ci: Update to v7 for checkout so we no longer get Node.js 20 warnings

andressbarajas:
Sorry, I thought I updated to the latest checkout version in my last PR. Turns out the latest version is 7.

### DIFF
--- a/.github/workflows/autobuild.yml
+++ b/.github/workflows/autobuild.yml
@@ -19,7 +19,7 @@ jobs:
       matrix: ${{ steps.setup-matrix.outputs.matrix }}
 
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v7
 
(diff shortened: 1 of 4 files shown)


pcercuei:
Kinda unrelated, but I wonder if we can tell curl to retry from a different mirror if possible, because I keep having to restart CI jobs by hand.

pcercuei (approved):

scribam:
@andressbarajas I think you missed this line:
https://github.com/KallistiOS/KallistiOS/blob/38eab48da57101f18916e19ba229a7912b64051a/.github/workflows/autobuild.yml#L44
Warnings can be seen here: https://github.com/KallistiOS/KallistiOS/actions/runs/27838863914

andressbarajas:
Thanks for catching that!

PR #1440, opened 2026-06-19
Introduce MIE-JVS bridge driver for NAOMI systems

DC-SWAT:
The Maple device driver for the NAOMI and NAOMI 2 Maple I/O emulator (MIE) on port A.
This port can operate in two modes: Maple and MIE; the type is detected and switched automatically.
The MIE mode polls bundled JVS responses (player switches, system test, panel DIP and service switches, coin meters, and analog channels) and exposes native state alongside a mapped cont_state_t for existing code.
Also Z80 firmware init (hot and cold), wheel and pedal calibration, outputs control, EEPROM read/write, input callbacks and example are included.

Tested SEGA JVS I/O Type 1 board on:
- NAOMI 2 King of Route 66 (Upright cabinet) for wheel and pedals.
- NAOMI Universal Cabinet for 2 players.
- MP12-IONA-SB JVS->USB adapter for USB gamepads.

### DIFF
--- a/examples/dreamcast/mie/basic/mie_basic.c
+++ b/examples/dreamcast/mie/basic/mie_basic.c
@@ -1,0 +1,431 @@
+/* KallistiOS ##version##
+
+   mie_basic.c
+   Copyright (C) 2026 Ruslan Rostovtsev
+
+   Poll MIE/JVS inputs, run wheel/pedal calibration, blink JVS outputs,
+   and print mapped controller state plus native JVS fields.
+*/
+
+#include <stdio.h>
+#include <string.h>
+
+#include <kos.h>
+#include <kos/init.h>
+
+#if defined(_arch_sub_naomi)
+#include <dc/minifont.h>
+#define bfont_draw(a,b,c,d) minifont_draw((a),(b),(d))
+#else
+#include <dc/biosfont.h>
+#endif
+#include <dc/maple.h>
+#include <dc/maple/controller.h>
+#include <dc/maple/keyboard.h>
+#include <dc/maple/mie.h>
+#include <dc/maple/mouse.h>
+#include <dc/pvr.h>
+
+KOS_INIT_FLAGS(INIT_DEFAULT | INIT_MIE);
+
+#define UI_FONT_W   12
+#define UI_FONT_H   24
+#define UI_LINE_MAX 80
+#define UI_ROW_MAX  19
+
+static pvr_ptr_t ui_font_tex;
+
+static const char *port_mode_name(mie_port0_mode_t mode) {
+    switch(mode) {
+    case MIE_PORT0_JVS:
+        return "JVS";
+    case MIE_PORT0_MAPLE:
+        return "Maple";
+    default:
+        return "unknown";
+    }
+}
+
+static void ui_font_init(void) {
+    uint16_t *vram;
+    int x, y;
+
+    ui_font_tex = pvr_mem_malloc(256 * 256 * 2);
+    vram = (uint16_t *)ui_font_tex;
+
+    for(y = 0; y < 8; y++) {
+        for(x = 0; x < 16; x++) {
+            bfont_draw(vram, 256, 0, y * 16 + x);
+            vram += 16;
+        }
+        vram += 23 * 256;
+    }
+}
+
+static void ui_draw_char(float x1, float y1, int c) {
+    pvr_vertex_t vert;
+    int ix, iy;
+    float u1, v1, u2, v2;
+
+    if(c == ' ') {
+        return;
+    }
+
+    if(!(c > ' ' && c < 127)) {
+        return;
+    }
+
+    ix = (c % 16) * 16;
+    iy = (c / 16) * 24;
+    u1 = ix * 1.0f / 256.0f;
+    v1 = iy * 1.0f / 256.0f;
+    u2 = (ix + 12) * 1.0f / 256.0f;
+    v2 = (iy + 24) * 1.0f / 256.0f;
+
+    vert.flags = PVR_CMD_VERTEX;
+    vert.x = x1;
+    vert.y = y1 + UI_FONT_H;
+    vert.z = 1.0f;
+    vert.u = u1;
+    vert.v = v2;
+    vert.argb = PVR_PACK_COLOR(1.0f, 1.0f, 1.0f, 1.0f);
+    vert.oargb = 0;
+    pvr_prim(&vert, sizeof(vert));
+
+    vert.x = x1;
+    vert.y = y1;
+    vert.u = u1;
+    vert.v = v1;
+    pvr_prim(&vert, sizeof(vert));
+
+    vert.x = x1 + UI_FONT_W;
+    vert.y = y1 + UI_FONT_H;
+    vert.u = u2;
+    vert.v = v2;
+    pvr_prim(&vert, sizeof(vert));
+
+    vert.flags = PVR_CMD_VERTEX_EOL;
+    vert.x = x1 + UI_FONT_W;
+    vert.y = y1;
+    vert.u = u2;
+    vert.v = v1;
+    pvr_prim(&vert, sizeof(vert));
+}
+
+static void ui_draw_str(float x, float y, const char *str) {
+    pvr_poly_cxt_t cxt;
+    pvr_poly_hdr_t poly;
+    int i, len;
+
+    if(!str || !str[0]) {
+        return;
+    }
+
+    len = (int)strlen(str);
+
+    pvr_poly_cxt_txr(&cxt, PVR_LIST_TR_POLY,
+                     PVR_TXRFMT_ARGB1555 | PVR_TXRFMT_NONTWIDDLED,
+                     256, 256, ui_font_tex, PVR_FILTER_NONE);
+    pvr_poly_compile(&poly, &cxt);
+    pvr_prim(&poly, sizeof(poly));
+
+    for(i = 0; i < len; i++) {
+        ui_draw_char(x, y, str[i]);
+        x += UI_FONT_W;
+    }
+}
+
+static void ui_draw(const char lines[][UI_LINE_MAX], int count) {
+    float x = 8.0f;
+    float y = 8.0f;
+    int i;
+
+    pvr_wait_ready();
+    pvr_scene_begin();
+    pvr_list_begin(PVR_LIST_TR_POLY);
+
+    for(i = 0; i < count; i++) {
+        ui_draw_str(x, y, lines[i]);
+        y += UI_FONT_H;
+    }
+
+    pvr_list_finish();
+    pvr_scene_finish();
+}
+
+static volatile uint32_t coin_cb_count = 0;
+
+static void mie_coin_cb(mie_jvs_input_t input, uint32_t mask) {
+    (void)input;
+    (void)mask;
+    coin_cb_count++;
+}
+
+static void mie_service_cb(mie_jvs_input_t input, uint32_t mask) {
+    (void)input;
+    (void)mask;
+    mie_coin_add(0, 1, true);
+}
+
+static void outputs_blink(void) {
+    static uint64_t last_ms;
+    static uint8_t step;
+    uint64_t now;
+
+    if(mie_port0_mode() != MIE_PORT0_JVS) {
+        return;
+    }
+
+    now = timer_ms_gettime64();
+    if(now - last_ms < 500) {
+        return;
+    }
+
+    last_ms = now;
+    mie_jvs_set_outputs(MIE_JVS_OUTPUT_MASK(step), false);
+    step = (uint8_t)((step + 1) % MIE_JVS_OUTPUT_COUNT);
+}
+
+static const char *calib_step_name(mie_analog_calib_step_t step) {
+    switch(step) {
+    case MIE_ANALOG_CALIB_WHEEL:
+        return "turn wheel fully both ways, press A";
+    case MIE_ANALOG_CALIB_WHEEL_CENTER:
+        return "hold wheel centered, press A";
+    case MIE_ANALOG_CALIB_ACCEL:
+        return "press accel fully, press A";
+    case MIE_ANALOG_CALIB_BRAKE:
+        return "press brake fully, press A";
+    default:
+        return "idle";
+    }
+}
+
+static uint32_t ui_poll_aux_buttons(void) {
+    uint32_t btns = 0;
+
+    MAPLE_FOREACH_BEGIN(MAPLE_FUNC_CONTROLLER, cont_state_t, c)
+        if(c) {
+            btns |= c->buttons;
+        }
+    MAPLE_FOREACH_END()
+
+    MAPLE_FOREACH_BEGIN(MAPLE_FUNC_KEYBOARD, kbd_state_t, kbd)
+        if(kbd) {
+            if(kbd->key_states[KBD_KEY_A].is_down) {
+                btns |= CONT_A;
+            }
+            if(kbd->key_states[KBD_KEY_B].is_down) {
+                btns |= CONT_B;
+            }
+            if(kbd->key_states[KBD_KEY_ESCAPE].is_down) {
+                btns |= CONT_START;
+            }
+        }
+    MAPLE_FOREACH_END()
+
+    MAPLE_FOREACH_BEGIN(MAPLE_FUNC_MOUSE, mouse_state_t, m)
+        if(m) {
+            if(m->buttons & MOUSE_LEFTBUTTON) {
+                btns |= CONT_A;
+            }
+            if(m->buttons & MOUSE_RIGHTBUTTON) {
+                btns |= CONT_B;
+            }
+            if(m->buttons & MOUSE_SIDEBUTTON) {
+                btns |= CONT_START;
+            }
+        }
+    MAPLE_FOREACH_END()
+
+    return btns;
+}
+
+static void ui_buttons_text(char *linebuf, size_t len, uint32_t btns) {
+    size_t n = 0;
+
+    linebuf[0] = '\0';
+
+    if(btns & CONT_START)
+        n += snprintf(linebuf + n, len - n, "START ");
+    if(btns & CONT_A)
+        n += snprintf(linebuf + n, len - n, "A ");
+    if(btns & CONT_B)
+        n += snprintf(linebuf + n, len - n, "B ");
+    if(btns & CONT_X)
+        n += snprintf(linebuf + n, len - n, "X ");
+    if(btns & CONT_Y)
+        n += snprintf(linebuf + n, len - n, "Y ");
+    if(btns & CONT_Z)
+        n += snprintf(linebuf + n, len - n, "Z ");
+    if(btns & CONT_C)
+        n += snprintf(linebuf + n, len - n, "C ");
+    if(btns & CONT_D)
+        n += snprintf(linebuf + n, len - n, "D ");
+    if(btns & CONT_DPAD_UP)
+        n += snprintf(linebuf + n, len - n, "UP ");
+    if(btns & CONT_DPAD_DOWN)
+        n += snprintf(linebuf + n, len - n, "DOWN ");
+    if(btns & CONT_DPAD_LEFT)
+        n += snprintf(linebuf + n, len - n, "LEFT ");
+    if(btns & CONT_DPAD_RIGHT)
+        n += snprintf(linebuf + n, len - n, "RIGHT ");
+
+    if(n == 0) {
+        snprintf(linebuf, len, "(none)");
+    }
+}
+
+static void ui_sys_line(char *linebuf, size_t len,
+                        const mie_jvs_system_t *sys) {
+    size_t n;
+
+    n = snprintf(linebuf, len, "sys=0x%02x", sys->raw);
+    if(sys->test) {
+        n += snprintf(linebuf + n, len - n, " test");
+    }
+    if(sys->raw == 0) {
+        snprintf(linebuf + n, len - n, " (idle)");
+    }
+}
+
+static void ui_panel_line(char *linebuf, size_t len,
+                            const mie_jvs_panel_t *panel) {
+    snprintf(linebuf, len,
+             "panel dip=0x%02x psw=0x%02x test=%d service=%d",
+             panel->dip.raw, panel->psw.raw,
+             panel->psw.test, panel->psw.service);
+}
+
+static int ui_build(mie_state_t *st, char lines[][UI_LINE_MAX]) {
+    int n = 0;
+
+    snprintf(lines[n++], UI_LINE_MAX, "MIE basic example");
+    snprintf(lines[n++], UI_LINE_MAX, "port A mode: %s",
+             port_mode_name(mie_port0_mode()));
+
+    if(!st) {
+        snprintf(lines[n++], UI_LINE_MAX, "No MIE device (waiting...)");
+        snprintf(lines[n++], UI_LINE_MAX,
+                 "outputs=0x%08lx (running light 0..%d)",
+                 (unsigned long)mie_jvs_get_outputs(),
+                 MIE_JVS_OUTPUT_COUNT - 1);
+        snprintf(lines[n++], UI_LINE_MAX,
+                 "Press START on pad or Esc to exit");
+        return n;
+    }
+
+    snprintf(lines[n++], UI_LINE_MAX, "Mapped controller (st->cont):");
+    ui_buttons_text(lines[n++], UI_LINE_MAX, st->cont.buttons);
+    snprintf(lines[n++], UI_LINE_MAX,
+             "joyx=%d ltrig=%d rtrig=%d",
+             st->cont.joyx, st->cont.ltrig, st->cont.rtrig);
+    snprintf(lines[n++], UI_LINE_MAX, "Native JVS (st->jvs):");
+    ui_sys_line(lines[n++], UI_LINE_MAX, &st->jvs.system);
+    snprintf(lines[n++], UI_LINE_MAX, "p1=0x%04x p2=0x%04x",
+             st->jvs.p1.raw, st->jvs.p2.raw);
+    ui_panel_line(lines[n++], UI_LINE_MAX, &st->jvs.panel);
+    snprintf(lines[n++], UI_LINE_MAX,
+             "meter0=%u meter1=%u coin callbacks: %lu",
+             mie_get_coin_meter(0), mie_get_coin_meter(1),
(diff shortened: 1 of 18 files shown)


kernel/arch/dreamcast/exports.txt
```diff
@@ -236,6 +237,37 @@ vmu_block_read
...
+mie_analog_norm_wheel
+mie_analog_norm_accel
+mie_analog_norm_brake
+mie_get_id
+mie_get_eeprom
+mie_set_eeprom
+mie_eeprom_crc16
+mie_eeprom_fix_crc
+mie_init_fw
+
```
pcercuei, line 271: I think those exports should be in `exports-naomi.txt`, no?

kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
```diff
@@ -218,6 +219,10 @@ void maple_wait_scan(void) {
     /* Wait for it to finish */
     thd_poll((thd_cb_t)maple_scan_done, &maple_state, 0);
 
+    if(mie_port0_mode() != MIE_PORT0_MAPLE) {
+        mie_init_scan();
```
pcercuei: I think you need to add a `KOS_INIT_FLAG()` + `KOS_INIT_FLAG_WEAK()` + `KOS_INIT_FLAG_CALL()` for it, otherwise you are unconditionally linking the MIE driver even when `INIT_MIE` is not set.

pcercuei (commented):
This is a huge PR so I'll do a more in-depth review later, but here are two comments I have on the code after a quick look at it.

kernel/arch/dreamcast/exports.txt
DC-SWAT: No, this also will work for pristine builds, that run via DreamShell on NAOMI.
exports-naomi.txt is for NAOMI with original BIOS.

kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
DC-SWAT: Yes, you're right, I did.

kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
```diff
@@ -212,12 +214,16 @@ static int maple_scan_done(maple_state_t *state) {
     return state->scan_ready_mask == 0xf;
 }
 
+KOS_INIT_FLAG_WEAK(mie_init_scan, true);
```
pcercuei: The boolean argument should be "true" only if `INIT_MIE` is in the default init flags, which in this case is true for Naomi and false for pristine.

This is an unprecedented case but I think it is still possible to fix it cleanly.

You could do `KOS_INIT_FLAG_WEAK(mie_init_scan, __is_defined(_arch_sub_naomi))`;
Here and for the other KOS_INIT_FLAG_WEAK() you added.

The `__is_defined()` macro requires `_arch_sub_naomi` to be defined to 1 (if defined at all), so you would have to update `environ_base.sh` to replace the `-D_arch_sub_${KOS_SUBARCH}` with `-D_arch_sub_${KOS_SUBARCH}=1`

DC-SWAT: But `_arch_sub_naomi` is not defined for `pristine` build. I need this for both `pristine` and `naomi` builds.

Edit:
Ah, I got it.

kernel/arch/dreamcast/hardware/maple/maple_init_shutdown.c
```diff
@@ -88,6 +91,18 @@ static void maple_dev_reset_all(void) {
...
 }
 
+static uint8_t maple_dma_addr_prot_byte(uint32_t addr) {
+    return ((((addr) & MEM_AREA_CACHE_MASK) >> 20) - 0x80) & 0xff;
+}
+
+static uint32_t maple_dma_mem_protection(void) {
+    uint32_t ram_lo, ram_hi;
+
+    ram_lo = maple_dma_addr_prot_byte(0x0c000000);
```
QuzarDC, line 101: Is this the same scheme used by the g2dma just with a different magic protection for the top 16 bits? It looks quite similar but just arrived at in a different way than what's in g2dma.c

DC-SWAT: Yes, G1, G2 and Maple DMAs has protection with the same scheme.
But they are configured slightly differently. G1 and G2 unlock for the entire SH4 memory range, while Maple DMA only unlocks for the system memory area, because it doesn't transfer to/from external devices.

DC-SWAT:
The first force push is a fix for coin handling on the original hardware.
The second force push is a rebase.

DC-SWAT:
Force-push for improving docs.

DC-SWAT:
Just fixed commit message.

kernel/arch/dreamcast/include/dc/maple.h
```diff
@@ -95,19 +97,19 @@ __BEGIN_DECLS
...
+#define MAPLE_SPEED_4MBPS       0x0200          /**< \brief 4Mbps bus speed */
+#define MAPLE_SPEED_8MBPS       0x0300          /**< \brief 8Mbps bus speed */
 #define MAPLE_SPEED_TIMEOUT(n)  ((n) << 16)     /**< \brief Bus timeout macro */
 
-#ifndef _arch_sub_naomi
-#define MAPLE_RESET1_MAGIC      0x6155404f      /**< \brief First reset value */
-#else
-#define MAPLE_RESET1_MAGIC      0x6155405f
-#endif
+#define MAPLE_DMA_PROT_MAGIC    0x6155          /**< \brief Key in bits 31-16; lo/hi bytes set the allowed range */
```
pcercuei: I think it would be better to have the "magic" as a 32-bit value, aka. 0x61550000, then you can OR the other bits.

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+    MIE_DEF_CMD_PENDING = 0,
+    MIE_DEF_CMD_OK = 1,
+    MIE_DEF_CMD_ERR = -1
+} mie_def_cmd_state_t;
+
+static maple_driver_t mie_drv;
+static maple_frame_t mie_cmd_frame;
+static mie_drv_state_t mie_static_state;
+static maple_device_t mie_static_dev;
+static uint8_t mie_io_tx_arr[32] __attribute__((aligned(32)));
```
pcercuei: Nitpick here, but I'd prefer the C variant if possible, `alignas(32) static uint8_t mie_io_tx_arr[32];` (with the `<stdalign.h>` include)

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+    irq_mask_t save;
+    bool drained;
+
+    save = irq_disable();
+    drained = mie_static_state.io_state == MIE_IO_IDLE &&
+              !mie_static_dev.frame.queued;
+    irq_restore(save);
+
+    return drained;
+}
```
pcercuei, line 150: This can be written much simpler with `irq_disable_scoped()`:
```c
static bool mie_bus_drained(void) {
    irq_disable_scoped();

    return mie_static_state.io_state == MIE_IO_IDLE && !mie_static_dev.frame_queued;
}
```

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+        return true;
+    }
+
+    /* Wait until periodic finishes the current poll cycle. */
+    genwait_wait(&mie_bus_wq, "mie_bus", 2000);
+
+    /* Re-check after wake; periodic signals mie_bus_wq when drained. */
+    if(mie_bus_drained()) {
+        return true;
+    }
```
pcercuei: I think this code block could be replaced with a `thd_poll(mie_bus_drained, NULL, 2000);` (if "mie_bus_drained" prototype is updated accordingly for the extra parameter)

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+    strncpy(mie_static_dev.info.product_license, "SEGA ENTERPRISES,LTD.",
+            sizeof(mie_static_dev.info.product_license) - 1);
+    mie_static_dev.info.area_code = 0xff;
+
+    mie_drv.functions = MAPLE_FUNC_MIE;
+    mie_drv.name = "MIE Driver";
+    mie_drv.periodic = mie_periodic;
+    mie_drv.status_size = sizeof(mie_drv_state_t);
+    mie_drv.attach = mie_attach;
+    mie_drv.detach = mie_detach;
```
pcercuei: All these initializations for `mie_static_dev` and `mie_drv` could be static initializations instead.

kernel/arch/dreamcast/hardware/maple/mie/mie_calib.c
```diff
@@ -0,0 +1,257 @@
...
+    if(c->wheel.min >= c->wheel.center || c->wheel.center >= c->wheel.max) {
+        return false;
+    }
+    if(c->accel.min >= c->accel.max) {
+        return false;
+    }
+    if(c->brake.min >= c->brake.max) {
+        return false;
+    }
+    return true;
```
pcercuei:
```c
return c->wheel.min < c->wheel.center
    && c->wheel.center < c->wheel.max
    && c->accel.min < c->accel.max
    && c->brake.min < c->brake.max;
```

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+    for(i = 0; i < 200; i++) {
+        resp = mie_cmd_resp(MIE_MAPLE_IDLE_PROBE, NULL, 0);
+        if(resp &&
+           resp->response != MAPLE_RESPONSE_AGAIN &&
+           resp->response != MAPLE_RESPONSE_NONE) {
+            return true;
+        }
+
+        thd_sleep(10);
+    }
```
pcercuei: This should be a `thd_poll()` with a timeout of 2000 ms.

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+    maple_response_t *resp;
+
+    resp = mie_cmd_resp(cmd, send_buf, length);
+    if(!resp) {
+        return NULL;
+    }
+
+    if((uint8_t)resp->response != (uint8_t)expect_resp) {
+        return NULL;
+    }
```
pcercuei, line 730: You can factorize here,
```c
if (!resp || (uint8_t)resp->response != (uint8_t)expect_resp) {
    return NULL;
}
```

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+    if(!out || max_slots < 1 || !mie_z80_init_done) {
+        return false;
+    }
+
+    slots = max_slots;
+    if(slots > MIE_JVS_COIN_SLOTS) {
+        slots = MIE_JVS_COIN_SLOTS;
+    }
+
+    save = irq_disable();
```
pcercuei: irq_disable_scoped() here

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+}
+
+bool mie_coin_decrease(uint8_t slot, uint16_t amount, bool block) {
+    if(!mie_z80_init_done) {
+        return false;
+    }
+
+    if(amount == 0) {
+        return false;
+    }
```
pcercuei, line 768: Factorize here,
```c
if(!mie_z80_init_done || amount == 0) {
    return false;
}
```

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+}
+
+bool mie_coin_add(uint8_t slot, uint16_t amount, bool block) {
+    if(!mie_z80_init_done) {
+        return false;
+    }
+
+    if(amount == 0) {
+        return false;
+    }
```
pcercuei, line 779: Factorize here

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,991 @@
...
+
+    return mie_jvs_coin_io(JVS_CMD_COINADD, slot, amount, block);
+}
+
+static inline uint8_t mie_jvs_output_bit(uint8_t index) {
+    return (uint8_t)(1 << (7 - index));
```
pcercuei: `return BIT(7 - index)`

kernel/arch/dreamcast/hardware/maple/mie/mie_fw.c
```diff
@@ -0,0 +1,582 @@
...
+    return (uint16_t)(n >> 16);
+}
+
+void mie_eeprom_fix_crc(uint8_t *eeprom) {
+    uint16_t crc;
+    uint8_t size;
+
+    if(!eeprom) {
+        return;
+    }
```
pcercuei: All those checks that "should never happen" should be asserts instead.

pcercuei (commented):
I did another review, a little bit more in-depth. Overall I find the driver way too complex, it is very hard to see what's doing what, there are almost no comments, no documentation.

kernel/arch/dreamcast/hardware/maple/mie/mie.c
DC-SWAT: Ah, I keep forgetting about this.

DC-SWAT:
> I did another review, a little bit more in-depth. Overall I find the driver way too complex, it is very hard to see what's doing what, there are almost no comments, no documentation.

It's complicated because the JVS system itself is complex. And it was not my first goal to describe it in detail.
I've made all the changes you suggested and a few more. You can continue the review, but I'll test it on my hardware tomorrow to make sure nothing breaks.

DC-SWAT:
Tested on real hardware and rebased.

DC-SWAT:
- Fixed `attach()` and `detach()` handlers after rebase.
- Fixed `mie_poll_idle()` after refactor to `thd_poll()`.
- Tested on real hardware.

DC-SWAT:
Added more documentation to the code.

DC-SWAT:
Fixed and improved coins handling.

DC-SWAT:
Function check, dynamic poll config, 24-bit outputs, 4 coin slots with auto COINDEC.
Also some cleanup. Tested on real hardware.

DC-SWAT:
Just rebased.

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+    }
+
+    frm->state = MAPLE_FRAME_VACANT;
+    frm->queued = 0;
+}
+
+static int mie_bus_drained(void *data) {
+    (void)data;
+
+    irq_disable_scoped();
```
pcercuei: That's a callback for `thd_poll()` so it already executes in interrupt context, you can drop the irq_disable_scoped().

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+        if(ms->io_state == MIE_IO_NEED_FETCH) {
+            return mie_queue_jvs_fetch(dev);
+        }
+
+        return -1;
+    }
+
+    mie_io_tick(dev);
+    rv = mie_poll(dev);
+    return rv;
```
pcercuei: Just `return mie_poll(dev);`, you're not using "rv" anywhere else.

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+        return;
+    }
+
+    maple_driver_foreach(drv, mie_periodic_dev);
+}
+
+static int mie_bus_idle_check(void *data) {
+    (void)data;
+
+    irq_disable_scoped();
```
pcercuei: Same as above, this is a thd_poll() callback so it's already running in IRQ context, no need to disable IRQs here.

DC-SWAT:
Just cleanup `dist` target for mie basic example Makefile, sorry.

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+    if(!out || max_slots < 1 || !mie_z80_init_done) {
+        return false;
+    }
+
+    slots = max_slots;
+    if(slots > MIE_JVS_COIN_SLOTS) {
+        slots = MIE_JVS_COIN_SLOTS;
+    }
+
+    irq_disable_scoped();
```
pcercuei: Not sure why you need to disable IRQs here?

kernel/arch/dreamcast/hardware/maple/mie/mie_input.c
```diff
@@ -0,0 +1,498 @@
...
+}
+
+static int mie_callback_register(mie_callback_params_t *params) {
+    const kthread_attr_t thd_attr = {
+        .stack_size = MIE_CALLBACK_THD_STACK_SIZE,
+        .prio = PRIO_DEFAULT,
+        .label = "mie_callback"
+    };
+
+    params->worker = thd_worker_create_ex(&thd_attr, &mie_cb_thread, params);
```
pcercuei: I notice that you create one worker thread per callback registered, that sounds expensive.
What I would suggest you to do instead, is to create a single workqueue (from `<kos/workqueue.h>`, then your `mie_callback_params_t` can have a `workqueue_job_t job;` field, initialized to `{ .cb = mie_cb }`.

Then your callback becomes
```c
static void mie_cb(workqueue_t *wq, struct workqueue_job *job) {
    mie_callback_params_t *params = __containerof(job, mie_callback_params_t, job);
    /* Handle stuff */
}
```

And instead of waking up the worker thread you just add the job to the workqueue:
```c
workqueue_enqueue(mie_wq, &params->job);
```

The caveat is that with the workqueue all the callbacks run in the same thread, I don't know if that's a problem for you or not.

kernel/arch/dreamcast/hardware/maple/mie/mie_input.c
```diff
@@ -0,0 +1,498 @@
...
+    uint32_t prev_val;
+    uint32_t cur_val;
+
+    if(!baseline) {
+        return;
+    }
+
+    if(mutex_trylock(&cbs_mtx)) {
+        return;
+    }
```
pcercuei, line 429: If you use trylock() because it's running in IRQ context, better use `mutex_lock_irqsafe()`.

kernel/arch/dreamcast/hardware/maple/maple_irq.c
```diff
@@ -183,8 +184,16 @@ static void vbl_autodet_callback(maple_state_t *state, maple_frame_t *frm) {
...
+    maple_device_t *dev;
+    maple_frame_t *frm;
+
+    if(state->detect_port_next == 0 && state->port0_mie) {
+        state->detect_port_next = 1;
+        return;
+    }
+
+    dev = maple_state.ports[state->detect_port_next].units[0];
+    frm = (dev != NULL) ? &dev->frame : &detect_frame;
```
pcercuei: You can move the variable declarations after the if() check:
```c
if (state-> ...) {
    return;
}

maple_device_t *dev = maple_state.ports[state->detect_port_next].units[0];
```

quzar has been actively updating the code to do this. It's allowed (and recommended) now.

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+static int mie_queue_buttons_request(maple_device_t *dev) {
+    mie_drv_state_t *ms = dev->status;
+
+    if(ms->io_state != MIE_IO_IDLE) {
+        return -1;
+    }
+
+    if(ms->io_backoff > 0) {
+        return -1;
+    }
```
pcercuei, line 252: You can factorize that a bit,
```c
if(ms->io_state != MIE_IO_IDLE || ms->io_backoff > 0) {
    return -1;
}
```

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+}
+
+static void mie_periodic(maple_driver_t *drv) {
+    (void)drv;
+
+    if(port0_mode != MIE_PORT0_JVS || !mie_jvs_ready) {
+        return;
+    }
+
+    maple_driver_foreach(drv, mie_periodic_dev);
```
pcercuei, line 569: I would invert the condition to simplify the code,
```c
if(port0_mode == MIE_PORT0_JVS && mie_jvs_ready) {
    maple_driver_foreach(drv, mie_periodic_dev);
}
```

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+            mie_port0_set_mode(MIE_PORT0_MAPLE);
+            return;
+        }
+    }
+
+    if(port0_mode == MIE_PORT0_MAPLE) {
+        return;
+    }
+
+    mie_init_common(NULL);
```
pcercuei: Invert the condition,
```c
if(port0_mode != MIE_PORT0_MAPLE) {
    mie_init_common(NULL);
}
```

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+    }
+    else if(mie_z80_try_active(&mie_static_dev)) {
+        mie_z80_init_done = 1;
+    }
+
+    if(!mie_z80_init_done) {
+        return;
+    }
+
+    mie_scan_complete();
```
pcercuei: Invert the condition,
```c
if(mie_z80_init_done) {
    mie_scan_complete();
}
```

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+}
+/* Unblocks periodic input polling after Z80 firmware is verified. */
+void mie_scan_complete(void) {
+    if(port0_mode != MIE_PORT0_JVS) {
+        return;
+    }
+
+    mie_jvs_ready = 1;
```
pcercuei: `mie_jvs_ready = port0_mode == MIE_PORT0_JVS;`

kernel/arch/dreamcast/hardware/maple/mie/mie.c
```diff
@@ -0,0 +1,1028 @@
...
+        mie_init();
+    }
+
+    if(!mie_drv_registered) {
+        return false;
+    }
+
+    mie_init_common(fw_path);
+
+    return mie_z80_init_done != 0;
```
pcercuei: Why isn't your `mie_z80_init_done` a bool in the first place? Same for all the other variables that are used as booleans but aren't the bool type.

kernel/arch/dreamcast/hardware/maple/mie/mie_fw.c
```diff
@@ -0,0 +1,567 @@
...
+        }
+        thd_sleep(20);
+    }
+
+    dbglog(DBG_ERROR, "mie: maple dev reset failed\n");
+    return false;
+}
+
+static maple_response_t *mie_io_retry(const void *req, int len,
+                                           int attempts, int delay_ms) {
```
pcercuei: Your `attempts` and `delay_ms` parameters are always the same, so they could just be hardcoded in the function's body.

pcercuei (changes requested):

kernel/arch/dreamcast/hardware/maple/mie/mie_input.c
DC-SWAT: I simply copied this mechanism from the controller. OK, I'll improve it.

pcercuei: Ah, I did not notice that the controller used the same pattern.
So I guess you can keep the worker here and I'll refactor that later alongside the controller driver.

DC-SWAT: I created a single worker thread with a list. The workqueue doesn't work from IRQ.

DC-SWAT:
I've made all the requested fixes, but I had to do something slightly different for the callbacks.
I also reverted back to two players and two coin acceptors per board, as more than two have never been used on a single board (though there seems to have been some potential). In such cases, the boards themselves are connected in series, but I don't yet support this in the driver; it's a rare case.

DC-SWAT:
Just better naming for few defines and rebase.

pcercuei (dismissed):

examples/dreamcast/mie/basic/mie_basic.c
```diff
@@ -0,0 +1,426 @@
...
+   and print mapped controller state plus native JVS fields.
+*/
+
+#include <stdio.h>
+#include <string.h>
+
+#include <kos.h>
+#include <kos/init.h>
+
+#include <dc/biosfont.h>
```
QuzarDC, line 20: Could you please adjust to something like the below? This allows correct building and running for the old subarch NAOMI.

```
#if defined(_arch_sub_naomi)
#include <dc/minifont.h>
#define bfont_draw(a,b,c,d) minifont_draw((a),(b),(d))
#else
#include <dc/biosfont.h>
#endif

QuzarDC (changes requested):
Finally got this to run on my NAOMI!

There's one small change for the example in order to have it also work with the subarch naomi. This is the first KOS thing I've gotten to run on the NAOMI, looking forward to testing it out with the steering wheel in my Crazy Taxi cab at some point :)

examples/dreamcast/mie/basic/mie_basic.c
DC-SWAT: Done

DC-SWAT:
Fix example for subarch NAOMI and rebase.

QuzarDC (approved):

PR #1441, opened 2026-06-19
New useful macros + UDIV API

pcercuei:
This PR adds two useful macros:
- `__container_of()`, which can be used to retrieve a pointer to a container struct from a pointer to one of its fields;
- `BITLL()`, which is the same as BIT() but will create a 64-bit mask.

The third commit introduces a new header, `<kos/udiv.h>`. This header contains a small API that can be used to create a pre-processed `udiv_t` divider value from a regular unsigned int. This udiv_t can then be used to divide another unsigned int very quickly, without the cost of a full division operation. This is very useful when a rarely updated divider value is used often.

None of these macros / APIs are used right now, but they will all be used by the AICA port. Also, some people (Falco) were already complaining about us missing the `__container_of()` macro, and some others (Ian) reportedly used the udiv code in their own projects with great success.

### DIFF
--- a/include/kos/regfield.h
+++ b/include/kos/regfield.h
@@ -19,12 +19,19 @@
 #include <kos/cdefs.h>
 __BEGIN_DECLS
 
-/** \brief  Create a mask with a bit set
+/** \brief  Create a 32-bit mask with a bit set
 
     \param  bit             The bit to set (from 0 to 31)
     \return                 A 32-bit mask with the corresponding bit set
  */
 #define BIT(bit)        (1u << (bit))
+
+/** \brief  Create a 64-bit mask with a bit set
+
+    \param  bit             The bit to set (from 0 to 63)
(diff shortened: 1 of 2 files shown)


gyrovorbis (dismissed):
Everything looks good to me.

1) container_of() is super useful, especially with how often we use intrusively linked list containers and related data structures. This is one that all the big, serious C codebases include.

2) `BITLL()` is useful af for obvious reasons.

3) Great to see the `libudiv` stuff here. Personally used it, and it's gainzy AF. Very useful.

pcercuei:
Sorry, had to force-push as I had a stray bogus doxygen line in the udiv header.

gyrovorbis (dismissed):

pcercuei:
Sorry... force-pushed again. It turns out that Newlib already provides `__containerof()`, so I dropped the first commit.

QuzarDC (approved):

PR #1443, opened 2026-06-20
kos-chain: Add backup GNU mirror for source downloads

andressbarajas:
Installing the toolchain sometimes fails to download binutils/GCC because ftpmirror.gnu.org redirects us to a bad mirror, and we can't control which one. This also causes random CI failures that have to be manually re-run.

This PR adds mirrors.kernel.org as a backup: if the primary GNU mirror fails, the downloader retries the same file from the backup. Works for both curl and wget.

### DIFF
--- a/utils/kos-chain/scripts/download.mk
+++ b/utils/kos-chain/scripts/download.mk
@@ -2,6 +2,7 @@
 
 # Set default mirror if not specified
 gnu_mirror ?= ftpmirror.gnu.org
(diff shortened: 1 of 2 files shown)


QuzarDC (approved):
Since `gnu_mirror` was already redefinable it was simple to test by overriding it to a fake mirror that would fail to demonstrate that the backup would kick in. It would be a nice update in the future to have it as a list rather than two independent ones. That way we could add as many as we'd want.

PR #1444, opened 2026-06-20
kos: Add NULL checks for unchecked allocations

andressbarajas:
cppcheck and clang found a few allocations that were used without checking for NULL, plus a zero-length underflow in pthread_getname_np. This adds guards for them.


kernel/arch/dreamcast/sound/snd_stream.c
```diff
@@ -173,6 +173,9 @@ void snd_stream_filter_add(snd_stream_hnd_t hnd, snd_stream_filter_t filtfunc, v
     CHECK_HND(hnd);
 
     f = malloc(sizeof(filter_t));
+    if(f == NULL)
```
QuzarDC: Perhaps this one should be updated to add a return value, or turned into an assert? And if that doesn't seem worthwhile that's fine too. Rest looks good.

QuzarDC (approved):

PR #1446, opened 2026-06-20
genwait bugfix and cleanup

QuzarDC:
Fixes a bug where `genwait_wait(NULL, anything, timeout >0)` would cause a hang. Details in the first commit. From there performed a few other small cleanups.

It seemed pretty straightforward, but if anyone thinks that `obj == NULL` should be *reserved* for the unset state, then we could instead fail to wait if asked to wait on `NULL`, assert if passed to the wake functions, and set `__nonnull` on those params. I couldn't think of any real benefit to doing that though

### DIFF
--- a/kernel/thread/genwait.c
+++ b/kernel/thread/genwait.c
@@ -106,7 +106,17 @@ int genwait_wait(void *obj, const char *mesg, unsigned int timeout) {
 }
 
 /* Removes a thread from its wait queue; assumes ints are disabled. */
-static void __nonnull_all genwait_unqueue(kthread_t *thd) {
+static void __nonnull_all genwait_unqueue(kthread_t *thd, int err) {
+
+    /* Set the wake return value */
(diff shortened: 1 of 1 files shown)


pcercuei (approved):
On one hand I think that genwait() was designed to be used with any *valid* pointer, and NULL cannot be considered valid in 99.9% of cases.

On the other hand, this check on `thd->wake_obj` seems to be useless, as it will never contain anything else than an address we're waiting on.

NGL, `genwait(NULL)` makes me a bit uneasy, but there's no valid reason not to support it.

QuzarDC:
> NGL, `genwait(NULL)` makes me a bit uneasy, but there's no valid reason not to support it.

Same. Hence my preamble of 'maybe block it?' in case someone else had an idea. Maybe we'll come up with a good reason in the future, but at least the current bad behavior will be gone :)

PR #1449, opened 2026-06-21
speedtest: Update example to push the limits of network stack

andressbarajas:
KOS stack has improved a bunch since I first created this example.  We can now send and receive more data at a time. I increased the payloads for download and upload to 2MB. I also increased the input buffer so less recv calls are needed to receive the full payload. I also display the url the user should visit in order to run the tests.

@Bruceleeto: I know you've edited this example before to get better output. Did I miss anything?

### DIFF
--- a/examples/dreamcast/network/speedtest/speedtest.c
+++ b/examples/dreamcast/network/speedtest/speedtest.c
@@ -53,6 +53,9 @@ int main(int argc, char **argv) {
             return 0;
 
         MAPLE_FOREACH_END()
+
+        /* Poll the controller once per frame. */
+        thd_sleep(17);
     }
 
     return 0;
--- a/examples/dreamcast/network/speedtest/speedtest.c
+++ b/examples/dreamcast/network/speedtest/speedtest.c
@@ -17,6 +17,7 @@
 
 #include <kos/init.h>
 #include <kos/dbgio.h>
+#include <kos/net.h>
 
 #include <dc/video.h>
 #include <dc/biosfont.h>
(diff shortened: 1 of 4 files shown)


Bruceleeto:
yea looks fine,  should of done this myself but I forgot.

andressbarajas:
I was comparing data rate speed between this example and ftp example and made additional changes:

1. Cleaned up redudant code
2. Poll controller once per frame (Busy waiting literally made it 2+ MB/s SLOWER than ftp example)
3. Remove printfs for now because it caused crashes. Probably because dcload syscalls disable IRQs which net stack uses while moving data and causes corruption somewhere. Need to look deeper into that issue but its outside the scope of this example.

examples/dreamcast/network/speedtest/speedtest.c
```diff
@@ -28,9 +31,22 @@
...
+    /* Show the URL the user should open in their browser. */
+    if(net_default_dev) {
+        uint8_t *ip = net_default_dev->ip_addr;
+        snprintf(url, sizeof(url), "Open http://%d.%d.%d.%d/ in a web browser",
+                 ip[0], ip[1], ip[2], ip[3]);
+    }
+    else {
+        snprintf(url, sizeof(url), "No network device found");
+    }
+    bfont_draw_str(vram_s + 44 * 640 + 20, 640, 0, url);
```
QuzarDC: How's about just `bfont_draw_str_fmt` for the `net_defaut_dev` case?

QuzarDC (commented):
Just one possible update, otherwise looks good. Does it rely on #1450 at all?

andressbarajas:
> Just one possible update, otherwise looks good. Does it rely on #1450 at all?

It relies on it to get good upload speeds. To me it doesn't matter which order. Im about to knock out though. Work early tomorrow morning so I probably wont get to follow up on my PRs till tomorrow early afternoon MAYBE but tomorrow for sure.

QuzarDC (approved):

PR #1451, opened 2026-06-22
Update fs_dclsocket & fix fs_rewinddir

andressbarajas:
This PR updates fs_dclsocket to support dc-load-ip v2 packet sizes/port, adds rewinddir syscall to fs_dclsocket, and fixes a VFS regression I tripped over on the way. Also saved over 14 kb by using bitfields instead of full bytes to represent chunks. All tested on real hardware over dc-load-ip without -l flag.

This addresses issue: https://github.com/KallistiOS/KallistiOS/issues/1092

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_dclsocket.c
+++ b/kernel/arch/dreamcast/fs/fs_dclsocket.c
@@ -419,7 +419,13 @@ static size_t dcls_total(void *hnd) {
 }
 
 static dirent_t their_dir;
-static dirent_t our_dir;
+
+/* The host fills the readdir reply as a POSIX struct dirent, not a KOS
+   dirent_t, so receive it as one */
+static union {
+    struct dirent de;
+    char storage[sizeof(struct dirent) + NAME_MAX + 1];
+} our_dir;
 
 static const dirent_t *dcls_readdir(void *hnd) {
     uint32_t fd = (uint32_t) hnd;
@@ -435,25 +441,33 @@ static const dirent_t *dcls_readdir(void *hnd) {
 
     memcpy(cmd->id, "DC18", 4);
     cmd->value0 = htonl(fd);
-    cmd->value1 = htonl((uint32_t)(&our_dir));
-    cmd->value2 = htonl(sizeof(dirent_t));
+    cmd->value1 = htonl((uint32_t)(&our_dir.de));
+    cmd->value2 = htonl(sizeof(our_dir));
 
     send(dcls_socket, cmd, sizeof(command_3int_t), 0);
 
     dcls_recv_loop();
 
     if(retval) {
-        char fn[strlen(dcload_path) + strlen(our_dir.name) + 1];
         command_t *cmd2 = (command_t *)pktbuf;
         dcload_stat_t filestat;
 
-        strcpy(their_dir.name, our_dir.name);
+        /* Verify dcload won't overflow us */
+        if(strlen(our_dir.de.d_name) + 1 > NAME_MAX) {
+            mutex_unlock(&mutex);
+            errno = EOVERFLOW;
+            return NULL;
+        }
+
+        char fn[strlen(dcload_path) + strlen(our_dir.de.d_name) + 1];
+
+        strcpy(their_dir.name, our_dir.de.d_name);
         their_dir.size = 0;
         their_dir.time = 0;
         their_dir.attr = 0;
 
         strcpy(fn, dcload_path);
-        strcat(fn, our_dir.name);
+        strcat(fn, our_dir.de.d_name);
 
         memcpy(cmd2->id, "DC13", 4);
         cmd2->address = htonl((uint32_t) &filestat);
@@ -467,6 +481,7 @@ static const dirent_t *dcls_readdir(void *hnd) {
         if(!retval) {
             if(filestat.st_mode & S_IFDIR) {
                 their_dir.size = -1;
+                their_dir.attr = O_DIR;
             }
             else {
                 their_dir.size = filestat.st_size;
--- a/kernel/arch/dreamcast/fs/fs_dclsocket.c
+++ b/kernel/arch/dreamcast/fs/fs_dclsocket.c
@@ -40,8 +40,12 @@
 #include <dc/fs_dcload.h>
 #include <dc/dcload.h>
 
-#define DCLOAD_PORT 31313
+#define DCLOAD_PORT 53535
+#define DCLOAD_PACKET_SIZE 1440
 #define NAME "dcload-ip over KOS sockets"
+
+/* Number of DCLOAD_PACKET_SIZE chunks needed to cover n bytes. */
+#define DCLOAD_PACKETS(n)  (((n) + DCLOAD_PACKET_SIZE - 1) / DCLOAD_PACKET_SIZE)
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


kernel/arch/dreamcast/fs/fs_dclsocket.c
```diff
@@ -40,9 +40,13 @@
 #include <dc/fs_dcload.h>
 #include <dc/dcload.h>
 
-#define DCLOAD_PORT 31313
+#define DCLOAD_PORT 53535
+#define DCLOAD_PACKET_SIZE 1440
 #define NAME "dcload-ip over KOS sockets"
 
+/* Number of DCLOAD_PACKET_SIZE chunks needed to cover n bytes. */
+#define DCLOAD_PACKETS(n)  (((n) + DCLOAD_PACKET_SIZE - 1) / DCLOAD_PACKET_SIZE)
```
pcercuei, line 49: `__align_up(n, DCLOAD_PACKET_SIZE)`

kernel/arch/dreamcast/fs/fs_dclsocket.c
```diff
@@ -47,6 +47,10 @@
 /* Number of DCLOAD_PACKET_SIZE chunks needed to cover n bytes. */
 #define DCLOAD_PACKETS(n)  (((n) + DCLOAD_PACKET_SIZE - 1) / DCLOAD_PACKET_SIZE)
 
+/* PARTBIN receive map: one bit per chunk, sized to cover 16 MB of RAM. */
+#define DCLOAD_MAP_ENTRIES DCLOAD_PACKETS(16 * 1024 * 1024)
+#define DCLOAD_MAP_BYTES   (((DCLOAD_MAP_ENTRIES) + 7) / 8)
```
pcercuei, line 53: `__align_up(DCLOAD_MAP_ENTRIES, 8)`

kernel/arch/dreamcast/fs/fs_dclsocket.c
```diff
@@ -82,10 +86,19 @@ static uint8_t pktbuf[DCLOAD_PACKET_SIZE + sizeof(command_t)];
 
 static int dcls_socket = -1;
 
+/* Mark / test receipt of chunk `index` in the PARTBIN map. */
+static inline void bin_map_set(int index) {
+    bin_info.map[index >> 3] |= (1 << (index & 7));
```
pcercuei: `bin_info.map[index >> 3] |= BIT(index & 7);`
"index" should probably also be unsigned.

kernel/arch/dreamcast/fs/fs_dclsocket.c
```diff
@@ -575,6 +575,30 @@ static int dcls_stat(vfs_handler_t *vfs, const char *fn, struct stat *rv,
     return -1;
 }
 
+static int dcls_rewinddir(void *hnd) {
+    uint32_t fd = (uint32_t) hnd;
+    command_int_t *cmd = (command_int_t *)pktbuf;
+
+    if(fd < 100) {
```
pcercuei: I see that magic value in `dcls_close()` and `dcls_readdir()`, any idea what it corresponds to?
A macro for it would be better IMHO.

kernel/fs/fs.c
```diff
@@ -890,7 +890,7 @@ int fs_rewinddir(file_t fd) {
         return -1;
     }
 
-    h->idx = 0;
+    h->idx = -2;
```
pcercuei, line 893: Note that the high-level `fs_readdir()` function will explicitly handle the `.` and `..` paths, so if it makes things easier for you, feel free to not handle those.

kernel/arch/dreamcast/fs/fs_dclsocket.c
pcercuei: `return bin_info.map[index >> 3] & BIT(index & 7);`

pcercuei (changes requested):

kernel/arch/dreamcast/fs/fs_dclsocket.c
andressbarajas: DCLOAD_PACKET_SIZE is not a power of two so __align_up doesnt work here.

kernel/arch/dreamcast/fs/fs_dclsocket.c
andressbarajas: With __align_up() it would be: 

`#define DCLOAD_MAP_BYTES   (__align_up(DCLOAD_MAP_ENTRIES, 8) / 8)`

and thats more verbose

kernel/arch/dreamcast/fs/fs_dclsocket.c
andressbarajas: It's a leftover from how old dc-tool handed back directory handles. Up until 2013 opendir() on the host just returned the raw DIR * pointer cast to an int, and closedir/readdir cast it straight back and dereferenced it. File handles were just the small open() fd. So a directory handle was always a big heap address and a file handle was always small. 100 is just a cutoff to tell the two apart.

Newer dc-tool stopped returning the pointer (it broke on 64-bit) and now returns a small table index offset by 1337 instead, but it's still always large, so the same check works.

Im going to put this PR back in draft and take the same approach fs_dcload.c takes.

pcercuei: I mean just having a `#define FS_DCLSOCKET_FD_DIR_THRESH 100` would be enough to please me here :)

andressbarajas: Awesome. I'll do a follow up PR in the next couple of days.

andressbarajas:
Force pushed the BIT field change requests and did a commit to replace magic value with macro.

kernel/fs/fs.c
andressbarajas: It's actually relying on that. Since fs_open() opens a fresh handle at -2 (where fs_readdir() emits ./..), fs_rewinddir() resetting to 0 meant those two entries were skipped after a rewind. Matching -2 just keeps a rewound dir consistent with a freshly-opened one.

pcercuei: Right, sorry, I missed the fact that you were actually modifying the high-level FS code and not fs_dclsocket here.

pcercuei (approved):

QuzarDC:
My apologies for the delay. The holdup here is that it cuts off support for our hosted dcload-ip to use the functionality. That's fine since it's barely maintained and this isn't core to it, but I'm just delaying a bit more to see about the feasibility of fast-forwarding it some (and if not finally marking it as 'don't use'). It should be fairly reasonable as it's all stuff that lives on the PC side.

andressbarajas:
> My apologies for the delay. The holdup here is that it cuts off support for our hosted dcload-ip to use the functionality. That's fine since it's barely maintained and this isn't core to it, but I'm just delaying a bit more to see about the feasibility of fast-forwarding it some (and if not finally marking it as 'don't use'). It should be fairly reasonable as it's all stuff that lives on the PC side.

@QuzarDC KOS's dcload-ip was left in a broken state I think. Doesn't compile from what I remember. Mainly my fault. Let me know if you need help with that. I think loader side changes are also required unfortunately to support 1440 sized packets.

PR #1453, opened 2026-06-23
Collection of misc cleanups

QuzarDC:
- sq_unlock had some dead code after the last changes to it.
- sem_wait_timed had a variable being set and never used.
- hardware had a switch with intentional fallthroughs without the appropriate macro to denote them.
- stack was using `asm` rather than `__asm__`.
- video had the `cable_type` value passed around as int8, int16, and uint16.

Aside from the last one's change to the size of the vid_mode struct, I don't think there should be any observable functional differences as the first two gcc should have been able to discard anyways, the second two are compatibility/warning related, and the last didn't have any case where it could have caused a problem from the type changes.

### DIFF
--- a/kernel/arch/dreamcast/hardware/hardware.c
+++ b/kernel/arch/dreamcast/hardware/hardware.c
@@ -101,11 +101,11 @@ void hardware_shutdown(void) {
             g2_dma_shutdown();
             spu_shutdown();
             vid_shutdown();
-            /* fallthru */
+            __fallthrough;
         case 1:
             vblank_shutdown();
             asic_shutdown();
-            /* fallthru */
+            __fallthrough;
         case 0:
             return;
     }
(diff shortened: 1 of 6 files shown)


kernel/libc/koslib/select.c
```diff
@@ -68,7 +68,6 @@ int select(int nfds, fd_set *readfds, fd_set *writefds, fd_set *errorfds,
         return rv;
     }
 
-    rv = 0;
```
andressbarajas, line 71: select() is actually broken. I found out 2 days ago when I came across this statement. Gonna create a PR for it right now so you can undo this change because the fix actually requires rv = 0;

andressbarajas: https://github.com/KallistiOS/KallistiOS/pull/1454.  Everything else in your branch looks good!

pcercuei (approved):

PR #1454, opened 2026-06-23
select: Return the correct ready-fd count

andressbarajas:
j holds the number of descriptors handed to poll(), which is why it is used as the loop bound when scanning the results. The bug is that the same loop increments j for each ready condition it finds -- so it both walks past the entries poll() filled and returns that inflated value instead of the real count.


QuzarDC (approved):

QuzarDC:
Good catch. I guess nobody's been using select for this to have stayed like that since it was added. I had figured the setting rv to 0 was just habitual with the rest, rather than to reuse it for the actual return.

PR #1456, opened 2026-06-24
fat: Fix fs_fat_mount() crash when mounting an exFAT volume

andressbarajas:
exFAT carries the same 0x55AA boot signature as FAT16/32, so it slipped past fat_read_boot()'s signature check and into fat_parse_boot(). There its zeroed BPB fields (bytes_per_sector, sectors_per_cluster) are used as divisors, so instead of failing cleanly the parse hit a divide-by-zero and crashed.

Fix is to require a non-zero FAT count before accepting the volume: a real FAT filesystem always has at least one FAT.

Verified fix on hardware with SDXC exFAT formatted card. It fails cleanly now instead of crashing.

### DIFF
--- a/addons/libkosfat/bpb.c
+++ b/addons/libkosfat/bpb.c
@@ -191,7 +191,8 @@ int fat_read_boot(fat_superblock_t *sb, kos_blockdev_t *bd) {
     }
 
     /* Make sure it looks sane... */
-    if(bb.ebpb.fat16.boot_sig[0] != 0x55 || bb.ebpb.fat16.boot_sig[1] != 0xAA) {
+    if(bb.ebpb.fat16.boot_sig[0] != 0x55 || bb.ebpb.fat16.boot_sig[1] != 0xAA ||
+       bb.bpb.num_fats == 0) {
         return -EINVAL;
     }
 


pcercuei (approved):

QuzarDC:
Think there'd be any value in signaling that it's not merely invalid but it's a specific known sort of invalid? Maybe debuglog output saying `exFAT not supported`? Or a differentiated errno?

andressbarajas:
> Think there'd be any value in signaling that it's not merely invalid but it's a specific known sort of invalid? Maybe debuglog output saying `exFAT not supported`? Or a differentiated errno?

I can do that. More info:

Before even mounting the SD card using fs_fat_mount() the user has to first sd_blockdev_for_partition(). The partition number output from that identifies the type of filesystem(https://en.wikipedia.org/wiki/Partition_type). 0x7 in this case. NTFS formatted card would also produce the same crash with the same partition number. This PR handles both exFAT and NTFS.

How do you want me to continue?

QuzarDC:
> > Think there'd be any value in signaling that it's not merely invalid but it's a specific known sort of invalid? Maybe debuglog output saying `exFAT not supported`? Or a differentiated errno?
> 
> I can do that. More info:
> 
> Before even mounting the SD card using fs_fat_mount() the user has to first sd_blockdev_for_partition(). The partition number output from that identifies the type of filesystem(https://en.wikipedia.org/wiki/Partition_type). 0x7 in this case. NTFS formatted card would also produce the same crash with the same partition number. This PR handles both exFAT and NTFS.
> 
> How do you want me to continue?

Ah, ok. I thought it was specific to exFAT and that seemed like a more common kind of confusion that and end-user might have and a front-end dev would want to relay, compared to a more general 'formatted, but to a different fs'. If someone wants to separate them out in the future it can easily be done so no need for further changes.

PR #1458, opened 2026-06-24
CMake cleanups + update to support Makefile-based kos-ports

pcercuei:
This PR brings some fixes to the CMake toolchain file. It also updates the paths in KallistiOS to support the new Makefile-based kos-ports, that is in a PR in kos-ports: https://github.com/KallistiOS/kos-ports/pull/165


andressbarajas:
I pulled this PR and kos-ports 2.0.0, built the tsunami port, and sourced the new environ.sh. Compiling a tsunami example works, but linking fails:

'ld: cannot find -ltsunami: No such file or directory'

Looks like the headers are found but the linker can't find the port libraries.

I use make and not cmake.

pcercuei:
@andressbarajas Thanks. I had updated the include path in the environ file, but not the linker path. It works now.

pcercuei:
Added another CMake fix. GLdc would use CMake's `INTERPROCEDURAL_OPTIMIZATION` to enable LTO, which caused CMake to pass `-fno-fat-lto-objects` during building and linking GLdc, and in turn causing apps compiled with `-fno-lto` to not be able to link with GLdc.

QuzarDC:
Could you test these as well for `v2.2x`? There would have to be a bit of legwork to try to help make sure that it's not a mysterious hard break for someone to build `kos-ports` on our stable version. I think they should be able to cleanly cherry-pick in.

pcercuei:
Everything rebases cleanly as long as 927dcc9d8 is merged too.

pcercuei:
Everything builds fine too, after merging https://github.com/KallistiOS/kos-ports/pull/163 as well.

pcercuei:
Force-pushed to rebase on master, + a tiny fix for the "distclean" rule to remove the sysroot recursively (`rm -f` -> `rm -rf`).

pcercuei:
Rebased to master.

Issue #1176, opened 2025-08-31
Provide resource-sharing functionality in and to dcload


QuzarDC:
Currently dcload can cause issues in resource sharing on the g2 bus due to not having any mechanism to share between it and kos:

```
I just looked at the BBA driver in dcload and realized that they are really cruel to the G2 bus.
 There are no locks, FIFOs or DMA pauses at all. There is only hope that when the dcload system
call is called, interrupts are turned off (which, by the way, leads to lags) and here there is a wait for
FIFO, however, once, and not every 8x32-bit words.

Now it is clear why audio streaming does not work with dcload. We need to add to
g2_lock(), g2_read() and g2_write() inside dcload, like in KOS, and not turn off interrupts when 
calling dcload syscall, just sync g2_lock() with KOS by some way.
```

_Originally posted by @DC-SWAT in https://github.com/KallistiOS/KallistiOS/pull/1110#discussion_r2249511225_

I think the generic way to do this would be to provide a command that KOS can call over to dcload similar to the assignment of worker memory, to provide the location of a KOS `mutex_t` for the hardware resource that dcload will be using. For dcload-ip it would be the g2 mutex.

Issue #788, opened 2024-10-04
Watchdog Timer timeout rate is unexpectedly low


gh-sh4:
```cpp
#include <cstdlib>
#include <stdio.h>
#include <atomic>

#include <arch/types.h>
#include <arch/timer.h>
#include <arch/wdt.h>

#include <dc/maple.h>
#include <dc/maple/controller.h>

#define MSEC 1000
#define WDT_INTERVAL (500 * MSEC)

static void
wdt_timeout(void *data)
{
  std::atomic<uint32_t> *wdt_count = (std::atomic<uint32_t> *)data;
  wdt_count->fetch_add(1);
}

bool
is_start_pressed()
{
  maple_device_t *cont = maple_enum_type(0, MAPLE_FUNC_CONTROLLER);
  if (cont != NULL) {
    cont_state_t *state = (cont_state_t *)maple_dev_status(cont);

    if (state != NULL) {
      if (state->buttons & CONT_START)
        return true;
    }
  }
  return false;
}

int
main(int argc, char **argv)
{
  int finished   = 0;
  uint64 timer   = timer_ms_gettime64(), start, end;
  uint32 counter = 0;

  /* Enable the WDT in interval counter mode, with a period of 500ms,
     an interrupt priority level of 15 (highest), and give it our
     timeout callback plus pass it our counter as its userdata pointer. */
  std::atomic<uint32_t> wdt_count = 0;
  wdt_enable_timer(0, 1000, 15, wdt_timeout, &wdt_count);

  uint64_t last_timer = 0;

  for (uint32_t iteration = 0;; iteration++) {
    if (iteration % 1024 == 0) {
      const uint64_t now   = timer_us_gettime64();
      const uint32_t delta = (uint32_t)(now - last_timer);
      printf("delta micros %u cummulative wdt_count %d\n", delta, wdt_count.load());
      last_timer = now;
    }
    if (is_start_pressed()) {
      finished = 1;
    }
  }

  return 0;
}
```

Eventually gives
```
delta micros 5294 cummulative wdt_count 2375
delta micros 6090 cummulative wdt_count 2379
delta micros 6328 cummulative wdt_count 2382
delta micros 5315 cummulative wdt_count 2386
delta micros 5242 cummulative wdt_count 2390
```
This setting gives 3 or 4 updates to the atomic counter per 5-6 milliseconds. By varying the microseconds I could get various discrepancies. Setting this to 500 ms I never got an update..

[...]

PR #1459, opened 2026-06-25
dcload: Don't block IRQs across the dcload syscall

andressbarajas:
A dcload syscall is a slow (millisecond) round-trip to the host, and the old code disabled IRQs for its whole duration to guard against re-entrancy. Holding IRQs off that long starves other modules that rely on interrupts to make progress. I hit this running the speedtest example over dc-load-serial: an upload would hang the machine because the example's printf logging kept IRQs disabled long enough for the BBA's 16 KB RX ring to overflow before its interrupt could drain it.

Use a mutex instead, it gives the same re-entrancy protection while leaving IRQs enabled, and the syscall now refuses to run from interrupt context, where blocking on a mutex isn't allowed anyway.

This takes care of a portion of https://github.com/KallistiOS/KallistiOS/issues/1176 and I think https://github.com/KallistiOS/KallistiOS/issues/788 since WDT depends on interrupts to update and the user is printfing while checking the time.

### DIFF
--- a/kernel/arch/dreamcast/hardware/dcload.c
+++ b/kernel/arch/dreamcast/hardware/dcload.c
@@ -30,7 +30,7 @@ static int dcload_syscall(dcload_cmd_t cmd, void *param1, void *param2, void *pa
 
     /* Ensure that the FIFO buffer is clear */
     /* XXX - Is this needed? It seems like something only for serial. */
-    while(FIFO_STATUS & FIFO_SH4)            ;
+    while(FIFO_STATUS & FIFO_SH4);
 
     /* Make the call */
     return syscall(cmd, param1, param2, param3);
--- a/kernel/arch/dreamcast/hardware/dcload.c
+++ b/kernel/arch/dreamcast/hardware/dcload.c
@@ -12,9 +12,12 @@
 #include <dc/dcload.h>
 #include <dc/memory.h>
 #include <kos/irq.h>
+#include <kos/mutex.h>
 
 /* This is the address where the function pointer for the dcload syscall is fetched from */
 #define VEC_DCLOAD        (MEM_AREA_P1_BASE | 0x0C004008)
+
+static mutex_t mutex = RECURSIVE_MUTEX_INITIALIZER;
 
 /*
     This is the single syscall dcload provides. It is then multiplexed out based on the `cmd`
@@ -24,16 +27,22 @@
 static int dcload_syscall(dcload_cmd_t cmd, void *param1, void *param2, void *param3) {
     uintptr_t *syscall_ptr = (uintptr_t *)VEC_DCLOAD;
     int (*syscall)() = (int (*)())(*syscall_ptr);
+    int ret;
 
-    /* Disable IRQs until the syscall returns */
-    irq_disable_scoped();
+    /* Serialize callers */
+    if(mutex_lock_irqsafe(&mutex))
+        return -1;
 
     /* Ensure that the FIFO buffer is clear */
     /* XXX - Is this needed? It seems like something only for serial. */
     while(FIFO_STATUS & FIFO_SH4);
 
     /* Make the call */
-    return syscall(cmd, param1, param2, param3);
+    ret = syscall(cmd, param1, param2, param3);
+
+    mutex_unlock(&mutex);
+
+    return ret;
 }
 
 ssize_t dcload_read(uint32_t hnd, uint8_t *data, size_t len) {


pcercuei (dismissed):
Great cleanup.

DC-SWAT:
This was for older versions of dcload-ip that used the g2 bus without disabling IRQ.

kernel/arch/dreamcast/hardware/dcload.c
```diff
@@ -25,8 +28,11 @@ static int dcload_syscall(dcload_cmd_t cmd, void *param1, void *param2, void *pa
     uintptr_t *syscall_ptr = (uintptr_t *)VEC_DCLOAD;
     int (*syscall)() = (int (*)())(*syscall_ptr);
 
-    /* Disable IRQs until the syscall returns */
-    irq_disable_scoped();
+    /* A dcload call is slow. Never run one from an interrupt handler. */
+    if(irq_inside_int())
```
QuzarDC: Wouldn't this prevent all output when on dcload dbgio during interrupts? Certainly we shouldn't have just normal printing or file transfers going on there, but I'm thinking of the need to be able to send `assert` messages.

Not only that but I thought gdb communication was all inside interrupts as well, wouldn't this entirely break that?

kernel/arch/dreamcast/hardware/dcload.c
```diff
@@ -12,10 +12,13 @@
 #include <dc/dcload.h>
 #include <dc/memory.h>
 #include <kos/irq.h>
+#include <kos/mutex.h>
 
 /* This is the address where the function pointer for the dcload syscall is fetched from */
 #define VEC_DCLOAD        (MEM_AREA_P1_BASE | 0x0C004008)
 
+static mutex_t mutex = MUTEX_INITIALIZER;
```
QuzarDC: I believe this would need to be recursive in order to account for the possibility of needing debug output during the calls to mutex_lock/unlock.

QuzarDC (changes requested):
This seems like it would break a lot, but maybe I'm missing something. comments inline.

andressbarajas:
> This seems like it would break a lot, but maybe I'm missing something. comments inline.

You're right. Totally forgot about GDB. I'll dig into the issues you listed.

andressbarajas:
@QuzarDC Okay, to support GDB and asserts we just do business as usually and do the syscall. Otherwise in a thread context we stick with the mutex approach which is now recursive(thanks for that). 

I ran hardware tests against speedtest uploads, and tested asserts and GDB and they worked fine with these changes.

kernel/arch/dreamcast/hardware/dcload.c
andressbarajas: Yes it would. Now instead just make the syscall since we are already in an IRQ

andressbarajas:
I ran further tests and there is a rare case that a thread is mid-syscall and can get interrupted. In that case I early exit with:

```
if(mutex_is_locked(&mutex))
            return -1;
```

DC-SWAT:
What does this mutex provide? How does it solve the G2 bus access problem?
Have you tried working with sound while the data is being read?

andressbarajas:
> What does this mutex provide? How does it solve the G2 bus access problem? Have you tried working with sound while the data is being read?

The mutex only provides re-entrancy protection and does not solve the G2 bus access problem. When I referenced issue #1176 I only ment this portion: "not turn off interrupts when calling dcload syscall". I had a similar bus access problem for PS2 and the only way to fix it is to make changes on the dc-load-ip and KOS side. This will require a dependency on the new dc-load-ip which Im putting off until I can squeeze out all the fixes that are backwards compatible. I will get to it though! I promise.

EDIT: Im gonna try reading sound file from /pc today with this change and see how it sounds. Maybe it will just require dc-load-ip loader update like you said: 

"Now it is clear why audio streaming does not work with dcload. **We need to add to g2_lock(), g2_read() and g2_write() inside dcload, like in KOS**, and not turn off interrupts when calling dcload syscall, just sync g2_lock() with KOS by some way."

andressbarajas:
@DC-SWAT With this PR and this PR: https://github.com/KallistiOS/KallistiOS/pull/1451

I was able to run the multi-stream example using dc-load-ip and it worked!

andybarajas@Andys-Mac-mini multi-stream % /opt/toolchains/dc-utils/kos-tool -m . -t dhcp -x multistream.elf

Code change in main.c:

```
 wav_stream_hnd_t faucet = wav_create("/pc/romdisk/faucet.wav", LOOP);
 wav_volume(faucet, faucet_volume);

 wav_stream_hnd_t brushing = wav_create("/pc/romdisk/brushing.wav", LOOP);
 wav_volume(brushing, brushing_volume);
```

EDIT: @DC-SWAT Seems like all I needed was this PR https://github.com/KallistiOS/KallistiOS/pull/1451 to get the above example working. Do you have other things I can test?

DC-SWAT:
> EDIT: @DC-SWAT Seems like all I needed was this PR #1451 to get the above example working. Do you have other things I can test?

Other than the sound, there are no complaints. But I'm not sure a simple example can cover everything. It's better to try running the game.

kernel/arch/dreamcast/hardware/dcload.c
```diff
@@ -25,12 +28,24 @@ static int dcload_syscall(dcload_cmd_t cmd, void *param1, void *param2, void *pa
     uintptr_t *syscall_ptr = (uintptr_t *)VEC_DCLOAD;
     int (*syscall)() = (int (*)())(*syscall_ptr);
 
-    /* Disable IRQs until the syscall returns */
-    irq_disable_scoped();
+    /* Interrupt context can't block on a mutex, so call directly. */
+    if(irq_inside_int()) {
+        /* A thread is mid-transaction; drop rather than corrupt it. */
+        if(mutex_is_locked(&mutex))
```
QuzarDC: What exactly was the result from the corruption? Just like my commentary before about asserts, it would seem here that the most likely scenario for an interrupt needing to call dcload would be something critical where we're going to end up aborting. If this would lead just to the original syscall failing, then it may make sense to just allow it anyways since otherwise you'd end up with silent/mysterious aborts/exits. I know GDB does a lot more in an interrupt context that isn't that, so if that was the concern perhaps that command could just be filtered for or something?

Ideally the syscall would be bolstered to be able to support such a case. It wouldn't be unrealistic to allow an overlapping command to cleanly override a previous in-progress one when it would explicitly be *one* rather than arbitrary reentrancy.

pcercuei: Also, just on the code itself: You can do
```c
if(mutex_lock_irqsafe(&mutex))
    return -1;
```

Then you can factorize the code below (while loop + syscall()) and unlock the mutex afterwards.

andressbarajas: @QuzarDC Good questions. The corruption is a hang because it desyncs the dcload transport, so it's really drop-vs-hang, not drop-vs-failed-call. Right now the code only drops in the narrow window where it'd collide.

I totally agree that bolstering the syscall to support an override is the way to go but that needs changes on the kos-tool side to support which I dont mind adding later.

@pcercuei Thanks. Force pushed.

Issue #625, opened 2024-06-11
httpd example issues and possible bugs.


QuzarDC:
Testing the httpd example with dcload serial I found a number of possible problems:
1) All folders appear to show -1 as size. Is this normal? I can't recall if that's a change from previous behavior. We may want to consider changing the example to use standard file interaction methods rather than our internal `fs_xxx` ones for this and other examples. The method it's using, accessing the kos dirent_t should be discouraged in favor of more posix-style methods (though that in itself isn't an issue in need of 'fixing' necessarily).
2) When attempting to navigate into /pc/ I got an immediate segmentation fault in dc-tool. Additionally the DC became unresponsive (did not allow shutdown via START, nor further browsing). This was with dc-tool-ser 1.0.6 built from the repo and a disc made from the image from here: https://dreamcast.wiki/File:Dcload-2023-06-22.zip . This may strictly be a dc-tool/load bug but KOS also shouldn't be able to crash due to bad behavior on dc-load's side.
3) Related to 1), it seems something is wrong either with the code behind the /dev devices, or the way that the example distinguishes between folders and files. /dev/null, /dev/random, and /dev/urandom appear as directories that can be browsed to rather than special files. This likely just reflects how the example predates /dev.
4) Currently the example leaks any client threads that were listening still at the time of shutdown.
5) I've found that it seems to almost always end with some threads still open, despite the individual requests seeming to be closed. This might be related to the behavior I had previously found in #213 .

See this sample of output where sock 5 seems to have gotten stuck:
```
  httpd: listening for connections on socket 4
  httpd: connect from 6401a8c0, port 20444, socket 5
  httpd: client thread started, sock 5
  httpd: client requested '/dev/'
  httpd: closed client connection 5
  httpd: connect from 6401a8c0, port 20700, socket 5
  httpd: client thread started, sock 5
  httpd: connect from 6401a8c0, port 22492, socket 6
  httpd: client thread started, sock 6
  httpd: client requested '/cd/'
  fs_iso9660: disc change detected
  httpd: closed client connection 6
```

andressbarajas:
For # 1,
Yes, thats how KOS distinguishes folders from files.
For # 3, https://github.com/KallistiOS/KallistiOS/blob/f7ca4fc012173af8d4f0ed0ea992a485463624e2/kernel/fs/fs_dev.c#L48. This line makes all the contents of the /dev directory a folder.
For # 5,
I think this may be fixed because of zcrc's Mutex fix.

PR #1460, opened 2026-06-26
fs_dev: Bugfix and cleanup

andressbarajas:
I was looking through the issues and came across this one(https://github.com/KallistiOS/KallistiOS/issues/625) that had an easy bug fix. Plus I mentioned in the issue what was causing the bug so here is the PR to fix it....almost two years later.  Also did some white space cleanup.


QuzarDC (approved):
Fantastic, thanks. I had repeatedly just skipped over that issue when skimming through them for more todos. I'll make sure to properly update it soon as it does seem like most of it has been fixed up already.

PR #1461, opened 2026-06-27
  pvr: add sized render-to-texture API and Example

GPF:
Add PVR render-to-texture examples covering both the existing fullscreen-sized
  RTT path and the new sized RTT API.

  The new pvr_txr_targettexture example provides a raw KOS benchmark for
  pvr_scene_begin_txr(), comparing two common approaches:

  - rendering into a large RTT backing and copying the logical region into a
    normal texture
  - sampling the RTT backing directly with clipped UVs

  The new pvr_txr_targettexture_ex example demonstrates pvr_scene_begin_txr_ex()
  with small RGB565 render targets. It includes selectable RTT modes, onscreen
  dcplib text showing the active RTT size and PVR RAM usage, and safe runtime
  texture resizing that waits for the PVR before freeing the previous target.

  Also tighten the sized RTT API contract by validating the 16-bit stride
  alignment requirement and documenting that stride_px must be a multiple of
  4 pixels. RTT clipping explicitly packs a zero start coordinate for clarity.


kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -887,6 +887,33 @@ void pvr_scene_begin(void);
...
+    Unlike pvr_scene_begin_txr(), this function does not require a screen-sized
+    backing texture. The PVR will render into a region of render_w by render_h
+    pixels, using stride_px pixels as the backing memory pitch.
+
+    render_w and render_h describe the area to draw. stride_px describes the
+    number of pixels between rows in memory and must be greater than or equal
+    to render_w. For the initial 16-bit render target implementation,
+    stride_px must also be a multiple of 4 pixels.
+
+    Initial support is intended for 16-bit render targets, matching the
```
gyrovorbis:
```suggestion
    \note Initial support is intended for 16-bit render targets, matching the
```

kernel/arch/dreamcast/hardware/pvr/pvr_scene.c
gyrovorbis, line 7: Add your copyright here. This is a nontrivial feature and API addition. :)

examples/dreamcast/pvr/pvr_txr_targettexture/Makefile
```diff
@@ -0,0 +1,27 @@
+#
+# pvr_txr_targettexture
+# Standalone render-to-texture test for pvr_scene_begin_txr().
+#
```
gyrovorbis: Add your copyright info.

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
+/*
+   pvr_txr_targettexture
+
```
gyrovorbis: Add your name + copyright year.

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+      B      Toggle copy/direct mode
+      X      Toggle visible draw enabled
+      Y      Toggle stress draw enabled
+      START  Exit
+*/
+
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
```
gyrovorbis:
```suggestion
#include <stdbool.h>
```
We're moving KOS away from using `int` to denote true/false and towards using actual boolean types to make code more modern and clearer.

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+#define SIZE_COUNT ((int)(sizeof(logical_sizes) / sizeof(logical_sizes[0])))
+
+static pvr_ptr_t rtt_backing;
+static pvr_ptr_t logical_texture;
+static int size_index = 0;
+static int logical_w = 64;
+static int logical_h = 64;
+static bench_mode_t bench_mode = BENCH_MODE_COPY;
+static int visible_draw_enabled = 1;
```
gyrovorbis:
```suggestion
static bool visible_draw_enabled = true;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+static int stress_draw_enabled = 0;
```
gyrovorbis:
```suggestion
static int stress_draw_enabled = false;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+static int exit_requested = 0;
```
gyrovorbis:
```suggestion
static int exit_requested = false;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+static int printed_rtt_dims = 0;
```
gyrovorbis:
```suggestion
static int printed_rtt_dims = false;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+    uint32_t ry = RTT_H;
+
+    pvr_poly_cxt_col(&cxt, PVR_LIST_OP_POLY);
+    pvr_poly_compile(&hdr, &cxt);
+
+    pvr_scene_begin_txr(rtt_backing, &rx, &ry);
+    if(!printed_rtt_dims) {
+        printf("pvr_scene_begin_txr returned rx=%lu ry=%lu\n",
+               (unsigned long)rx, (unsigned long)ry);
+        printed_rtt_dims = 1;
```
gyrovorbis:
```suggestion
        printed_rtt_dims = true;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+        stats_frames = 0;
+        stats_frame_ms = 0;
+        stats_rtt_ms = 0;
+        stats_copy_ms = 0;
+        stats_screen_ms = 0;
+        stats_copy_bytes = 0;
+    }
+}
+
+int main(int argc, char **argv) {
```
gyrovorbis:
```suggestion
int main(int argc, const char **argv) {
```
I think LTO might warn if this isn't `const`, cuz it's declared `const` internal to KOS's `init.c` which calls it.

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+    (void)argc;
+    (void)argv;
+
+    pvr_init(&pvr_params);
+
+    rtt_backing = pvr_mem_malloc(RTT_W * RTT_H * RGB565_BYTES);
+    if(!rtt_backing) {
+        printf("Failed to allocate %dx%d RGB565 RTT backing\n", RTT_W, RTT_H);
```
gyrovorbis:
```suggestion
        fprintf(stderr, "Failed to allocate %dx%d RGB565 RTT backing\n", RTT_W, RTT_H);
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+        return 1;
```
gyrovorbis:
```suggestion
        return EXIT_FAILURE;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+      X      Toggle visible draw enabled
+      Y      Toggle stress draw enabled
+      START  Exit
+*/
+
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+#include <arch/cache.h>
```
gyrovorbis:
```suggestion
#include <stdlib.h>
#include <arch/cache.h>
```
For `EXIT_SUCCESS`/`EXIT_FAILURE`.

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+    stats_screen_ms = 0;
+    stats_copy_bytes = 0;
+}
+
+static void poll_controls(void) {
+    uint32_t buttons = read_buttons();
+    uint32_t pressed = buttons & ~last_buttons;
+
+    if(pressed & CONT_START)
+        exit_requested = 1;
```
gyrovorbis:
```suggestion
        exit_requested = true;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+static bench_mode_t bench_mode = BENCH_MODE_COPY;
+static int visible_draw_enabled = 1;
+static int stress_draw_enabled = 0;
+static int exit_requested = 0;
+static int printed_rtt_dims = 0;
+static uint32_t last_buttons = 0;
+static float player_x = 10.0f;
+static float player_dx = 2.0f;
+
+static uint64_t stats_start_ms;
```
gyrovorbis: Would you mind bumping the resolution here from `ms` to `ns`? It's the same API and same types, so there's really no reason not to use the highest resolution we have.

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+    { 256, 256 },
+    { 320, 240 },
+    { 640, 480 }
+};
+
+#define SIZE_COUNT ((int)(sizeof(logical_sizes) / sizeof(logical_sizes[0])))
+
+static pvr_ptr_t rtt_backing;
+static pvr_ptr_t logical_texture;
+static int size_index = 0;
```
gyrovorbis:
```suggestion
static size_t size_index = 0;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+static int logical_w = 64;
```
gyrovorbis:
```suggestion
static size_t logical_w = 64;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+static int logical_h = 64;
```
gyrovorbis:
```suggestion
static size_t logical_h = 64;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+static uint64_t stats_screen_ms;
+static uint64_t stats_copy_bytes;
+
+static pvr_init_params_t pvr_params = {
+    { PVR_BINSIZE_16, PVR_BINSIZE_0, PVR_BINSIZE_0,
+      PVR_BINSIZE_0, PVR_BINSIZE_0 },
+    256 * 1024, 0, 0, 0, 0, 0
+};
+
+static int copy_bytes_per_frame(void) {
```
gyrovorbis:
```suggestion
static size_t copy_bytes_per_frame(void) {
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+    logical_h = logical_sizes[size_index].h;
+
+    if(player_x > (float)logical_w - 20.0f)
+        player_x = (float)logical_w - 20.0f;
+    if(player_x < 0.0f)
+        player_x = 0.0f;
+}
+
+static void reset_benchmark_stats(void) {
+    stats_start_ms = timer_ms_gettime64();
```
gyrovorbis:
```suggestion
    stats_start_ms = timer_ns_gettime64();
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+    rtt_backing = pvr_mem_malloc(RTT_W * RTT_H * RGB565_BYTES);
+    if(!rtt_backing) {
+        printf("Failed to allocate %dx%d RGB565 RTT backing\n", RTT_W, RTT_H);
+        return 1;
+    }
+
+    logical_texture = pvr_mem_malloc(LOGICAL_TXR_W * LOGICAL_TXR_H *
+                                     RGB565_BYTES);
+    if(!logical_texture) {
+        printf("Failed to allocate %dx%d RGB565 logical texture backing\n",
```
gyrovorbis:
```suggestion
        fprintf(stderr, "Failed to allocate %dx%d RGB565 logical texture backing\n",
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+               LOGICAL_TXR_W, LOGICAL_TXR_H);
+        pvr_mem_free(rtt_backing);
+        return 1;
```
gyrovorbis:
```suggestion
        return EXIT_FAILURE;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+    printf("logical sizes: 64x64, 128x128, 256x256, 320x240, 640x480\n");
+    printf("controls: A=size B=copy/direct X=visible Y=stress START=exit\n");
+
+    pvr_set_bg_color(0.0f, 0.0f, 0.0f);
+    apply_size();
+    last_buttons = read_buttons();
+    reset_benchmark_stats();
+
+    while(!exit_requested) {
+        uint64_t t0 = timer_ms_gettime64();
```
gyrovorbis:
```suggestion
        uint64_t t0 = timer_ns_gettime64();
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+        uint64_t t1;
+        uint64_t t2;
+        uint64_t t3;
+
+        render_target_scene();
+        t1 = timer_ms_gettime64();
```
gyrovorbis:
```suggestion
        t1 = timer_ns_gettime64();
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+
+        if(bench_mode == BENCH_MODE_COPY)
+            copy_logical_region_from_rtt();
+        t2 = timer_ms_gettime64();
```
gyrovorbis:
```suggestion
        t2 = timer_ns_gettime64();
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+
+        if(visible_draw_enabled) {
+            if(bench_mode == BENCH_MODE_COPY)
+                render_screen_scene_copy();
+            else
+                render_screen_scene_direct();
+        }
+        t3 = timer_ms_gettime64();
```
gyrovorbis:
```suggestion
        t3 = timer_ns_gettime64();
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+        update_motion();
+        update_benchmark_stats(t0, t1, t2, t3);
+        poll_controls();
+    }
+
+    pvr_mem_free(logical_texture);
+    pvr_mem_free(rtt_backing);
+    pvr_shutdown();
+
+    return 0;
```
gyrovorbis:
```suggestion
    return EXIT_SUCCESS;
```

examples/dreamcast/pvr/pvr_txr_targettexture/pvr_txr_targettexture.c
```diff
@@ -0,0 +1,457 @@
...
+    stats_rtt_ms += t1 - t0;
+    stats_copy_ms += t2 - t1;
+    stats_screen_ms += t3 - t2;
+    if(bench_mode == BENCH_MODE_COPY)
+        stats_copy_bytes += (uint64_t)copy_bytes_per_frame();
+
+    elapsed = now - stats_start_ms;
+    if(elapsed >= 1000 && stats_frames > 0) {
+        double frames = (double)stats_frames;
+        double elapsed_s = (double)elapsed / 1000.0;
```
gyrovorbis:
```suggestion
        double elapsed_s = (double)elapsed / 1000000000.0;
```
ms to ns.

gyrovorbis (commented):
Still have a lot more to look at, but I'll go ahead and leave these initial comments here.

Overall, looks amazing, super stoked to support this. Still going through driver changes, but went through the first example and use-case THOROUGHLY. Sorry to be bitchy, but I am kinda anal about all new KOS examples being like the finest quality C/C++ code humanly possible, because it's how people learn a lot of this stuff and reflects our expertise as devs...  just lots of lil nitpicks.

kernel/arch/dreamcast/include/dc/pvr.h
GPF: updated PR commit.

kernel/arch/dreamcast/hardware/pvr/pvr_scene.c
GPF: Add copyright headers, use bool/EXIT_* in the RTT samples, switch benchmark
  timing to nanoseconds, and wait for render completion before freeing RTT
  texture memory.

examples/dreamcast/pvr/pvr_txr_targettexture/Makefile
GPF: Add copyright headers, use bool/EXIT_* in the RTT samples, switch benchmark
  timing to nanoseconds, and wait for render completion before freeing RTT
  texture memory.

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -887,6 +887,32 @@ void pvr_scene_begin(void);
...
+    \param  txr             The texture to render to.
+    \param  render_w        Width of the render area (in pixels).
+    \param  render_h        Height of the render area (in pixels).
+    \param  stride_px       Backing texture pitch (in pixels).
+
+    \retval 0               On success.
+    \retval -1              If the specified arguments are invalid.
+*/
+int pvr_scene_begin_txr_ex(pvr_ptr_t txr, uint32_t render_w,
+                           uint32_t render_h, uint32_t stride_px);
```
pcercuei: I wouldn't add a new function. Just update `pvr_scene_begin_txr()` instead.

Or if we want to retain compatibility with old code, which we may not (KOS master is probably going to be KOS 3.0 so it's semi-OK to break API)  `pvr_scene_begin_txr()` can be renamed to a deprecated `pvr_scene_begin_txr_legacy`, and we can have `pvr_scene_begin_txr()` be a `_Generic()` macro that resolves to the _ex or the _legacy function accordingly.

kernel/arch/dreamcast/hardware/pvr/pvr_scene.c
```diff
@@ -164,6 +170,26 @@ void pvr_scene_begin_txr(pvr_ptr_t txr, uint32_t *rx, uint32_t *ry) {
...
+    if(!txr || render_w == 0 || render_h == 0 || stride_px < render_w)
+        return -1;
+
+    /* The existing render-to-texture path programs a 16-bit render pitch in
+       64-bit units. Keep that behavior explicit for the sized RTT API. */
+    pvr_state.next_to_txr_rp = stride_px * 2 / 8;
+    pvr_state.next_to_txr_w = render_w;
+    pvr_state.next_to_txr_h = render_h;
+    pvr_state.next_to_txr_stride_px = stride_px;
+    pvr_state.next_to_txr_addr = (uint32_t)(txr) - PVR_RAM_INT_BASE;
```
pcercuei, line 163: I might do this cleanup later, so in the meantime it is OK (unless you feel like doing it yourself), but I think this could be factorized.

Right now `pvr_begin_queued_render()` will pick the 'to_txr' variables if doing RTT, otherwise will pick values for non-RTT. We could very well have those variables set to RTT values in `pvr_scene_begin_txr_ex` and to regular values in `pvr_scene_begin`. In which case `pvr_begin_queued_render()` would not need to know whether or not we're doing RTT.

pcercuei (changes requested):
See the comments.

Also you should separate the example from the PVR driver update in two distinct commits.

kernel/arch/dreamcast/include/dc/pvr.h
gyrovorbis: I 100% agree with you here in terms of how to handle the API for it. Would much rather this be the new `pvr_scene_begin_txr()`, and your proposal would keep the existing behavior while also letting us implement the new behavior with the same API. Fantastic.

But... bad news on the implementation of this proposal. It would have to be done _extremely_ carefully, because in their infinite wisdom, the C++ standards committee has blatantly refused to add support to the language for the `_Generic()` keyword for almost a decade now, so you cannot ever use that construct within a public header which can be consumed by both C and C++...

What you have to do in such a scenario is what I'm doing for the generic versions of my vector API in SH4ZAM, which you can see here: https://github.com/gyrovorbis/sh4zam/blob/2f343ed2d4f7ed090a0c53986abf235f0e9a656a/include/sh4zam/shz_vector.h#L710.

I basically have to do the following pattern:
```cpp
#ifndef __cplusplus
    /* C type-generic vector absolute value, resolving the call via C11's `_Generic()` selection mechanism. */
#   define shz_vec_abs(vec) \
        _Generic((vec), \
                 shz_vec2_t: shz_vec2_abs, \
                 shz_vec3_t: shz_vec3_abs, \
                 shz_vec4_t: shz_vec4_abs)(vec)
#else

    //! C++ type-generic vector absolute value, resolving the call via C++'s function overloading mechanism. */
    SHZ_INLINE shz_vec2_t shz_vec_abs(shz_vec2_t vec) SHZ_NOEXCEPT { return shz_vec2_abs(vec); }
    SHZ_INLINE shz_vec3_t shz_vec_abs(shz_vec3_t vec) SHZ_NOEXCEPT { return shz_vec3_abs(vec); }
    SHZ_INLINE shz_vec4_t shz_vec_abs(shz_vec4_t vec) SHZ_NOEXCEPT { return shz_vec4_abs(vec); }
#endif
```

I'm totally okay if we want to do this, and am a fan of function overloading from an ergonomics perspective, but we would have to be VERY careful here, as we've never done such a thing in KOS's public API before, and we would need to establish a good precedent for how to do this moving forward. :)

kernel/arch/dreamcast/hardware/pvr/pvr_scene.c
gyrovorbis: That would be fantastic, but yeah, you can prolly do that in a follow-up cleanup once this makes it in.

gyrovorbis:
Yo, @pcercuei, have you had a chance to validate that from a HW/driver perspective, he's properly:

1) Configuring the right regs for explicitly setting the render-target resolution. I have not had a chance to research how the HW configuration has to work there, specifically in terms of what registers need to be set to what.

2) Not racing between when he finishes his first RTT pass and when he starts the main render pass to the framebuffer. Does he need a `pvr_wait_render_done()` or something to basically be a HW barrier to ensure the PVR is done drawing to the texture before swapping render-targets and trying to use it as a texture? We don't wanna race right there, but we also don't wanna stall waiting for the PVR to complete, unless we absolutely need to.

Basically I've been focusing on the API, examples, and docs, but have yet to validate the driver changes and that one particular aspect of how it should be used in the examples.

kernel/arch/dreamcast/include/dc/pvr.h
gyrovorbis: Oh... sorry, I'm an idiot too. You don't even need to use `_Generic()` to implement this style of function-overloading based on _argument count_. That's only for when you are doing so with the same number of parameters but are overloading based on the type of the argument, which the user passes at the call-site.

Refer back to how I handled making `sem_init()` support both the POSIX behavior and the existing KOS behavior, based on number of args passed at the call-site: https://github.com/KallistiOS/KallistiOS/blob/584293bb488615322cae9c52f84e36a3069f7e0d/include/semaphore.h#L51. 

NOTE HOWEVER: If we're going to implement this same function-overloading-by-argument-count behavior for `pvr_scene_begin_txr()`, that I  would rather start moving the selectors used to implement overloading for both into a common header, so that it becomes resuable macro infrastructure that we can leverage to implement such overloading in the API in the future.

I do exactly that within libGimbal's API, and it's not an uncommon thing to do in high-end advanced C codebases: 
https://github.com/gyrovorbis/libgimbal/blob/d1e1b2f1746c9b2702c7f6fa8446fd7b9563c503/lib/api/gimbal/preprocessor/gimbal_macro_composition.h#L316

QuzarDC: I don't see why we would jump through such hoops, is the name `pvr_scene_begin_txr` so fundamental and singular that no other would suffice? I get entirely the desire to not keep around that one as the limited way to use the functionality, especially if the newer would be more useful so why not something like `pvr_scene_begin_rtt` or such as the new name and deprecate the older limited function?

That way users of the existing one can get a deprecated message instructing on the use of the new API, nothing's broken, we don't litter in another `_ex` function, and after a version we drop the deprecated one?

Is there some benefit to these overloadings that I'm missing?

GPF: ok renamiing to pvr_scene_begin_rtt , and marking pvr_scene_begin_txr as deprecated. removing  examples and will upload new example as seperate PR.

pcercuei: The examples can stay in this PR, I would just prefer if the driver changes and the examples wouldn't be mixed in the same commits.

pcercuei (approved):

Issue #806, opened 2024-10-10
Plugging in a VMU disconnects the controller and any other subunits


QuzarDC:
Found while debugging for #804, I found this to be repeatable in any situation where the set of maple drivers is available for the devices in question.

Either by setting `MAPLE_IRQ_DEBUG` or removing those protections from the debug prints for attach/detach in https://github.com/KallistiOS/KallistiOS/blob/a11364a33bbd9fd5bbd40ba76743d2a8f0eb9cc0/kernel/arch/dreamcast/hardware/maple/maple_irq.c#L51-L54 and https://github.com/KallistiOS/KallistiOS/blob/a11364a33bbd9fd5bbd40ba76743d2a8f0eb9cc0/kernel/arch/dreamcast/hardware/maple/maple_irq.c#L153-L156 you can be notified of each time a new device/device removal is detected and get a message.

With that set, simply plug a vmu into an attached controller to see that the process of operations will be that the controller and any other devices currently connected (such as another VMU or rumble) are disconnected then they will all be reattached with the new VMU. Tested with multiple combinations of VMUs, rumble, and 3rd party VMS (which reports as a VMU). Only the gameshark/cdx 'VMU' (which is just a memory unit but reports as a VMU) does not trigger this behavior.

From what I can tell, it is that we legitimately receive a 'no response' from the probe of the controller when this happens, so it may be a purely hardware-side quirk. It may also be related though to some of the issues noted in #597 and that have been otherwise observed as happening when there's 'too many maple commands'. Alternatively, it may be that there is a bit of the attach/detach process that we're not currently handling properly (for instance, it might be that a subdev with more than one function always triggers 'no response' on the host, which is meant to flag 'check me for new subdevs' rather than 'disconnect me').

EDIT: Found mention to this same issue in #40 as well. Apparently commercial games have the same issue.

QuzarDC:
Took some time to test and #1258 did indeed work around this issue. The VMU still causes the controller to not respond for a beat, but now we don't detach and reattach everything needlessly.

PR #1463, opened 2026-06-27
maple: Fix detach/shutdown bugs, tidy IRQ code and docs

andressbarajas:
While trying to chase a maple shutdown issue that I ultimately could not reproduce I found a few things that needed to be fixed. Mainly:

1. examples/dreamcast/maple.elf example does not detect attach/detach of controllers that have VMU/Rumble in them
2. Fixed a dangling pointer
3. Moved an if-block of code to protect against NULL dereferences.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_irq.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_irq.c
@@ -16,6 +16,7 @@
 #include <dc/asic.h>
 #include <dc/pvr.h>
 #include <kos/dbglog.h>
+#include <kos/regfield.h>
 #include <kos/thread.h>
 
 /*********************************************************************/
@@ -66,12 +67,12 @@ static void vbl_chk_next_subdev(maple_state_t *state, maple_frame_t *frm, int p)
 
     if(dev && dev->probe_mask) {
         int u = __builtin_ffs(dev->probe_mask);
-        dev->probe_mask &= ~(1 << (u - 1));
+        dev->probe_mask &= ~BIT(u - 1);
 
         vbl_send_devinfo(frm, p, u);
     } else {
         /* Nothing else to probe on this port */
-        state->scan_ready_mask |= 1 << p;
+        state->scan_ready_mask |= BIT(p);
     }
 }
 
@@ -79,18 +80,18 @@ static void vbl_dev_probed(int p, int u) {
     maple_device_t *dev = maple_enum_dev(p, 0);
 
     if(dev)
-        dev->dev_mask |= 1 << (u - 1);
+        dev->dev_mask |= BIT(u - 1);
 }
 
 /* Check the sub-devices for a top-level port */
 static void vbl_chk_subdevs(maple_state_t *state, int p, uint8_t newmask) {
     maple_device_t *dev = maple_enum_dev(p, 0);
 
-    newmask &= (1 << (MAPLE_UNIT_COUNT - 1)) - 1;
+    newmask &= BIT(MAPLE_UNIT_COUNT - 1) - 1;
 
     /* Disconnect any device that disappeared */
     for(size_t u = 1; u < MAPLE_UNIT_COUNT; u++) {
-        if(dev->dev_mask & ~newmask & (1 << (u - 1))) {
+        if(dev->dev_mask & ~newmask & BIT(u - 1)) {
             vbl_chk_disconnect(state, p, u);
         }
     }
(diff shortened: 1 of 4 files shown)


kernel/arch/dreamcast/hardware/maple/maple_irq.c
```diff
@@ -128,10 +128,16 @@ static void vbl_autodet_callback(maple_state_t *state, maple_frame_t *frm) {
     dev = maple_enum_dev(p, u);
 
     if(resp->response == MAPLE_RESPONSE_NONE) {
-        /* No device, or not functioning properly; check for removal */
+        /* Top-level device is removed */
         if(u == 0) {
-            if(dev && dev->dev_mask == 0)
+            if(dev) {
```
QuzarDC: I believe this would cause the problem fixed by #1258 ( #806 ) to reemerge though. It's probably possible to account for both, but generally the idea is that the controller might stop responding briefly but the subdevs continue to respond and we didn't want to disconnect the whole tree in those cases. What should be happening is that the subdevs get their own probed NONE responses and get disconnected and detach.

andressbarajas: I'll look into it. I dont think I got that experience with this change against the Maple example. I remember that issue from long ago and tested it earlier by having a controller with a VMU plugged in and just removing the VMU. The maple example just reported the removal of the VMU and not the controller. Same with re-attachment of the VMU. Double checking now.

andressbarajas: Yes, its functioning okay in maple.elf. If I remove or attach a VMU or Rumble, the callback only reports the sub device. If I detach/attach a controller(top device) it detaches/attaches both devices . I will look at the #1258 more closely and see if I can reproduce.

pcercuei:
I can confirm that there's something fishy going on with the maple.

However, hotplug did absolutely work properly after #1258.

pcercuei:
Nevermind. #1258 is the source of the problem.

kernel/arch/dreamcast/hardware/maple/maple_irq.c
andressbarajas: @QuzarDC Removed this commit. @pcercuei said he fixed the issue.

gyrovorbis (approved):
LMFAO @ the driver deref happening OUTSIDE the null check... good fixes, nice cleanup, and the docs on the bugs look good. Nice!

kernel/arch/dreamcast/include/dc/maple.h
```diff
@@ -21,6 +21,15 @@
...
     functionality that is implemented in their header files, as well.
 
+    \bug    Sending a rumble (PuruPuru / Jump Pack) command can cause VMUs
+            plugged into any controller to beep. This is a hardware-level side
+            effect and not something KOS can prevent in software.
+
+    \bug    Inserting a VMU can cause its parent controller to briefly disconnect
+            and re-enumerate on the Maple bus, producing a detach/attach event for
+            the controller as well as the VMU. This is another hardware-level side
+            effect and not something KOS can prevent in software.
```
QuzarDC, line 31: I think it's more like this, right?

```
Inserting a VMU can cause its parent controller to briefly disconnect
and re-enumerate on the Maple bus. This is another hardware-level side
effect and KOS addresses it by delaying the disconnect of the first unit
on a port (controller) until it's confirmed that there are no subunits
still responding. This prevents the insertion of a VMU From triggering
an attach/detach of its controller.
```

And thanks for getting these in here :) I don't think the first had ever been described in the repo before and the second was kind of buried away in #806  (linked just to create better trackback)

QuzarDC: Hrm, looks like #1464 reintroduced that behavior so maybe it had never actually been fixed at all but had accidentally stopped as we were not triggering controller detaches at all.

QuzarDC: Hopefully #1466 will test fine with others. If so, based on that it would be something like:
```
Inserting a VMU can cause its parent controller to briefly disconnect
and re-enumerate on the Maple bus. This is another hardware-level
bug that KOS addresses by waiting for a device to fail multiple attempts
to detect it before disconnecting it. This prevents the insertion of a
VMU from triggering an attach/detach of all devices on the same port.

andressbarajas: @QuzarDC I think it would be a better idea for the docs change to be a part of the PR that fixes it instead of before. That should get this PR moving. What do you think?

PR #1464, opened 2026-06-27
maple: Fix unplug detection

pcercuei:
When plugging out a controller with peripherals, the maple driver would not properly report the event.

The top device (unit 0) would be reported as disconnected only after all sub-devices were reported as disconnected, but:
- at that point the sub-devices were not probed anymore, as they are only probed when the unit 0 device still responds;
- any sub-device being marked as unplugged would not update the device mask in the parent device.

Fix this issue by updating the parent device's mask when sub-devices are unplugged, and by making sure that we probe any sub-devices when the top device is detected as unplugged.

PR #1466, opened 2026-06-29
maple: Controller and accessories detach triggering

QuzarDC:
See #1464 .

This first reverts #1258 as it was only preventing the issue from #806 by accidentally suppressing controller detaching entirely when there were subunits connected to it. Devices attached to a disconnected controller would never be probed again once it was disconnected, and it would never get detached because it would wait for all attached devices to report that they were disconnected as well.

In order to work around the issue instead add a counter to the autodetect failures, requiring that a device be unresponsive for multiple successive autodetects before being actually disconnected.

Originally I had set the timeout count to 10 and that was sufficient for plugging in a VMU to a controller with a puru also in it, but an empty controller would still trip it. At 16-17 it would *usually* work, but only at 18 did I stop seeing it ever happen.

### DIFF
--- a/kernel/arch/dreamcast/include/dc/maple.h
+++ b/kernel/arch/dreamcast/include/dc/maple.h
@@ -262,6 +262,11 @@ typedef struct maple_response {
     uint8_t   data[];     /**< \brief Data (if any) */
 } maple_response_t;
 
+/* \cond */
+/* Number of failed autodetects before detaching a device. */
+#define MAPLE_DEV_VALID_TIMEOUT 30
+/* \endcond */
+
 /** \brief   One maple device.
     \ingroup maple
 
@@ -272,7 +277,7 @@ typedef struct maple_response {
 */
 typedef struct maple_device {
     /* Public */
-    bool            valid;  /**< \brief Is this a valid device? */
+    uint8_t         valid;  /**< \brief Is this a valid device? 0 for no */
     int             port;   /**< \brief Maple bus port connected to */
     int             unit;   /**< \brief Unit number, off of the port */
     maple_devinfo_t info;   /**< \brief Device info struct */
--- a/examples/dreamcast/rumble/rumble.c
+++ b/examples/dreamcast/rumble/rumble.c
@@ -52,8 +52,8 @@ void print_rumble_fields(purupuru_effect_t fields) {
   printf("  .inc    =  %u,\n", fields.inc);
 }
 /* This blocks waiting for a specified device to be present and valid */
-void wait_for_dev_attach(maple_device_t **dev_ptr, unsigned int func) {
-    maple_device_t *dev = *dev_ptr;
+void wait_for_dev_attach(maple_device_t **cont_dev_ptr, maple_device_t **puru_dev_ptr) {
+    maple_device_t *dev = *puru_dev_ptr;
     point_t w = {40.0f, 200.0f, 10.0f, 0.0f};
 
     /* If we already have it, and it's still valid, leave */
@@ -69,20 +69,20 @@ void wait_for_dev_attach(maple_device_t **dev_ptr, unsigned int func) {
 
     plx_fcxt_begin(cxt);
     plx_fcxt_setpos_pnt(cxt, &w);
-    if(func == MAPLE_FUNC_CONTROLLER)
-        plx_fcxt_draw(cxt, "Please attach a controller!");
-    else if(func == MAPLE_FUNC_PURUPURU)
-        plx_fcxt_draw(cxt, "Please attach a rumbler!");
+    plx_fcxt_draw(cxt, "Please attach a rumbler!");
     plx_fcxt_end(cxt);
 
     pvr_scene_finish();
 
     /* Repeatedly check until we find one and it's valid */
     while((dev == NULL) || !dev->valid) {
-        *dev_ptr = maple_enum_type(0, func);
-        dev = *dev_ptr;
+        *puru_dev_ptr = maple_enum_type(0, MAPLE_FUNC_PURUPURU);
+        dev = *puru_dev_ptr;
         usleep(50);
     }
+
+    /* Now that we have a rumbler, grab its parent */
+    *cont_dev_ptr = maple_enum_dev(dev->port, 0);
 }
 
 
@@ -136,8 +136,7 @@ int main(int argc, char *argv[]) {
 
         /* Before drawing the screen, trap into these functions to be
            sure that there's at least one controller and one rumbler */
-        wait_for_dev_attach(&contdev, MAPLE_FUNC_CONTROLLER);
-        wait_for_dev_attach(&purudev, MAPLE_FUNC_PURUPURU);
+        wait_for_dev_attach(&contdev, &purudev);
 
         /* Start drawing and draw the header */
         pvr_scene_begin();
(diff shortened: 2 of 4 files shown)


kernel/arch/dreamcast/include/dc/maple.h
```diff
@@ -262,6 +262,11 @@ typedef struct maple_response {
     uint8_t   data[];     /**< \brief Data (if any) */
 } maple_response_t;
 
+/* \cond */
+/* Number of failed autodetects before detaching a device. */
+#define MAPLE_DEV_VALID_TIMEOUT 18
+/* \endcond */
```
pcercuei, line 268: I see no reason to have that in the maple.h header if it's only used internally.

pcercuei (commented):
> Originally I had set the timeout count to 10 and that was sufficient for plugging in a VMU to a controller with a puru also in it, but an empty controller would still trip it. At 16-17 it would _usually_ work, but only at 18 did I stop seeing it ever happen.

You could push it to 20 or even 30, that would be about a third of a second / half a second, still acceptable IMHO. We're trying to detect a human unplugging a cable so the time frames can be on the human scale.

We're dealing with electronics here, stuff like temperature and age of capacitors can have an effect, so it's better to have some margin of error.

andressbarajas:
@QuzarDC Using maple example, when I remove a VMU it correctly says that the vmu has been removed. When I reinsert the VMU it says the controller and the VMU have been attached. Also Im getting weird results with testing with the rumble example too. If I detach the controller, the example says insert rumbler instead of insert controller.

kernel/arch/dreamcast/include/dc/maple.h
QuzarDC: To maintain it the same across `maple_irq.c` and `maple_driver.c` without creating a new `maple_internal.h`. I suppose though, it might make sense to not use it in the driver attach code and instead just have a flat 1 there and it will get updated to the timeout on its first autodetect.

QuzarDC:
> > Originally I had set the timeout count to 10 and that was sufficient for plugging in a VMU to a controller with a puru also in it, but an empty controller would still trip it. At 16-17 it would _usually_ work, but only at 18 did I stop seeing it ever happen.
> 
> You could push it to 20 or even 30, that would be about a third of a second / half a second, still acceptable IMHO. We're trying to detect a human unplugging a cable so the time frames can be on the human scale.
> 
> We're dealing with electronics here, stuff like temperature and age of capacitors can have an effect, so it's better to have some margin of error.

Good point, I was forgetting the context. Especially how this isn't meant to be something that's just happening over and over, quickly, or with low latency.

QuzarDC:
> @QuzarDC Using maple example, when I remove a VMU it correctly says that the vmu has been removed. When I reinsert the VMU it says the controller and the VMU have been attached. Also Im getting weird results with testing with the rumble example too. If I detach the controller, the example says insert rumbler instead of insert controller.

I went ahead and upped the timeout to 30 per @pcercuei 's suggestion. If that still doesn't work for you, think you could up it further to verify if it's just the timeout sometimes being *really* long or the logic being wrong?

Regarding the rumble demo, that should have always been possible to happen, and I think is just reflective of some of the sloppiness of the demo. Since the two wait checks are sequential rather than nested it's always been possible to get notified about the rumble . It's also been subject to other bad behavior on master too. Disconnecting the controller can cause the text on the screen (aside from the value) to disappear and for it to crash on reinsert. 

I dug into it a bit, and finally was able to see the mistake in it that was causing the bad behavior, and corrected that along with simplifying/correcting the detection behavior.

andressbarajas:
@QuzarDC Upping the timeout didnt help. I think I have something working with a different approach. I will follow up tomorrow.

andressbarajas:
@QuzarDC I had to make two little changes to get it to work. Here is diff:

```diff --git a/kernel/arch/dreamcast/hardware/maple/maple_irq.c b/kernel/arch/dreamcast/hardware/maple/maple_irq.c
index 4bff799a..641f59d6 100644
--- a/kernel/arch/dreamcast/hardware/maple/maple_irq.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_irq.c
@@ -204,8 +204,9 @@ static void vbl_autodet_callback(maple_state_t *state, maple_frame_t *frm) {
 }
 
 static void vbl_autodetect(maple_state_t *state) {
-    maple_device_t  *dev = maple_state.ports[state->detect_port_next].units[0];
-    maple_frame_t   *frm = (dev != NULL) ? &dev->frame : &detect_frame;
+    /* Probe on the shared detect_frame, not dev->frame: sharing it with
+       status polling stops a glitching port from recovering until teardown. */
+    maple_frame_t   *frm = &detect_frame;
 
     /* Queue a detection on the next device */
     bool queued = vbl_send_devinfo(frm,
diff --git a/kernel/arch/dreamcast/include/dc/maple.h b/kernel/arch/dreamcast/include/dc/maple.h
index 2617724f..46c3a615 100644
--- a/kernel/arch/dreamcast/include/dc/maple.h
+++ b/kernel/arch/dreamcast/include/dc/maple.h
@@ -273,7 +273,7 @@ typedef struct maple_response {
 
 /* \cond */
 /* Number of failed autodetects before detaching a device. */
-#define MAPLE_DEV_VALID_TIMEOUT 30
+#define MAPLE_DEV_VALID_TIMEOUT 50
 /* \endcond */
 
 /** \brief   One maple device.
```

pcercuei:
50 vblanks, that's close to a second, it's possible to have a maple device unresponsive for so long?

andressbarajas:
> 50 vblanks, that's close to a second, it's possible to have a maple device unresponsive for so long?

Yes, I actually think it was around 40 vblanks but upped it to 50 to be safe. This only happens with VMU getting plugged in :^/

QuzarDC:
> 50 vblanks, that's close to a second, it's possible to have a maple device unresponsive for so long?

I'm sort of imagining now this same conversation happening at SEGA in '98. With such a limited sample it could well be that others take even longer. Perhaps it's worth it to just eat the disconnects, or keep a short timeout to help avoid the issue with purus disconnecting and to protect sometimes. The count is not the number of vblanks, but the number of autodetects that have failed for that device. So it's 4x as many vblanks putting 50 at 3.3-4s. 

It also occurs to me that with a human-interactable timeout time, we risk the device changing when it becomes responsive again, which isn't accounted for with the current code. If I unplug a controller and plug in a keyboard within the timeout period, the result would be a broken maple dev. I think it would just be a matter of testing the maple devinfo in the response against what was already in the dev, but the current code does the exact opposite, updating the devinfo as if it can legitimately change so not sure if there's something to that rather than just being needless.

pcercuei:
I wonder, when we detect a unit0 disconnect, can we force it to re-appear by resetting the maple bus? (either write the magic reset1/reset2 values again or maple_bus_disable() + maple_bus_enable())

pcercuei (approved):

QuzarDC:
> I wonder, when we detect a unit0 disconnect, can we force it to re-appear by resetting the maple bus? (either write the magic reset1/reset2 values again or maple_bus_disable() + maple_bus_enable())

This might be possible, but I'm quite certain that would cause all other devices on the bus to disappear too. Though it might be for a shorter or fixed time rather than the variable time we're currently seeing. Would need testing.

Along a similar line of thought, it *might* reduce the recovery time to briefly pause autodetect or all maple traffic on a disconnect, as it may be that other bus traffic slows it down.

Hopefully we can do some further experimenting, for new methods in the future, and/or refine this one a bit more.

Issue #116, opened 2023-02-27
Performance and reliability issues with the current network stack


darcagn:
Opening up this issue to track the current performance/reliability problems with the KOS network stack. 

Examples include:
- I updated the httpd-ack server (https://gotwalls.com/) from KOS 1.x+lwIP to KOS 2.x with its native networking stack, and now transfer from DC to PC tops out around 700kB/s and shows rapid up/down fluctuations as if the system is pausing/unpausing the transfer rapidly (possibly waiting on buffers mentioned below?)
- @DC-SWAT wrote an FTP server module for DreamShell; transferring from DC to PC does not exceed 750kB/s and transferring from PC to DC results in ~1.5kB/s speeds and dropped connections.
- @ljsebald wrote a simple speed testing utility to test connecting a DC and PC over the network and transferring a small buffer. In my test run using the BBA, it showed 10485760 bytes received by the Dreamcast console in 261479 milliseconds, or 40.10188kB/s. 

During a conversation with @ljsebald, he suggested several things which need improvement upon:
- > there's definitely a bug if you use recvmsg and specify the MSG_WAITALL flag... and the connection gets dropped in the middle of waiting... that will just end up waiting forever :/ 
- There may be a bug with connection close
- Allocating network buffers ahead of time would allow increased performance at the expense of memory use, so this preferably could be a tuneable option. There's no way to do a zero-copy network stack in KOS's architecture, but it should be possible to drastically reduce the number of copies by pre-allocating buffers. Optimally, the only copies should be when one calls send() to copy into an internal buffer, and when one calls recv() to copy out of the internal buffer into the specified destination. 

To quote @ljsebald:
> one of the biggest bottlenecks of the current design of the network stack is that it is constantly doing memcpy()
> like... all over the place
> [...]
> right now, when you do a tcp socket and call send, this is what happens:
> - net_tcp_sendto - memcpy into internal buffer, call tcp_send_data
> - tcp_send_data potentially shifts around said internal buffer with some more memcpys, calls net_ipv6_checksum_pseudo/net_ipv4_checksum to checksum the data (which should always be aligned, at least, so takes the fast-path), then calls net_ipv6_send
> - net_ipv6_send does another memcpy before calling the network adapter's transmit
> -- If using a bba, this does a g2_write_block_32 (which is basically another memcpy, but this is unavoidable)
> -- if using a lan adapter, you have a similar story, but with an even slower 8-bits at a time write
> -- if using a modem, it's... worse
[...]

PR #1472, opened 2026-07-02
net: TCP/IP stack speedup enhancements & bug fixes (+ IP_TOS)

andressbarajas:
This PR adds speedup enhancements and bug fixes to the KOS TCP/IP stack and the Broadband Adapter driver, plus a new `IP_TOS` socket option.

## Speedups:

- **IPv4 checksum**: Bring back @gyrovorbis deferred carry folding approach to the current algorithm. The IPv4 checksum is ~2.2× faster at typical sizes — 1.55× on a 20-byte IP header, 2.22× at the 1460-byte TCP MSS.

- **Zero-copy TCP/IPv4 send**: Build the IP/eth headers in place in front of the TCP segment (`net_ipv4_send_inplace`) instead of copying header+payload into a staging buffer. Header-less links (PPP), loopback, and fragmentation still take the copy path. This increased TX speed by 4%. Mentioned here (https://github.com/KallistiOS/KallistiOS/issues/116)

- **Raise the `SO_SNDBUF` cap 64 KB → 1 MB** (default unchanged): It's only a local staging buffer with no protocol limit (unlike the receive window, which stays clamped at 64 KB). The old cap starved TX because the buffer has to hold onto packets that have been sent but haven't been acked yet; a 512 KB buffer measured ~+20% download.

- **BBA TX via G2 DMA.**: `bba_rtx()`copied each frame into the adapter TX FIFO with PIO (`g2_write_block_32`); use the G2 DMA instead. Measured DC→PC download: 64 KB `SO_SNDBUF` 3.75 → 6.41 MB/s (+71%); 512 KB 3.75 → 6.62 MB/s (+76%).

## Bug fixes
- **MSG_WAITALL recv no longer blocks forever**: When the connection drops mid-receive it returns what it has instead of hanging. Mentioned here (https://github.com/KallistiOS/KallistiOS/issues/116)

- **Defer `poll()` wakeups until after the socket / UDP mutex is released** (`net_tcp`, `net_udp`): Waking waiters under the lock inverted the socket-vs-poll lock order (ABBA), hanging `poll()`. This is what caused hangs when trying to use libsmb2.

## New Option
- **IP_TOS socket option**: For DSCP/ToS marking (IPv4 ToS byte + IPv6 traffic class). It lets apps mark some of its internet traffic as “high priority” so that a network may handle it faster or with less delay.

## Testing
- HW-measured on real Dreamcast over the BBA and WIZnet W5500(SCIF).
- Validated against all network examples, Meese's network testsuite (5000 attempts), libsmb2, and FTP (liblftpd).

### DIFF
--- a/kernel/arch/dreamcast/hardware/network/broadband_adapter.c
+++ b/kernel/arch/dreamcast/hardware/network/broadband_adapter.c
@@ -504,7 +504,7 @@ static void bba_dma_cb(void *p) {
     if(next_len) {
         g2_dma_transfer(next_dst, next_src, next_len, 0,
                         bba_dma_cb, 0,  /* callback */
-                        1,  /* dir = 1, we're *reading* from the g2 bus */
+                        G2_DMA_TO_SH4,
                         0, G2_DMA_CHAN_BBA, 0);
         next_len = 0;
     }
@@ -536,7 +536,7 @@ static int bba_copy_packet(uint8_t *dst, uint32_t s, int len) {
             dma_used = 1;
             g2_dma_transfer(dst, src, len, 0,
                             bba_dma_cb, 0,  /* callback */
-                            1,  /* dir = 1, we're *reading* from the g2 bus */
+                            G2_DMA_TO_SH4,
                             0, G2_DMA_CHAN_BBA, 0);
         }
         else if(next_len) {
--- a/kernel/arch/dreamcast/hardware/network/broadband_adapter.c
+++ b/kernel/arch/dreamcast/hardware/network/broadband_adapter.c
@@ -8,6 +8,7 @@
 
  */
 
+#include <stdatomic.h>
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -16,7 +17,6 @@
 #include <dc/net/broadband_adapter.h>
 #include <dc/asic.h>
 #include <dc/g2bus.h>
-#include <dc/sq.h>
 #include <dc/flashrom.h>
 #include <dc/memory.h>
 #include <kos/cache.h>
@@ -439,7 +439,7 @@ static uint8_t *rxbuff;
 static uint32_t rxbuff_pos;
 static int rxin;
 static int rxout;
-static int dma_used;
+static atomic_int dma_used;
 
 static uint32_t rx_size;
 
@@ -492,7 +492,9 @@ static void rx_finish_enq(int room) {
     if(room > 0 && (((rxin + 1) % MAX_PKTS) != rxout)) {
         rxin = (rxin + 1) % MAX_PKTS;
         thd_worker_wakeup(rx_worker);
-        thd_schedule(true);
+        /* Only in IRQ context (unsafe from the TX drain). */
+        if(irq_inside_int())
+            thd_schedule(true);
     }
 }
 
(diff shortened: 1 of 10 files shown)


DC-SWAT:
Why isn't DMA used for TX?

andressbarajas:
> Why isn't DMA used for TX?

I just looked at the XXXX suggestion comment and implemented that. Also code-wise it was an easier change since I didn't have to touch any of the RX dma handling to get it working. I can look into it this weekend see what the results are.

QuzarDC:
> > Why isn't DMA used for TX?
> 
> I just looked at the XXXX suggestion comment and implemented that. Also code-wise it was an easier change since I didn't have to touch any of the RX dma handling to get it working. I can look into it this weekend see what the results are.

I would advocate checking DMA here as well. I'm wary of any internal opaque use of SQs as they're used for rendering and the expectation from the user is generally that they're manually controlled. Though I do recognize that here they get locked then _pretty much_ immediately irqs are disabled for the `g2_lock`, the gap there could hurt if it ends up forcing a `pvr_list_begin` to wait.

kernel/arch/dreamcast/hardware/network/broadband_adapter.c
```diff
@@ -599,12 +599,44 @@ static int bba_rtx(const uint8_t *pkt, int len, int wait)
...
+        g2_fifo_wait();
+
+        sq_cpy((void *)dst, pkt, aligned);
+
+        /* We have some free time here to finish up the SQs work
+           before we unlock G2 and enable IRQ. So we'll unlock it first. */
+        sq_unlock();
+        sq_wait();
+
+        g2_unlock(ctx);
```
QuzarDC: I know this is just taken from `spu_memload_sq` but shouldn't the unlock be after doing the leftovers? The `g2_write_block_32` will have to lock/unlock again anyways and it seems to give the opportunity for getting sent back to the queue before finishing the last bit of transfer.

Since there's the possible memset and final write after the if/else block it might even make sense to only unlock right at the end.

DC-SWAT:
> > > Why isn't DMA used for TX?
> > 
> > 
> > I just looked at the XXXX suggestion comment and implemented that. Also code-wise it was an easier change since I didn't have to touch any of the RX dma handling to get it working. I can look into it this weekend see what the results are.
> 
> I would advocate checking DMA here as well. I'm wary of any internal opaque use of SQs as they're used for rendering and the expectation from the user is generally that they're manually controlled. Though I do recognize that here they get locked then _pretty much_ immediately irqs are disabled for the `g2_lock`, the gap there could hurt if it ends up forcing a `pvr_list_begin` to wait.

In addition to the fact that SQs is used in rendering, it is also generally better to work with the G2 bus through DMA for data transfer.

kernel/arch/dreamcast/hardware/network/broadband_adapter.c
DC-SWAT: Need to wait until the SQs has finished transmitting before enabling interrupts.

QuzarDC: > Need to wait until the SQs has finished transmitting before enabling interrupts.

Right, I'm wondering why we don't continue to hold the g2 lock until after all g2 writes in `bba_rtx()` are completed and not just the ones under this `if`. Currently it will g2_lock/unlock either 2 or 3 times depending on `len` without any substantial time between them. It seems that `_raw` versions of the g2 operations could be leveraged like elsewhere.

andressbarajas:
@QuzarDC Force pushed G2 DMA replacement and added new cleanup commit to replace hardcoded 1 with `G2_DMA_TO_SH4`

kernel/arch/dreamcast/hardware/network/broadband_adapter.c
andressbarajas: Code uses G2 DMA now instead of SQs.

QuzarDC (approved):
The TOS change would be an API break, but I don't think anyone would/should be directly using those functions (and it should be a simple update to adjust).

PR #1474, opened 2026-07-04
Newlib interface and errno cleanup

QuzarDC:
This focuses on the reent syscall functions we provide to newlib and does the following:
- Match function prototypes to newlib's headers. Most only differed in relatively trivial ways (`int` vs `ssize_t` for instance) but others were quite wrong and probably would have caused problems if ever actually used.
- Include the correct `<reent.h>` rather than `<sys/reent.h>`. The sys header only contained the system-specific portions of the reent code like the definition of the reent struct but the root-level header has the function prototypes for all the functions. Now we'd get errors if the function prototypes didn't match again.
- Update `isatty` to better protect against invalid file handles. It was only previously testing against negative values and not also out-of-bounds values, this could have lead to OoB memory access.
- Drop `env_lock.c`. I was going to replace the usage of newlib lock wrappers with a kos mutex, but it seems as though this is being provided entirely by newlib in a way that's identical to our own already.
- Replace direct access to errno in the reent params with assignment to standard errno. These might have predated errno being properly wired up to the thread structure. There should be no difference functionally but newlib doesn't guarantee this sort of direct access (even if we were modifying in the reent struct it should be done via a macro) and it implies that it's required.
- Update errno values to better match documentation. Specifically in the `execve` and `wait` stubs `EAGAIN` was being set which is both not defined in any standard I could find and misleading as it implies that these might work on retry. These were updated to return `EINVAL` and `ECHILD` which are values that would be tested for. For `times` also updated the overflow path to properly set `EOVERFLOW`.

Finally, as a followup to the other errno corrections, removed the kthread struct member `thd_errno`. This seems to have predated our integration with newlib. When setting an errno in an interrupt as was being done by genwait, we were sending the errno value to `thd_errno` despite it being entirely independent from errno generally which is stored in the thread's `thd_reent`. Now when a `genwait_wait` times out errno actually gets set, and when `thd_get_errno` is called you can read and set an errno that is actually readable via `errno`.

This can be trivially tested with something like:
```
errno = 0;
printf("genwait_wait returned %d, setting errno to %d [%s]\n", genwait_wait(NULL, NULL, 10), errno, strerror(errno));
```
which used to return:
`genwait_wait returned -1, setting errno to 0 [Success]`
and now returns:
`genwait_wait returned -1, setting errno to 11 [No more processes]`
(`EAGAIN == 11` the string newlib provides is just kind of weird for it)


QuzarDC:
As a followup to this, we may want to set up a way to detect errno being set in an interrupt. This could probably be done by setting up a fake reent struct that `impure_ptr` gets switched to and asserting that the errno there is 0. The current behavior would just set the errno in whatever thread was interrupted, though that would likely not be the intended behavior and could cause an interrupt to override an errno before it gets read.

kernel/thread/genwait.c
```diff
@@ -111,7 +111,7 @@ static void __nonnull_all genwait_unqueue(kthread_t *thd, int err) {
     /* Set the wake return value */
     if(err) {
         CONTEXT_RET(thd->context) = -1;
-        thd->thd_errno = err;
+        __errno_r(&thd->thd_reent) = err;
```
pcercuei: ~I think you should be able to just write `errno = err;`, no?~
Nevermind.

QuzarDC: Right, just for the benefit of onlookers, the whole issue is that `errno` is per-thread in order to track the last error set by a function in that context. Our scheduler swaps out where `errno` points to when changing to a new thread via swapping which `thd_reent` `_impure_ptr` points. `genwait_wake` will be happening either in the context of a thread not the thread being woken, or in an interrupt context, where in either case plain `errno` will not belong to that thread. In order to set the errno of a target thread, we use a macro from newlib to access the errno within the thread's reent structure.

For clarity's sake, I went ahead and updated this to `*thd_get_errno(thd) = err` so we minimize needing to directly interact with the newlib macro.

kernel/libc/newlib/newlib_wait.c
```diff
@@ -9,7 +9,9 @@
 #include <errno.h>
 
 int _wait_r(struct _reent *reent, int *status) {
+    (void)reent;
     (void)status;
-    reent->_errno = EAGAIN;
+
+    errno = ECHILD;
```
pcercuei, line 15: Note that `errno` is basically defined as `get_reent_ptr()->errno`, so using `reent->_errno` avoids the extra function call.

pcercuei (approved):
ACK with a small "FYI" comment below.

kernel/libc/newlib/newlib_wait.c
QuzarDC: Yea, I considered that. I would hope that gcc would usually be clever enough to inline such a thing, but also figured that it since these aren't part of any kind of critical paths the small difference would be well outweighed by the code clarity.

The way these were previously made it seem very much like it was required to set the errno like this directly into that reent param as if it was outside of the normal usage.

Perhaps just generally for usage throughout the codebase it might be worthwhile to look into overriding some part of the chain that leads to errno being translated that way into a define for `thd_current->reent->errno`

PR #1378, opened 2026-05-10
Make load code agnostic

andressbarajas:
As some of you know I have been working on the new dc-load/dc-tool replacement(kos-tool) that adds support for other consoles besides Dreamcast. I recently started porting KOS to another console and I found myself duplicating a lot of dcload related code that I know could be shared. So I pulled all the dcload related code outside of the Dreamcast paths and made it available for all future consoles. There is no functional change. This still works with the current dc-load-ip/serial w/ dc-tool-ip/serial. This change just makes it really easy to add future consoles. All they have to do is create a kernel/arch/CONSOLENAME/include/arch/kosload.h file and specify these like I do for Dreamcast:

```
/** \brief  Address of the dc-load syscall function pointer */
#define VEC_KOSLOAD      (MEM_AREA_P1_BASE | 0x0C004008)

/** \brief  Address of the dc-load magic value in BIOS RAM */
#define KOSLOADMAGICADDR ((unsigned int *)0x8c004004)

/** \brief  Flush the SH4 FIFO before issuing a syscall (serial mode) */
#define KOSLOAD_ARCH_FIFO_FLUSH()  do { while(FIFO_STATUS & FIFO_SH4); } while(0)
```

PR #1475, opened 2026-07-05
fs_dclsocket: Cleanup/fix pt2

andressbarajas:
This PR is mainly a follow up to close out this conversation: https://github.com/KallistiOS/KallistiOS/pull/1451#discussion_r3470300202 (To mirror how fs_dcload manages open handles in fs_dclsocket). Remove dead code too.

### DIFF
--- a/kernel/arch/dreamcast/fs/fs_dclsocket.c
+++ b/kernel/arch/dreamcast/fs/fs_dclsocket.c
@@ -659,7 +659,30 @@ static int dcls_writebuf(const uint8_t *buf, int len, int xlat) {
         return -1;
 
     memcpy(cmd.id, "DD02", 4);
-    cmd.value0 = htonl(1);
+    cmd.value0 = htonl(STDOUT_FILENO);
+    cmd.value1 = htonl((uint32_t) buf);
+    cmd.value2 = htonl(len);
+
+    send(dcls_socket, &cmd, sizeof(cmd), 0);
+
+    dcls_recv_loop();
+
+    mutex_unlock(&mutex);
+
+    return retval;
+}
+
+static int dcls_readbuf(uint8_t *buf, int len) {
+    command_3int_t cmd;
+
+    if(initted < 2)
+        return -1;
+
+    if(mutex_lock(&mutex))
+        return -1;
+
+    memcpy(cmd.id, "DC03", 4);
+    cmd.value0 = htonl(STDIN_FILENO);
     cmd.value1 = htonl((uint32_t) buf);
     cmd.value2 = htonl(len);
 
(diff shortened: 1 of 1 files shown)


kernel/arch/dreamcast/fs/fs_dclsocket.c
```diff
@@ -222,12 +224,21 @@ static void dcls_recv_loop(void) {
 static void *dcls_open(struct vfs_handler *vfs, const char *fn, int mode) {
     int hnd, dcload_mode = 0;
     int mm = (mode & O_MODE_MASK);
+    dcl_obj_t *entry;
     command_t *cmd = (command_t *)pktbuf;
 
     (void)vfs;
 
-    if(mutex_lock_irqsafe(&mutex))
+    entry = calloc(1, sizeof(dcl_obj_t));
```
QuzarDC, line 232: If this has to be irq safe then you need to also test if it's safe to malloc. I *think* though that it doesn't really need to be, and it seems like the mutex locking may be able to shrink.

QuzarDC:
I should have mentioned this with the first set of fixes, but it would probably make sense to refactor this into the dcload syscall. Part of my notion with refactoring fs_dcload away from the syscall was to be able to swap the net-based version in at the syscall driver level, since that's more accurate to what's happening. After all there shouldn't be any difference in the vfs or dbgio handling for them, just the way the data is retrieved from the host.

Absolutely not a requirement for here, but something to consider especially in how it should substantially shrink #1378

andressbarajas:
> I should have mentioned this with the first set of fixes, but it would probably make sense to refactor this into the dcload syscall. Part of my notion with refactoring fs_dcload away from the syscall was to be able to swap the net-based version in at the syscall driver level, since that's more accurate to what's happening. After all there shouldn't be any difference in the vfs or dbgio handling for them, just the way the data is retrieved from the host.
> 
> Absolutely not a requirement for here, but something to consider especially in how it should substantially shrink #1378

Thats a good plan. I'll work on that in a follow-up PR.

kernel/arch/dreamcast/fs/fs_dclsocket.c
andressbarajas: I looked at the other filesystem implementations, and it looks like mutex_lock_irqsafe() is only needed for the path used by dbgio, which is dcls_writebuf(). I updated the code and put the change in a new commit.

andressbarajas:
@QuzarDC Downgraded the file ops to a plain mutex_lock(), keeping mutex_lock_irqsafe() only on dcls_writebuf() (the dbgio handler). Also implemented console reading via dcls_readbuf(), mirroring fs_dcload's dcload_read_buffer().

kernel/arch/dreamcast/fs/fs_dclsocket.c
pcercuei: I'd be in favor of enforcing that in the high-level FS code. Doing I/O in interrupt context is a bad idea and I think the only case where we want to allow it is for pty.

pcercuei:
@andressbarajas could you split your 3rd commit ("give each open handle its own object") into two commits? The first one that introduces and uses the `dcl_obj_t`, and the second one that moves the path/dirent to it.

andressbarajas:
> @andressbarajas could you split your 3rd commit ("give each open handle its own object") into two commits? The first one that introduces and uses the `dcl_obj_t`, and the second one that moves the path/dirent to it.

Done

pcercuei (approved):

PR #1477, opened 2026-07-08
fs_dcload: Backend refactor

andressbarajas:
This is a followup PR to address this post: https://github.com/KallistiOS/KallistiOS/pull/1475#issuecomment-4886980608

Adds a swappable dcload syscall transport plus a network backend, so the existing fs_dcload serves /pc and the console over sockets on dcload-ip (when INIT_NET is used) retiring the duplicate fs_dclsocket.


kernel/arch/dreamcast/hardware/dcload_syscalls_net.c
```diff
@@ -210,9 +210,29 @@ static void dcls_recv_loop(void) {
     escape = false;
 }
 
+static int dcload_net_open(const char *fn, uint32_t flags, uint32_t umask) {
+    command_t *cmd = (command_t *)pktbuf;
+
+    if(mutex_lock(&mutex))
+        return -1;
```
pcercuei, line 220: `mutex_lock()` will never return non-zero here. The doc is wrong and it won't fail when called inside an interrupt (it asserts instead). If you need that, use `mutex_lock_irqsafe()`. Otherwise, just use `mutex_lock_scoped()`.

pcercuei (dismissed):

kernel/arch/dreamcast/hardware/dcload_syscalls_net.c
andressbarajas: Done and force pushed

andressbarajas:
@pcercuei @QuzarDC See any more changes I need to make?

QuzarDC:
> @pcercuei @QuzarDC See any more changes I need to make?

I was testing with the browse example and ran into unexpected hangs when browsing `/pc/` hadn't come up with a reproducer yet. Also ran into a bug in the browse example to do with handling not being able to open a dir leading to a blank screen ( I think it still tries to update the screen with entries from the dir which it has no info on) and hadn't made the time to track that down yet.

Haven't forgotten this though, and hope to clear it in the next few days as I couldn't find any reason that the above would be caused by the changes here.

QuzarDC (approved):

QuzarDC:
Tracked down the bug in the browse example, and will put in a PR for it. It was indeed entirely unrelated.

Thanks for working on this one, it's something I've had in mind for some time :)

PR #1480, opened 2026-07-10
Scoped asic disable, vblank cleanup

QuzarDC:
This was meant to be merely a cleanup of vblank, but while doing so I had the thought for a pattern that might be helpful elsewhere. Added a scoped asic disable to better encourage and support the temporary disabling of an event rather than the older more brute-force method of just disabling IRQs. In vlbank, it is used in order to prevent the vblank handler from triggering while a new entry is being added/removed from it. It should also be useful for bba/gdrom/pvr code where we are currently disabling irqs wholesale.

The changes for vblank are:
- Instead of disabling IRQs when adding/removing, take a lock and disable the vblank asic event until done.
- Instead of having numbered IDs for the handlers, pass out pointers to them.
- Whitespace cleanups
- Use FOREACH_SAFE macro rather than replicating it from scratch.
- Drop the return value for removing the handler as there's no longer a failure state without intentional tampering.

Additionally, as offered by @pcercuei , moving asic to use atomics in order to allow it to be safe to use without IRQs disabled.

### DIFF
--- a/kernel/arch/dreamcast/hardware/asic.c
+++ b/kernel/arch/dreamcast/hardware/asic.c
@@ -93,6 +93,7 @@
 
  */
 
+#include <stdatomic.h>
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
@@ -103,10 +104,6 @@
 #include <kos/irq.h>
 #include <kos/regfield.h>
 #include <kos/worker_thread.h>
-
-/* XXX These based on g1ata.c and pvr.h and should be replaced by a standardized method */
-#define IN32(addr)         (* ( (volatile uint32_t *)(addr) ) )
-#define OUT32(addr, data)  IN32(addr) = (data)
 
 /* The set of asic regs are spaced by 0x10 with 0x4 between each sub reg */
 #define ASIC_EVT_REG_ADDR(irq, sub) (ASIC_IRQD_A + ((irq) * 0x10) + ((sub) * 0x4))
@@ -157,8 +154,8 @@ static void handler_irq9(irq_t source, irq_context_t *context, void *data) {
     /* Go through each event register and look for pending events */
     for(reg = 0; reg < ASIC_EVT_REGS; reg++) {
         /* Read the event mask and clear pending */
-        uint32_t mask = IN32(ASIC_ACK_A + (reg * 0x4));
-        OUT32(ASIC_ACK_A + (reg * 0x4), mask);
+        uint32_t addr = ASIC_ACK_A + (reg * 0x4);
+        uint32_t mask = atomic_fetch_add((atomic_uint *)addr, 0);
 
         /* Short circuit going through the table if none on this reg */
         if(mask == 0) continue;
@@ -179,7 +176,7 @@ void asic_evt_disable_all(void) {
 
     for(irq = 0; irq < ASIC_IRQ_MAX; irq++) {
         for(sub = 0; sub < ASIC_EVT_REGS; sub++) {
-            OUT32(ASIC_EVT_REG_ADDR(irq, sub), 0);
+            atomic_store((atomic_uint *)ASIC_EVT_REG_ADDR(irq, sub), 0);
         }
     }
 }
@@ -194,8 +191,7 @@ void asic_evt_disable(uint16_t code, uint8_t irqlevel) {
     evt = code & 0xff;
 
     uint32_t addr = ASIC_EVT_REG_ADDR(irqlevel, evtreg);
-    uint32_t val = IN32(addr);
-    OUT32(addr, val & ~BIT(evt));
+    atomic_fetch_and((atomic_uint *)addr, ~BIT(evt));
 }
 
 /* Enable a particular G2 event */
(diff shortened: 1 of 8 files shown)


pcercuei:
I like the general idea of using `asic_evt_disable()` as a synchronization routine vs. the interrupts, but I don't think it can work as-is, because that function is not thread-safe.

You would first need to make it (and `asic_evt_enable()`) thread-safe, using the C11 atomics API to update the registers.

QuzarDC:
> I like the general idea of using `asic_evt_disable()` as a synchronization routine vs. the interrupts, but I don't think it can work as-is, because that function is not thread-safe.
> 
> You would first need to make it (and `asic_evt_enable()`) thread-safe, using the C11 atomics API to update the registers.

Good call. Currently not all calls to it are actually thread safe so I wasn't considering it, but they're all during startup/shutdown so it just doesn't cause problems.

QuzarDC:
Force-pushed as it had become conflicted from the big whitespace header cleanup.

kernel/arch/dreamcast/hardware/asic.c
```diff
@@ -157,8 +154,8 @@ static void handler_irq9(irq_t source, irq_context_t *context, void *data) {
     /* Go through each event register and look for pending events */
     for(reg = 0; reg < ASIC_EVT_REGS; reg++) {
         /* Read the event mask and clear pending */
-        uint32_t mask = IN32(ASIC_ACK_A + (reg * 0x4));
-        OUT32(ASIC_ACK_A + (reg * 0x4), mask);
+        uint32_t addr = ASIC_ACK_A + (reg * 0x4);
+        uint32_t mask = atomic_fetch_add((atomic_uint *)addr, 0);
```
gyrovorbis, line 158:
```suggestion
        uint32_t mask = atomic_load((atomic_uint *)addr);
```
Was there any reason specifically why you chose to do a `fetch/add` of `0` here, or was it just that you didn't notice the `atomic_load()`? Should be equivalent?

kernel/arch/dreamcast/include/dc/vblank.h
```diff
@@ -31,6 +31,13 @@ __BEGIN_DECLS
     @{
 */
 
+struct vblhnd;
+
+/** \struct  vblhnd_t
+    \brief   Opaque structure describing one vblank handler.
+*/
+typedef struct vblhnd vblhnd_t;
```
gyrovorbis, line 39: Sooooooooooooooooooooooo much nicer as an opaque type-punned pointer to the SLIST entry than being a fucking index... GREAT job.

gyrovorbis (commented):
Just had one teensy little comment about the `atomic_load()` vs `atomic_fetch_add(,0)`. Otherwise this is a fantastic PR that is a cleanup, perf gainz, plus slight memory gainz, plus the opaque pointer type added to the vblank API is much better API design as well.

Big fan of this PR. Also tested against DKR, and everything is fine.

kernel/arch/dreamcast/hardware/asic.c
pcercuei: It's not equivalent though. The `atomic_fetch_add(addr, 0)` will read the 32-bit value then write it again right after.

gyrovorbis: Oh yeah, no shit. Sorry, rusty on atomics at this point. lol.

gyrovorbis (approved):
Given that @pcercuei answered my dumb comment about not using `atomic_load()`, which was my only reservation, I now approve.

pcercuei:
I remember voicing a few concerns about it on Discord some time ago (probably should have done it here). Only disabling a particular IRQ means that the thread disabling it might be preempted, and the IRQ might be re-enabled very late. It sounds like a micro-optimization that can very well backfire.

gyrovorbis:
> I remember voicing a few concerns about it on Discord some time ago (probably should have done it here). Only disabling a particular IRQ means that the thread disabling it might be preempted, and the IRQ might be re-enabled very late. It sounds like a micro-optimization that can very well backfire.

Ah, is your concern that the perf gainz could backfire if it's preempted and less about correctness?

QuzarDC:
> I like the general idea of using `asic_evt_disable()` as a synchronization routine vs. the interrupts

> I remember voicing a few concerns about it on Discord some time ago (probably should have done it here). Only disabling a particular IRQ means that the thread disabling it might be preempted, and the IRQ might be re-enabled very late. It sounds like a micro-optimization that can very well backfire.

Just to be sure I understand, this is an update to your previous assessment, as a general issue that would be faced with `asic_evt_disable()` (and really by an individual irq masking), or is it something specific to it's usage here? In either case it could be then that the way such a thing should work would be to have a pairing where what is disabled is preemption + single IRQ in order to not accidentally skip other important IRQs during the operation but not have the subject IRQ fire nor get preempted.

For this PR, I could just tear out the asic.h changes and flip vblh to use irq_disable again easily enough. Aside from the cleanups the whole point was to get that `asic_evt_disable()` right and if it needs more in the general case, then it can happen in a followup (probably better anyways since it was kind of a big thing to be incidentally combined in here).

QuzarDC:
Rather than going backwards and taking out the functionality figured to just correct it. I believe this approach of simply stopping the timer would address the concerns, thought I'm not sure if there are negative consequences to doing it this way compared to something like having the handler just not trigger the preemption or adding new functionality in the threading API to handle this.

It would probably be good to also have an arch-agnostic method to simply pause preemption for the same kind of purposes (replacing `irq_disables` that are *just* for the purpose of threading) but it didn't seem like that whole layer would be required for this dc-specific use.

pcercuei:
> > I like the general idea of using `asic_evt_disable()` as a synchronization routine vs. the interrupts
> 
> > I remember voicing a few concerns about it on Discord some time ago (probably should have done it here). Only disabling a particular IRQ means that the thread disabling it might be preempted, and the IRQ might be re-enabled very late. It sounds like a micro-optimization that can very well backfire.
> 
> Just to be sure I understand, this is an update to your previous assessment, as a general issue that would be faced with `asic_evt_disable()` (and really by an individual irq masking), or is it something specific to it's usage here? In either case it could be then that the way such a thing should work would be to have a pairing where what is disabled is preemption + single IRQ in order to not accidentally skip other important IRQs during the operation but not have the subject IRQ fire nor get preempted.

It is a general issue with the approach, yes - it would only work (I think) if we also disable preemption, as you said.
 
> For this PR, I could just tear out the asic.h changes and flip vblh to use irq_disable again easily enough. Aside from the cleanups the whole point was to get that `asic_evt_disable()` right and if it needs more in the general case, then it can happen in a followup (probably better anyways since it was kind of a big thing to be incidentally combined in here).

Works for me.

PR #1483, opened 2026-07-13
fs: Correct fs shutdown order and have prevent vfs from leaving orphaned FDs

QuzarDC:
The current ordering has the vfs shut down first, then the upper-level fs system, which is just wrong. In that order, a vfs will clear out all the backing handles for files that are still open, destroy its mutexes, unload it's memory, etc but has no way to inform the upper-level fs about its departure. When the upper-level fs shuts down, its FDs still have their vfs listed and it will attempt to call the `fs_close` on each resulting in issues for any files still open.

Address this by making two adjustments:
First: the upper-level fs will now shut down prior to any individual vfs. This means that each FD will have the functionality of its vfs available to do whatever is needed for `fs_close`. (Subject to review, an alternative behavior would be to clear out the handler and otherwise mark the fd as not having a handler. That would keep the fd # in use until the user calls `fs_close` or the whole fs system is shut down. It would be more complex and require some modification as fs assumes all fds have a vfs, but would have the benefit of preventing access of the wrong file if a new file is opened and takes the fd)

Second: a new function is added to request from a vfs that the upper-level fs close out all its files. With this in place the order of the shutdowns is actually irrelevant, which is what would be expected as the user should be able to init/shutdown individual vfs as needed. This allows for the removal of the custom cleanups of dynamically allocated handle lists in the vfs as they should all have been removed when calling `fs_vfs_shutdown()`.

A quick place to test this is with the cpp/modplug_test example (prior to #1482 ). Without the change, on shutdown it will assert because the mutex for the romdisk vfs will have been closed out during `romdisk_shutdown` but `fs_shutdown` tries to close out the orphaned FD that still points to the romdisk vfs.

I had mentioned this as needing to be done in #1007 and just dropped it from discussion at the time.

The one vfs not updated with this is vmu as its shutdown behavior differs with the caching and does not write out its files on shutdown currently. It will continue to behave as it always has and could perhaps be a subject of a future update.

### DIFF
--- a/addons/libkosext2fs/fs_ext2.c
+++ b/addons/libkosext2fs/fs_ext2.c
@@ -2073,7 +2073,8 @@ int fs_ext2_shutdown(void) {
     while(i) {
         next = LIST_NEXT(i, entry);
 
-        /* XXXX: We should probably do something with open files... */
+        /* Request open files be closed */
+        fs_vfs_shutdown(i->vfsh);
         nmmgr_handler_remove(&i->vfsh->nmmgr);
         ext2_fs_shutdown(i->fs);
         free(i->vfsh);
--- a/addons/libkosfat/fs_fat.c
+++ b/addons/libkosfat/fs_fat.c
@@ -1501,7 +1501,8 @@ int fs_fat_shutdown(void) {
     while(i) {
         next = LIST_NEXT(i, entry);
 
-        /* XXXX: We should probably do something with open files... */
+        /* Request open files be closed */
+        fs_vfs_shutdown(i->vfsh);
         nmmgr_handler_remove(&i->vfsh->nmmgr);
         fat_fs_shutdown(i->fs);
         free(i->vfsh);
--- a/include/kos/fs.h
+++ b/include/kos/fs.h
@@ -663,10 +663,20 @@ file_t fs_open_handle(vfs_handler_t *vfs, void *hnd);
     code, as it is meant for use internally.
 
     \param  fd              The file descriptor to retrieve the handler for.
-    
+
     \return                 The VFS' handler structure.
 */
 vfs_handler_t *fs_get_handler(file_t fd);
+
+/** \brief   Close all FDs for a VFS Handler.
+
+    This function goes through all open fds and closes any that use the given
+    VFS Handler. This should be called during the shutdown of a vfs to ensure
+    the top-level fs is aware that the handler is gone.
+
+    \param  vfs             The VFS that is being shut down.
+*/
+void fs_vfs_shutdown(vfs_handler_t *vfs);
 
 /** \brief   Retrieve the internal handle for a file descriptor.
 
--- a/include/kos/fs.h
+++ b/include/kos/fs.h
@@ -665,6 +665,16 @@ file_t fs_open_handle(vfs_handler_t *vfs, void *hnd);
 */
 vfs_handler_t *fs_get_handler(file_t fd);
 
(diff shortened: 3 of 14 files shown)


andressbarajas:
This may be the old PR you are looking for: https://github.com/KallistiOS/KallistiOS/pull/1007

QuzarDC:
> This may be the old PR you are looking for: #1007

Yea, I think that may have been it, thank you. With the further commits I added the top-level and subvfs can now be shut down in any order as would be expected (as they could be shutdown individually by the user).

I'm thinking some more still on the architecture and perhaps others will have more feedback. Additionally there's a gap in implementing this for the vmu vfs as that one has different closing behavior during shutdown. If a file is closed regularly, the vfs attempts to write it out if it hasn't yet been written to the vmu. On shutdown however, it just discards data written to unclosed files.

QuzarDC:
Force pushed to resolve a conflict with the (now) deleted fs_dclsocket as well as to update the addon vfs with the new function.

pcercuei:
So I think the idea is OK overall, however the implementation is too naive.

Right now your `fs_vfs_shutdown()` does race against pretty much everything.

- Between your fd_table check and your fs_close(), the table entry may have changed to something else. I think that for each table entry, you should read it once, check that the handle matches, and if it does, do an atomic-compare-and-swap to NULL. If the atomic compare succeeded, unref the handle.

- Nothing prevents another thread from opening new files while you're shutting down the VFS. There should be a locking mechanism in the top-level FS code between fs_hnd_open() and the fs_vfs_shutdown() + the various nmmgr_handler_remove() calls.

QuzarDC:
> So I think the idea is OK overall, however the implementation is too naive.
> 
>     * Between your fd_table check and your fs_close(), the table entry may have changed to something else. I think that for each table entry, you should read it once, check that the handle matches, and if it does, do an atomic-compare-and-swap to NULL. If the atomic compare succeeded, unref the handle.

Ok, that makes sense. Especially for low-index FDs it could be that a different file got opened into that slot.

>     * Nothing prevents another thread from opening new files while you're shutting down the VFS. There should be a locking mechanism in the top-level FS code between fs_hnd_open() and the fs_vfs_shutdown() + the various nmmgr_handler_remove() calls.

Hrmm, with this consideration it seems like it should be that `nmmgr_handler_remove()` is the first thing called in any of the vfs' shutdowns rather than the last. That would narrow the gap, and just make more sense (after all that doesn't *destroy* the handler, it merely unregisters it from searches).

Something would have to be done for in-progress opens though, and it can't just be at `fs_hnd_open()` but up in `fs_open()` because the former could be completed but without `fs_hnd_assign()` being called yet and you'd still end up with an orphaned file.

> Right now your `fs_vfs_shutdown()` does race against pretty much everything.
To be sure, the first one is entirely in my code, but the other is preexisting ;)

There's one other set of races that I'd been thinking on for a while in these regarding how the `mutex_destroy()`s are being used. Right now shutting down while a vfs operation is happening but before it has locked the internal mutex could lead to the mutex being destroyed out from under it, and an assert (or bad UB). The vfs handlers are almost the only place where we're destroying mutexes in this way, and it should probably be removed but that would still leave the gap of a vfs op that was in progress either hanging on the mutex being perma-locked or going into UB. The best solution I'd imagined was having an 'initted' check after taking the vfs internal mutexes and bailing. Seems like a *lot* to change, especially since any number of the handlers attempt to do as much as possible before taking the mutex.

PR #1484, opened 2026-07-14
kos-chain: Add PS2/IOP/DVP toolchain support

andressbarajas:
Add toolchains for the three PlayStation 2 processors:

- playstation2: mips64r5900el-ps2-elf for the Emotion Engine.
- iop: mipsel-elf for the I/O Processor.
- dvp: dvp-elf, a binutils-only assembler/linker for the VUs.

### DIFF
--- a/utils/kos-chain/.gitignore
+++ b/utils/kos-chain/.gitignore
@@ -1,10 +1,12 @@
 logs
 *.tar.*
-binutils-*
-gcc-*
-newlib-*
-gdb-*
-build-*
+# Anchored to this directory: these ignore the unpacked source trees and build
+# dirs, and must not swallow the tracked patches under patches/targets/.
+/binutils-*
+/gcc-*
+/newlib-*
+/gdb-*
+/build-*
 *.stamp
 config.guess
 config.sub
--- a/utils/kos-chain/Makefile.aica.cfg
+++ b/utils/kos-chain/Makefile.aica.cfg
@@ -5,9 +5,12 @@
 #########################
 
 # One of the supported platforms:
-# - dreamcast:    Build a toolchain for the SEGA Dreamcast
-# - aica:         Build a toolchain for the Dreamcast's AICA co-processor
-# - gamecube:     Build a toolchain for the Nintendo GameCube
+# - dreamcast:      Build a toolchain for the SEGA Dreamcast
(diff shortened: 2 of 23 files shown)


QuzarDC:
A first quick pass:
- Where did the newly added files for the dvp come from? `dvp-opc.c`, `dvp-dis.c`, `tc-dvp.c` etc. Were they in an older version of binutils and dropped or something?
- It looks like the diff for `newlib-4.6.0.20260123-kos/newlib/Makefile.in` is bigger than probably needed, generating hundreds of lines where it seems like it's just changing the 'append' value but without having inserted a new thing? There are other places in the newlib diff like for assert as well where it's just changing the timestamps. 
- This should be split up into some different commits. At the very least just the changes unrelated to PS2 like the host patch for macos, should be separate. At best possibly each of the three new targets as their own.
- Are there some examples or trivial pieces to test with this? I have a PS2 and a PS2 slim set up with freemcboot, so I can run current general homebrew.

andressbarajas:
@QuzarDC Im working on redoing this PR so Im converting it to draft right now. If you want to play with this version of the PS2 toolchain you can use it to compile ps2-load-ip and then after that the ps2 examples in this repo: https://github.com/andressbarajas/kos-tool. It has no visual examples, besides ps2-load-ip, but its something. After installing the toolchain and cloning the kos-tool repo:

cd kos-tool
make
Grab ps2-load-ip.elf from kos-tool/build directory. Put it on USB drive and load it through freemcboot. Then you can use kos-tool to upload the examples (kos-tool/build/examples/ps2): kos-tool -t dhcp -x syscall-test.elf

andressbarajas:
@QuzarDC I should have an update for this PR by this weekend. Did you have a chance to play with kos-tool PS2 examples?

EDIT: Gonna need another couple of days. 100% I'll update by next weekend(Aug 16th) if not sooner.

andressbarajas:
@QuzarDC The new files for binutils were from a patched version that never got official merged into binutils. This patched version was floating around the internet and I believe was used for official PS2 linux. Just so there is no issue, I created a new dvp assembler patch that was done through cleanroom methods. I thoroughly tested it for 2 weeks with other VU related programs I have been working on vs the original version I had PR'd. That also seemed to clear the newlib diff too. I also split it into multiple commits like you asked and force pushed.

The whitespace/tab failures are inside the *.diff files and can be ignored.

PR #1485, opened 2026-07-14
examples: Fix non-building examples, overall warnings reduction, and loadable cleanup.

QuzarDC:
A collection of changes for the examples, with some going a few steps deeper:
- Broken GCC 9.5 examples. These didn't have anything necessary that was preventing them from being built in GCC 9.5 (and so were never marked with a min GCC version) but had either anonymous parameters or issues with labels that weren't supported back then and have been updated to work. When we drop support we can go ahead and clean up all the void casts at once.
- The SDL sound example was not building due to a new default dependence on libGL upstream.
- Use `pvr_scene_begin_rtt()` and `strftime()` rather than the deprecated `pvr_scene_begin_txr()` and `ctime()`.
- Clean up unnecessarily mismatched sign comparison warnings.
- Clean up the loadable example and prefab makefile
  - Provide explicit start address during test first run, preventing a warning about the symbol not being found.
  - Apply `-fno-lto` in the prefab rather than the example (which as added in #1127 to prevent build failure). It's expected this be the case for any use and not unique to the example.
  - Use standard kos cflags rather than an explicitly enumerated set of our defaults.

### DIFF
--- a/examples/dreamcast/basic/exec/exec.c
+++ b/examples/dreamcast/basic/exec/exec.c
@@ -9,8 +9,7 @@
 
 int main(int argc, char **argv) {
     file_t f;
-    size_t s;
-    ssize_t rv;
+    ssize_t s, rv;
     void *subbin;
 
     /* Print a hello */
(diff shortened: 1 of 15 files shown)


QuzarDC:
@DC-SWAT I'd appreciate your feedback on the parts related to the loadable Makefile.prefag. Especially for the use of the standard CFLAGS I wanted to be sure there wasn't some specific reason to set it with the flags you did.

pcercuei (approved):

PR #1486, opened 2026-07-16
Misc cleanups

pcercuei:
- Cleanup overly complex inline ASM in IRQ code
- Correct doxygen documentation for mutex API
- Rework assert() function to avoid hiding the body
- Use the `net_crc16ccitt()` function in the flashrom / vmu_pkg drivers to factorize code and use the faster algorithm.


andressbarajas:
Looks good to me.

QuzarDC (approved):

PR #1487, opened 2026-07-19
Header cleanups

QuzarDC:
A newer version of curl was yelling at me because they have `-Wtrailing-whitespace` set. While nitpicky, we do already try to avoid this so figured I'd go ahead and do a pass clearing them out.

Additionally add a missing include of `<sys/cdefs.h>`


pcercuei (approved):
That's great!

PR #1488, opened 2026-07-19
examples: Add alternative header locations for kos-ports includes.

QuzarDC:
In the new kos-ports scheme (ref KallistiOS/kos-ports#165 ) headers will be available in their intended top-level locations rather than the traditional portname subfolder that had. This set of changes tests if the headers exist in their old location and if not uses the new (more correct) locations.

This should reduce issues related to the kos-ports update rather than just changing the headers by not requiring as many changes to be hard-cutovers.

### DIFF
--- a/examples/dreamcast/sound/hello-opus/opustest.c
+++ b/examples/dreamcast/sound/hello-opus/opustest.c
@@ -17,7 +17,12 @@
 #include <kos/init.h>
 #include <kos/thread.h>
 
+/* Allow both old kos-ports location and standard */
+#if __has_include("opusplay/opusplay.h")
 #include <opusplay/opusplay.h>
+#else
+#include <opus/opusplay.h>
+#endif
 
 int main(int argc, char **argv) {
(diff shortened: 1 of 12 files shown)


examples/dreamcast/sound/hello-opus/opustest.c
```diff
@@ -17,7 +17,12 @@
 #include <kos/init.h>
 #include <kos/thread.h>
 
+/* Allow both old kos-ports location and standard */
+#if __has_include("opusplay/opusplay.")
```
pcercuei: + `h`

QuzarDC: Fixed it but I for sure have to give it a run through both current and new scheme to test again and make sure the paths are right (or if you have that set up to check, that too). I had done that originally, but clearly I must have missed clearing out the sysroot when switching back to test the current standard or this would have failed.

pcercuei (approved):

PR #1490, opened 2026-07-25
kos-chain: Fix parallel libgcc build race in SH GCC patches

andressbarajas:
On macOS it is easy to get compile errors when trying to build any GCC version when makejobs is not 1. 

The reason is the SH-specific unwind-dw2-Os-4-200.o rule in libgcc/config/sh/t-sh compiles unwind-dw2.c, which #includes the generated libgcc_tm.h, but does not list it as a prerequisite. Under a parallel build (-j) the object can be compiled before mkheader.sh has generated the header, failing with "libgcc_tm.h: No such file or directory".

Add libgcc_tm.h as a prerequisite to that rule, matching the dependency the generic libgcc Makefile.in already applies to every other object. Applied to the GCC 13.2.0 through 17.0.0 target patches.

### DIFF
--- a/utils/kos-chain/patches/targets/gcc-13.2.0-kos.diff
+++ b/utils/kos-chain/patches/targets/gcc-13.2.0-kos.diff
@@ -145,3 +145,15 @@ diff --color -ruN gcc-13.2.0/libstdc++-v3/configure gcc-13.2.0-kos/libstdc++-v3/
      mcf)	thread_header=config/i386/gthr-mcf.h ;;
  esac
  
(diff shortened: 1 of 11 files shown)


QuzarDC:
Just letting the CI rerun and will merge. In case you hadn't considered it this would be a good kind of thing to try upstreaming since it's such a clear bug. I've never looked into gcc's process for that but there's multiple users in the discord dc-toolchain channel who might be able to do so easily.

QuzarDC (approved):

andressbarajas:
> Just letting the CI rerun and will merge. In case you hadn't considered it this would be a good kind of thing to try upstreaming since it's such a clear bug. I've never looked into gcc's process for that but there's multiple users in the discord dc-toolchain channel who might be able to do so easily.

Great Idea. I reached out.

PR #1491, opened 2026-07-25
Fix inline assembly for IRQ and dcache operations

DC-SWAT:
- Use +m for ocbp/ocbwb so GCC cannot drop stores before write-back, and mark the line asms volatile so they survive when the CPU never reads the buffer again.
- Add a memory clobber to stc/ldc SR so GCC cannot move memory accesses out of a critical section around irq_disable()/irq_restore().

### DIFF
--- a/kernel/arch/dreamcast/include/arch/cache.h
+++ b/kernel/arch/dreamcast/include/arch/cache.h
@@ -2,7 +2,7 @@
 
    arch/dreamcast/include/arch/cache.h
    Copyright (C) 2001 Megan Potter
-   Copyright (C) 2014, 2016, 2023 Ruslan Rostovtsev
+   Copyright (C) 2014, 2016, 2023, 2026 Ruslan Rostovtsev
    Copyright (C) 2023 Andy Barajas
    Copyright (C) 2025 Eric Fradella
    Copyright (C) 2026 Falco Girgis
@@ -67,7 +67,7 @@ static inline void arch_dcache_pref_line(const void *src) {
 static inline void arch_dcache_alloc_line_with_value(void *src, uintptr_t value) {
     uintptr_t *ptr = (uintptr_t *)src;
 
-    __asm__ ("movca.l r0, @%8\n\t"
+    __asm__ __volatile__("movca.l r0, @%8\n\t"
              : "=m"(ptr[0]),
                "=m"(ptr[1]),
                "=m"(ptr[2]),
@@ -83,7 +83,7 @@ static inline void arch_dcache_alloc_line_with_value(void *src, uintptr_t value)
 static inline void arch_dcache_alloc_line(void *src) {
     uintptr_t *ptr = (uintptr_t *)src;
 
-    __asm__ ("movca.l r0, @%8\n\t"
+    __asm__ __volatile__("movca.l r0, @%8\n\t"
              : "=m"(ptr[0]),
                "=m"(ptr[1]),
                "=m"(ptr[2]),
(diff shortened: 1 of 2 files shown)


gyrovorbis (dismissed):
OH MY GOD.

I think @DC-SWAT is correct here.

With only output variable clobbers, GCC is free to reorder instructions here, changing when the cache lines get impacted, and can even COMPLETELY OPTIMIZE AWAY these entire operations if it sees you aren't even accessing the pointer at the address you've passed to these routines. 

`volatile` fixes the instruction reordering and optimization potential issues.

I'm not so sure about `"memory"`, though. If that's ACTUALLY needed or if that's too heavy-handed.

gyrovorbis:
OH. I think this also needs to be done for `arch_dcache_alloc_line()`, which will have the same issue with `MOVCA.L`, if @pcercuei agrees that these are warranted.

pcercuei:
> At least it fixes G1ATA DMA data buffer invalidation.

What's the bug?

DC-SWAT:
> > At least it fixes G1ATA DMA data buffer invalidation.
> 
> What's the bug?

Random heap corruption. Also this changes is not enough, I completely rollback cache changes to old asm version to solve crashes. Look in contributors chat.

pcercuei:
Could you maybe provide the disasm of a function where one of these is called, to show the before/after?
(Or I'll test this when I'm home)

Also I think that if you add the memory clobbers you can just drop the output constraints, which become redundant.

pcercuei:
> With only output variable clobbers, GCC is free to reorder instructions here, changing when the cache lines get impacted, and can even COMPLETELY OPTIMIZE AWAY these entire operations if it sees you aren't even accessing the pointer at the address you've passed to these routines.

Yes, GCC is free to reorder instructions, and that's by design. It will still respect the I/O ordering as dictated by the constraints, and it won't completely optimize away unless you do something stupid like alloc a cache line on a stack address before returning from the function.

@DC-SWAT can you try this? https://github.com/pcercuei/KallistiOS/commit/64308da5633175a8eb6d5800bc88b476499d5c75

gyrovorbis:
> > With only output variable clobbers, GCC is free to reorder instructions here, changing when the cache lines get impacted, and can even COMPLETELY OPTIMIZE AWAY these entire operations if it sees you aren't even accessing the pointer at the address you've passed to these routines.
> 
> Yes, GCC is free to reorder instructions, and that's by design. It will still respect the I/O ordering as dictated by the constraints, and it won't completely optimize away unless you do something stupid like alloc a cache line on a stack address before returning from the function.
> 
> @DC-SWAT can you try this? [pcercuei@64308da](https://github.com/pcercuei/KallistiOS/commit/64308da5633175a8eb6d5800bc88b476499d5c75)

I don't see how GCC could possibly know not to optimize away something it only sees being written to and never read from... Isn't that how we're doing the `dcache_alloc_with_buffer()` now, or whatever? Don't we just have a static buffer used to alias the cache lines, which is a write-only taget?

DC-SWAT:
> > With only output variable clobbers, GCC is free to reorder instructions here, changing when the cache lines get impacted, and can even COMPLETELY OPTIMIZE AWAY these entire operations if it sees you aren't even accessing the pointer at the address you've passed to these routines.
> 
> Yes, GCC is free to reorder instructions, and that's by design. It will still respect the I/O ordering as dictated by the constraints, and it won't completely optimize away unless you do something stupid like alloc a cache line on a stack address before returning from the function.
> 
> @DC-SWAT can you try this? [pcercuei@64308da](https://github.com/pcercuei/KallistiOS/commit/64308da5633175a8eb6d5800bc88b476499d5c75)

Yeah, that helped overall, but I still need add `volatile` to it and `memory` to `IRQ` inline asm. I’ve pushed all changes, everything is working fine for me with it.

QuzarDC (approved):

PR #1495, opened 2026-07-31
Misc pvr updates

QuzarDC:
A collection of small updates based on reports from users in the simulant discord:
- Enable vertex fog color setting. This has been part of the API forever but always set to assert. Repeated reports from users to one another are 'just remove the assert and it works fine'. So doing that and cleaning it up as the alpha value isn't used. This changes the API, but considering the original function would do nothing but assert that seems reasonable.
- Correct `pvr_txr_load_kimg` dma blocking. It was inverted, as reported by @DC-SWAT .
- Add `PVR_PT_ALPHA_REF` . As reported by @jnmartin84 
- Make `pvr_init_defaults()` params public. Since this was private and not well documented, users would have to look into `pvr_init_shutdown.c` in order to see what the defaults were. By moving it to `pvr.h` now they're documented publicly as well as being available to do something like: `memcpy(&my_params, &pvr_default_params, sizeof(my_params)); myparams.fsaa_enabled = 1; pvr_init(&my_params);` where only a single setting is changed and the rest are ensured to be default.
- Minor cleanups in pvr_init_shutdown. Dropped some commented out debug code, some pointless init, de-magic'd some values, and cleaned up some comments.

### DIFF
--- a/kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
+++ b/kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
@@ -211,7 +211,7 @@ void pvr_txr_load_kimg(const kos_img_t *img, pvr_ptr_t dst, uint32_t flags) {
             if(flags & PVR_TXRLOAD_DMA) {
                 sem_wait((semaphore_t *)&pvr_state.dma_lock);
                 pvr_txr_load_dma(img->data, dst, img->byte_count,
-                                 (flags & PVR_TXRLOAD_NONBLOCK) ? 1 : 0, NULL, 0);
+                                 !(flags & PVR_TXRLOAD_NONBLOCK), NULL, 0);
                 sem_signal((semaphore_t *)&pvr_state.dma_lock);
             }
             else if(flags & PVR_TXRLOAD_SQ) {
--- a/kernel/arch/dreamcast/include/dc/pvr.h
+++ b/kernel/arch/dreamcast/include/dc/pvr.h
@@ -659,6 +659,7 @@ Striplength set to 2 */
     You essentially fill one of these in, and pass it to pvr_init().
 
     \headerfile dc/pvr.h
+    \sa pvr_default_params
 */
 typedef struct {
     /** \brief  Bin sizes.
@@ -714,6 +715,19 @@ typedef struct {
 
 } pvr_init_params_t;
 
+/** \brief   PVR initialization structure defaults
+    \ingroup pvr_init
+
+    These are the default values for pvr_init_params_t used by pvr_init_defaults().
+
+    \sa pvr_init_defaults()
+*/
+static const pvr_init_params_t pvr_default_params = {
+    .opb_sizes = { PVR_BINSIZE_16, PVR_BINSIZE_0, PVR_BINSIZE_16, PVR_BINSIZE_0, PVR_BINSIZE_0 },
+    .vertex_buf_size = 512 * 1024,
+    .opb_overflow_count = 3
+};
+
 /** \brief   Initialize the PVR chip to ready status.
     \ingroup pvr_init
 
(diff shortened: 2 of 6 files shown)


andressbarajas:
Looks good to me!

kernel/arch/dreamcast/include/dc/pvr.h
```diff
@@ -714,6 +715,19 @@ typedef struct {
...
 } pvr_init_params_t;
 
+/** \brief   PVR initialization structure defaults
+    \ingroup pvr_init
+
+    These are the default values for pvr_init_params_t used by pvr_init_defaults().
+
+    \sa pvr_init_defaults()
+*/
+static const pvr_init_params_t pvr_default_params = {
```
andressbarajas, line 725: Actually, maybe this can be moved to pvr_init_shutdown.c and externed here

QuzarDC: But that would hide the contents of it, which is meant to be one of the benefits of adding it to the header. That way someone referring to the dox can see what the settings are rather than the way it was before where it was hidden in pvr_init_shutdown.c. Would there be some other benefit to making it an extern that outweighs that? The values could be copied into the dox themselves if so to maintain that visibility.

andressbarajas: Because it's defined (not just declared) in a public header, every source file that includes the header and actually uses pvr_default_params ends up with its own private copy.

QuzarDC: Right, but why is that a negative that outweighs the documentation? The extra memory required? Cause I could see how that would be a problem if it were something that would be used a lot in different files, but I wouldn't expect this to be used more than once in most cases.

If it really is a problem then perhaps an alternative would be to have dox on it inside the .c and just have it grouped with the rest that are in the .h, but I'm not sure if that'd make the docs messier.

andressbarajas: No, it doesn't out weight the documentation. What you have is good.

gyrovorbis: This is where C23's `contstexpr` would kick ass... lol.

gyrovorbis: Well, actually... you would have to actually *use it* within the translation unit to get a duplicate copy, so it's not that big of a deal. GCC isn't going to just emit a symbol with static internal linkage if it's not actually used.

Worst case, it could be potentially used once within KOS and once outside, MAYBE if the user wanted to manually memcpy() from it and override portions of it or something...

QuzarDC: > No, it doesn't out weight the documentation. What you have is good.

Alrighty, I was worried there might have been some sorts of other side effects I was missing.

> Well, actually... you would have to actually _use it_ within the translation unit to get a duplicate copy, so it's not that big of a deal. GCC isn't going to just emit a symbol with static internal linkage if it's not actually used.
> 
> Worst case, it could be potentially used once within KOS and once outside, MAYBE if the user wanted to manually memcpy() from it and override portions of it or something...

If the user's using it manually they likely wouldn't also call `pvr_init_defaults()` which uses it internally though ;)

gyrovorbis: ....GOOD POINT. Even better.

kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
```diff
@@ -211,7 +211,7 @@ void pvr_txr_load_kimg(const kos_img_t *img, pvr_ptr_t dst, uint32_t flags) {
             if(flags & PVR_TXRLOAD_DMA) {
                 sem_wait((semaphore_t *)&pvr_state.dma_lock);
                 pvr_txr_load_dma(img->data, dst, img->byte_count,
-                                 (flags & PVR_TXRLOAD_NONBLOCK) ? 1 : 0, NULL, 0);
+                                 !(flags & PVR_TXRLOAD_NONBLOCK), NULL, 0);
```
gyrovorbis, line 214: EX-FUCKING-SCUUUUSE ME?

LOL. HOW LONG HAS THAT BEEN INVERTED????

gyrovorbis (approved):
Everything looks good here.

kernel/arch/dreamcast/hardware/pvr/pvr_texture.c
QuzarDC: Just goes to show how many people use kimg I guess.

PR #1496, opened 2026-08-03
[v2.2.x backport] kos-chain: Force C++14 to build GCC

pcercuei:
Newer host compiler default to C++23, which cannot compile some versions of GCC properly. Force C++14 to fix that.

Note that the -std= argument is passed to the CC/CXX variables. This means that CFLAGS/CXXFLAGS are not overriden, which caused problems when building libcody.


QuzarDC (approved):

QuzarDC:
For tracking this is #1479 .

PR #1497, opened 2026-08-06
dc: Report call sites in Dreamcast stack traces

andressbarajas:
SH-4 call instructions save the address following the delay slot as the return address. Subtract four bytes so traces point to the call instruction.

@pcercuei Here you go.


pcercuei (approved):
Awesome, thanks!

PR #1499, opened 2026-08-07
libc: Implement setpriority()

pcercuei:
Implement setpriority() by changing the priority to KOS' priority scale, and applying it to the thread ID passed as argument, or to the current thread if the thread ID is 0.

Note that while setpriority() was designed more than 30 years ago to change the priority of processes vs. one another, in modern OSes each thread has its unique PID, and setpriority() can then be used to change the priority of individual threads.

### DIFF
--- a/kernel/libc/posix/resource.c
+++ b/kernel/libc/posix/resource.c
@@ -5,29 +5,59 @@
 */
 
 #include <errno.h>
+#include <stdint.h>
 #include <kos/thread.h>
 #include <kos/timer.h>
 #include <sys/resource.h>
 
-static int priorities[PRIO_USER + 1] = { 0 };
-
 int getpriority(int which, id_t who) {
-    if(who || (which > PRIO_USER) || (which < PRIO_PROCESS)) {
+    kthread_t *thd;
+
+    if(which > PRIO_USER || which < PRIO_PROCESS) {
         errno = EINVAL;
         return -1;
     }
 
-    return priorities[which];
+    if(!who)
+        thd = thd_get_current();
+    else
+        thd = thd_by_tid((tid_t)who);
+    if(!thd) {
+        errno = ESRCH;
+        return -1;
+    }
+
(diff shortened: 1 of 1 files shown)


pcercuei:
Now that I think of it - it might be a good idea to make sure that a priority of 0 in the POSIX scale `[-20:19]` corresponds to PRIO_DEFAULT.

pcercuei:
Updated with a better scaling algorithm. POSIX' 0 now maps to KOS' PRIO_DEFAULT.

QuzarDC:
So, I think it's a good idea but it does map process and thread in a way that we don't do elsewhere in the codebase and, more importantly, elsewhere in our POSIX implementation. In the rest we describe KOS as having a single process `KOS_PID` that has multiple threads rather than as each thread being its own 'process'. I think if it were to change here it should be for all cases. I don't think there'd be anything that would really hinder that or any complex situations where the distinction would really matter fortunately, just a few extra files that would need updating.

As it stands now though, just updating this function would mean that `setpriority(PRIO_USER, getpid(), 5);` wouldn't change the current thread's priority for instance (unless coincidentally on thread with `tid==KOS_PID`)

gyrovorbis (approved):
Okay... I'm going to approve this despite agreeing with what @QuzarDC is saying--that we're being inconsistent here with how KOS treats threads vs processes--ONLY because I found this:

Turns out Linux does the exact same thing... and treats the `who` argument as a `tid`, and associates these priority values with individual threads, rather than with the entire process.

It also looks like the example given by @QuzarDC, `setpriority(PRIO_USER, getpid(), 5);`, would also have the same issue on Linux, with the routine expecting a thread ID and not actually updating properly with the process ID as the thread ID.

This is what it looks like Linux expects: `setpriority(PRIO_PROCESS, 0, new_priority)`.

So basically I'm okay with the inconsistency here specifically, because it's actually totally in-line with Linux + pthreads... I am also fine if @QuzarDC disagrees, though. :)

PR #1500, opened 2026-08-09
    browse: Correct screen blanking on open fails.

QuzarDC:
Don't override the content count if opening the new dir fails.
Without this, on the next redraw, the content count may have stayed
overwritten with 0 from the failure and blank the screen.

Additionally cleaned up an include.


andressbarajas:
Looks good!

pcercuei (approved):

PR #1504, opened 2026-08-19
kos-chain: Add XBox toolchain support

andressbarajas:
Add a toolchain for the original Xbox:
- xbox: i686-pc-xbox for building Xbox programs.


pcercuei:
I'm all for supporting more platforms, but I'd kinda like seeing a bit of KallistiOS running on the Xbox before ACKing that.

andressbarajas:
> I'm all for supporting more platforms, but I'd kinda like seeing a bit of KallistiOS running on the Xbox before ACKing that.

I sent you a DM of working xbox examples. @cypressru  ported/created these examples from Dreamcast.

PR #1505, opened 2026-08-20
workqueue: Race fix

andressbarajas:
@QuzarDC reported a hang on shutdown with dcload-ip + net. Tracking it down turned up these race conditions; fixing them clears the hang.

**Only unlink a cancelled job when it is queued** — the worker dequeues a job
before running it, so an in-flight job is no longer on `wq->jobs`, and
`STAILQ_REMOVE()` of a non-member walks off the end of the list. Now checks
membership first.

**Release blocked cancellers when the thread exits** — during shutdown the
thread could exit without clearing `curr_job` or signalling, leaving a caller
blocked forever on a condvar that `workqueue_destroy()` then frees. The thread
now broadcasts on its way out.

### DIFF
--- a/kernel/thread/workqueue.c
+++ b/kernel/thread/workqueue.c
@@ -115,9 +115,18 @@ void workqueue_enqueue(workqueue_t *wq, workqueue_job_t *job) {
 }
 
 void workqueue_cancel(workqueue_t *wq, workqueue_job_t *job) {
+    workqueue_job_t *cur;
+
     mutex_lock_scoped(&wq->lock);
 
-    STAILQ_REMOVE(&wq->jobs, job, workqueue_job, entry);
+    /* Only unlink it if it is actually queued: a job in flight is no longer
+       on the list, and STAILQ_REMOVE() of a non-member walks off the end. */
+    STAILQ_FOREACH(cur, &wq->jobs, entry) {
+        if(cur == job) {
+            STAILQ_REMOVE(&wq->jobs, job, workqueue_job, entry);
+            break;
+        }
+    }
 
     if(wq->curr_job == job) {
         /* The job's callback is being executed. Wait until it's done. */
--- a/kernel/thread/workqueue.c
+++ b/kernel/thread/workqueue.c
@@ -128,8 +128,9 @@ void workqueue_cancel(workqueue_t *wq, workqueue_job_t *job) {
         }
     }
 
-    if(wq->curr_job == job) {
-        /* The job's callback is being executed. Wait until it's done. */
+    if(wq->curr_job == job && !wq->quit) {
+        /* The job's callback is being executed. Wait until it's done, unless
+           we are shutting down: do not block on a queue that is going away. */
         cond_wait(&wq->cond, &wq->lock);
     }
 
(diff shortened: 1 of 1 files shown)


kernel/thread/workqueue.c
```diff
@@ -61,6 +61,13 @@ static void *workqueue_thread(void *d) {
         job->cb(d, job);
     }
 
+    /* Nothing is executing any more; wake up any blocked workqueue_cancel(),
+       as nobody will be left to do it once we exit. */
+    mutex_lock(&wq->lock);
+    wq->curr_job = NULL;
+    cond_broadcast(&wq->cond);
+    mutex_unlock(&wq->lock);
+
```
pcercuei: I think this does not need to be done here. It is done already in workqueue_kill(). However workqueue_kill() does it without holding the lock which is wrong.

andressbarajas: Removed and updated workqueue_kill() to hold the lock. Force pushed.

kernel/thread/workqueue.c
```diff
@@ -115,12 +115,22 @@ void workqueue_enqueue(workqueue_t *wq, workqueue_job_t *job) {
...
     mutex_lock_scoped(&wq->lock);
 
-    STAILQ_REMOVE(&wq->jobs, job, workqueue_job, entry);
+    /* Only unlink it if it is actually queued: a job in flight is no longer
+       on the list, and STAILQ_REMOVE() of a non-member walks off the end. */
+    STAILQ_FOREACH(cur, &wq->jobs, entry) {
+        if(cur == job) {
+            STAILQ_REMOVE(&wq->jobs, job, workqueue_job, entry);
+            break;
```
QuzarDC, line 127: Should it be that either of the followup lines are only needed if there is a job removed from the queue? The `cond_signal(&wq->cond);` specifically notes that it's due to the queue being modified, so perhaps it and the others should be nested here with the remove?

cc: @pcercuei

pcercuei: Probably. I don't think it matters that much.

QuzarDC: Ah, ok, I thought it'd be a noticeable hit to requeue the waiting wq thread just to have it go back to sleep. Though it should be exceedingly rare for the race to show up.

pcercuei: Actually, answering your original question, it would be wrong - a job that is currently being processed (`wq->curr_job` set) is not in the queue anymore, so what you suggested wouldn't have worked. You could do it the other way around, and only do the following lines if no job was removed from the queue.

Also the code here does not "requeue the waiting wq thread just to have it go back to sleep": it's a cond_wait, not a cond_signal.

QuzarDC: > Actually, answering your original question, it would be wrong - a job that is currently being processed (`wq->curr_job` set) is not in the queue anymore, so what you suggested wouldn't have worked. You could do it the other way around, and only do the following lines if no job was removed from the queue.
> 
> Also the code here does not "requeue the waiting wq thread just to have it go back to sleep": it's a cond_wait, not a cond_signal.

`    /* We modified the queue, so notify the thread that it should parse the
     * list once again. */
    cond_signal(&wq->cond);`

pcercuei: Ah, I blame github hiding some lines before the "cond_wait" and the end of the function :)

QuzarDC (approved):

PR #1507, opened 2026-08-22
build: Fix build system issues

andressbarajas:
While working on porting KOS to other consoles there are a few places that can be made more sturdy or better handle multiple archs. These have no behavioral/negative-effects on our current one arch build system. They show up when a library output directory does not exist yet, which is what adding a new arch looks like, when an archive keeps a member nothing builds anymore, and when a second environ.sh is sourced in a shell that already has one.

I plan to do a follow up branch next week. After I experiment with building for multiple archs without having to clean the kernel in between, and that builds on these.

### DIFF
--- a/addons/Makefile.prefab
+++ b/addons/Makefile.prefab
@@ -19,6 +19,7 @@
 defaultall: $(OBJS) subdirs linklib
 
 linklib: $(OBJS) $(LIB_OBJS)
+	@mkdir -p $(KOS_BASE)/addons/lib/$(KOS_ARCH)
 	rm -f $(KOS_BASE)/addons/lib/$(KOS_ARCH)/$(TARGET)
 	$(KOS_AR) rcs $(KOS_BASE)/addons/lib/$(KOS_ARCH)/$(TARGET) $(OBJS) $(LIB_OBJS)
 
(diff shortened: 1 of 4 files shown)


darcagn:
The third commit says
> Nothing else assigns to KOS_INC_PATHS, so its better to assign it than append to it.

But the `environ.sh.sample` implies that it's intended to be configurable:
```Makefile
# Resets build flags. You can also initialize them to some preset
# default values here if you wish.
#
export KOS_INC_PATHS=""
```
and if they aren't configured, they should have been reset to empty via this mechanism. Are you resetting it to empty in your PS2 `environ.sh`?

QuzarDC:
> The third commit says
> 
> > Nothing else assigns to KOS_INC_PATHS, so its better to assign it than append to it.
> 
> But the `environ.sh.sample` implies that it's intended to be configurable:
> 
> ```makefile
> # Resets build flags. You can also initialize them to some preset
> # default values here if you wish.
> #
> export KOS_INC_PATHS=""
> ```
> 
> and if they aren't configured, they should have been reset to empty via this mechanism. Are you resetting it to empty in your PS2 `environ.sh`?

Was gonna note similar. It occurred to me though ... is there a good reason to have this? It would seem there's probably better places to be slipping in extra include paths.

andressbarajas:
> 
> and if they aren't configured, they should have been reset to empty via this mechanism. Are you resetting it to empty in your PS2 `environ.sh`?

Oops, I am not. I'll let you two @darcagn @QuzarDC decide to keep or drop that commit.

PR #1510, opened 2026-08-29
libc: Treat C11 timeouts as absolute deadlines

andressbarajas:
These fixes were initially made by @cypressru. I cleaned them up and made them PR ready.

C11 says these functions take a *deadline* (a point in time), but KOS read it as a *duration*.

A new `c11_timeout_ms()` subtracts the current time to get the duration KOS actually wants. Following the spec:

- [`mtx_timedlock()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/mtx_timedlock.html) tries the lock before checking the clock. A free lock is yours even if the deadline passed — C11 §7.26.4.4 only reports a timeout if you *didn't* get the lock.
- [`cnd_timedwait()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/cnd_timedwait.html) still unlocks and relocks your mutex when the deadline has passed, as §7.26.3.5 requires.

Also, `cond_wait_timed()` relocks your mutex after timing out. If another thread held it, that overwrote `ETIMEDOUT` with `EBUSY`, hiding real timeouts.

### DIFF
--- a/kernel/thread/cond.c
+++ b/kernel/thread/cond.c
@@ -15,6 +15,7 @@
 
 #include <kos/thread.h>
 #include <kos/cond.h>
+#include <kos/errno.h>
 #include <kos/genwait.h>
 
 #include <kos/dbglog.h>
@@ -59,6 +60,9 @@ int cond_wait_timed(condvar_t *cv, mutex_t *m, int timeout) {
 
     if(rv < 0 && errno == EAGAIN)
         errno = ETIMEDOUT;
+
+    /* Caching errno since a contended mutex sets EBUSY. */
+    errno_save_scoped();
 
     /* Re-lock our mutex */
     mutex_lock(m);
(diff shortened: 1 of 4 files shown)


kernel/thread/cond.c
```diff
@@ -60,9 +60,15 @@ int cond_wait_timed(condvar_t *cv, mutex_t *m, int timeout) {
     if(rv < 0 && errno == EAGAIN)
         errno = ETIMEDOUT;
 
+    /* Caching errno since a contended mutex sets EBUSY. */
+    err = (rv < 0) ? errno : 0;
```
QuzarDC: Pretty sure this could be an `errno_save_scoped()` from `kos/errno.h` like we used in pthreads since it shouldn't be a problem to unconditionally save/restore errno.

andressbarajas: Thanks. I'll give it try later today when I'm home.

andressbarajas: Fixed

QuzarDC (approved):

PR #1512, opened 2026-08-30
mm: Avoid stdio recursion when sbrk() reports OOM

andressbarajas:
Another fix made by @cypressru I cleaned up and made PR ready.

mm_sbrk() is called from inside malloc(), so malloc's lock can already be held when we reach the out-of-memory path. The old dbglog() call goes through printf, and printf can itself call malloc. Trying to report the failure could therefore run right back into the allocator that just failed.

Use dbgio_write_str() instead. It writes straight to the debug console without going through stdio, so it is safe to call from here.

The message loses its detail as a result: dbgio_write_str() takes a plain string and does no formatting, so it can no longer report the requested and previous sbrk_base or the size of the failed request. Printing those values would mean using printf.

### DIFF
--- a/kernel/mm/mm.c
+++ b/kernel/mm/mm.c
@@ -15,14 +15,14 @@
 
 #include <arch/arch.h>
 #include <arch/stack.h>
-#include <kos/dbglog.h>
+#include <kos/dbgio.h>
 #include <kos/mm.h>
 #include <kos/linker.h>
 #include <errno.h>
-#include <inttypes.h>
 #include <stdatomic.h>
 #include <stdint.h>
-#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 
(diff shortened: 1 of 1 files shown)


QuzarDC (commented):
Absolutely makes sense. Think though you could check out the debug in malloc ( https://github.com/KallistiOS/KallistiOS/blob/b0eeadec379081d9874734ac81ab80ba39bdd06c/kernel/libc/koslib/malloc.c#L1702-L1709 ) so that we could continue printing out the memory details? (Or spin your own version) I'm pretty sure those details being provided immediately is appreciated by users.

If it seems too much and the existing values aren't needed that's fine too

andressbarajas:
> If it seems too much and the existing values aren't needed that's fine too

I'll look into it and try and have something by tomorrow.

QuzarDC (approved):

PR #1517, opened 2026-09-06
feat(kos-chain): add optional GDB support to Docker image

maishuji:
This change updates the kos-chain Docker image so GDB can be included when requested without changing the default build behavior.

By default, the Dockerfile still builds the standard toolchain with make build. When include_gdb=1 is passed as a Docker build argument, it switches to make all, which includes sh-elf-gdb in the resulting image. The Docker documentation was updated to describe this new build option.

The Dockerfile was also cleaned up by replacing the deprecated MAINTAINER instruction with LABEL and normalizing the FROM ... AS syntax.

### Validation:

Built the Docker image with --build-arg include_gdb=1
Verified sh-elf-gdb is present and runnable in the image
Confirmed reported version: GNU gdb (GDB) 17.1

### DIFF
--- a/utils/kos-chain/docker/Dockerfile
+++ b/utils/kos-chain/docker/Dockerfile
@@ -9,8 +9,8 @@
 #       docker run -it --name containername dcchain:stable /bin/bash
 
 # Stage 1
-FROM alpine:latest as build
-MAINTAINER KallistiOS <cadcdev-kallistios@lists.sourceforge.net>
+FROM alpine:latest AS build
+LABEL maintainer="KallistiOS <cadcdev-kallistios@lists.sourceforge.net>"
 
 # Installing prerequisites
 RUN apk --update add --no-cache \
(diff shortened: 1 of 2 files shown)


pcercuei:
> This change updates the kos-chain Docker image so GDB can be included when requested without changing the default build behavior.
> 
> By default, the Dockerfile still builds the standard toolchain with make build. When include_gdb=1 is passed as a Docker build argument, it switches to make all, which includes sh-elf-gdb in the resulting image. The Docker documentation was updated to describe this new build option.
> 
> The Dockerfile was also cleaned up by replacing the deprecated MAINTAINER instruction with LABEL and normalizing the FROM ... AS syntax.

Can you just copy-paste what I quoted here above into the commit itself?
(`git commit --amend` to edit the text then `git push -f` to force-push the new commit and the PR will be updated automatically)

maishuji:
> > This change updates the kos-chain Docker image so GDB can be included when requested without changing the default build behavior.
> > By default, the Dockerfile still builds the standard toolchain with make build. When include_gdb=1 is passed as a Docker build argument, it switches to make all, which includes sh-elf-gdb in the resulting image. The Docker documentation was updated to describe this new build option.
> > The Dockerfile was also cleaned up by replacing the deprecated MAINTAINER instruction with LABEL and normalizing the FROM ... AS syntax.
> 
> Can you just copy-paste what I quoted here above into the commit itself? (`git commit --amend` to edit the text then `git push -f` to force-push the new commit and the PR will be updated automatically)

added description in the commit message itself

maishuji:
@pcercuei maybe worth to add a CONTRIBUTING.md file in the project root (or else) to describe the need to add the description in each commit and the format of commit ect
Maybe also in an AGENTS.md to provide it directly to aigents

pcercuei (approved):

PR #1518, opened 2026-09-07
libc: Give std::chrono its full 80ns resolution

andressbarajas:
Redo of #1387 by @gyrovorbis, which fixes `std::chrono`'s useless 1-second resolution. I've reworked the approach to address the review feedback there and rebased onto current master.

### DIFF
--- a/examples/dreamcast/cpp/clock/Makefile
+++ b/examples/dreamcast/cpp/clock/Makefile
@@ -1,18 +1,22 @@
 #
 # DC Clock App
 # Based on Peter Hatch's Font Test
-# (c)2002 Megan Potter
-#   
+#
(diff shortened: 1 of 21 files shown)


gyrovorbis (approved):
All changes by @andressbarajas look good to me.

Tested the `clock` C++ example WITHOUT upgrading the toolchain, and everything still works fine, as it always did, just at shitty `1 sec` resolution.

Built toolchain `gcc-17.0.0-exp`, which is the LRA toolchain, and lo' and behold, we have full 80ns tick resolution.

gyrovorbis:
https://github.com/user-attachments/assets/e886c021-d104-4194-b770-edf6a2d7582b

PR #1519, opened 2026-09-07
irq: Init new thread contexts with the kernel thread's FPSCR

andressbarajas:
"Non-kernel threads init the fpu denorm handling bit opposite of what the kernel thread inits it to when created. So they hit exception handler and just skip instruction and pretend like nothing happened if they touch denorms, which isn't good" - jn64


kernel/arch/dreamcast/kernel/irq.c
```diff
@@ -341,6 +348,9 @@ int irq_init(void) {
     /* Make sure interrupts are disabled */
     irq_disable();
 
+    /* Inherit the kernel thread's FP mode for later threads. */
+    irq_fpscr_init = __builtin_sh_get_fpscr() & ~FPSCR_STATUS_BITS;
```
QuzarDC, line 352: Would it be more expected for this to be inherited from the parent thread rather than static tied to the initial kernel thread?

@gyrovorbis @pcercuei - Perhaps you'd have an opinion or idea on that too.

pcercuei: The main thread does get a context too, which is memset'd as well, so I would be enclined to drop the code that sets FPSCR in startup.S, and instead do a `__builtin_sh_set_fpscr(context->fpscr)` in the main thread's creation.

That does not answer your question here though. I do not know. I guess using the parent thread's fpscr setting would work better.

QuzarDC: Falco wasn't sure either. Found in the linux man pages though that their pthreads_create implementation inherits fenv from its parent, which matches your guess and makes sense as something to follow.

@andressbarajas - Think you could switch it to do this read at context creation time?

Regarding the change to remove the setting in startup.S, that would leave a gap during initialization prior to threading being enabled (during `arch_auto_init()` or `__kos_init_early_fn()` notably) where fpscr would just be in an unknown state guided by whatever ran previously which seems like it could cause really hard to track down trouble.

andressbarajas: Sure thing! I'll try and knock it out today or tomorrow

andressbarajas: Sure thing! I'll try and knock it out today or tomorrow

gyrovorbis: Good point on the unset `FPSCR` values during KOS early initialization.... yeah, not ideal...

andressbarajas: @QuzarDC Done

QuzarDC (approved):

PR #1524, opened 2026-09-10
Simplify dma blocking checks

QuzarDC:
Unless I'm missing something the usage of the static bools to track dma blocking are superfluous because you can simply test if the semaphore is waiting before signaling it. Update it in the three locations.

### DIFF
--- a/kernel/arch/dreamcast/hardware/g1ata.c
+++ b/kernel/arch/dreamcast/hardware/g1ata.c
@@ -177,7 +177,6 @@ static uint8_t orig_dev = 0x00;
 
 /* Variables related to DMA. */
 static int dma_in_progress = 0;
-static int dma_blocking = 0;
 static uint8_t dma_cmd = 0;
 static size_t dma_nb_sectors = 0;
 static uint64_t dma_sector = 0;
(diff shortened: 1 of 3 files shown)


kernel/arch/dreamcast/hardware/g1ata.c
```diff
@@ -240,10 +239,9 @@ static void g1_ata_set_sector_and_count(uint64_t sector, size_t count, int lba28
 
 static void g1_dma_done(void) {
     /* Signal the calling thread to continue, if it is blocking. */
-    if(dma_blocking) {
+    if(sem_count(&dma_done)) {
```
pcercuei, line 242: This is trickier than the first commit because dma_common() is outside the critical section. But I think it still works because it's a semaphore and not a mutex+cond, so if the IRQ handler runs first, dma_common() won't wait.

pcercuei (approved):

kernel/arch/dreamcast/hardware/g1ata.c
QuzarDC: I hadn't noticed that race between the transfer start and the sem_wait like that. It seems to me like it would cause it to wait though, since the dma_done wouldn't have signaled the sem, finding none waiting on a signal.

 Something like this should close the hole I suppose, as we can check if the transfer has completed prior to trying to wait:
```
if(block) {
    irq_mask_t tmp = irq_disable();
    if(dma_in_progress)
        sem_wait(&dma_done);
    irq_restore(tmp);`

pcercuei: It would close the hole, but with your previous solution it's not needed.

PR #1525, opened 2026-09-10
Misc threading cleanups and fixes

QuzarDC:
The majority of these changes have to do with removing code that was attempting to enforce strict ordering of the thread queue, which was required prior to #1296 when we switched to always scanning through the full run queue vs the old behavior of maintaining ordered by priority and only searching until a good thread was found to switch to.

Contains the following changes:
- Don't reschedule a mutex holder thread when priority boosting. This was necessary when the queue was strictly ordered, but now is not.
- When adding to runnable, only add to front or end, as the strict ordering does not apply anymore. Front and end is still required in order to maintain virtual 'priority groups' judged by their order in the queue as the scheduler takes the first it finds of a certain priority.
- Refactor of `thd_schedule()` based on the new queue structure. Rather than inserting the current thread before searching the queue, just check if the current thread should be chosen after searching the full queue.
- Clean up `thd_schedule()` to allow debug to be ignored when building with `NDEBUG`
- Update `thd_destroy()` to account for `thd_poll()`'s reuse of the `kthread::wait_obj`. Instead of testing for it existing, test the state to know if the thread is waiting on anything. Also use `FOREACH_SAFE` rather than inventing it from components.
- Drop dead function in genwait, update header to stop exposing documentation for internal function.
- Add static reent struct so that `errno` (and potentially other reent struct internals from newlib) can function correctly before/after threading is initialized/shutdown.
- Clean up usage of noreturn and unreachable, avoiding the use of `abort()` to signal unreachable paths.
- Use atomics to track `tid_highest` in order to shrink the irq disable required for `thd_create_ex()`
- Clean up `thd_create_ex()`: unnest its main body, make the initialization with custom/default attrs simpler, replace param error return with assert.

Marking as draft solely because these haven't been tested much and all have the possibility for significant negative impact. Wanted to get them public early though for others to review.

### DIFF
--- a/include/kos/genwait.h
+++ b/include/kos/genwait.h
@@ -123,27 +123,10 @@ void genwait_wake_one_err(const void *obj, int err);
 */
 int genwait_wake_thd(const void *obj, kthread_t *thd, int err) __nonnull((2));
 
-/** \brief  Look for timed out genwait_wait() calls.
-
-    There should be no reason you need to call this function, it is called
-    internally by the scheduler for you.
-
-    \param  now             The current system time, in milliseconds since boot
-*/
+/** \cond */
+/* Look for timed out genwait_wait() calls */
 void genwait_check_timeouts(uint64_t now);
 
-/** \brief  Look for the next timeout event time.
-
-    This function looks up when the next genwait_wait() call will timeout. This
-    function is for the internal use of the scheduler, and should not be called
-    from user code.
-
-    \return                 The next timeout time in milliseconds since boot, or
-                            0 if there are no pending genwait_wait() calls
-*/
-uint64_t genwait_next_timeout(void);
-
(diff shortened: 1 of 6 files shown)


QuzarDC:
Wow, hadn't tested in gcc9. Seems like (unfortunately) this is an issue with newlib rather than what I wrote. Could possibly be avoided by using the runtime initialization like we already do during thread creation.

image

QuzarDC:
Updated as I described by using the runtime initialization direct ahead of setting `_impure_ptr` rather than the static initializer that triggers a warning.

> Wow, hadn't tested in gcc9. Seems like (unfortunately) this is an issue with newlib rather than what I wrote. Could possibly be avoided by using the runtime initialization like we already do during thread creation.
> image

PR #1526, opened 2026-09-12
maple: read the devinfo response once when attaching

Bruceleeto:
So in Bloom, when I turn off the threaded compiler and load from a save state, I get no controller input at all.
Maple was attaching a device with functions `0` and marking it as valid, so the port looks taken and autodetect never retries it.
TL;DR: `recv_buf` gets read twice, and the DMA might not be done yet. So the function mask the driver matched against isn’t necessarily the same mask that ends up being stored.

this change in kos is what worked for me, but I dont know  maple that well so if anyone got a better solution that's all good and i'd be happy to test it. 

```
KallistiOS v2.3.0 [dreamcast/pristine]
  Git revision: 4fdde8b2-dirty
  Fri Sep 11 10:02:40 AM BST 2026
  bruce@bruce-HP-Z240-Tower-Workstation:/opt/toolchains/dc/kos
  sh-elf-gcc (GCC) 17.0.0 20260703 (experimental)
maple: active drivers:
    Dreameye (Camera): Camera
    Sound Input Peripheral: Microphone
    PuruPuru (Vibration) Pack: JumpPack
    VMU Driver: Clock, LCD, MemoryCard
    Mouse Driver: Mouse
    Keyboard Driver: Keyboard
    Controller Driver: Controller
    Lightgun: LightGun
dc-load console support enabled
maple: attached devices:
  A0:                                (00000000: LightGun)

Found IDE partition 0
Mounted IDE partition 0 to /ide
maple: attached devices:
  A0:                                (00000000: LightGun)    <---- Problem
  A1: Visual Memory                  (0e000000: Clock, LCD, MemoryCard)
RAM mapped
Scratchpad / IO mapped
Parallel port mapped
BIOS mapped
Memory-map succeeded.

```

If you wanna recreate the conditions then I can probably help someone set up the right scenario with bloom.

### DIFF
--- a/kernel/arch/dreamcast/hardware/maple/maple_driver.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_driver.c
@@ -69,18 +69,24 @@ int maple_driver_unreg(maple_driver_t *driver) {
 int maple_driver_attach(maple_frame_t *det) {
     maple_driver_t      *i;
     maple_response_t    *resp;
-    maple_devinfo_t     *devinfo;
+    maple_devinfo_t     devinfo;
     maple_device_t      *dev = maple_state.ports[det->dst_port].units[det->dst_unit];
     bool                dev_allocated = false;
 
     /* Resolve some pointers first */
     resp = (maple_response_t *)det->recv_buf;
-    devinfo = (maple_devinfo_t *)resp->data;
+
+    /* recv_buf is uncached: read the response once, the DMA may still be
+       filling it. A zero mask means the payload has not landed. */
+    devinfo = *(maple_devinfo_t *)resp->data;
+
+    if(!devinfo.functions)
+        return -1;
 
     /* Go through the list and look for a matching driver */
     LIST_FOREACH(i, &maple_state.driver_list, drv_list) {
         /* For now we just pick the first matching driver */
-        if(i->functions & devinfo->functions) {
+        if(i->functions & devinfo.functions) {
 
             /* Driver matches. Alloc a device if needed. */
             if(!dev) {
@@ -95,7 +101,7 @@ int maple_driver_attach(maple_frame_t *det) {
                 dev_allocated = true;
             }
 
-            memcpy(&dev->info, devinfo, sizeof(maple_devinfo_t));
+            dev->info = devinfo;
 
             /* Now lets allocate a new status buffer */
             if(i->status_size && !dev->status) {


pcercuei:
I think this commit is a band-aid that hides a deeper problem.

When maple_driver_attach() is called, the frame is locked so nothing else should be using. The data should be fully available, and not disappear under your feet...

pcercuei:
@Bruceleeto Does it also happen when you have no VMU plugged?

Bruceleeto:
band aid ye.
still a issue with vmu disconnected. yes.  tried moving controller to different ports.  
was kinda hoping someone else would figure it out on the info I gave.. But as i have the repro.. guess i'll look into it more. :cry:

pcercuei:
I can have a look, however I can't reproduce the issue that you're having...

Bruceleeto:
```c 
void maple_frame_init(maple_frame_t *frame) {

    assert(frame->state == MAPLE_FRAME_UNSENT);
    assert(!frame->queued);

    /* Setup the buffer pointer (32-byte align and force -> P2) */
    uint32_t buf_ptr = __align_up((uint32_t)frame->recv_buf_arr, 32);

    if(__is_defined(MAPLE_DMA_DEBUG))
        buf_ptr += 512;

    buf_ptr = (buf_ptr & MEM_AREA_CACHE_MASK) | MEM_AREA_P2_BASE;
    frame->recv_buf = (uint8_t *)buf_ptr;

    /* The buffer is only ever touched through P2, but it lives in BSS, which
       was zeroed through cached addresses. A line still dirty from that gets
       written back over whatever the DMA delivered. */
    dcache_purge_range((uintptr_t)frame->recv_buf_arr,
                       sizeof(frame->recv_buf_arr));

    /* Clear out the receive buffer */
    if(__is_defined(MAPLE_DMA_DEBUG))
        maple_sentinel_setup(frame->recv_buf - 512, 1024 + 1024);
    else
        memset(frame->recv_buf, 0, 1024);

    /* Initialize other state stuff */
    frame->cmd = -1;
    frame->dst_port = frame->dst_unit = 0;
    frame->length = 0;
    frame->queued = 0;
    frame->dev = NULL;
    frame->send_buf = (uint32_t *)frame->recv_buf;
    frame->callback = NULL;
}
```



tldr so the controller's reply lands in is normal memory that got zeroed at boot. That zeroing went through the cache, so the cache is still holding a copy of "this memory is zeros" that hasn't been written out yet.

Then the maple hardware writes the real reply straight into RAM, behind the cache's back. Nothing tells the cache its copy is out of date.
Later the cache flushes that old copy out, and the zeros land on top of the reply. Same address reads fine one moment and zeros the next, with no code touching it.

so throw away the cache's copy of that buffer before the hardware writes into it. like we do elsehwere for anything DMA writes to KOS just never did it for maple.

pcercuei:
What about this then?
```diff
diff --git a/kernel/arch/dreamcast/hardware/maple/maple_driver.c b/kernel/arch/dreamcast/hardware/maple/maple_driver.c
index 244a3590a..9de29de42 100644
--- a/kernel/arch/dreamcast/hardware/maple/maple_driver.c
+++ b/kernel/arch/dreamcast/hardware/maple/maple_driver.c
@@ -7,6 +7,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <dc/maple.h>
+#include <kos/cache.h>
 
 void maple_attach_callback(uint32_t functions, maple_user_callback_t cb, void *user_data) {
     maple_driver_t *i;
@@ -93,6 +94,12 @@ int maple_driver_attach(maple_frame_t *det) {
                 dev->unit = det->dst_unit;
                 dev->frame.state = MAPLE_FRAME_VACANT;
                 dev_allocated = true;
+
+                /* This buffer will be accessed through P2 area - uncached.
+                 * We need to purge anything the cache may have for this
+                 * area. */
+                dcache_purge_range((uintptr_t)dev->frame.recv_buf_arr,
+                                  sizeof(dev->frame.recv_buf_arr));
             }
 
             memcpy(&dev->info, devinfo, sizeof(maple_devinfo_t));

```

Bruceleeto:
I suspect that wont help.. it purges `dev->frame.recv_buf_arr`, but the zero mask comes out of `det->recv_buf` (detect_frame)

anyway i'll test in a few days.. as I already burnt too much time chasing this issue and want to move on. lol.

pcercuei:
> I suspect that wont help.. it purges `dev->frame.recv_buf_arr`, but the zero mask comes out of `det->recv_buf` (detect_frame)

The detect_frame is statically allocated so it's not memset'd. It should not give any problems. The `dev->frame.recv_buf_arr` however is memset'd when the "dev" is allocated since it's calloc'd. The `recv_buf` is just the P2 area (uncached) pointer to the `recv_buf_arr` array.

pcercuei:
Actually I just had that thought that the BSS section is memset'd too, so the detect_frame being statically allocated does not change anything. It should also have its `recv_buf_arr` flushed.

gyrovorbis:
> Actually I just had that thought that the BSS section is memset'd too, so the detect_frame being statically allocated does not change anything. It should also have its `recv_buf_arr` flushed.

Yeah, the BSS is initialized with a `memset(0)`, so the cache lines could be potentially dirty... 

So do you think @Bruceleeto's first proposal is correct? To purge the dcache within `maple_frame_init()`, which should presumably handle all of the frames?

pcercuei:
Doing it on `maple_frame_init()` would be wasteful since this is called every time a frame is reused. It should be done on the frame alloc like in my previous diff. We could have a `maple_frame_alloc()` that takes care of that, and also dynamically alloc the detect frame. Or call `dcache_purge_range` on the detect frame's buffer too.

gyrovorbis:
> Doing it on `maple_frame_init()` would be wasteful since this is called every time a frame is reused. It should be done on the frame alloc like in my previous diff. We could have a `maple_frame_alloc()` that takes care of that, and also dynamically alloc the detect frame. Or call `dcache_purge_range` on the detect frame's buffer too.

Oops. Sorry, I didn't follow the control flow fully. Thought `maple_frame_init()` was just a one-time call, like your `maple_frame_alloc()` suggestion would be...

I would totally be down with heap allocating them...

Wouldn't that essentially be saving some RAM in the long-run, by only allocating the frames that are actually needed lazily as well? 1KB is kinda big too for these buffers... nothing to scoff at, imo.

QuzarDC:
I think I managed to hit the same or similar problem as well:
```
KallistiOS v2.3.0 [dreamcast/pristine]
  Git revision: 15f69a785
  Tue Sep 22 06:13:04 PM EDT 2026
  quzar@Centurion-IV.localdomain:/home/quzar/dcdev/KallistiOS
  sh-elf-gcc (GCC) 16.2.0
maple: active drivers:
    Dreameye (Camera): Camera
    Sound Input Peripheral: Microphone
    PuruPuru (Vibration) Pack: JumpPack
    VMU Driver: Clock, LCD, MemoryCard
    Mouse Driver: Mouse
    Keyboard Driver: Keyboard
    Controller Driver: Controller
    Lightgun: LightGun
dc-load console support enabled
maple: attached devices:
  A0: Dreamcast Controller           (01000000: Controller)
  A1: Visual Memory                  (0e000000: Clock, LCD, MemoryCard)
  A2: Visual Memory                  (0e000000: Clock, LCD, MemoryCard)
  B0: Dreamcast Camera Flash  Device (01000000: Controller)
  B1:                                (00000000: Controller)
  B2: Dreamcast Camera Flash LDevice (00080000: Camera)
  B3: Dreamcast Camera Flash LDevice (00080000: Camera)
  B4: Dreamcast Camera Flash LDevice (00080000: Camera)
  B5: Dreamcast Camera Flash LDevice (00080000: Camera)
```
and eventually 
```
*** ASSERTION FAILURE ***
Assertion "!frame->queued" failed at maple_queue.c:153 in `maple_frame_init'
```
Which might happen when sending commands (testing via MapleTest, it's allinfo commands) or on shutdown when `maple_frame_init()` is called for `maple_dev_reset()`

With the cache purge in either place, I can't retrigger the assertion (though it's not strictly reproducible so it may be coincidence). With it in the frame init, the initial banner print is fixed, unlike when it's in driver attach.

To note, I could not trigger any of this behavior on gcc 13.4.0, only ever see it with 16.2.0. Also, not that it was much a mystery but, the identification of the bad device is due to `maple_pcaps()` being sloppy. It never blanks the static buffer it uses, so on functions == 0 it can return whatever happened to still be in that buffer.

PR #1528, opened 2026-09-13
Ability to get the current SPI interface for the SD card and the W5500 network adapter.

DC-SWAT:
Introduced `sd_get_interface()` and `w5500_adapter_interface()` to check the current  interface.


QuzarDC:
It makes absolute sense to need this kind of functionality once there's multiple devices and multiple interfaces. But since the interfaces are finite and the devices not, would it meet the same needs and be more flexible to instead have this functionality as part of the interface? That way you could also find out if, for instance, `scif` is currently connected for serial or dc-load-serial.

Something like `const char *sci/scif_get_device_name()`

DC-SWAT:
> It makes absolute sense to need this kind of functionality once there's multiple devices and multiple interfaces. But since the interfaces are finite and the devices not, would it meet the same needs and be more flexible to instead have this functionality as part of the interface? That way you could also find out if, for instance, `scif` is currently connected for serial or dc-load-serial.
> 
> Something like `const char *sci/scif_get_device_name()`

However, the interface driver hierarchy lacks feedback from the devices that use it; this requires some refactoring and device registration.

PR #1529, opened 2026-09-13
Fix for uploading big texture using pvr_txr_load

nextgeniuspro:
This PR is fixing incorrect uploading for big textures because there is mistake on how sq_copy function calculate offset 

```C++
        const uint32_t *s = src;
        ...
        else { /* If src is 8-byte aligned, fast path */
            sq_fast_cpy(d, s, nb);
            s += nb * 32;
        }
```
in this piece of code `s` supposed to be incremented to `nb * 32` bytes but because `s` declared as `uint32_t*` real shift will be `nb * 32 * 4`

In the attached archive is a modified version of png kos example app, that loads 1024x1024 ARGB1555 twidled texture, that can be used for testing fix
[big_texture_issue.zip](https://github.com/user-attachments/files/32164726/big_texture_issue.zip)


QuzarDC (approved):
Thank you, good catch. A bad regression from #779 . It seems it went unnoticed so long just because it requires that the data being SQ'd be both 8-byte aligned AND over 1MiB which would be relatively rare.

PR #1533, opened 2026-09-18
Added support for middle button on Jap  mice.

### LINKED ISSUE #1532: Maple mouse driver masks out the middle button (bit 0)
`kernel/arch/dreamcast/hardware/maple/mouse.c` cooks the raw button word with

```c
cooked->buttons = (~raw->buttons) & 14;
```

Bit 0 is dropped, and `dc/maple/mouse.h` has no define for it. But alas, bit 0 is the middle button!

In adding mouse support to the upcoming openMenu release...

I also found Marcus Comstedt's [maple mouse notes](http://mc.pp.se/dc/mouse.html) list the same layout: bit 0 middle (C), bit 1 right (B), bit 2 left (A), bit 3 thumb (START). 

On the Japanese Dreamcast mouse (good old HKT-9900), the wheel click reports bit 0, which KOS currently doesn't see. The western HKT-4200 thumb button reports bit 3 and works, of course.

Looking back, I see the mask started as `& 6` (left and right only). Commit `529bdb61` (2006, "Fix bitmask for mouse buttons so 3rd button is usable") widened it to `& 14` to expose the thumb button. Bit 0 has never been passed through, though.

All that said, the fix is obviously pretty simple...

```diff
--- a/kernel/arch/dreamcast/include/dc/maple/mouse.h
+++ b/kernel/arch/dreamcast/include/dc/maple/mouse.h
@@ -38,6 +38,7 @@
     @{
 */
+#define MOUSE_MIDDLEBUTTON  BIT(0)  /**< \brief Middle mouse button */
 #define MOUSE_RIGHTBUTTON   BIT(1)  /**< \brief Right mouse button */
 #define MOUSE_LEFTBUTTON    BIT(2)  /**< \brief Left mouse button */
 #define MOUSE_SIDEBUTTON    BIT(3)  /**< \brief Side mouse button */
--- a/kernel/arch/dreamcast/hardware/maple/mouse.c
+++ b/kernel/arch/dreamcast/hardware/maple/mouse.c
@@ -56,7 +56,7 @@
     /* Fill the "nice" struct from the raw data */
     cooked = (mouse_state_t *)(frm->dev->status);
-    cooked->buttons = (~raw->buttons) & 14;
+    cooked->buttons = (~raw->buttons) & 15;
     cooked->dx = raw->dx - MOUSE_DELTA_CENTER;
     cooked->dy = raw->dy - MOUSE_DELTA_CENTER;
     cooked->dz = raw->dz - MOUSE_DELTA_CENTER;
```

gyrovorbis:
Fixes issue: https://github.com/KallistiOS/KallistiOS/issues/1532

Bit 0 of the button mask returned from the maple getcond query was not being stored by the mouse maple driver, nor did the driver have any indication that this bit corresponded to any actual button on any unit.

Turns out it does... it corresponds to the middle mouse button state on Japanese DC mice! Added a #define to the public interface for the bit mask of the middle button, plus fixed the mouse_reply() routine to store the full mask of all button state bits within the mouse_state_t.

Credit to ateam/Derek Pascarella for the bug report, fix, and testing.


kernel/arch/dreamcast/hardware/maple/mouse.c
```diff
@@ -56,7 +56,7 @@ static void mouse_reply(maple_state_t *st, maple_frame_t *frm) {
 
     /* Fill the "nice" struct from the raw data */
     cooked = (mouse_state_t *)(frm->dev->status);
-    cooked->buttons = (~raw->buttons) & 14;
+    cooked->buttons = (~raw->buttons) & 15;
```
andressbarajas: Can you replace the magic number 15 with some defined macros

gyrovorbis: Good call. Will do when I get home.

QuzarDC (approved):
At first I thought "Shouldn't there be more changes?" but then realized you hadn't done the API modernizing on the mouse like on the controller to have a struct with named bitfield members like in controller. Some day!

